PR #4, opened 2022-02-01
Feature/uw 19

ope-adejumo:


.github/workflows/CodeQC.yaml
```diff
@@ -0,0 +1,20 @@
...
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
+      run: |
+        find . -name '*.py' -exec pylint {} \;
```
christinaholtNOAA, line 20: We may want to consider a different approach. I believe this line will run pylint individually on each file, and could take quite a while. It will also skip the ability to tell you information about a package (based on the inclusion of an __init__.py file). 

I ran it on a package I currently maintain (already linted), and it took  51s to go through 12 files. Running the following command on that same package takes 23s and also gives information about duplicate code found in the package.

```
pylint [package name]
```

Until we have an actual package structure, we may want to use:

```
pylint *.py
```

This last option also runs pretty fast (11 s) and tells you the duplicate information, but you only get the files at the current directory level.

christinaholtNOAA (changes requested):
Ope, this looks good. I have just one concern about the pylint command below.

.github/workflows/CodeQC.yaml
ope-adejumo: @christinaholtNOAA I made changes based on your suggestion. I'm getting errors with exit code, which i'm trying to troubleshoot at the moment. Any insight from you on the error will be welcomed as well

jprestop: @ope-adejumo 
In case you are still needing help with this, I thought I'd try to help.  Clicking on the "Actions" tab and then "Feature/uw 19" workflow run and then "Code Quality Check", I can see the following errors (see attached screenshot):
Screen Shot 2022-02-03 at 5 03 00 PM


> helloWorld.py:4:0: C0304: Final newline missing (missing-final-newline)
It looks like you need a newline after "print("Good Bye World!!!")"

> helloWorld.py:1:0: C0103: Module name "helloWorld" doesn't conform to snake_case naming style (invalid-name)
It looks like "helloWorld.py" needs to be "hello_world.py" to follow snake_case naming conventions.

> helloWorld.py:1:0: C0114: Missing module docstring (missing-module-docstring)
It looks like you need to add a docstring. [See PEP 257 - Docstring Conventions](https://www.python.org/dev/peps/pep-0257/)

Please let me know if, as a stakeholder I should not be commenting on things like this - just wanted to help out if possible, but I will understand if that is not my role.

benjamin-cash: I don't think we're in the business of turning down helpful suggestions. :)

christinaholtNOAA (approved):

PR #9, opened 2022-02-15
removed duplicate gh actions

ope-adejumo:


fgabelmannjr (approved):

fgabelmannjr:
This looks good to me @ope-adejumo , but I'd like to also have @christinaholtNOAA or @aerorahul review as well.

.github/workflows/CodeQC.yaml
```diff
@@ -2,19 +2,23 @@ name: Run Pylint
...
 
     steps:
-    - uses: actions/checkout@v2
+    - name: Checkout
+      uses: actions/checkout@v2
+    
     - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
```
jprestop, line 15: @ope-adejumo, I see that you listed Python 3.8.  In [this Google Slides presentation](https://docs.google.com/presentation/d/1HCobC8eLu7qFiVVayNNj6Gkr2zBlaUO1JutYF67W1Uw/edit#slide=id.gf5c8a5e628_0_20) it lists Python 3.7+ .  I haven't been to all of the meetings, so maybe we switched to Python 3.8?  I listed Python 3.7 in the [.readthedocs.yaml](https://github.com/ufs-community/workflow-tools/pull/7/files#diff-03efc769b870804394632e45d7885272b44c16939517fb31c9d7c614d2ffae57) and in [docs.yaml](https://github.com/ufs-community/workflow-tools/pull/7/files#diff-d54d69dbb27e75dae25cb4b2384310cb57707e419377cf572d5cb0ecc1f16877) and am wondering if I should change it.  Can someone please confirm?

christinaholtNOAA: I think we settled on 3.7 being fine, but I think in the next sprint we have a card that will help us manage this a bit more uniformly. @jprestop Does this need to be 3.7 to work for your purposes, or will your GH Actions workflow set up the version you expect?

christinaholtNOAA (approved):
@ope-adejumo LGTM!

.github/workflows/CodeQC.yaml
jprestop: @christinaholtNOAA, I think it should all work fine, I just mainly wanted to ensure consistency so that we didn't have various versions spread throughout.  I'm also happy to change mine to 3.8 if that is desired for consistency.

PR #19, opened 2022-04-04
Uw 44 generate generic scheduler tool

ope-adejumo:
**Description**

**Type of change**

**How Has This Been Tested?**

**Checklist**


christinaholtNOAA:
@ope-adejumo Thanks for addressing some of the concerns we'd discussed in person here! It's really shaping up. Just because our future selves may thank us for being more present in the PR conversation, I am going to attempt to summarize a few offline conversations. Thanks for bearing with us as we figure it all out. 

- The Jira ticket was a wall of text and would have been more understandable with the addition of a UML diagram, provided [here](https://lucid.app/documents/view/ad6d2795-4464-49e6-8088-f470719c260a) after our conversations. The latest push indicates that it did seem to be helpful in communicating what we were looking for. :)
- @aerorahul Provided us a code example of the in the feature/scheduler branch. We will plan to use that as a guide for these types of tools as we grow our Factories. 
- In news unrelated to the implementation of the scheduler tool, we may need to look more closely at our pylint test. It may not be highlighting all the things we want...or it may not be finding all the python code. I will open a GH issue for that.

src/uwtools/generic_scheduler/slurm.py
```diff
@@ -1,14 +1,44 @@
...
+        if 'ntasks_per_node' in specs:
+            strings.append(f"{self._DIRECTIVE} --ntasks_per_node={specs.ntasks_per_node}")
+        if 'join' in specs:
+            strings.append(f"{self._DIRECTIVE} --error {specs.join}")
+        else:
+            if 'stdout' in specs:
+                strings.append(f"{self._DIRECTIVE} --output {specs.stdout}")
+            if 'stderr' in specs:
+                strings.append(f"{self._DIRECTIVE} --error {specs.stderr}")
+        return "\n".join(strings)
```
christinaholtNOAA, line 44: I think this section could be abstracted quite a bit to reduce duplicate information. Just spit balling, but I imaging something like (with the mod that I'm suggestion for map_flags above):

```
for flag, setting in self.map_flags.items():
    strings.append(f"{self._DIRECTIVE} --{flag}={setting}")
```

This takes advantage of the class already having these values set and the caller wouldn't need to pass them in again.

src/uwtools/generic_scheduler/slurm.py
```diff
@@ -1,14 +1,44 @@
...
-def create_directive_list():
-    pass
\ No newline at end of file
+    _DIRECTIVE = '#SBATCH'
+
+    def __init__(self, scheduler: str, job_name: str, partition: list, qos: str, output: Path, error: Path, walltime: datetime, account: str, nodes: int, ntasks_per_node: int, ntasks: int, cpus_per_task: int, reservation: str, join: bool, native_flags: list, run_command: str):
+        super().__init__(scheduler, job_name, partition, qos, output, error, walltime, account, nodes, ntasks_per_node, ntasks, cpus_per_task, reservation, join, native_flags, run_command)
+        
+    def map_flags():
+        mappings = {'account': '-A', 'partition': '-p', 'wallclock': '-t', 'job_name': '-J', 'nodes': '-N', 'ntasks_per_node': '-n', 'output': '-o', 'error': '-e'}
```
christinaholtNOAA, line 14: Here, I _think_ it could be useful to map in the other direction, so then you could map the value of the flag directly to the flag itself. 

You might also consider making it an object attribute in the __init__ method (instead of its own method here) so that you don't have to repeat the list of info, and all the values are stored in a convenient dict for retrieval elsewhere if needed.

christinaholtNOAA:
I just pushed a small batch of changes after @ope-adejumo and I talked through some of the fuzzy concepts of what should/shouldn't be in the base class and sub class. We also discussed some of the implementation details of providing native flags, joining stderr/stdout among diff schedulers, and what it means to write a job card.

christinaholtNOAA:
We should probably close this PR.

PR #23, opened 2022-04-14
update to recursively collect all python files

### LINKED ISSUE #22: GHA for pylint may be broken
**Expected behavior**

The GHA test for pylint should catch items like "unused imports", and it doesn't seem to be doing that.

**Current behavior**
On PR #19, I couldn't find a reference to `ast.List` in `src/uwtools/generic_scheduler/scheduler.py` after being interested in how it was being used. I checked to see if all the tests passed, mainly the linter after expecting that it should, and found that pylint reported a 10/10. I expected that it should fail with something < 10.

**Machines affected**
GitHub Actions

ryanlong1004:
**Description**

The original command ```pylint *py``` does not search recursively.  

Updated per https://stackoverflow.com/questions/36873096/run-pylint-for-all-python-files-in-a-directory-and-all-subdirectories.

Fixes #22 

**Type of change**
Update pylint command to act recursively.

**How Has This Been Tested?**
Running the pylint command via the CLI.


.github/workflows/CodeQC.yaml
```diff
@@ -6,19 +6,19 @@ jobs:
...
+          python-version: 3.8
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pylint
+
+      - name: Analysing the code with pylint
+        run: |
+          find . -type f -not -path "./venv/*" -name "*.py" | xargs pylint
```
christinaholtNOAA, line 24: Should we expect this call to pylint to handle linting all the files at once , or does it run the linter on each individual file in sequence?

ryanlong1004: @christinaholtNOAA attached sample output.

It will lint each file consecutively.
[pylint_sample_output.txt](https://github.com/ufs-community/workflow-tools/files/8491396/pylint_sample_output.txt)

aerorahul:
@ryanlong1004 
Thanks for this fix.

Please open the PR against `develop`.

What is the best way to run `pylint` in a project?
The [documentation](https://pylint.pycqa.org/en/latest/user_guide/run.html) shows multiple ways.  The approach here is applying pylint recursively file-by-file.

.github/workflows/CodeQC.yaml
christinaholtNOAA: We should probably be linting the whole package together. It will catch things we may want to be aware of at a package level like duplicate code and such.

ryanlong1004: @christinaholtNOAA in UW-44 branch, since packages were introduced here, you would want to call ```pylint src```.

image

PR #24, opened 2022-04-14
run pylint recursively in github actions

ryanlong1004:
Description

The original command pylint *py does not search recursively.

Updated per https://stackoverflow.com/questions/36873096/run-pylint-for-all-python-files-in-a-directory-and-all-subdirectories.

Fixes https://github.com/ufs-community/workflow-tools/issues/22

Type of change
Update pylint command to act recursively.

How Has This Been Tested?
Running the pylint command via the CLI.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,38 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+# Docstrings and comments use max_line_length = 79
+[*.py]
+max_line_length = 88
+
+# Use 2 spaces for the HTML files
+[*.html]
+indent_size = 2
+
+# The JSON files contain newlines inconsistently
+[*.json]
+indent_size = 2
+insert_final_newline = ignore
+
+# Makefiles always use tabs for indentation
+[Makefile]
+indent_style = tab
+
+# Batch files use tabs for indentation
+[*.bat]
+indent_style = tab
+
+[docs/**.txt]
+max_line_length = 79
+
+[*.yml]
+indent_size = 2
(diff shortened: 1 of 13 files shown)


aerorahul (approved):
looks good.

Need to figure out what best practices are for linting a project.
The [documentation](https://pylint.pycqa.org/en/latest/user_guide/run.html) shows multiple ways.  The approach here is applying pylint recursively file-by-file.

aerorahul:
Next obvious question is, should we be linting `docs/*.py` or restrict to `src/`

ryanlong1004:
Once everything is moved in ```src```, calling ```pylint src``` will lint all python files  in the src package.

We're recursively looking for python files now because packages won't be introduced until the UW-44 branch is merged into develop.

christinaholtNOAA (dismissed):
I think we should be linting all our python code, but that's just my preference. I think it's outside the scope of this PR to clean up the docs python code. Maybe we can open an issue for that one?

.editorconfig
```diff
@@ -0,0 +1,38 @@
+# https://editorconfig.org/
```
christinaholtNOAA, line 1: I think it's fine that this is added, but it does seem outside the scope of this PR, and will only help a subset of our code contributors: those using IDEs or those who have downloaded a plugin.

christinaholtNOAA (approved):

.editorconfig
ryanlong1004: @christinaholtNOAA sorry about that.  I was adding to my fork without realizing it would come up in the PR.

ryanlong1004:
@christinaholtNOAA agreed.  

Starting a few [discussions](https://github.com/ufs-community/workflow-tools/discussions) to start talking about a project charter and best practices.

christinaholtNOAA:
Ooh! I haven't used that tool yet. Very nice.

PR #28, opened 2022-04-21
Per discussion Today

ryanlong1004:
I added pytest and pytest coverage so we can see what our continued code coverage is.  Brief instructions on how to use added to README.md.

image

### DIFF
--- a/src/uwtools/scheduler.py
+++ b/src/uwtools/scheduler.py
@@ -1,0 +1,117 @@
+"""
+Job Scheduling
+author: Ryan Long <ryan.long@noaa.gov>
+"""
+
+import logging
+import collections
+from typing import Dict
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s",
+    handlers=[logging.StreamHandler()],
+)
+
+
+class JobCard(collections.UserList):
+    """represents a job card to submit to a scheduler"""
+
+    def content(self, line_separator: str = "\n") -> str:
+        """returns the formatted content of the job card"""
+        return line_separator.join(self)
+
+
+class JobScheduler(collections.UserDict):
+
+    _map = {}
+    prefix = ""
+
+    def __getattr__(self, name):
+        if name in self:
+            return self[name]
+
+    @property
+    def job_card(self) -> JobCard:
+        """returns a job card representation"""
+        return JobCard(
+            [
+                f"{self.prefix} {self._map[key] if key in self._map else key + '='}{value}"
+                for (key, value) in self.items()
+            ]
+        )
+
+    @classmethod
+    def get_scheduler(cls, props: Dict[str, str]) -> "JobScheduler":
+        """returns the appropriate scheduler
+
+        TODO: map_schedulers should be hoisted up out of the method
+        """
+        if "scheduler" not in props:
+            raise KeyError(
+                f"no scheduler defined in props: [{', '.join(props.keys())}]"
+            )
+
+        map_schedulers = {"slurm": Slurm, "pbs": PBS, "lsf": LSF, object: None}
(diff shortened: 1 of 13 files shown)


aerorahul (dismissed):
looks good.

aerorahul:
I would suggest breaking this PR into separate PR's addressing 1 specific issue.
This PR conflates too many in a single PR.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+        "queue": "-q",
+        "account": "-P",
+        "wall_time": "-W",
+        "total_nodes": "-n",
+        "cpus": "-R affinity[core()]",  # TODO
+        "number_tasks": "-R span[ptile=]",  # TODO
+        "change_dir": "-cwd /tmp",
+    }
+
+    prefix = "#BSUB"
```
christinaholtNOAA, line 113: Super picky comment, but maybe move this up above the `_map` definition for consistency?

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+    def __getattr__(self, name):
+        if name in self:
+            return self[name]
+
+    @property
+    def job_card(self) -> JobCard:
+        """returns a job card representation"""
+        return JobCard(
+            [
+                f"{self.prefix} {self._map[key] if key in self._map else key + '='}{value}"
```
christinaholtNOAA, line 39: The `else` branch here prevents the use of flags that don't have values. We'd discussed `--exclusive` as an example this morning.  Would just having `else key` work here? The user would be required to provide the `=` if it is needed (precedence in the implementation of required flags)?

christinaholtNOAA (approved):
I left a couple of comments below on easy items. The whole scheduler looks great! Thanks for addressing this, and walking us through it earlier. 

I think it's fine to do some of the clean-up in this PR -- removing old placeholders now that we're adding real code. I'm not so sure that so many other changes should go in as part of the same PR -- config file changes, etc. I'll let others weigh in on that, though.

christinaholtNOAA:
Oh! Ha! @aerorahul beat me to it!

ryanlong1004:
@aerorahul @christinaholtNOAA 

Thanks, I'm going to break it apart today.

src/uwtools/scheduler.py
ryanlong1004: It doesn't prevent values, but allows keys that are not listed in the map to not be translated (hence the adding of the '=' instead of appending to its value in the _map table.

It's not the final solution as I think we're going to have to build some intelligence into how the parsing takes the raw data -> JobCard, but it will allow tests to pass in the interim.

The tests illustrate what I mean.

christinaholtNOAA: I suppose I was suggesting that instead of `else key + '='` just having `else key`. That would allow a test like this to work:

```
expected = """#SBATCH scheduler=slurm
#SBATCH --job-name=abcd
#SBATCH --exclusive"""
props = {"scheduler": "slurm", "job_name": "abcd", "--exclusive": ""}
```

When user wanted to provide a key that requires an `=` then they'd need to put that as part of the key in props. The existing test would then turn out like this:

```
expected = """#SBATCH scheduler=slurm
#SBATCH --job-name=abcd
#SBATCH extra_stuff=12345"""

    props = {"scheduler": "slurm", "job_name": "abcd", "extra_stuff=": "12345"}
```

ryanlong1004: I understand now, thanks.  Updating.

aerorahul: doesn't that mean the user needs to know what needs an `=` and what doesn't?

ryanlong1004: Yes, the way it is now, yes.  We have to make smarter.

PR #29, opened 2022-04-21
Tool configs

ryanlong1004:
This unobtrusively adds three configuration files to the repo.  It's unobtrusive because they only need to be present to work.

[tox.ini](https://tox.wiki/en/latest/index.html) is a tool to validate multiple versions of uwtools.

[bump2version](https://github.com/c4urself/bump2version) has been instrumental in keeping releases and tag syncing and publishing very easy.

[editorconfig](https://editorconfig.org/) will ensure that all formatting across all IDE's is the same.  Not just for Python, but YAML files, JSON files etc.  

My IDE autoformats on save.  This can lead to files being changed due to formatting only.  This will prevent that to anyone using the repo.


aerorahul (changes requested):
I am fine with the 3 files you mention in the description. But, why are the scheduler files being added in this PR then?

PR #31, opened 2022-04-21
Refactor scheduler

ryanlong1004:
Refactor scheduler as discussed with tests.


src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s",
+    handlers=[logging.StreamHandler()],
+)
+
+
+class JobCard(collections.UserList):
+    """represents a job card to submit to a scheduler"""
```
aerorahul, line 18: We follow `numpy` documentation protocol.
You can tell your IDE to use that, and it will automagically format the doc block.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
+"""
+Job Scheduling
+author: Ryan Long <ryan.long@noaa.gov>
+"""
```
aerorahul, line 4: `git log` will save the name of the developer.
Also, it becomes unwieldy when many developers add their names to the list as time progresses.
The doc block (if present) should follow numpy protocol.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+
+import logging
+import collections
+from typing import Dict
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s",
+    handlers=[logging.StreamHandler()],
+)
```
aerorahul, line 14: I am glad you are thinking about logging. 
There was a PR #20 that was closed because, we were not ready to discuss logging  yet.

tests/test_scheduler.py
```diff
@@ -0,0 +1,47 @@
+# pylint: disable=all
+import pytest
+
+from src.uwtools.scheduler import JobScheduler
+
+
+def test_dummy():
+    """Skip the test. Used only as a placeholder."""
+    return True
```
aerorahul, line 9: Can be deleted

tests/test_scheduler.py
```diff
@@ -0,0 +1,47 @@
+# pylint: disable=all
+import pytest
+
+from src.uwtools.scheduler import JobScheduler
```
aerorahul, line 4: Whats `src`?
This should be `uwtools.scheduler`.

aerorahul (dismissed):
Apart from the fact that 3 files from PR #30 are showing up again, there are a few comments.

tests/test_scheduler.py
aerorahul: I believe that is why the pytest is failing.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+
+class PBS(JobScheduler):
+    """represents a PBS based scheduler"""
+
+    prefix = "#PBS"
+
+    _map = {
+        "bash": "-S",
```
aerorahul, line 83: this should be `shell` not `bash`.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,117 @@
...
+        "place": "-l place=",
+        "debug": "-l debug=",
+    }
+
+
+class LSF(JobScheduler):
+    """represents a LSF based scheduler"""
+
+    _map = {
+        "bash": "-L",
```
aerorahul, line 101: Needs to be `shell` not `bash`.

tests/test_scheduler.py
ryanlong1004: @aerorahul thanks for pointing this out; needed to make an update to the github action.

aerorahul: I don't think this is correct. The install should not be src.uwtools

.bumpversion.cfg
```diff
@@ -0,0 +1,6 @@
+bumpversion]
```
christinaholtNOAA, line 1: Is this missing an open bracket?

christinaholtNOAA (commented):
Please also see my comments in PR #28.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,128 @@
...
+        logging.debug("getting scheduler type %s", map_schedulers[props["scheduler"]])
+        return map_schedulers[props["scheduler"]](props)
+
+
+class Slurm(JobScheduler):
+    """represents a Slurm based scheduler"""
+
+    prefix = "#SBATCH"
+
+    _map = {
```
aerorahul, line 76: what happens if I don't provide a key in the `props`? e.g. a `job_name`.
The job-card is still a valid job card and the scheduler will accept it.
Are we enforcing that these `keys` must be provided?

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,128 @@
...
+        "shell": "-S",
+        "job_name": "-N",
+        "output": "-o",
+        "job_name": "-j",  # TODO 2 job name defs
+        "queue": "-q",
+        "account": "-A",
+        "wall_time": "-l walltime=",
+        "total_nodes": "-l select=",
```
aerorahul, line 101: what happens when the resource request is total_nodes, total_tasks_per_node, threads, mpi_procs.  All this synthesizes into:
```
-l select=TOTAL_NODES:mpiprocs=MPIRANKS:ompthreads=THREADS:ncpus=TOTAL_RANKS
```

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class PBS(JobScheduler):
+    """represents a PBS based scheduler"""
+
+    prefix = "#PBS"
+
+    _map = {
+        "shell": "-S",
+        "job_name": "-N",
```
aerorahul, line 95: a `job_name` is usually optional to the scheduler, so is the stdout and stderr file names.

src/uwtools/scheduler.py
ryanlong1004: Right now, it doesn't get parsed but passes the K/V pair as is.

src/uwtools/scheduler.py
ryanlong1004: Previous comment on making this part 'smarter'.

tests/test_scheduler.py
ryanlong1004: It works both ways, as we're specifying to look undre 'src' in 'setup.cfg'.

Removing the prefix 'src'.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,128 @@
...
+
+class JobScheduler(collections.UserDict):
+    """object that creates JobCard"""
+
+    _map = {}
+    prefix = ""
+
+    def __getattr__(self, name):
+        if name in self:
+            return self[name]
```
aerorahul, line 38: can you explain where is this used?

ryanlong1004: This gives us the ability to use dot notation for any props.

object["job_name"] and object.job_name are both equivalent.

aerorahul: Understood.
So if one were to use this `.` notation, we would have to implement the `getter` in every place we wish to use this.

src/uwtools/scheduler.py
```diff
@@ -0,0 +1,128 @@
...
+    level=logging.DEBUG,
+    format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s",
+    handlers=[logging.StreamHandler()],
+)
+
+
+class JobCard(collections.UserList):
+    """represents a job card to submit to a scheduler"""
+
+    def content(self, line_separator: str = "\n") -> str:
```
benjamin-cash, line 19: content includes the "schedulerslurm" information as its first entry, which I believe will cause the job to be rejected on submission. Is the idea that the user will skip that entry when writing out the content to a file?

src/uwtools/scheduler.py
ryanlong1004: no, just here.  All the subclasses will inherit it.

christinaholtNOAA (approved):

fgabelmannjr (approved):

PR #35, opened 2022-04-26
implement remaining schedulers and update tests

ryanlong1004:
Remaining schedulers implementation (lsf and pbs).

* Added solution to remove 'scheduler` and other keys from being output to `JobCard`.
* Added a `key_value_separator` for each concrete so they can all share the base property `job_card `.

Extending `JobScheduler` is very lean.  See the full PBS implementation below:

```python
class PBS(JobScheduler):
    """represents a PBS based scheduler"""

    prefix = "#PBS"
    key_value_separator = " "

    _map = {
        "shell": "-S",
        "job_name": "-N",
        "output": "-o",
        "job_name2": "-j",  # TODO 2 job name defs
        "queue": "-q",
        "account": "-A",
        "wall_time": "-l walltime=",
        "total_nodes": "-l select=",
        "cpus": "cpus=",
        "place": "-l place=",
        "debug": "-l debug=",
    }
```

Updated tests.


aerorahul (commented):
Can you provide a `props` dictionary that creates this PBS directive exactly:

```
#PBS -S /bin/bash
#PBS -N MyJob
#PBS -j oe
#PBS -q  batch
#PBS -A MyAccounting
#PBS -l walltime=01:30:00
#PBS -l select=112:mpiprocs=24:ompthreads=5:ncpus=120
#PBS -l place=vscatter:excl
#PBS -l debug=true
```

The output is explicitly joined into `MyJob.job_id`, where `job_id` will result from the job id assigned by the scheduler.

src/uwtools/scheduler.py
```diff
@@ -60,7 +84,12 @@ def get_scheduler(cls, props: Dict[str, str]) -> "JobScheduler":
                 f"no scheduler defined in props: [{', '.join(props.keys())}]"
             )
 
-        map_schedulers = {"slurm": Slurm, "pbs": PBS, "lsf": LSF, object: None}
+        map_schedulers = {
+            "slurm": Slurm,
+            "pbs": PBS,
+            "lsf": LSF,
+            object: None,
+        }  # TODO Does this need a default?
```
aerorahul, line 92: No. there is no concept of a "default" scheduler on an HPC.

src/uwtools/scheduler.py
```diff
@@ -60,7 +84,12 @@ def get_scheduler(cls, props: Dict[str, str]) -> "JobScheduler":
...
         logging.debug("getting scheduler type %s", map_schedulers[props["scheduler"]])
         return map_schedulers[props["scheduler"]](props)
```
aerorahul, line 94:
```suggestion
        try:
             return map_schedulers[props["scheduler"]](props)
        except KeyError:
             raise KeyError(f"{props['scheduler']} is not a supported scheduler" +
                           "Currently supported schedulers are:\n" + 
                           f'{" | ".join(map_schedulers.keys())}"')
```

tests/test_scheduler.py
```diff
@@ -18,6 +17,32 @@ def test_scheduler():
     assert actual == expected
 
 
+def test_scheduler_lfs():
```
aerorahul, line 20:
```suggestion
def test_scheduler_lsf():
```

src/uwtools/scheduler.py
aerorahul: Might even want a `logging.critical` message here.

aerorahul (changes requested):
A comment, a suggestion and a typo.

ryanlong1004:
@aerorahul since they're not required by the scheduler and they all share the `-l`, they shouldn't be included in the `_map`.  

The user can pass them as a key/value and leave the validation up to the scheduler.

src/uwtools/scheduler.py
```diff
@@ -34,15 +32,41 @@ class JobScheduler(collections.UserDict):
...
+        raise AttributeError(name)
 
     @property
-    @abstractmethod
-    def job_card(self) -> JobCard:
-        raise NotImplementedError
+    def job_card(self):
+        """returns the job card to be fed to external scheduler"""
+        ignored_keys = ["scheduler"]
+
```
christinaholtNOAA, line 46: Is it appropriate to pop scheduler from self._map and set it as an attribute in __init__ (or just pop it where ever it might be used)? That would reduce the logic here a bit.

src/uwtools/scheduler.py
```diff
@@ -34,15 +32,41 @@ class JobScheduler(collections.UserDict):
 
     _map = {}
```
christinaholtNOAA, line 33: Should `_map` be an instance variable instead of a class variable? It is different for all objects, right?

christinaholtNOAA (approved):
A few questions. Otherwise LGTM.

christinaholtNOAA:
@ryanlong1004 @aerorahul I agree. Complex resource mappings like that should probably be left to the user to provide the appropriate string as a key/value pair.

src/uwtools/scheduler.py
ryanlong1004: It's the same of all objects of that class.

src/uwtools/scheduler.py
christinaholtNOAA: That makes sense.

aerorahul:
> @ryanlong1004 @aerorahul I agree with Ryan. Complex resource mappings like that should probably be left to the user to provide the appropriate string as a key/value pair.

Consider a user request with the following resources.
```
nodes = 5
tasks_per_node = 4
threads = 7
```
i.e.  a user wishes to run a job using 5 nodes, with 4 tasks on each node, and each task consuming 7 threads.

The lines for the different schedulers are:
```
#PBS -l select 5:mpiprocs=4:omp_threads=7:ncpus=28
```

```
#BSUB
#BSUB -n 20
#BSUB -R affinity[core(7)]
#BSUB -R span[ptile=4]
```

```
#SBATCH --nodes=5
#SBATCH --tasks-per-node=4
#SBATCH --cpus-per-task=7
```

The user should provide to the `Scheduler` class only the information from above.   It is the responsibility of the `Scheduler` to map into the (complex) set of directives.

aerorahul:
> I will say that, as a user, one of the more appealing features of this tool is the capability that @aerorahul describes - namely not needing to track down details of the scheduler flags for my specific resource request. I realize that will get complicated to implement, and may need its own separate ticket and PR, but it is a capability I would definitely want the tool to have.

> I will say that, as a user, one of the more appealing features of this tool is the capability that @aerorahul describes - namely not needing to track down details of the scheduler flags for my specific resource request. I realize that will get complicated to implement, and may need its own separate ticket and PR, but it is a capability I would definitely want the tool to have.

It doesn't have to be complicated.
But, without that, this "tool" is useless if a user needs to know the scheduler specific flags.

benjamin-cash:
So would you say then that the implementation of the schedulers is not actually complete without this capability?

aerorahul:
> So would you say then that the implementation of the schedulers is not actually complete without this capability?

Correct.

benjamin-cash:
> > So would you say then that the implementation of the schedulers is not actually complete without this capability?
> 
> Correct.

Ok, that answers the question I posed on the slack channel then. :)

ryanlong1004:
Resolved by #41

PR #36, opened 2022-04-26
Feature slurm validation

ryanlong1004:
**Description**

Job card validation.

https://jira-epic.woc.noaa.gov/browse/UW-142

**Type of change**

**How Has This Been Tested?**
On hera.  1st execution passed as expected.  2nd failed with error 'acccount' as expected.
image

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -1,0 +1,1 @@
+python -m pytest --cov-report html:htmlcov --cov=uwtools tests/
--- a/src/uwtools/scheduler.py
+++ b/src/uwtools/scheduler.py
@@ -173,7 +173,7 @@ def submit_job_card(content) -> int:
     stdout, stderr = p.communicate(input=content.encode())
     job_id = stdout.decode().split(" ")[-1]
     if stderr:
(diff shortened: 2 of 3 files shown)


coverage.sh
```diff
@@ -0,0 +1 @@
+python -m pytest --cov-report html:htmlcov --cov=uwtools tests/
```
christinaholtNOAA, line 1: Had no idea it would be this easy to get a coverage report!

src/uwtools/scheduler.py
```diff
@@ -144,5 +161,33 @@ class LSF(JobScheduler):
     }
 
 
+@shelve_it("cache.shelve")
+def submit_job_card(content) -> int:
+    """sends the job card to the scheduler for processing"""
+    p = subprocess.Popen(
+        ["sbatch", "--deadline=now+1"],
```
christinaholtNOAA, line 168: This is a slurm-specific submission. We will likely need the submission string to be provided for us from the scheduler subclasses.

PR #37, opened 2022-04-27
Feature uniform logging

ryanlong1004:
**Description**
https://jira-epic.woc.noaa.gov/browse/UW-118

**Type of change**

**How Has This Been Tested?**
feature_logging
feature_logging2

### DIFF
--- a/src/uwtools/log_base.py
+++ b/src/uwtools/log_base.py
@@ -29,7 +29,7 @@ LEVELS = {
 
 def handle_logging(**kwargs):  # pylint: disable=unused-variable
     """handles default logging"""
-    level = LEVELS["DEFAULT"]
+    level = LEVELS["default"]
     if "args" in kwargs:
         level = LEVELS.get(kwargs["args"].log.lower())
     if level is None:
--- a/src/uwtools/logging.py
+++ b/src/uwtools/logging.py
@@ -1,0 +1,44 @@
+"""
+Encapsulates logging uniformity and exposes convenience methods
+to enforce requirements while not coupling.
+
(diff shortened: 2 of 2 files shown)


src/uwtools/log_base.py
```diff
@@ -0,0 +1,44 @@
...
+}
+
+
+def handle_logging(**kwargs):  # pylint: disable=unused-variable
+    """handles default logging"""
+    level = LEVELS["default"]
+    if "args" in kwargs:
+        level = LEVELS.get(kwargs["args"].log.lower())
+    if level is None:
+        raise ValueError(f"log level must be one of: {' | '.join(LEVELS.keys())}")
```
aerorahul, line 36: I don't think this needs to be an error.  If its `None`, then use the `default` level.

src/uwtools/log_base.py
```diff
@@ -0,0 +1,44 @@
...
+
+    logging.basicConfig(
```
aerorahul, line 38: How does this handle logging to `stdout` and `file`?  Specifically, we should provide user option for a `FileHandler` and ensure that it is the same file it will be written to throughout the execution.

aerorahul (commented):
Looks good.
My only request is to enable logging to a file and provide an easy way to initialize that.

Other than that, I would think it would log with the `Format` and `name` would be the appropriate module it is currently logging in.

aerorahul:
Oh, its not in this ticket, but a decorator `@log_it` would be wonderful to have.

ryanlong1004:
closing as per discussion with @aerorahul

PR #45, opened 2022-05-04
44 refactor attributemap

ryanlong1004:
AttributeMap was convoluted and good be removed.

### DIFF
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -6,11 +6,10 @@ from uwtools.scheduler import JobScheduler
 from uwtools.loaders import load_yaml
 
 
-
 def test_scheduler_dot_notation():
     props = load_yaml(pathlib.Path("tests/fixtures/simple.yaml"))
 
-    js = JobScheduler.get_scheduler(props)
+    js = JobScheduler.get_scheduler(dict(props))
     expected = "user_account"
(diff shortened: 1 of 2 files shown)


tests/test_scheduler.py
```diff
@@ -391,16 +386,16 @@ def test_lsf2():
     expected = """#BSUB -P account_name
 #BSUB -q batch
 #BSUB -W 00:01:00
-#BSUB -R affinity[core(1)]
+#BSUB -n 12
```
aerorahul, line 389: This test shouldn't be passing.
The number here should be 144, based on the `props` input.

ryanlong1004: created issue

tests/test_scheduler.py
```diff
@@ -415,9 +410,9 @@ def test_lsf3():
     expected = """#BSUB -P account_name
 #BSUB -q batch
 #BSUB -W 00:01:00
-#BSUB -R affinity[core(1)]
+#BSUB -n 2
```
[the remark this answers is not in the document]
ryanlong1004: created issue #48

aerorahul (approved):
Looks good.
Issues have been created to address the LSF test incorrectness.

PR #51, opened 2022-05-05
Create codeql-analysis.yml

ryanlong1004:
Please do not close or merge.

Trying to see if we want to adopt it.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ develop ]
+  schedule:
+    - cron: '30 19 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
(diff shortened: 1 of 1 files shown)


ryanlong1004:
Results are integrated into the repository security tab.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+    # 📚 See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
+
+    #   If the Autobuild fails above, remove it and uncomment the following three lines. 
+    #   modify them (or add more) to build your code if your project, please refer to the EXAMPLE below for guidance.
+
+    # - run: |
+    #   echo "Run, Build Application using script"
+    #   ./location_of_script_within_repo/buildscript.sh
+
+    - name: Perform CodeQL Analysis
```
ope-adejumo, line 71: Does this type of Static Code analysis cover also Syntactic code analysis? Reading through this action I only see Semantic

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ develop ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ develop ]
+  schedule:
```
ope-adejumo, line 20: Why would we need a cron job to trigger this workflow? Would this be like for Reporting purposes to run at that time once a week?

.github/workflows/codeql-analysis.yml
ryanlong1004: Pylint takes care of that.  Should never reach this point before.

aerorahul (approved):
Good to have code quality analysis.
It would be good to see report of this analysis. 
Perhaps add a little blurb on how to read said analysis.

.github/workflows/codeql-analysis.yml
ryanlong1004: I **believe** this has to do with database maintenance and perhaps syncing with Github.

Issue #48, opened 2022-05-05
LSF Flags Issue


ryanlong1004:
something is not right here either.
The `-n` is `nodes * tasks_per_node`
The `tasks_per_node` is the `ptile` line.

_Originally posted by @aerorahul in https://github.com/ufs-community/workflow-tools/pull/45#discussion_r865156224_

aerorahul:
#56 resolved this issue.

Issue #47, opened 2022-05-05
This test shouldn't be passing.


ryanlong1004:
This test shouldn't be passing.
The number here should be 144, based on the `props` input.

_Originally posted by @aerorahul in https://github.com/ufs-community/workflow-tools/pull/45#discussion_r865155152_

aerorahul:
#56 fixed this.

PR #56, opened 2022-05-07
Fix lsf jobcard issues

### LINKED ISSUE #43: LSF Memory Logic
In LSF, the memory needs to be represented in terms of kilobytes.

*Example*
The user will request memory: 1MB, but the scheduler directive would be #BSUB -R rusage[mem=1000]

aerorahul:
Fixes #43 #47 #48

**Description**
LSF tests were showing incorrect nodes.  This PR fixes them.

**Type of change**

- [X] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/scheduler.py
+++ b/src/uwtools/scheduler.py
@@ -30,7 +30,7 @@ class OptionalAttribs:  # pylint: disable=too-few-public-methods
     """key for optional attributes"""
 
     SHELL = "shell"
-    JOB_NAME = "job_name"
+    JOB_NAME = "jobname"
     STDOUT = "stdout"
     STDERR = "stderr"
     JOIN = "join"
@@ -39,7 +39,6 @@ class OptionalAttribs:  # pylint: disable=too-few-public-methods
     MEMORY = "memory"
     DEBUG = "debug"
     EXCLUSIVE = "exclusive"
-    CPUS = "cpus-per-task"
     PLACEMENT = "placement"
 
 
@@ -197,12 +196,13 @@ class PBS(JobScheduler):
         RequiredAttribs.NODES: lambda x: f"-l select={x}",
         RequiredAttribs.QUEUE: "-q",
         RequiredAttribs.WALLTIME: "-l walltime=",
-        RequiredAttribs.TASKS_PER_NODE: ":mpiprocs=",
+        RequiredAttribs.TASKS_PER_NODE: "mpiprocs",
         OptionalAttribs.SHELL: "-S",
         OptionalAttribs.JOB_NAME: "-N",
         OptionalAttribs.STDOUT: "-o",
-        OptionalAttribs.DEBUG: "-l debug=",
-        OptionalAttribs.THREADS: ":ompthreads=",
+        OptionalAttribs.DEBUG: lambda x: f"-l debug={str(x).lower()}",
+        OptionalAttribs.THREADS: "ompthreads",
+        OptionalAttribs.MEMORY: "mem",
     }
 
     def pre_process(self) -> Dict[str, Any]:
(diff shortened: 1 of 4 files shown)


src/uwtools/scheduler.py
```diff
@@ -222,20 +222,23 @@ def pre_process(self) -> Dict[str, Any]:
     def select(self, items) -> Dict[str, Any]:
         """select logic"""
         total_nodes = items.get(RequiredAttribs.NODES, "")
-        cores_per_node = items.get(RequiredAttribs.TASKS_PER_NODE, "")
-        threads_per_core = items.get(OptionalAttribs.THREADS, "")
+        tasks_per_node = items.get(RequiredAttribs.TASKS_PER_NODE, "")
+        # Set default threads=1 to address job variability with PBS
+        threads = items.get(OptionalAttribs.THREADS, 1)
```
ryanlong1004, line 227: Nice.

tests/test_scheduler.py
```diff
@@ -231,9 +231,6 @@ def test_pbs8():
     js = JobScheduler.get_scheduler(props)
     actual = js.job_card.content()
 
-    print(actual)
```
ryanlong1004, line 234: Thanks.

tests/test_scheduler.py
```diff
@@ -215,8 +215,8 @@ def test_pbs8():
     expected = """#PBS -A account_name
 #PBS -q batch
 #PBS -l walltime=00:01:00
-#PBS -l debug=True
-#PBS -l select=1:mpiprocs=1"""
+#PBS -l debug=true
```
ryanlong1004, line 218: good idea `tolower()`

ryanlong1004 (approved):

src/uwtools/scheduler.py
```diff
@@ -222,20 +222,23 @@ def pre_process(self) -> Dict[str, Any]:
...
         memory = items.get(OptionalAttribs.MEMORY, "")
 
-        items["-l select"] = f"={total_nodes}:mpiprocs={cores_per_node}"
-        if threads_per_core not in NONEISH:
-            items[
-                "-l select"
-            ] = f"{items['-l select']}:ompthreads={threads_per_core}:ncpus={int(cores_per_node) * int(threads_per_core)}"
+        select = [f"{total_nodes}",
```
christinaholtNOAA, line 230: This variable has the same name as the function. Please choose a different variable name so that the outer scope isn't redefined.

src/uwtools/scheduler.py
```diff
@@ -269,33 +272,28 @@ class LSF(JobScheduler):
...
 
-    def pre_process(self):
+    def pre_process(self) -> Dict[str, Any]:
         items = self._data
-        items[self._map[OptionalAttribs.THREADS](items[OptionalAttribs.THREADS])] = ""
+        # LSF requires threads to be set (if None is provided, default to 1)
+        items[OptionalAttribs.THREADS] = items.get(OptionalAttribs.THREADS, 1)
+        memory = items.get(OptionalAttribs.MEMORY, "")
+        if memory not in NONEISH:
+            items[self._map[OptionalAttribs.MEMORY](items[OptionalAttribs.MEMORY])] = ""
```
christinaholtNOAA, line 290: I'm confused by the logic here. I read it as:
- Get memory
- If memory is set to a value, save it to the config as an empty string

Is that right? Is it all about the key in this case because we're filling in the value with a `lambda` function?

Also, do you need to interrogate `items[OptionalAttribs.MEMORY]` again, or can you just reference `memory` where it's already stored?

src/uwtools/scheduler.py
```diff
@@ -269,33 +272,28 @@ class LSF(JobScheduler):
...
-        ] = ""
-        return items
-
-    def select(self, items: Dict[str, Any]):
-        """select logic"""
-
-        nodes = items.get(RequiredAttribs.NODES, "")
-        tasks_per_node = items.get(RequiredAttribs.TASKS_PER_NODE)
-        items[RequiredAttribs.NODES] = int(nodes) * int(tasks_per_node)
+        items[RequiredAttribs.NODES] = int(items[RequiredAttribs.TASKS_PER_NODE] * int(items[RequiredAttribs.NODES]))
```
christinaholtNOAA, line 296: Should `items[RequiredAttribs.TASKS_PER_NODE]` be a guaranteed `int` here?

christinaholtNOAA (commented):
I have a couple of nit picky comments below. Looks like a good set of additions to me, otherwise.

src/uwtools/scheduler.py
aerorahul: You got the 1st part right:
1. Get memory.

If it was provided, the `items` dictionary now contains a new element with the map of what that memory line translates to for the chosen scheduler, the value of which is an empty string.
e.g. in this case (LSF), the new element in `items` will be:
```
items[-R rusage[mem=1MB] = ""
```

src/uwtools/scheduler.py
aerorahul: Does it matter? doesn't `int( )` take care of that?

src/uwtools/scheduler.py
aerorahul: Yes. That would be my oversight.

src/uwtools/scheduler.py
christinaholtNOAA: It took me a minute to realize that in this "special" case, we'd be returning a lambda function and passing it the input value of memory. And that once you do that, the key will be the value for this item (thanks to a brush up on this data structure from @ryanlong1004).

src/uwtools/scheduler.py
christinaholtNOAA: If it's a string, what happens in the multiplication step?

aerorahul: good eye.  The `int( )` was misplaced.

christinaholtNOAA (approved):
LGTM!

PR #64, opened 2022-05-11
initial commit with tests

ryanlong1004:
Memory conversion for LSF.

Needed for #? LSF Memory Fix.


aerorahul (approved):
Serves the purpose for conversion needed in the scheduler.

PR #66, opened 2022-05-12
43 lsf memory logic

### LINKED ISSUE #43: LSF Memory Logic

ryanlong1004:
Fixes #43


aerorahul (approved):
looks good.
I can look later about the subtleties of the memory units on the different schedulers.
They can be picky.

PR #69, opened 2022-05-16
54 use GitHub actions to post pytest cov report to GitHub pages

ope-adejumo:
**Description**

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/.github/workflows/pytest-cov.yaml
+++ b/.github/workflows/pytest-cov.yaml
@@ -38,10 +38,9 @@ jobs:
           name: coverage-report
           path: ./cov_html
 
-      - name: Download Reports for GitHub pages
-        if: ${{ always() }}
-        uses: actions/download-artifact@v2
+      - name: Deploy Coverage Reports to GitHub pages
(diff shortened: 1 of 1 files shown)


aerorahul (changes requested):
@ope-adejumo 
I don't think you need to run the coverage for a range of python versions.
Either a test is there or it is not.  It does not depend (yet) on the version of python.
I think you could just choose the minimum version of python we are currently supporting; 3.7

Please update L14 of pytest-cov.yaml to:
```
python: ["3.7"]
```

.github/workflows/pytest-cov.yaml
```diff
@@ -42,5 +42,5 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@4.1.4
         with:
           branch: gh-pages
-          folder: cov_html
+          folder: coverage-report
```
aerorahul, line 45: I think this needs to be `cov_html`, no?

aerorahul (approved):
looks good.
lets see if we can see the output on the pages

PR #70, opened 2022-05-16
Create hotfix_template.md

ryanlong1004:
This would introduce a hotfix reporting template, for reporting issues that require immediate attention as the are breaking production.


aerorahul (approved):
Looks good

PR #71, opened 2022-05-16
53 create git actions workflow pytest cov

ryanlong1004:
image

Will need to remove the [develop, 53...] once merged.


.github/workflows/pytest-cov.yaml
```diff
@@ -0,0 +1,43 @@
...
+          name: coverage-report
+          path: docs
+
+      - name: Upload to pages
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add -f .
+          git commit -m "update code coverage"
+          git push
```
aerorahul, line 43: I don't this is necessary if we use the action from the marketplace. 
Also, gh_pages should only contain the coverage, not the entire repo.

aerorahul (changes requested):
I don't think this is the right way to do this. 
Also, a lot of files that shouldn't be committed are here. 
May be this is work in progress.

PR #72, opened 2022-05-18
53 create git actions workflow pytest cov

ryanlong1004:
Uses github actions to post code-coverage to GH pages using the gh-pages branch as its source.


ryanlong1004:
image

.github/workflows/pytest-cov.yaml
```diff
@@ -1,49 +1,56 @@
 name: Pytest Coverage
-on: 
+on:
   push:
-    branches: [ develop ]
+    branches: [develop, 53-create-git-actions-workflow-pytest-cov]
```
christinaholtNOAA, line 4: I think this line should be reverted.

.github/workflows/pytest-cov.yaml
```diff
@@ -1,49 +1,56 @@
...
+      - name: Install App
         run: |
-          pip install --upgrade pip
+          python -m pip install --upgrade pip
           pip install -e .[dev]
 
       - name: Run pytest-cov
         run: |
-          pytest --cov=src/uwtools/ --cov-fail-under=5 --cov-report=html:cov_html
+          pytest --cov-fail-under=5 --cov-report=html --cov=uwtools
```
christinaholtNOAA, line 27: Is this 5% as a min? The docs are not very clear on what that unit is. 

If 5%, do we want to shoot for something a little higher since we're seeing pretty good coverage already? Maybe we could just set it to like 80% so we get informed if it drops too much lower than the current levels?

ope-adejumo (commented):
I think the "53 create git actions workflow pytest cov" branch might need a little bit up clean up. I see some redundant folders ex. coverage, e.t.c that we don't need committed to develop.

christinaholtNOAA (changes requested):
I have a couple of comments below for changes, but before I dig too deep, I have a bigger one.

I'm worried about seeing all the additional files showing up. Do the .pytest_cache, assets, css, and coverage dirs need to be in the repo? I am assuming that they are all "artifacts" of running coverage/tests.

.github/workflows/pytest-cov.yaml
christinaholtNOAA: I was just suggesting a placeholder that would be a tad more useful now. If we choose something that is just below our current level, we won't be dropping too low in the interim before we know what the true desired value will be.

ryanlong1004: Changed to 70%

christinaholtNOAA (changes requested):
I think there are still a lot of files committed here that shouldn't be in the repo. Everything under `.pytest_cache` and `__pycache__` for sure. I'm less sure about `src/uwtools.egg-info`, `assets`, and such.

ryanlong1004:
Closing.

PR #74, opened 2022-05-20
add str method

### LINKED ISSUE #73: UW-144
As the scheduling tool, I want a simple method for outputting formatted job card content to a file

ryanlong1004:
Implements https://jira-epic.woc.noaa.gov/browse/UW-144

Closes #73

### DIFF
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -445,3 +445,29 @@ def test_lsf4():
     actual = js.job_card.content()
 
     assert actual == expected
+
+
+def test_string_output():
+    expected = """#BSUB -P account_name
+#BSUB -q batch
+#BSUB -W 00:01:00
+#BSUB -n 6
+#BSUB -R span[ptile=3]
+#BSUB -R affinity[core(2)]
+#BSUB -R rusage[mem=1000KB]
+""".strip()
+
+    props = {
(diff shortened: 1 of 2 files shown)


tests/test_scheduler.py
```diff
@@ -445,3 +445,29 @@ def test_lsf4():
...
+        "walltime": "00:01:00",
+        "nodes": 2,
+        "tasks_per_node": 3,
+        "threads": 2,
+        "memory": "1MB",
+    }
+
+    js = JobScheduler.get_scheduler(props)
+    actual = js.job_card.content()
+    assert actual.strip() == expected.strip()
```
christinaholtNOAA, line 473: It seems that `actual` was "stripped" above, I'm curious about the need to do that twice. Also,  should the return from `__str__` be stripped instead of doing it here? 

It's interesting that this test uses `strip()` and others do not. If that is the part being tested, would you mind adding a doc block for this test that the `strip()` is the important part?

ryanlong1004: It's so that I don't have to worry about the """ """ formatting.

ryanlong1004: Removed.

christinaholtNOAA (approved):

PR #82, opened 2022-06-01
update tests

ryanlong1004:
Tests the 'print job card' functionality.


tests/test_scheduler.py
```diff
@@ -468,5 +468,7 @@ def test_string_output():
     }
 
     js = JobScheduler.get_scheduler(props)
-    actual = js.job_card.content()
+    jc = js.job_card
+    actual = str(jc)
+    print(jc)
```
christinaholtNOAA, line 473: I assume that we want to expect that `__str__` is actually returning a string, and don't want to hide that potential oversight (an inadvertent change to the `__str__` method) in the `str()` on line 472. Could it just be:
```
actual = js.job_card
assert actual == expected
```

Also, would it be okay to remove the `print()`?

christinaholtNOAA (changes requested):
Looks good.

aerorahul:
waiting on resolution from @christinaholtNOAA's comment.
And then we can close/merge this.

tests/test_scheduler.py
aerorahul, line 473:
```suggestion
```

aerorahul (approved):
looks good.

ryanlong1004:
@christinaholtNOAA bump

PR #83, opened 2022-06-01
77 Fix Pylint Errors

ryanlong1004:
Update README


README.md
```diff
@@ -21,3 +21,5 @@ Documentation is automatically generated through [Read the Docs](https://readthe
 [Developer Status](https://github.com/orgs/ufs-community/projects/1)
 
 [UW Tools Github Pages Site](https://ufs-community.github.io/workflow-tools/)
+ 
+ This is new stuff
```
christinaholtNOAA, line 25: The README should be reverted to its original state.

src/uwtools/logger.py
```diff
@@ -105,14 +106,14 @@ def add_handlers(cls, logger: logging.Logger, handlers: List[logging.Handler]):
         -------
         logger
         """
-        for hh in handlers:
-            logger.addHandler(hh)
+        for h_h in handlers:
+            logger.addHandler(h_h)
```
christinaholtNOAA, line 110: This choice isn't particularly more useful than the original. A lot of times I prefer the use of a singular/plural relationship between the items and the container of items. I'd suggest:

```
for handler in handlers:
    ...
```

.pylintrc
```diff
@@ -337,7 +337,7 @@ max-module-lines=1000
 # separator` is used to allow tabulation in dicts, etc.: {1  : 1,\n222: 2}.
 # `trailing-comma` allows a space between comma and closing bracket: (a, ).
 # `empty-line` allows space-only lines.
-no-space-check=trailing-comma,
+#no-space-check=trailing-comma,
```
christinaholtNOAA, line 340: I do think it's a good idea for us to continue checking white-space. Would you mind re-enabling this check and fixing the code?

Do others have opinions about that?

christinaholtNOAA (changes requested):
@andhogan Nice work. Thanks for taking on this Issue.

I left a couple of suggestions below for changes. I think it would also be good to change the description of the PR itself (in the GitHub interface).

.pylintrc
andhogan: image
The issue we're having with pylint is indicated in this image. Earlier documentation shows it's something vestigial and perhaps something that black maintains better than pylint. Ryan suggested this, but so far all I've been able to find are workarounds for this error as well.

christinaholtNOAA: Huh. Weird. Okay. In that case, maybe let's get rid of the whole section since it isn't usable. Looks like this option was removed in v 2.6. 😬

PR #88, opened 2022-06-28
Fix GitHub pages permissions

ryanlong1004:


christinaholtNOAA (commented):
@ryanlong1004 Do we need all these files for this PR? Would just the pytest-cov.yaml work to fix the github pages issue?

PR #104, opened 2022-08-15
Creating a Config class and its subclasses

christinaholtNOAA:
**Description**
Here, a set of configuration classes is added that allows several dictable file formats to be ingested with common methods and behavior.  Some JCSDA borrowed code has been removed as this functionality replaces what was adopted in those modules for this stage of development (treating configuration files as flat templates). The design of the code follows the [Config Design Doc](https://confluence-epic.woc.noaa.gov/pages/viewpage.action?pageId=17433951). 

The name of the namelist management tool was changed to reflect that it is meant to be a templater tool that will handle many types of configuration file templates, not just Fortran namelists. Several variables and methods within the tool were also changed similarly. I the new YAML config class is being used as an alternative to using the user's environment to supply the necessary values.

The test suite was updated to include unit tests for the concrete Config classes. The existing functional tests were modified to test the tool without the use of subprocesses. 

**Type of change**

**How Has This Been Tested?**

**Checklist**


src/uwtools/templater.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+    return parser.parse_args(argv)
+
+def set_template(argv):
+    '''Main section for set_namelist ingest utility'''
+
+    cla = parse_args(argv)
+
+    if cla.config_file:
+        cfg = config.YAMLConfig(cla.config_file)
```
benjamin-cash, line 69: Should there be a check to confirm the file really is a yaml file?

christinaholtNOAA: We are definitely checking whether the file exists. I am not sure how we'd check to see if it's a YAML other than try to parse it. If it can't be parsed, pyyaml will complain pretty quickly in the config constructor. 

Thoughts on what kind of checking we might do?

benjamin-cash: Letting the parser complain probably makes as much sense as anything. At least, nothing better is leaping to mind.

christinaholtNOAA: The stack trace usually tells you a line/column on where the YAML file isn't YAMLy enough, so I'm not sure I could do better than that. 😬

benjamin-cash: If it errors out with an informative message if it isn't YAMLy enough then I think that accomplishes what I was after with this comment. :)

src/uwtools/config.py
```diff
@@ -0,0 +1,198 @@
+'''
```
aerorahul, line 1: I can see that this is going to become unmanageable.  Even with the "dictable" types that have been added, it is already too big.

TerrenceMcGuinness-NOAA (dismissed):
Not bad.

src/uwtools/config.py
```diff
@@ -0,0 +1,198 @@
...
+This file contains the Config file and its subclasses for a variety of
+dicatable file types.
+'''
+
+import abc
+import collections
+import configparser
+
+import f90nml
```
aerorahul, line 10: So, if other "dictable" filetypes are added, all those will be imported here and not in the class that defines it?

src/uwtools/config.py
```diff
@@ -0,0 +1,198 @@
...
+        class for now.'''
+
+    def replace_templates(self):
+        ''' This method will be used as a method by which any Config
+        object can cycle through its key/value pairs recursively,
+        replacing Jinja2 templates as necessary. This is a placeholder
+        until more details are fleshed out in work scheduled for PI6.'''
+
+
+class YAMLConfig(Config):
```
aerorahul, line 81: This and other concrete classes should be in their own .py files.
The class(es) should import its dependencies.

src/uwtools/templater.py
```diff
@@ -0,0 +1,109 @@
...
+    if cla.dry_run:
+        print(nml)
+    # apply switch to print out required template values
+    # write out f90 name list
+    if not cla.dry_run and not cla.values_needed:
+        with open(cla.outfile, 'w+', encoding='utf-8') as nml_file:
+            f90nml.write(nml, nml_file)
+
+if __name__ == '__main__':
+    set_template(sys.argv[1:])
```
aerorahul, line 109: not sure I understand this.
Is it a "generic" template substitution or specific to Fortran namelist?

aerorahul (commented):
Since comments were requested, here are a few:

- This PR is too big to review and accomplishes more than a few things.
- code is added, code is removed and code is modified.
- There is configuration class added (see inline comments) but there is also a revamp of template substitution; in same PR.

src/uwtools/config.py
christinaholtNOAA: I'm not sure it would necessarily become unmanageable if we limit it to only parse standard file types. All the ones that fall outside that will use templates as is the current way.

We already need a way to parse YAML and bash configure files at a basic workflow level. This really just adds f90nml to the mix beyond that, and that is something that I feel will allow the Apps to couple more directly with the model regression tested files. The goal moving forward would hopefully be to persuade the developers of components to adopt _standard_ config languages to configure the UFS components so that we don't have an untenable number of languages to support in the workflow.

The proliferation of config types does need to be managed, for sure. We have the power in the UFS community not to let it get any worse at a component level!

src/uwtools/config.py
christinaholtNOAA: It seems like this one might be argued either way. This [Stack Overflow](https://softwareengineering.stackexchange.com/questions/209982/is-it-considered-pythonic-to-have-multiple-classes-defined-in-the-same-file) thread provides some nice reasons why you might keep all these types of classes together. One response references [The Zen of Python](https://peps.python.org/pep-0020/#the-zen-of-python) in support of keeping them in a single file:

>  Simple is better than complex.
> Flat is better than nested.

To which I'd add:

> Readability counts.

Since you can easily find all the concrete methods of subclasses in a single place.

[The Hitchhikers Guide to Python ](https://docs.python-guide.org/writing/structure/#structure-of-code-is-key) also suggests:

> a layer of a project can handle interfacing with user actions, while another would handle low-level manipulation of data. The most natural way to separate these two layers is to regroup all interfacing functionality in one file, and all low-level operations in another file. In this case, the interface file needs to import the low-level file. This is done with the import and from ... import statements.

In this code, all of these classes are interfaces to 3rd party, imported libraries.

christinaholtNOAA (commented):
@aerorahul Thanks for your thoughtful feedback. 

I left some responses below, and also want to apologize for the size of the contribution. I should have broken it into 3 PRs: remove unnecessary code, introduce Config classes, and hook Config classes into the user interface. I skipped steps on my end to save time and will make a better effort to not do that to our reviewers in the future.

src/uwtools/config.py
christinaholtNOAA: Let's hope we don't need that!

aerorahul:
> @aerorahul Thanks for your thoughtful feedback.
> 
> I left some responses below, and also want to apologize for the size of the contribution. I should have broken it into 3 PRs: remove unnecessary code, introduce Config classes, and hook Config classes into the user interface. I skipped steps on my end to save time and will make a better effort to not do that to our reviewers in the future.

I would also add; `Separation of Concerns` to make the code readable.
I am also familiar with both `Zen of Python` and `Hitchhikers guide to Python`. 

Anyways, these are comments FWIW.

christinaholtNOAA:
@aerorahul I think that separation of concerns is a very valid concern. 

After digging around for some guidance from the internet overlords, I feel like I'm finding that the rules are generally fairly arbitrary and are left to discretion of the team in large part.  [This article](https://levelup.gitconnected.com/understanding-separation-of-concerns-47f987f040d3) provides one nice example of what I mean. 

There, they offer some (albeit arbitrary) guidelines on when file sizes become too large (around 1000 lines), the overhead of separating code chunks with similar concerns into separate files, and how to balance the time it takes to search for code within a file and to search for code within a file system. Essentially translating to how many windows does it take to open the files you need to get the full scope of the problem at hand.

I personally feel that with this code, I have captured most, if not all the use cases we will need in the ~200 lines in one file, and the classes are group by their concern: interfaces to dictable configuration files. If we need to modify how we interact with these filetypes, this is the go-to file.

Code will inevitably be refactored as we learn more and continue to grow the code base. For now if you aren't worried about the implementation of the classes violating the separation of concerns in regard to their duties or methods being overloaded, I'd prefer to leave it organized in this way so that we may find all the configuration interfaces in one spot.

src/uwtools/templater.py
benjamin-cash: My understanding is the specific ticket this addresses is for the namelist, but it is anticipated that other tickets will be created to address other file types, and this will be extended to cover them in that work. @christinaholtNOAA do I have that correct?

christinaholtNOAA: That is correct. I totally missed responding to this one! 

We do have work to do here to make this a generic templater where it is not currently. We definitely did not want to grow this work any larger than it already was.

src/uwtools/config.py
laurenfrederick: I am inclined to agree with @aerorahul, but I think it's a personal preference thing. It's one of the things I noticed when first looking at it, but I think overall it's a style thing.

From my thinking, it would make it more clear looking at the file hierarchy what is supported rather than have to look in the code file. And also then there would be an individual test file for each. But again, not something I'd block the review on. It's more about consistency and what the team prefers in terms of readability and maintainability.

src/uwtools/templater.py
```diff
@@ -0,0 +1,109 @@
...
+import os
+import sys
+import argparse
+
+from jinja2 import Environment, FileSystemLoader, meta
+import f90nml
+
+from uwtools import config
+
+def file_exists(arg):
```
laurenfrederick, line 17: Nitpicky comment: `file_exists()` to me implies that a boolean would be returned and the function would just check if the file exists. Maybe `path_if_file_exists()` or something along those lines would be more descriptive.

src/uwtools/templater.py
```diff
@@ -0,0 +1,109 @@
...
+        '--values_needed',
+        action='store_true',
+        help='If provided, print a list of required configuration settings to stdout',
+        )
+    return parser.parse_args(argv)
+
+def set_template(argv):
+    '''Main section for set_namelist ingest utility'''
+
+    cla = parse_args(argv)
```
laurenfrederick, line 66: It's not clear to me what `cla` stands for here.

src/uwtools/templater.py
```diff
@@ -0,0 +1,109 @@
...
+
+    parser = f90nml.Parser()
+    # write out fully qualified and populated f90 namelist file
+    nml = parser.reads(rendered_template)
+
+    if cla.dry_run:
+        if cla.outfile:
+            print(f'warning file {cla.outfile} not written when using --dry_run or --values_needed')
+    # apply switch to allow user to view the results of namelist instead of writing to disk
+    if cla.dry_run:
```
laurenfrederick, line 100: I think lines 96-106 could be refactored as:

```
if cla.dry_run:
        if cla.outfile:
            print(f'warning file {cla.outfile} not written when using --dry_run or --values_needed')
        # apply switch to allow user to view the results of namelist instead of writing to disk
        print(nml)
    else: 
        # apply switch to print out required template values
        # write out f90 name list
        with open(cla.outfile, 'w+', encoding='utf-8') as nml_file:
            f90nml.write(nml, nml_file)
```

Which combines the two if the same `if` check, but I think `and not cla.values_needed` is redundant because of the `return` on line 87.

tests/test_templater.py
```diff
@@ -0,0 +1,89 @@
+#pylint: disable=unused-variable
```
laurenfrederick, line 1: From a test robustness perspective, I might make sure there are tests for:

The `file_exists` function itself - properly throws error, returns the path and the `parse_args()` function itself and that the correct environment variables (file or env) are used in `set_template()`

laurenfrederick (commented):
I have a few nitpicky comments in there that I will not be offended if you ignore!

src/uwtools/templater.py
christinaholtNOAA: I've used it as another acronym for command line args. I am considering user_args instead. What do you think?

src/uwtools/config.py
benjamin-cash: Since @laurenfrederick sees this as down to personal preferences and not a blocker my recommendation is that we leave it as is for now and move on. We will undoubtedly come back to this code in the future and can reassess then if it would be better to split it up.

benjamin-cash (approved):
Since @laurenfrederick sees the main points of difference as primarily down to style differences and minor issues I think we should press ahead. We can circle back to the questions raised here the next time we touch this code and see if we want to change the structure then.

benjamin-cash (approved):

src/uwtools/templater.py
laurenfrederick: Yep I think that works!

TerrenceMcGuinness-NOAA (changes requested):

src/uwtools/config.py
```diff
@@ -0,0 +1,198 @@
...
+
+import abc
+import collections
+import configparser
+
+import f90nml
+import yaml
+
+class Config(collections.UserDict):
+
```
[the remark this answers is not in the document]
christinaholtNOAA: This is an abstract base class as collections objects are derived from ABCs ([source](https://docs.python.org/3/library/abc.html)) 

> The [collections](https://docs.python.org/3/library/collections.html#module-collections) module has some concrete classes that derive from ABCs; these can, of course, be further derived. In addition, the [collections.abc](https://docs.python.org/3/library/collections.abc.html#module-collections.abc) submodule has some ABCs that can be used to test whether a class or instance provides a particular interface, for example, if it is hashable or if it is a mapping.


Just to double check the documentation wasn't sending me down a questionable path, I also followed the source code for collections through the definition of collections.UserDict [here](https://github.com/python/cpython/blob/230b630a767a457c6e7e4d797548a172c64ae735/Lib/collections/__init__.py) and it's base class defined in collections' Lib [here](https://github.com/python/cpython/blob/230b630a767a457c6e7e4d797548a172c64ae735/Lib/_collections_abc.py#L1).

Given that we are inheriting a wrapped ABC, I don't think we'll have problems.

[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: Yes you are close,  it is true  UserDict ⊂ collections and ABC ⊂ collections but by no means is collections.ABC ⊂ collections.UserDict and you don't get all the "base classes defined in collections" when inheriting collections.UserDict.  Having said that this is what should be done: Config(ABC)  -> childClass(Config, UserDict) which I had tested before I had this better understanding [[ref](https://stackoverflow.com/questions/66173332/python-abc-multiple-inheritance)]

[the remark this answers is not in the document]
benjamin-cash: This seems like a question with a concrete answer - is there a test that would/does confirm if this is behaving as expected?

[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: Oh and incidentally Config had been inheriting UserDict only because we where overloading the **__getattr__** method in order to support the dot as a value resolution operator and it appears you do not have that in this PR.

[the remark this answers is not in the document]
christinaholtNOAA: @TerrenceMcGuinness-NOAA the links I included don't imply the relationships you suggest here at all. They define that the `collections.UserDict` has ABCMeta as one of its base classes, therefore `Config` inherits ABCMeta as a base class.

@benjamin-cash There is indeed. Here I show the example as in the one in the StackOverflow Terry linked above:
```
from abc import ABC, abstractmethod
class BaseABC(ABC):
    @abstractmethod
    def hello(self):
        pass
class Child(BaseABC):
    pass

child = Child()
```

Just like it says in SO, this code will fail if the `Child` class does not implement the `hello` method. I tested it out by removing ABC as the base class and I found that the code will not complain at all about `Child` not implementing `hello`. We can test that with a version of YAMLConfig in a similar way. In my local code, I have commented out the YAMLConfig method `_load`, rebuilt with pip, and tried to instantiate a YAMLConfig object. Here is the code and the stack trace:

```
>>> import uwtools.config as config
>>> cfg = config.YAMLConfig()
TypeError: Can't instantiate abstract class YAMLConfig with abstract methods _load
```

Since `collections.UserDict` is derived from a `ABCMeta` as you can clearly follow in the source code, using the `@abstractmethod` decorator works just as expected.

To addres your follow up, Terry, `collections.UserDict` is not used "only because we where overloading the **getattr** method". That is a convenient feature that we _can_ use and I have chosen not to implement here because as Rahul pointed out, it was already a big push. It can always be added with a smaller PR in the future. The reason that we are using `collections.UserDict` as a base class is because we'd like to have some additional methods to manipulate these config dictionaries and this is the Pythonic way to achieve that goal -- extend the functionality of the built-in dictionary.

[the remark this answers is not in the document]
benjamin-cash: Ok, given that there is a concrete test showing that the code is behaving as expected, and this code was already approved once, I feel that this PR should be approved and any additional concerns can be addressed in future tickets and PRs.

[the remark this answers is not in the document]
benjamin-cash: @TerrenceMcGuinness-NOAA Are you satisfied with the test @christinaholtNOAA provided here?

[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: Yes will like to see dot resolution at all costs myself because from users perspective it had often been called out as a "favorite".  Still the "proper" way to establish a Abstract Base Class is to inherit the ABCMeta class directly and not to simply subsume it from a "derived" class.    -- end of suggestion --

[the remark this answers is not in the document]
TerrenceMcGuinness-NOAA: Yes Benjamin, I like you "common sense" approach.  Yes we all now have a good understanding of the issues and intent, as well as why I still have a relevant prototype branch, so let's move on shall we :)

TerrenceMcGuinness-NOAA (approved):

PR #113, opened 2022-09-27
Feature/yaml inc uw 149 151

TerrenceMcGuinness-NOAA:
**Description**

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -2,14 +2,16 @@
 This file contains the Config file and its subclasses for a variety of
 dicatable file types.
 '''
-
+import os
 import abc
+from typing import Any, IO
 import collections
 import configparser
 
 import f90nml
 import yaml
 
+#pylint: disable=unused-variable
 class Config(collections.UserDict):
 
     '''
@@ -95,7 +97,7 @@ class YAMLConfig(Config):
         object. '''
 
         with open(self.config_path, 'r', encoding="utf-8") as file_name:
-            cfg = yaml.load(file_name, Loader=yaml.SafeLoader)
+            cfg = yaml.load(file_name, self.Loader)
         return cfg
 
     def dump_file(self, output_path):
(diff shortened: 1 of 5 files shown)


src/uwtools/set_namelist.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
```
christinaholtNOAA, line 1: This file should not be included as a part of this PR as it is no longer a part of the repository and has been moved to `templater.py`.

src/uwtools/config.py
```diff
@@ -104,23 +106,47 @@ def dump_file(self, output_path):
...
+    class Loader(yaml.SafeLoader): #pylint: disable=too-many-ancestors
+        '''Yaml loader to add INCLUDE tag for yaml file concatenation'''
+
+        def __init__(self, stream: IO) -> None:
+            try:
+                self._root = os.path.split(stream.name)[0]
+            except AttributeError:
+                self._root = os.path.curdir
+
+            super().__init__(stream)
```
christinaholtNOAA, line 118: It is unclear to me the usefulness of this local class here. What benefit do we gain from making this class? It seems that you have added the constructor from a method of the YAMLConfig class. Why not remove the Loader class and use SafeLoader directly? For example:

```
yaml.add_constructor("!INCLUDE", add_include, Loader=yaml.SafeLoader)
```

src/uwtools/config.py
```diff
@@ -104,23 +106,47 @@ def dump_file(self, output_path):
...
+
+    def add_include(loader: Loader, node: yaml.Node) -> Any: #pylint: disable=no-self-argument
```
christinaholtNOAA, line 120: The pythonic way of achieving this result is to decorate it with `@staticmethod` instead of having the linter ignore it.

src/uwtools/config.py
```diff
@@ -104,23 +106,47 @@ def dump_file(self, output_path):
...
+        '''method to implement INCLUDE tag for yaml file concatenation'''
+        #pylint: disable=no-member
```
christinaholtNOAA, line 122: What is the linter complaining about here? There are only rare occasions where this disable is needed as this check helps us to make sure we aren't doing something we shouldn't be. 

If it is truly needed, we should document the need for it in a comment so that we don't have to guess at what's going on.

src/uwtools/config.py
```diff
@@ -104,23 +106,47 @@ def dump_file(self, output_path):
...
         ''' Load the user-provided Fortran namelist path into a dict
         object. '''
-        with open(self.config_path, 'r', encoding="utf-8") as file_name:
-            cfg = f90nml.read(file_name)
+        if self.string is not None:
+            cfg = f90nml.reads(self.string)
+        else:
+            with open(self.config_path, 'r', encoding="utf-8") as file_name:
+                cfg = f90nml.read(file_name)
+        self.config_obj = cfg
```
christinaholtNOAA, line 149: The changes to this class are outside the scope of the work, and it is unclear that it's needed. Since YAGNI applies here, please remove the modification.

christinaholtNOAA (changes requested):
Please make sure to include proper descriptions in the PR template when submitting a PR. 

Please address the change requests below.

PR #116, opened 2022-10-05
UW-191 Add Jinja Template Class and Tests

nbharwani11:
**Description**

Added test cases for reusable templating class, Jinja Template Class that can be used to enable a variety of templating needs

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/j2template.py
+++ b/src/uwtools/j2template.py
@@ -1,0 +1,123 @@
+'''
+Template classes
+
+'''
+
+from jinja2 import Environment, BaseLoader, FileSystemLoader, meta
+
+class J2Template():
+
+    '''
+    This class reads in Jinja templates from files or strings, and
+    renders the template given the user-provided configuration object.
+
+    Attributes
+    ----------
+
+    configure_obj : dict
+        The key/value pairs needed to fill in the provided template.
+        Defaults to the user's shell environment.
+
+    template
+        Jinja2 template object
+
+    undeclared_variables : list
+        List of variables needed given a Jinja2 template
+
+    Methods
+    -------
+
+    _load_file()
+        Wrapper around loading a Jinja2 template from a file.
+
+    _load_string()
+        Wrapper around loading a Jinja2 template from a string
+
+    dump_file(output_path)
+        Wrapper around writing the template to an output path
+
+    validate_config()
+        Checks to ensure that the provided configure_obj is sufficient
+        to meet the needs of the undeclared_variables in the template
+
+    '''
+
+    def __init__(self, configure_obj, template_path=None, template_str=None):
+
+        '''
+        Parameters
+        ----------
+        configure_obj : dict (See above)
+        template_path : Path
+            Path to a Jinja2 template file
+        template_str : str
+            A Jinja2 template string
+
+        '''
+        self.configure_obj = configure_obj
+        self.template_path = template_path
+        self.template_str = template_str
+        if template_path is not None:
+            self.template = self._load_file(template_path)
+        elif template_str is not None:
+            self.template = self._load_string(template_str)
+        else:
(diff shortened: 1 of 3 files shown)


src/uwtools/j2template.py
```diff
@@ -0,0 +1,123 @@
...
+    @property
+    def undeclared_variables(self):
+
+        ''' Generates a list of variables needed for self.template '''
+        if self.template_str is not None:
+            j2_parsed = self._j2env.parse(self.template_str)
+        else:
+            with open(self.template_path,encoding='utf-8') as __file:
+                j2_parsed = self._j2env.parse(__file.read())
+        return meta.find_undeclared_variables(j2_parsed)
```
christinaholtNOAA, line 123: As we'd discussed earlier, the test for this method fails like this:
```
self = <uwtools.j2template.J2Template object at 0x7f03b872b5d0>

    @property
    def undeclared_variables(self):
    
        ''' Generates a list of variables needed for self.template '''
        if self.template_str is not None:
>           j2_parsed = self._j2env.parse(self.template_str)
E           AttributeError: 'J2Template' object has no attribute '_j2env'

.tox/py/lib/python3.7/site-packages/uwtools/j2template.py:119: AttributeError
```

That's related to the object attribute `_j2env` not being set appropriately in the case of a string template. That means this is a really nice test since it caught a limitation of the code. I'm thinking that you could modify the `_load_string` method to set the attribute and then return the template object to mirror what is done for `_load_file`. Something like this untested code might do the trick:

```
def _load_string(self):

    self._j2env = Environment(loader=BaseLoader())
    return self._j2env.from_string(template_str)
  
```

Given that we are assuming the load functions set this attribute, it might be worth a similar test for undeclared variables in the existing file loading test you already have. That should help future developers make sure that they don't refactor `_load_file` and mess up `undeclared_variables`.

tests/test_template.py
```diff
@@ -151,3 +154,63 @@ def test_substitute_with_dependencies():
...
+Unit tests for testing J2Template Class
+"""
+
+
+def test_dump_file():
+    """
+         Write rendered template to the output_path provided
+         Parameters
+         ----------
+         output_path : Path
```
christinaholtNOAA, line 171: I don't think we need parameter lists in the doc blocks for the tests since they don't take any arguments. It may also be good to ensure that the doc block description is accurate based on the latest set of tests you have here. For example, this one tests that the template is rendered in memory from an input string template, and checks that the undeclared_variables returns the expected list.

I don't want to get ahead of myself too far, but this test could certainly be extended to test the dump method using a temporary file.

christinaholtNOAA (commented):
@elcarpenter @nbharwani11 I left a couple comments below. The code looks nice, especially for your first contribution. :)

christinaholtNOAA (dismissed):
The code looks good to me! 

I have a couple of formatting or cleanup requests below if you don't mind addressing those. I'm good otherwise.

tests/test_template.py
```diff
@@ -151,3 +154,58 @@ def test_substitute_with_dependencies():
...
+         Load the Jinja2 template from the string provided.
+         Returns
+         -------
+         Jinja2 Template object
+    """
+    template_str = 'Hello to the world'
+    test_config  = {
+    	'greeting': 'Hello',
+    	'the_world': 'the world'
+    }
```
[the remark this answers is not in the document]
nbharwani11: @christinaholtNOAA All of the formatting requests should have been updated. Let me know if there were any others.

christinaholtNOAA (dismissed):
This looks great. Thank you!

fgabelmannjr (approved):
This looks great.  Wonderful job on this PR!

christinaholtNOAA (approved):

PR #118, opened 2022-10-13
UW-192: Replace Direct Calls to Jinja2 with Jinja Template Class

elcarpenter:
**Description**

Updated the existing tool by replacing direct calls to Jinja2 with calls to the Jinja Template Class.

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/templater.py
+++ b/src/uwtools/templater.py
@@ -9,8 +9,7 @@ import os
 import sys
 import argparse
 
-from jinja2 import Environment, FileSystemLoader, meta, StrictUndefined
-
+from uwtools.j2template import J2Template
 from uwtools import config
 
 def dict_from_config_args(args):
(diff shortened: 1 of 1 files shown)


src/uwtools/templater.py
```diff
@@ -86,25 +85,19 @@ def set_template(argv):
...
-        loader=FileSystemLoader(user_args.input_template),
-        undefined=StrictUndefined,
-    )
-    template = env.get_template('')
+    template = J2Template(cfg, user_args.input_template)
 
     # Gather the undefined template variables
-    j2_parsed = env.parse(env.loader.get_source(env, ''))
-    undeclared_variables = meta.find_undeclared_variables(j2_parsed)
+    undeclared_variables = template.undeclared_variables
```
christinaholtNOAA, line 91: Does it make sense to move this line inside the `if user_args.values_needed` block? Maybe we don't have to spend the computational effort unless we need it.

The same principle holds for `rendered_template`.

christinaholtNOAA (approved):
The code looks nice, clean, and concise. 

My only comments are suggestions related to efficiency. In general, we may see speed up for the overall script if we only call methods when they're needed. It likely won't matter in our contrived easy tests here, but with big templates with many many variables it could make a difference.

christinaholtNOAA (approved):
This looks great!

PR #119, opened 2022-10-28
UW-200: Add method to Config class to print the configure contents.

nbharwani11:
**Description**
Adding methods to the Config base class that enable printing the configure contents. Included test cases for these new methods.

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -56,7 +56,8 @@ class Config(collections.UserDict):
         self.config_path = config_path
 
     def __repr__(self):
-        return f'Config("{self.config_path}")'
+        ''' This method will return configure contents'''
+        return json.dumps(self.data)
 
     @abc.abstractmethod
     def _load(self):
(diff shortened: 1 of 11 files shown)


src/uwtools/config.py
```diff
@@ -55,6 +56,9 @@ def __init__(self, config_path):
 
         self.config_path = config_path
 
+    def __repr__(self):
+        return f'Config("{self.config_path}")'
```
christinaholtNOAA, line 60: This line is a bit curious. I was definitely thinking this is where we'd see something like 

``` json.dumps(self.data))```

File: src/uwtools/config.py [hunk 2]
```diff
@@ -79,6 +83,9 @@ def replace_templates(self):
         replacing Jinja2 templates as necessary. This is a placeholder
         until more details are fleshed out in work scheduled for PI6.'''
 
+    def config_dump(self):
+        ''' This method will print configure contents'''
+        print(json.dumps(self.data))
```
christinaholtNOAA, line 88: I'm not entirely sure we need this method, but am not completely against it. The reason I'm thinking that is because I was hoping that if we printed a Config object, we'd see its data contents and maybe not have to call a method.

A test case that I was thinking for that scenario might be something like:

```
assert print(cfg) == print(json.dumps(expected))
```

christinaholtNOAA (commented):
I have just a couple of comments below. I'm sorry for the delay in getting these to you.

christinaholtNOAA (approved):
That looks super clean! Thank you!!!

Issue #121, opened 2022-11-29
UW-207


WeirAE:
https://jira-epic.woc.noaa.gov/browse/UW-207

WeirAE:
Reopening due to the commit reversion noted above. Problem has been replicated and identified, hopefully quick to correct.

PR #125, opened 2022-12-09
UW-207 update: As a user of the templater.py tool, I need to see useful logging messages (#121)

WeirAE:
**Description**
Added --quiet and --verbose arguments calling logger.py. Fixed string formatting issue for logging. Fixes issue https://github.com/ufs-community/workflow-tools/issues/121

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/logger.py
+++ b/src/uwtools/logger.py
@@ -73,8 +73,6 @@ class Logger:
         # Add console handler for logger
         _handler = Logger.add_stream_handler(
             level=self.level,
-            #_format=self.format,
-            #colored_log=self.colored_log,
             _format='%(message)s',
             colored_log= False,
             )
--- a/src/uwtools/logger.py
+++ b/src/uwtools/logger.py
@@ -73,8 +73,8 @@ class Logger:
         # Add console handler for logger
         _handler = Logger.add_stream_handler(
             level=self.level,
-            _format='%(message)s',
-            colored_log= False,
+            _format=self.format,
+            colored_log=self.colored_log,
             )
         _handlers.append(_handler)
         self._logger.addHandler(_handler)
--- a/src/uwtools/logger.py
+++ b/src/uwtools/logger.py
@@ -73,8 +73,10 @@ class Logger:
         # Add console handler for logger
         _handler = Logger.add_stream_handler(
             level=self.level,
-            _format=self.format,
-            colored_log=self.colored_log,
+            #_format=self.format,
+            #colored_log=self.colored_log,
+            _format='%(message)s',
+            colored_log= False,
             )
         _handlers.append(_handler)
         self._logger.addHandler(_handler)
--- a/src/uwtools/templater.py
+++ b/src/uwtools/templater.py
@@ -86,19 +86,33 @@ def set_template(argv):
     user_args = parse_args(argv)
 
     logfile = os.path.join(os.path.dirname(__file__), "templater.log")
-    log = Logger(level='info', logfile_path=logfile)
+    log = Logger(level='info',
+        _format='%(message)s',
+        colored_log= False,
+        logfile_path=logfile
+        )
     if user_args.verbose:
-        log = Logger(level='debug', logfile_path=logfile, colored_log=True)
-        log.debug('Finished setting up debug file logging in %(logfile)s')
+        log = Logger(level='debug',
+            _format='%(asctime)s - %(levelname)-8s - %(name)-12s: %(message)s',
+            colored_log= True,
+            logfile_path=logfile
+            )
+        log.debug(r"Finished setting up debug file logging in {logfile}".format(logfile=logfile))
     elif user_args.quiet:
+        log.handlers.clear()
         log.propagate = False
 
 
-    log.info('Running script templater.py with args:n '+('-' * 70)+'\n'+('-' * 70))
+
+    log.info(f"""Running script templater.py with args:
+{('-' * 70)}
+{('-' * 70)}""")
     for name, val in user_args.__dict__.items():
         if name not in ["config"]:
             log.info("{name:>15s}: {val}".format(name=name, val=val))
-    log.info(('-' * 70)+'\n'+('-' * 70))
+    log.info(f"""{('-' * 70)}
+{('-' * 70)}""")
+
 
     if user_args.config_file:
         cfg = config.YAMLConfig(user_args.config_file)
@@ -122,8 +136,7 @@ def set_template(argv):
 
     if user_args.dry_run:
         if user_args.outfile:
-            log.info('warning file %(user_args.outfile)s not written when using --dry_run')
-
+            log.info(r"warning file {outfile} not written when using --dry_run".format(outfile=user_args.outfile))
         # apply switch to allow user to view the results of rendered template
         # instead of writing to disk
         # Render the template with the specified config object
--- a/src/uwtools/templater.py
+++ b/src/uwtools/templater.py
@@ -11,6 +11,7 @@ import argparse
 
 from uwtools.j2template import J2Template
 from uwtools import config
+from uwtools import logger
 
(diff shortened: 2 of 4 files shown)


src/uwtools/templater.py
```diff
@@ -83,22 +83,22 @@ def parse_args(argv):
 
 def set_template(argv):
     '''Main section for rendering and writing a template file'''
-    logfile = os.path.join(os.path.dirname(__file__), "templater.log")
-    log = Logger(level='info', logfile_path=logfile)
-
-
     user_args = parse_args(argv)
 
+    logfile = os.path.join(os.path.dirname(__file__), ".logtemplater")
```
christinaholtNOAA, line 88: I think `templater.log` was a bit more standard. Any specific reason for the change?

src/uwtools/templater.py
```diff
@@ -83,22 +83,22 @@ def parse_args(argv):
...
         log.propagate = False
 
-    log.info("Running script templater.py with args:n", f"{('-' * 70)}\n{('-' * 70)}")
+
+    log.info('Running script templater.py with args:n '+('-' * 70)+'\n'+('-' * 70))
     for name, val in user_args.__dict__.items():
         if name not in ["config"]:
-            log.info(f"{name:>15s}: {val}")
-    log.info(f"{('-' * 70)}\n{('-' * 70)}")
+            log.info(format(name).rjust(15) + ': ' + str.format(val))
```
christinaholtNOAA, line 100: So I wasn't quite sure if you were having problems with f-strings (apologies if I missed that), but are you able to use formatting like this?
```
"{name:>15s}: {val}".format(name=name, val=val)
```

I'd expect that should render a string first, then pass it to the formatter as string without additional need to process.

WeirAE: The issue was that logging predates fstrings and there's some internal dev concerns about it being a security vulnerability so its never been updated. Hmm, maybe, I was trying to avoid repeating variables while maintaining the fixed format, but I'll see if it'll let me get away with it.

src/uwtools/templater.py
WeirAE: To be honest, I did see a good reason to change it, and now I can't for the life of me remember where or why,  so it can't be that good of a reason! I'll revert this

src/uwtools/templater.py
WeirAE: OK  confirmed - if this is the preferred syntax, I'll swap it over!

WeirAE:
Thats so weird - the 'conflicts' are the updates, I'm not quite sure how to resolve this per se

christinaholtNOAA:
I think it's just that you deleted/added them to the same lines. Perhaps just remove the `<<<<<<` and `>>>>>` directives and make the lines what they should be.

WeirAE:
Thanks - that's apparently what it wanted, although it made a bit of a mess of my code as a result. Working on resolving the resulting issues

WeirAE:
Looks like pylint is just mad about the .format() instead of f-string, even though f-strings are disabled for logging. Is there a way to make this acceptable to pylint? Otherwie we'll get this error every PR

src/uwtools/logger.py
```diff
@@ -73,8 +73,8 @@ def __init__(self, name: str = None,
         # Add console handler for logger
         _handler = Logger.add_stream_handler(
             level=self.level,
-            _format=self.format,
-            colored_log=self.colored_log,
+            _format='%(message)s',
+            colored_log= False,
```
christinaholtNOAA, line 77: I doubt you want to make this change for all Logging functionality, after taking a second look at this. I think having the original defaults is nice. You should be able to overwrite the default values for format and colored_log when you create the Logger object.

src/uwtools/templater.py
```diff
@@ -68,17 +69,36 @@ def parse_args(argv):
...
+    logfile = os.path.join(os.path.dirname(__file__), "templater.log")
+    log = Logger(level='info', logfile_path=logfile)
+    if user_args.verbose:
+        log = Logger(level='debug', logfile_path=logfile, colored_log=True)
+        log.debug('Finished setting up debug file logging in %(logfile)s')
+    elif user_args.quiet:
+        log.propagate = False
+
+
+    log.info('Running script templater.py with args:n '+('-' * 70)+'\n'+('-' * 70))
```
christinaholtNOAA, line 97: There is a typo in this line. there should be a newline after `args:`.

src/uwtools/templater.py
```diff
@@ -102,7 +122,8 @@ def set_template(argv):
 
     if user_args.dry_run:
         if user_args.outfile:
-            print(f'warning file {user_args.outfile} not written when using --dry_run')
+            log.info('warning file %(user_args.outfile)s not written when using --dry_run')
```
christinaholtNOAA, line 125: Is there a typo here with the s after `%(user_args.outfile)`?

christinaholtNOAA (changes requested):
We do want to ensure the linter is happy. Let's make sure that all the strings are formatted consistently in calls to the logger. I think it's also important that they're readable. I'd prefer to stay with f-strings where we can, but if we do need the printf style strings, that's fine.

Let's remove as much of the appending bits of strings together, though. Adding strings that need a whole variety of pairs of quotes gets pretty cumbersome.

Also, I am not sure the "quiet" flag works. If I use it, I still get output to the screen and to the log file, and this flag is meant to suppress all output.

src/uwtools/templater.py
WeirAE: Yep, I marked this as #127 as it's in the current version, but I can include it in this PR

src/uwtools/logger.py
WeirAE: OK, I can take care of #128 while I'm at it with that change

src/uwtools/templater.py
WeirAE: ah, a remnant of an earlier %s formatting version, I'll replace it to be consistent with the other changes

src/uwtools/templater.py
```diff
@@ -68,17 +69,36 @@ def parse_args(argv):
...
+    elif user_args.quiet:
+        log.propagate = False
+
+
+    log.info('Running script templater.py with args:n '+('-' * 70)+'\n'+('-' * 70))
     for name, val in user_args.__dict__.items():
         if name not in ["config"]:
-            print(f"{name:>15s}: {val}")
-    print(f"{('-' * 70)}\n{('-' * 70)}")
+            log.info("{name:>15s}: {val}".format(name=name, val=val))
```
WeirAE, line 100: This right here is unfortunately the sticking point for f-strings. In order to set the alignment for name, you have to call the variable, but the logging() can't handle it, and separating the padding to just operate on "" doesn't align it right. Maybe tighten it up like this?
`log.info("{name:>15s}: ".format(name=name), val)`

WeirAE:
Multiple changes added: 
logger defaults reverted
log formatting more like fstrings where possible
stream handling now set correctly by formatter
formatter test of verbose and quiet added
Verbose test still has a parsing bug due to having to clean timestamps

WeirAE:
Fixed the templater verbose test with a much cleaner line-in-line logic!

christinaholtNOAA (dismissed):
LGTM! Thanks for your diligence on this one! :)

Edit: of course, make sure the tests pass before merging.

WeirAE:
Don't worry, not doing that again! Haha - still going to have the linter issue going forward, though

WeirAE:
@christinaholtNOAA I found some extensive discussion about the linter and logger not getting along here:
https://github.com/PyCQA/pylint/issues/5039
There may be a workaround to add `disable=logging-fstring-interpolation` to pylint.rc?

christinaholtNOAA (approved):
LGTM

christinaholtNOAA (approved):

Issue #129, opened 2022-12-12
UW-209


WeirAE:
**Description**
https://jira-epic.woc.noaa.gov/browse/UW-209

**Requirements**
Test that the assumptions that the polymorphism built into the Config subclasses for YAML, namelists and INI files is sufficient to transform to another subclass. May require making the input file path optional in the '__init__' methods of each.

**Acceptance Criteria (Definition of Done)**
One additional test in test_config.py must test all transform combinations and complete successfully

WeirAE:
Test comparisons are as follows:
INI vs YAML
False
YAML ref: ['salad:', '  base: kale', '  fruit: banana', '  vegetable: tomato', '  how_many: 12', '  dressing: balsamic']
YAML out: ['salad:', '  base: kale', '  fruit: banana', '  vegetable: tomato', "  how_many: '12'", '  dressing: balsamic']

INI vs F90
False
F90 ref: ['&salad', "    base = 'kale'", "    fruit = 'banana'", "    vegetable = 'tomato'", '    how_many = 12', "    dressing = 'balsamic'", '/']
F90 out: ['&salad', "    base = 'kale'", "    dressing = 'balsamic'", "    fruit = 'banana'", "    how_many = '12'", "    vegetable = 'tomato'", '/']

YAML vs INI
False
INI ref: ['[salad]', 'base = kale', 'fruit = banana', 'vegetable = tomato', 'how_many = 12', 'dressing = balsamic', '']
INI out: ["salad={'base': 'kale', 'fruit': 'banana', 'vegetable': 'tomato', 'how_many': 12, 'dressing': 'balsamic'}"]

YAML vs F90
False
F90 ref: ['&salad', "    base = 'kale'", "    fruit = 'banana'", "    vegetable = 'tomato'", '    how_many = 12', "    dressing = 'balsamic'", '/']
F90 out: ['&salad', "    base = 'kale'", "    dressing = 'balsamic'", "    fruit = 'banana'", '    how_many = 12', "    vegetable = 'tomato'", '/']

F90 vs INI
False
INI ref: ['[salad]', 'base = kale', 'fruit = banana', 'vegetable = tomato', 'how_many = 12', 'dressing = balsamic', '']
INI out: ["salad=OrderedDict([('base', 'kale'), ('fruit', 'banana'), ('vegetable', 'tomato'), ('how_many', 12), ('dressing', 'balsamic')])"]

F90 vs YAML
False
[...]

PR #131, opened 2022-12-13
UW-209: As the set_config.py tool, I need to make sure any Config subclass an be transformed to any other Config subclass

WeirAE:
**Description**

Added test 'test_transform_config' that iterates through config options and converts to other options. Currently set to use Differ rather than assert as all seem to be showing differences currently. Changed simple.yaml to simple2.yaml for consistency in iteratable testing. Updated config.py __init__ to not require an input file so subclasses can be called without issue. Fixes issue #129 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Orion

**Checklist**


tests/test_config.py
```diff
@@ -212,3 +214,30 @@ def test_ini_config_bash():
...
+        cfgin = getattr(config, f"{test1}Config")
+        cfg = cfgin(test)
+        cfgout = getattr(config, f"{test2}Config")
+
+        with tempfile.TemporaryDirectory(dir='.') as tmp_dir:
+            out_file = f'{tmp_dir}/test_{test1.lower()}to{test2.lower()}_dump.{test2file.lower()}'
+            cfgout.dump_file(cfg, out_file)
+
+            #assert filecmp.cmp(ref, out_file)
+            print(f"{test1} vs {test2}")
```
WeirAE, line 238: going to have to replace this with logging once #122 gets pulled

tests/test_config.py
```diff
@@ -212,3 +214,30 @@ def test_ini_config_bash():
...
+            with open(ref) as file_1, open(out_file) as file_2:
+                differ = Differ()
+
+                for line in differ.compare(file_1.readlines(), file_2.readlines()):
```
WeirAE, line 242: May also consider changing this to match the 'for line in lines' from #122 as well, although I feel that this should be evaluated more strictly than there

tests/test_config.py
```diff
@@ -212,3 +214,30 @@ def test_ini_config_bash():
     cfg.update({'dressing': ['ranch', 'italian']})
     expected['dressing'] = ['ranch', 'italian']
     assert cfg == expected
+
+def test_transform_config():
+    '''Test that transforms work as intended.
```
christinaholtNOAA, line 219: This might be clearer as "...transforms config objects to objects of other config subclasses". It might not be, but I think we should attempt to give the doc block more meaning here to tell folks what the point of the test is.

tests/test_config.py
```diff
@@ -212,3 +214,30 @@ def test_ini_config_bash():
...
+
+    '''
+    for test1, test2 in itertools.permutations(["INI", "YAML", "F90"],2):
+        test1file = "NML" if test1 == "F90" else test1
+        test2file = "NML" if test2 == "F90" else test2
```
christinaholtNOAA, line 224: We had several discussions about this becoming a Rube Goldberg machine. I'm glad to see that we have the tools necessary to make that logic fairly compact and straightforward.

I think it could be helpful to add a quick comment before these two lines to mention that we want to transform Config subclass type to an appropriate file subscript. It took me a little while to understand why we needed this transformation.

christinaholtNOAA (commented):
@WeirAE I think this looks pretty good! It does what we want it to do. I left just a couple of requests for changes below that are super minor. Mainly just clarity in comments.

WeirAE:
Updated comments and disabled pylint error; changes to any dictionary creations I can add in #132

tests/test_config.py
```diff
@@ -1,14 +1,16 @@
 '''
 Set of test for loading YAML files using the function call load_yaml
 '''
-#pylint: disable=unused-variable
+#pylint: disable=unused-variable, unspecified-encoding
```
christinaholtNOAA, line 4: Is there a reason we can't or shouldn't use `encoding="utf-8"` in the open statement? In general, it's good to fall in line with what the linter wants. There are of course exceptions.

WeirAE: Ah thats very fair, I'll change over to that

tests/test_config.py
WeirAE: Hmm, I'm inclined to leave it on 'print' for now, may revisit this later

tests/test_config.py
WeirAE: Updated with most recent commit

WeirAE:
@christinaholtNOAA Thanks so much for the support, I don't see any issues after your cleanup. Its much better as well!

Issue #132, opened 2022-12-13
UW-210


WeirAE:
**Description**

https://jira-epic.woc.noaa.gov/browse/UW-210

**Requirements**
Add a set of tests that compare two config objects of any combination of 2 file types (9 tests) X and X,Y,Z for X, Y and Z for f90nml, INI and YAML, with a user output that shows the differences. Additionally it should only be supported by files with the same structure (depth of sections, etc.)

**Describe alternatives you've considered**
May be partly completed by UW-209

**Acceptance Criteria (Definition of Done)**
All combinations must be tested with similar structure and correctly fail on different structure

**Additional context**
An additional option may be to allow user comparison of any two optional files as a tool

WeirAE:
Issues are related to #129 where I've implemented the test as a method under Config, but reading dict objects. However, it breaks the key/value evaluations either as here or with an items difference. I'm still working on a workaround but have not solved it yet.
```
    @abc.abstractmethod
    def compare_config(self, user_dict, base_dict=None):
        '''
        Assuming a section, key/value structure of configuration types,
        compare the dictionary to the values stored in the external file.
        '''

        if base_dict is None:
            base_dict = self.data

        if len(copy.deepcopy(user_dict).items()) \
            != len(copy.deepcopy(base_dict).items()):
            raise KeyError('Given dictionaries have different ranges of items!')
        # Then, compare keys and values
        _miskey = {key for key in base_dict.keys() \
            & user_dict if base_dict[key] != user_dict[key]}
        _misval = {value for value in base_dict.values() \
            & user_dict if base_dict[value] != user_dict[value]}
        if len(_miskey) == 0 & len(_misval) == 0:
            print('All config values match!')
        else:
            if len(_miskey) != 0: print('The following keys are different: ',
                 ', '.join(map(str,_miskey)))
            if len(_misval) != 0: print('The following values are different: ',
                 ', '.join(map(str,_misval)))
```

WeirAE:
Update: after fixing dict objects in #131 , the remaining errors are about instantiation since I'm calling this in the concrete methods with super(). 
```
    def compare_config(self, user_dict, base_dict=None):
        '''
        Assuming a section, key/value structure of configuration types,
        compare the dictionary to the values stored in the external file.
        '''
        base_dict = base_dict or self.data

        super().compare_config(user_dict, base_dict=None)
        ```

PR #142, opened 2023-01-10
UW-210 update: Compare two config objects -- base and user-provided

WeirAE:
**Description**

Added a config comparison method that iterates through pairs of config objects and compares structure depth and content, as well as an associated test. Fixes issue #132 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Orion HPC

**Checklist**


tests/test_config.py
```diff
@@ -244,3 +244,22 @@ def test_transform_config():
...
+
+        basepath = os.path.join(uwtools_file_base,pathlib.Path("fixtures",f"simple.{basefile.lower()}"))
+        userpath = os.path.join(uwtools_file_base,pathlib.Path("fixtures",f"simple.{userfile.lower()}"))
+
+        cfgbaserun = getattr(config, f"{base}Config")(basepath)
+        cfguserrun = getattr(config, f"{user}Config")(userpath)
+
+        cfguserrun['salad']['dressing'] = 'italian'
+
+        cfgbaserun.compare_config(cfguserrun, cfgbaserun)
```
christinaholtNOAA, line 265: There doesn't seem to be a test here...

src/uwtools/config.py
```diff
@@ -80,6 +80,35 @@ def dump_file(self, output_path):
...
+    def compare_config(self, user_dict, base_dict=None):
+        #pylint: disable=multiple-statements
+        '''
+        Assuming a section, key/value structure of configuration types,
+        compare the dictionary to the values stored in the external file.
+        '''
+        if base_dict is None:
+            base_dict = self.data
+
+        mishdr = {}
```
christinaholtNOAA, line 93: It's unclear what `mis` is here, it may also be easier to understand if you spell out `header` and `value`.

src/uwtools/config.py
```diff
@@ -80,6 +80,35 @@ def dump_file(self, output_path):
         ''' Interface to write a config object to a file at the
         output_path provided. '''
 
+    @abc.abstractmethod
```
christinaholtNOAA, line 83: I don't _think_ this needs to be an abstract method since it is fully implemented here. 

If you can remove that, you should also be able to remove the additions of the methods to the sub-classes. Those should only need implementing if they somehow differ from this implementation.

christinaholtNOAA (changes requested):
The comparison looks pretty solid. I left a few comments below about things that aren't part of the algorithm. 

It might also be nice to see get a test in there that shows us expected output.

src/uwtools/config.py
WeirAE: Habit of keeping short names, I'll spell it out and add a comment here, thanks

src/uwtools/config.py
WeirAE: Hmm, I'll take a look - I debated that, but was keeping the existing behavior since we don't really call the base method outside of here. Should be able to make it a class method which will be a good starting before we add the args, thanks!

tests/test_config.py
WeirAE: Hmm, maybe use the same structure as def test_ini_config_simple(), iterating all three config files. First against an 'expected' dict, assert against stdout, then change the dict and assert against the expected difference in stdout?

WeirAE:
@christinaholtNOAA Thanks for taking the time to check through this! I can add these changes quickly - do you have any different preferences for formatting the output, or is the simple list here ok?

src/uwtools/config.py
WeirAE: Variable names extended!

src/uwtools/config.py
WeirAE: Class method applied without issue

tests/test_config.py
WeirAE: Assertion was a little tricky due to dictionary order; I updated the ini read to OrderedDict for some more consistency, but it was still unsorting so I modified the assertion to just look for the results.

christinaholtNOAA:
@WeirAE I think the output as a list is fine for now. We have future work in [UW-230](https://jira-epic.woc.noaa.gov/browse/UW-230) that does specify a bit more what we want to see from the UI. Let's leave that work for that ticket and keep the output as you have it here.

src/uwtools/config.py
```diff
@@ -80,6 +80,36 @@ def dump_file(self, output_path):
...
+        if base_dict is None:
+            base_dict = self.data
+
+        #Set up lists for missing headers, keys and values between dicts
+        misheader = {}
+        miskey = {}
+        misvalue = {}
+
+        if len(base_dict.items()) != len(user_dict.items()):
+            raise KeyError('Given dictionaries have different ranges of keys!')
```
christinaholtNOAA, line 98: I think that we'd want to know what those differences are. It's helpful to know if one config file contains a section that another does not. I think it may be useful to got through dict1, then dict2 to assess what exists/doesn't in each. Also _how_ they differ, I think is important.

I can make a couple small mods while you're away to capture that.

christinaholtNOAA (commented):
@WeirAE This looks nice. I'm going to just make a couple changes to hopefully get this through in this Sprint. Thanks for the refactor from earlier versions, and for working out the hard parts!

src/uwtools/config.py
WeirAE: Fair, and thanks! I had initially set it to just reject from the spec, but just giving the differences regardless is a better way for what we want to so

christinaholtNOAA:
So, as I was making my small mods, I got carried away, and started finding a couple of additional items that would be helpful. I think that perhaps, we should hang onto this ticket and have it carry over to next sprint. Let's chat about it more next week.

While I do still think it's fine to gather items in lists, I think it is important in the first pass to have a list of how the two files differ. Like which keys are different and how they differ, and which file has what key/value pair.

WeirAE:
Reopening to resolve merging issues

christinaholtNOAA (approved):
LGTM! Thanks for the iteration on this one.

WeirAE (commented):
Requests complete

PR #143, opened 2023-01-20
UW 188: Use Jinja2 templates for extending YAML references

christinaholtNOAA:
**Description**
These changes add a method to the Config class that dereferences templated fields. For now, this has been applied to YAML files, but can be easily extended to allow for references made in other supported Config subclasses (f90nml, INI, etc.) The name is totally up for discussion...

These changes ensure that the YAML parser will do the following:

- Read in a YAML file as a dictionary using pyyaml.
- Fill in references to other YAML key/value pairs, either at the same level, or a different level.
- Fill in references to environment variables, if provided. Best practices might indicate that environment variables are all caps, while keys in the yaml are lower case to avoid namespace collisions. Other options are available, though.
- Refrain from filling in values that are not defined.
- Apply the method repeatedly as long as there are still dereferencing tasks to be done so that order of settings does not matter (with at least one caveat that gets weird when multiplying numbers and strings.

It also adds a method that makes the best guess on whether strings that have templates should be changed to other data types -- mainly that integers, floats, and boolean values are set appropriately.

**Type of change**

**How Has This Been Tested?**

A set of tests have been included in pytest to check that all the above listed behaviors are maintained.

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -8,10 +8,14 @@ import collections
 import configparser
 import copy
 import os
+import re
 
+import jinja2
 import json
 import f90nml
 import yaml
+
+from uwtools.j2template import J2Template
 
 class Config(collections.UserDict):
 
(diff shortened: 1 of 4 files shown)


WeirAE (approved):
Very thorough! The extensive comments here really help. It looks like the test covers all of the added functionality. LGTM

fgabelmannjr (approved):

Issue #158, opened 2023-02-22
UW-228


j-derrico:
**Description**

Add dry-run flag to config_parser.

**Requirements**

Add additional CLI args that make the config parser more user friendly. Please reference the [Toolbox MVP](https://confluence-epic.woc.noaa.gov/display/UW/Toolbox+MVP) for specifics.

The  --dry-run feature shows the rendered final text, but doesn't write a file

**Acceptance Criteria (Definition of Done)**

GIVEN: The user wants to know what the final output will look like from the config parser, but doesn't want to create a file
WHEN: A --dry-run flag is provided via CLI
THEN: The rendered final content is printed to stdout/log.

PR #159, opened 2023-02-22
UW-228: Add dry-run flag to config_parser.

elcarpenter:
**Description**
Added a --dry-run feature that shows the rendered final text, but doesn't write a file  in set_config, as well as corresponding test. Fixes issue [#158](https://github.com/ufs-community/workflow-tools/issues/158#issue-1595795427).  

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -133,13 +133,13 @@ def create_config_obj(argv):
         out_object.dump_file(user_args.outfile)
 
     if user_args.dry_run:
+        log.info(config_obj)
         if user_args.outfile:
             out_object = user_args.outfile
             log.info(f'warning file {out_object} ',
                  r"not written when using --dry_run")
         # apply switch to allow user to view the results of config
         # instead of writing to disk
-    log.info(config_obj)
 
 if __name__ == '__main__':
     create_config_obj(sys.argv[1:])
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -82,10 +82,6 @@ def create_config_obj(argv):
     user_args = parse_args(argv)
     infile_type = get_file_type(user_args.input_base_file)
 
-    log.info(f"""Running script set_config.py:
-{('-' * 70)}
-{('-' * 70)}""")
-
     if infile_type in [".yaml", ".yml"]:
         config_obj = config.YAMLConfig(user_args.input_base_file)
         infile_type = ".yaml"
@@ -143,8 +139,6 @@ def create_config_obj(argv):
                  r"not written when using --dry_run")
         # apply switch to allow user to view the results of config
         # instead of writing to disk
-    elif user_args.outfile:
-        config_obj.dump_file(user_args.outfile)
     log.info(config_obj)
 
 if __name__ == '__main__':
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -94,8 +94,8 @@ def create_config_obj(argv):
         config_obj = config.F90Config(user_args.input_base_file)
 
     else:
-        config_obj = None
         log.info("Set config failure: bad file type")
+        return
 
 
     if user_args.config_file:
@@ -121,8 +121,8 @@ def create_config_obj(argv):
                  r"not written when using --dry_run")
         # apply switch to allow user to view the results of config
         # instead of writing to disk
-    if config_obj is not None:
         log.info(config_obj)
+        return
 
     if user_args.outfile:
         outfile_type = get_file_type(user_args.outfile)
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -94,9 +94,8 @@ def create_config_obj(argv):
         config_obj = config.F90Config(user_args.input_base_file)
 
     else:
-        print("Set config failure: bad file type")
-        log.info(f'warning file {user_args.input_base_file} ',
-                 r"has incorrect file type")
+        config_obj = None
+        log.info("Set config failure: bad file type")
 
 
     if user_args.config_file:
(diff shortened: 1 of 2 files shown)


src/uwtools/set_config.py
```diff
@@ -93,9 +93,22 @@ def create_config_obj(argv):
 
         config_obj.update_values(user_config_obj)
 
-    if user_args.outfile:
+    if user_args.dry_run:
+        logfile = os.path.join(os.path.dirname(__file__), "set_config.log")
+        log = Logger(level='info',
```
WeirAE, line 98: Hmm, I'm not sure we want to limit the logger use to only the dry run. I would be ok with either just rendering the output to stdout or initializing the logger at the start

elcarpenter:
We are failing test_config_field_table, from UW-16 for now, which is still "in progress." It fails regardless of the presence of the dry run flag and test.

WeirAE (changes requested):
I think your local DS_Store file got carried along from VS Code and the logger change would be good, but otherwise it looks fine

christinaholtNOAA (commented):
LGTM. Let's get the issues squared away with the test failure before merging, though.

tests/test_set_config.py
```diff
@@ -228,3 +228,23 @@ def test_incompatible_file_type():
...
+    will print an YAML config file'''
+
+    input_file = os.path.join(uwtools_file_base, pathlib.Path("fixtures/fruit_config.yaml"))
+
+    args = ['-i', input_file, '-d']
+
+    outstring = io.StringIO()
+    with redirect_stdout(outstring):
+        config.YAMLConfig(input_file)
+    expected = outstring.getvalue()
```
christinaholtNOAA, line 243: I think the issue with the test failure is somewhere in here. When you create this object, you don't really get any stdout information.

Would the test work if expected is defined simply like this?

```
expected = config.YAMLConfig(input_file)
```

christinaholtNOAA:
@elcarpenter It might be worth adding .DS_Store to the .gitignore if you wouldn't mind taking that on. (Here or separate PR is fine)

tests/test_set_config.py
elcarpenter: I have this assertion error when I try that:
assert '{"fruit": "papaya", "vegetable": "eggplant", "how_many": 17, "dressing": "ranch"}\n' == {"fruit": "papaya", "vegetable": "eggplant", "how_many": 17, "dressing": "ranch"}

christinaholtNOAA: Interesting...can you strip the new line from `expected`?

elcarpenter: Ok - stripping it from result and converting expected to a string worked.

src/uwtools/set_config.py
```diff
@@ -122,5 +132,14 @@ def create_config_obj(argv):
...
 
+    if user_args.dry_run:
+        if user_args.outfile:
+            out_object = user_args.outfile
+            log.info(f'warning file {out_object} ',
+                 r"not written when using --dry_run")
+        # apply switch to allow user to view the results of config
+        # instead of writing to disk
+    log.info(config_obj)
+
```
christinaholtNOAA, line 143: I think it could be good to move this logic up to just before the  `if user_args.outfile:` so that we don't do extra work if a user provides both an outfile and the dry-run flag.

christinaholtNOAA (commented):
Thanks for those mods! I have on more pretty easy change requested below, if you don't mind.

src/uwtools/set_config.py
christinaholtNOAA: Oh shoot! I meant the whole if-statement -- lines 135 - 142 -- could move to up before line 118 in your version.

src/uwtools/set_config.py
```diff
@@ -87,6 +95,8 @@ def create_config_obj(argv):
 
     else:
         print("Set config failure: bad file type")
+        log.info(f'warning file {user_args.input_base_file} ',
```
WeirAE, line 98: This f-string was giving me a hard time on my copy, but if it passes pytest its ok

christinaholtNOAA: This logged output is the culprit for the failing test because it's testing only for output that says "Set config failure: bad file type"

 I believe if you change the log to:

```
log.info("Set config failure: bad file type")
```
The test will pass.

Alternatively, if you want to keep the more verbose log message (modified to be one string message), the test will need to be updated to match the new output string.

I would suggest the latter since you've already gone to the trouble of introducing the logger and defining a verbose message introduced a logger.

WeirAE (approved):
Good resolution and update, LGTM

src/uwtools/set_config.py
```diff
@@ -86,7 +94,8 @@ def create_config_obj(argv):
         config_obj = config.F90Config(user_args.input_base_file)
 
     else:
-        print("Set config failure: bad file type")
+        config_obj = None
+        log.info("Set config failure: bad file type")
```
christinaholtNOAA, line 98: We likely want a `return` here.

src/uwtools/set_config.py
```diff
@@ -105,6 +114,16 @@ def create_config_obj(argv):
...
 
+    if user_args.dry_run:
+        if user_args.outfile:
+            out_object = user_args.outfile
+            log.info(f'warning file {out_object} ',
+                 r"not written when using --dry_run")
+        # apply switch to allow user to view the results of config
+        # instead of writing to disk
+    if config_obj is not None:
+        log.info(config_obj)
```
christinaholtNOAA, line 125: With a return above, you can likely remove the if-statement and just `log.info(config_obj)`, then `return`. For the dry-run portion, we don't even need to go through the next bit of code.

christinaholtNOAA (commented):
A couple of very minor behavioral cleanup bits here, but it looks really good.

christinaholtNOAA (approved):

Issue #152, opened 2023-01-30
UW-16


j-derrico:
**Description**

As a model user, I want to configure a field table file for the model with a YAML config object.
[https://jira-epic.woc.noaa.gov/browse/UW-16](https://jira-epic.woc.noaa.gov/browse/UW-16)
Scheduled for UW Sprint 7.2 ending 17 February 2023

The model requires a diag table file at run time. Its content is mostly dictable, but it doesn't follow a standard convention, instead it uses syntax like that described [here](https://confluence-epic.woc.noaa.gov/display/UW/UFS+Weather+Model+Configuration+File+Types). Include a class that will write this format from a Configuration object. There is no need to parse this content at this time.

**Acceptance Criteria (Definition of Done)**

GIVEN: A configuration object with exactly to correct entries.
WHEN: The user runs set_config.py with a field table output file (or syntax).
THEN: The output is formulated with syntax expected by a field table.

GIVEN: A configuration object with more entries than needed for the field table
WHEN: The user runs set_config.py with a field table output file (or dry run).
THEN: The tool knows which key/value pairs to translate to field table syntax for the final product.

GIVEN: A configuration object does not contain enough information needed for a field table.
WHEN: The user runs set_config.py with a field table output file (or dry-run).
THEN: The tool knows which key/value pairs to translate to field table syntax and does that as it creates the final product.

GIVEN: A user wants to see what fields are necessary for a field table
WHEN: The user provides a flag (maybe named show_me)
THEN: Then help is printed to the screen to give guidance on the fields necessary for the field table.

WeirAE:
Note that the initial tasks are complete in PR #16, however an additional AC to fulfill requirements of a user interface is needed to close this issue. An additional PR will be performed to complete that.

christinaholtNOAA:
Comments summarizing in person discussion:

For the CLI on this one, we'll modifying set_config.py.

I've opened an example branch with pseudocode here: https://github.com/ufs-community/workflow-tools/tree/fieldtable_set_config_example

This work will likely add 3 flags for in, config, out file type that are optional. If file type flags are not provided, get info from suffix. If they are different (suffix or flag), convert the Config object type to the output.

WeirAE:
Completed by #161

PR #161, opened 2023-02-23
UW-16 Update to use existing behavior under set_config.py

WeirAE:
**Description**
Additional changes move the FieldTableConfig() under set_config.py for command line access and consistency of file handling. Replication of the test_config.py test under test_set_config.py ensures consistent behavior. Completes expanded requirements under #152 

**Type of change**

**How Has This Been Tested?**

- [ ] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -42,11 +42,6 @@ def parse_args(argv):
     )
 
     parser.add_argument(
-        '--in_file_type',
-        help='Optional input file type',
-    )
-
-    parser.add_argument(
         '-o', '--outfile',
         help='Full path to output file. If different from input, will will perform conversion.\
             For field table output, specify model such as "field_table.FV3_GFS_v16"',
@@ -61,11 +56,6 @@ def parse_args(argv):
         '-c', '--config_file',
         help='Optional path to configuration file. Accepts YAML, bash/ini or namelist',
         type=path_if_file_exists,
-    )
-
-    parser.add_argument(
-        '--config_file_type',
-        help='Optional config file type override',
     )
 
     parser.add_argument(
@@ -99,16 +89,14 @@ def create_config_obj(argv):
     elif infile_type == ".nml":
         config_obj = config.F90Config(user_args.input_base_file)
 
-    elif infile_type == "field_table":
-        config_obj = config.FieldTableConfig()
-
     else:
         print("Set config failure: bad file type")
 
-    if user_args.config_file:
-        config_file_type = user_args.config_file_type or get_file_type(user_args.config_file)
 
-        if config_file_type in [".yaml", ".yml", "YML", "field_table"]:
+    if user_args.config_file:
+        config_file_type = get_file_type(user_args.config_file)
+
+        if config_file_type in [".yaml", ".yml"]:
             user_config_obj = config.YAMLConfig(user_args.config_file)
             config_file_type = ".yaml"
 
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -112,21 +112,17 @@ def create_config_obj(argv):
     if user_args.outfile:
         outfile_type = get_file_type(user_args.outfile)
         if outfile_type != infile_type:
-            if outfile_type in [".yaml", ".yml"] and infile_type != ".yaml":
+            if outfile_type in [".yaml", ".yml"]:
                 out_object = config.YAMLConfig()
-                out_object.update(config_obj.data)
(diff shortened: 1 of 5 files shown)


src/uwtools/set_config.py
```diff
@@ -39,50 +39,73 @@ def parse_args(argv):
         help='Path to a config base file.',
         required=True,
         type=path_if_file_exists,
-        )
+    )
+
+    parser.add_argument(
+        '--in_file_type',
+        help='Optional input file type',
+    )
```
christinaholtNOAA, line 47: These three args should have some options for what the user can choose, recalling that we can't parse an input Field Table.

src/uwtools/set_config.py
```diff
@@ -39,50 +39,73 @@ def parse_args(argv):
...
     else:
         print("Set config failure: bad file type")
 
-
     if user_args.config_file:
-        config_file_type = get_file_type(user_args.config_file)
+        config_file_type = user_args.config_file_type or get_file_type(user_args.config_file)
 
-        if config_file_type in [".yaml", ".yml"]:
+        if config_file_type in [".yaml", ".yml", "YML", "field_table"]:
```
christinaholtNOAA, line 108: We won't be able to parse a field_table.

src/uwtools/set_config.py
```diff
@@ -94,7 +117,13 @@ def create_config_obj(argv):
         config_obj.update_values(user_config_obj)
 
     if user_args.outfile:
-        config_obj.dump_file(user_args.outfile)
+        outfile_type = user_args.out_file_type or pathlib.Path(user_args.outfile).suffix
```
christinaholtNOAA, line 120: For consistency, maybe this one should also call `get_file_type`?

src/uwtools/set_config.py
```diff
@@ -94,7 +117,13 @@ def create_config_obj(argv):
...
+        if outfile_type == "field_table":
+            out_object = config.FieldTableConfig()
+            out_object.update(config_obj)
+        else:
```
christinaholtNOAA, line 124: When we chatted about this, we'd mentioned that this could be converted to any output object type. It would be nice to apply the same sort of logic as above to determine what sort of output config object should be created, not just handling the Field Table case.

That gives us a lot of flexibility, including what we need for the Field Table.

christinaholtNOAA (commented):
This looks pretty good! Just a few comments below.

I'm also concerned about the addition of files that have already been added, e.g., the fixtures, `config.py`, and `test_config.py`. Are you working from the same branch that was merged in the last PR?

Perhaps you could try to rebase on the current develop branch to trim this down to just the new additions since last time?

src/uwtools/set_config.py
WeirAE: It is a little tough to say, as the FieldTableConfig just handles field tables, but I think generalizing it here and adjusting as we add more table types later. There are some complexities in the logic, since the bash process reads INI and outputs SH, but I think I can resolve it fairly quickly

src/uwtools/set_config.py
WeirAE: Ah, that's a remnant of the original pseudocode, I had meant to remove it - thanks for catching me!

src/uwtools/set_config.py
WeirAE: I was going to put that into --values_needed, but you're right, the help description would be better

src/uwtools/set_config.py
WeirAE: There is a weird issue on outfile, as if the outfile is set to a temp directory it will fail os.path.exists(arg) - let me see if I can sort it under get_file_type

WeirAE: Got it - get_file_type doesn't need to call path_if_file_exists as that is already checked by the arg call

src/uwtools/set_config.py
WeirAE: Got it - fixing by standardizing expected infile_types with multiple options, so the config should now only trigger on intentional changes

src/uwtools/set_config.py
WeirAE: Also removing the type overrides as this is fixed by the other fix

src/uwtools/set_config.py
WeirAE: Fixed with improved logic

christinaholtNOAA (dismissed):
LGTM once the linter is made happy.

christinaholtNOAA (approved):

Issue #156, opened 2023-02-20
UW-197


WeirAE:
**Description**
As a model user, I want help in providing the necessary config object to generate a field table file
(https://jira-epic.woc.noaa.gov/browse/UW-197)

**Requirements**
FieldTableConfig (#152 ) must be updated to accept a help flag that returns the necessary table fields

**Acceptance Criteria (Definition of Done)**
Given: a user wants to see what fields are necessary for a field table
When: the user provides a flag
Then: Then help is printed to the screen to give guidance on the fields necessary for the field table

**Additional context**
#155 Needs to be completed first

WeirAE:
Completed by #163

PR #163, opened 2023-02-27
UW-197: As a model user, I want help in providing the necessary config object to generate a field table file

WeirAE:
**Description**

Used existing --values_needed arg as a help flag that returns the necessary table fields. Fixes issue #156 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -460,9 +460,18 @@ class INIConfig(Config):
                     file_name.write(f'{key}={value}\n')
 
 class FieldTableConfig(YAMLConfig):
-
-    ''' This class will exist only to write out field_table format given
-    that its configuration has been set by an input YAML file. '''
+    ''' This class exists to write out a field_table format given
+    that its configuration has been set by an input YAML file. 
+----------------------------------------------------------------------
+    Required YAML format for field files:
+----------------------------------------------------------------------
+sphum:
+  longname: specific humidity
+  units: kg/kg
+  profile_type: 
+    name: fixed
+    surface_value: 1.e30
+'''
 
     def __init__(self, config_path=None):
 
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -501,12 +501,12 @@ https://ufs-weather-model.readthedocs.io/en/ufs-v1.0.0/InputsOutputs.html#field-
 
 The example format for generating a field file is::
 
-sphum:
-  longname: specific humidity
-  units: kg/kg
-  profile_type: 
-    name: fixed
-    surface_value: 1.e30'''
+        sphum:
+          longname: specific humidity
+          units: kg/kg
+          profile_type: 
+            name: fixed
+            surface_value: 1.e30'''
 
         with open(output_path, 'w', encoding="utf-8") as file_name:
             file_name.write(self._format_output())
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -493,15 +493,20 @@ class FieldTableConfig(YAMLConfig):
 
     def dump_file(self, output_path):
         ''' Write the formatted output to a text file. 
+FMS field and tracer managers must be registered in an ASCII table called 'field_table'
+This table lists field type, target model and methods the querying model will ask for.
+
+See UFS documentation for more information:
+https://ufs-weather-model.readthedocs.io/en/ufs-v1.0.0/InputsOutputs.html#field-table-file
 
 The example format for generating a field file is::
 
-    sphum:
-      longname: specific humidity
-      units: kg/kg
-      profile_type: 
-        name: fixed
-        surface_value: 1.e30'''
+sphum:
+  longname: specific humidity
+  units: kg/kg
+  profile_type: 
+    name: fixed
+    surface_value: 1.e30'''
 
         with open(output_path, 'w', encoding="utf-8") as file_name:
             file_name.write(self._format_output())
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -461,17 +461,7 @@ class INIConfig(Config):
 
 class FieldTableConfig(YAMLConfig):
     ''' This class exists to write out a field_table format given
-    that its configuration has been set by an input YAML file. 
-----------------------------------------------------------------------
-    Required YAML format for field files:
-----------------------------------------------------------------------
-sphum:
-  longname: specific humidity
-  units: kg/kg
-  profile_type: 
-    name: fixed
-    surface_value: 1.e30
-'''
+    that its configuration has been set by an input YAML file. '''
 
     def __init__(self, config_path=None):
 
@@ -502,7 +492,16 @@ sphum:
         return "\n".join(outstring)
 
     def dump_file(self, output_path):
-        ''' Write the formatted output to a text file. '''
+        ''' Write the formatted output to a text file. 
+
+The example format for generating a field file is::
+
+    sphum:
+      longname: specific humidity
+      units: kg/kg
+      profile_type: 
+        name: fixed
+        surface_value: 1.e30'''
 
         with open(output_path, 'w', encoding="utf-8") as file_name:
             file_name.write(self._format_output())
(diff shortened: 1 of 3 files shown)


src/uwtools/set_config.py
```diff
@@ -122,5 +122,33 @@ def create_config_obj(argv):
...
+            if outfile_type not in convert_list:
+                print(f"""Required YAML format for field files:
+{('-' * 70)}
+sphum:
+  longname: specific humidity
+  units: kg/kg
+  profile_type: 
+    name: fixed
+    surface_value: 1.e30
+{('-' * 70)}""")
```
christinaholtNOAA, line 146: I think all this type of information goes into the doc block of the FieldTableConfigure object. Then this script can call on `help()` to print that information on any of our Config objects if a user provides a flag like "show-more" or whatever makes sense.

christinaholtNOAA (changes requested):
@WeirAE I was thinking we'd talked about a different direction on how to accomplish the goal of this task.

src/uwtools/set_config.py
WeirAE: Ah nice, thats much better! Changing now, thanks!

WeirAE: help() was a bit much, so I went with print(out_object.__doc__) which will let me add every other output object later

src/uwtools/set_config.py
```diff
@@ -122,5 +128,25 @@ def create_config_obj(argv):
...
         out_object.dump_file(user_args.outfile)
 
+    ##begin show_format
+    # if --show_format, print a list of required configuration settings to stdout
+
+    if user_args.show_format:
+        # first ensure all required args are present
+        assert(
+            user_args.show_format and user_args.outfile
+        ), "args: --show_format also requires -outfile for reference"
```
christinaholtNOAA, line 138: `assert` should be reserved for tests and non "production" code (see [ref](https://towardsdatascience.com/2-caveats-to-avoid-when-using-python-assert-statement-8060e123c26a)).

src/uwtools/set_config.py
```diff
@@ -122,5 +128,25 @@ def create_config_obj(argv):
...
+        if outfile_type != infile_type:
+            # next, provide needed format if not performing a simple conversion
+            convert_list = [".yaml", ".yml", ".bash", ".sh", ".ini", ".IN", ".nml"]
+            if outfile_type not in convert_list:
+                print(str(out_object.__doc__))
+            else:
+                print(f"Performing direct conversion from {infile_type} to {outfile_type}, \
+                    no additional formatting is required!")
+        else:
+            print("If not converting, no additional formatting is required!")
```
christinaholtNOAA, line 149: I would simplify this logic even more and check to see whether the out_object has a __doc__ option. If so, log it (since logger has been implemented at this point). If it doesn't have the __doc__ option then, log something like "no additional information is provided for this output file type"

christinaholtNOAA (changes requested):

src/uwtools/set_config.py
WeirAE: Replaced with `raise argparse.ArgumentError` which is much better, thanks!

src/uwtools/set_config.py
WeirAE: Hmm, I had put this in originally essentially to return an answer rather than error if a user checked the config on say .yaml to .ini. I like checking for doc, but we have doc statements on all of the Config classes. I've tried a few tricks using decorators or subheadings in docstrings, but its not very good.

WeirAE: Sorted this out and figured it out, now will print if the file_write docs are longer than default

src/uwtools/set_config.py
```diff
@@ -141,5 +147,20 @@ def create_config_obj(argv):
...
+            raise argparse.ArgumentError(user_args.outfile, \
+                "args: --show_format also requires -outfile for reference")
+
+        ref = out_object.__doc__
+        if outfile_type != infile_type and '\n' in ref:
+            # next, provide needed format if not performing a simple conversion
+            log.info(help(out_object.dump_file))
+        else:
+            log.info(r"No additional information is provided for this output file type")
+
```
christinaholtNOAA, line 164: Why not just log whatever the doc string is? Can you not always call `help()` on a method? What if you just wanted to show the user the one-liner included with the out_object??

tests/test_set_config.py
```diff
@@ -267,3 +267,37 @@ def test_set_config_dry_run():
...
+    The example format for generating a field file is::
+    
+        sphum:
+          longname: specific humidity
+          units: kg/kg
+          profile_type: 
+            name: fixed
+            surface_value: 1.e30
+
+None
```
christinaholtNOAA, line 290: What does this "None" here represent?

src/uwtools/set_config.py
```diff
@@ -141,5 +147,20 @@ def create_config_obj(argv):
             out_object = config_obj
         out_object.dump_file(user_args.outfile)
 
+    ##begin show_format
+    # if --show_format, print a list of required configuration settings to stdout
+    if user_args.show_format:
+        if user_args.outfile is None:
+        # first ensure all required args are present
+            raise argparse.ArgumentError(user_args.outfile, \
```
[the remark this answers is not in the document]
WeirAE: I was having line-too-long linter errors without it, oddly enough

src/uwtools/set_config.py
```diff
@@ -141,5 +147,20 @@ def create_config_obj(argv):
             out_object = config_obj
         out_object.dump_file(user_args.outfile)
 
+    ##begin show_format
```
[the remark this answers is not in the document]
WeirAE: Gotcha, habit, thanks

src/uwtools/set_config.py
WeirAE: I can call help(out_object) though its very long, unless there's a setting I'm not using correctly?

christinaholtNOAA: I wasn't saying that you should call it on the object, just call it on the dump_file method, whether or not the __doc__ has a newline. It seems arbitrary to decide based on that.

src/uwtools/set_config.py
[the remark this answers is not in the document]
christinaholtNOAA: It's fine to separate the arguments onto separate lines. I'm saying that you don't need the backslash when you do that.

src/uwtools/config.py
```diff
@@ -493,7 +492,16 @@ def _format_output(self):
...
+        ''' Write the formatted output to a text file. 
+
+The example format for generating a field file is::
+
+    sphum:
+      longname: specific humidity
+      units: kg/kg
+      profile_type: 
+        name: fixed
+        surface_value: 1.e30'''
```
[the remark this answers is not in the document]
WeirAE: OK,  I was placing it indented within the example header as per Sphinx, though I went back and forth - code indent is preferable for sure though. Will add links and summary!

src/uwtools/config.py
```diff
@@ -493,7 +492,21 @@ def _format_output(self):
...
+https://ufs-weather-model.readthedocs.io/en/ufs-v1.0.0/InputsOutputs.html#field-table-file
+
+The example format for generating a field file is::
+
+sphum:
+  longname: specific humidity
+  units: kg/kg
+  profile_type: 
+    name: fixed
+    surface_value: 1.e30'''
```
christinaholtNOAA, line 509: Please do push this whole block over to be inline with the beginning of the doc block in column 8.

christinaholtNOAA (commented):
LGTM. Let's merge after the small whitespace adjustment I included below.

src/uwtools/config.py
WeirAE: Moved over 8 spaces

christinaholtNOAA: I meant the entire section here. We want the first 8 columns of the whole doc block clear.

Issue #167, opened 2023-03-06
UW-202


j-derrico:
**Description**
As a user, I would like to know which template variables will be filled in, which will be left as jinja2 templates, and a full list of all that are needed (https://jira-epic.woc.noaa.gov/browse/UW-202)

**Requirements**
Add a "values_needed" type command line argument for the config parser that provides a list of template variables referenced, variables that can be filled in with their values, and those that will remain in templates upon completion.  Maybe this is just extra output when run in debug mode?  Formatting can be left to the developer, but make sure it's clear and readable.

**Acceptance Criteria (Definition of Done)**
GIVEN: A user would like additional information from the templater
WHEN: In debug mode (dry-run or not)
THEN: The templater will print to screen/log the values that are required by the template, the values provided by any config object provided.
WHEN: There is no value provided
THEN: The output will clearly state that -- we must differentiate between no value provided and None/Null values.

**Additional context**
How much additional work would it be to have the output contain the line number of the template? 
CRH thinks it would require an extra step of cycling through each of the undefined variables and searching it in the text-based template.

PR #171, opened 2023-03-13
UW 202

j-derrico:
**Description**
Adding a values_needed flag: as a user, I would like to know which template variables will be filled in, which will be left as jinja2 templates, and a full list of all that are needed (https://jira-epic.woc.noaa.gov/browse/UW-202)

Fixes issue #167 

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -301,6 +301,30 @@ class Config(collections.UserDict):
             else:
                 dict_to_update[key] = new_val
 
+    def iterate_values(self, config_dict, set_var, jinja2_var, undeclared_var):
+        '''
+        Given a config dictionary, parse which variables have been set, which are left as 
+        jinja2 templates, and which are undeclared
+        '''
+
+        for key, val in config_dict.items():
+            if isinstance(val, dict):
+                set_var.append(key)
+                self.iterate_values(val, set_var, jinja2_var, undeclared_var)
+            elif isinstance(val, list):
+                set_var.append(key)
+                for item in val:
+                    self.iterate_values(item, set_var, jinja2_var, undeclared_var)
+            elif "{{" in str(val) or "{%" in str(val):
+                jinja2_var.append(key)
+            elif val == "" or not val:
+                undeclared_var.append(key)
+
+            else:
+                set_var.append(key)
+
+        return config_dict, set_var, jinja2_var, undeclared_var
+
 
 class YAMLConfig(Config):
 
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -1,3 +1,4 @@
+#pylint: disable=too-many-arguments
 '''
 This file contains the Config file and its subclasses for a variety of
 dicatable file types.
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -301,32 +301,29 @@ class Config(collections.UserDict):
             else:
                 dict_to_update[key] = new_val
 
-    def iterate_values(self, config_dict, set_var, jinja2_var, empty_var):
+    def iterate_values(self, config_dict, set_var, jinja2_var, empty_var, parent):
         '''
-        Given a config dictionary, recursively parse which keys are complete, which keys have 
-        unfilled jinja templates, and which keys are set to empty.  Takes in and returns a dict 
-        object and three seperate lists.
+        Recursively parse which keys in the object are complete (set_var), which keys have 
+        unfilled jinja templates (jinja2_var), and which keys are set to empty (empty_var). 
         '''
 
         for key, val in config_dict.items():
             if isinstance(val, dict):
-                set_var.append(f'    {key}')
-                self.iterate_values(val, set_var, jinja2_var, empty_var)
+                set_var.append(f'    {parent}{key}')
+                new_parent = f'{parent}{key}.'
+                self.iterate_values(val, set_var, jinja2_var, empty_var, new_parent)
             elif isinstance(val, list):
-                set_var.append(f'    {key}')
+                set_var.append(f'    {parent}{key}')
                 for item in val:
-                    self.iterate_values(item, set_var, jinja2_var, empty_var)
+                    self.iterate_values(item, set_var, jinja2_var, empty_var, parent)
             elif "{{" in str(val) or "{%" in str(val):
-                jinja2_var.append(f'    {key}')
-            # elif val == "" or not val:
+                jinja2_var.append(f'    {parent}{key}')
             elif not val:
-                empty_var.append(f'    {key}')
+                empty_var.append(f'    {parent}{key}')
 
             else:
-                set_var.append(f'    {key}')
-
-        return config_dict, set_var, jinja2_var, empty_var
-
+                set_var.append(f'    {parent}{key}')
+        return config_dict, set_var, jinja2_var, empty_var, parent
 
 class YAMLConfig(Config):
 
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -303,9 +303,9 @@ class Config(collections.UserDict):
 
     def iterate_values(self, config_dict, set_var, jinja2_var, empty_var):
         '''
-        Given a config dictionary, recursively parse which keys are complete, which keys have unfilled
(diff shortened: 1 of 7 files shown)


src/uwtools/config.py
```diff
@@ -301,6 +301,30 @@ def update_values(self, new_dict, dict_to_update=None):
             else:
                 dict_to_update[key] = new_val
 
+    def iterate_values(self, config_dict, set_var, jinja2_var, undeclared_var):
+        '''
+        Given a config dictionary, parse which variables have been set, which are left as 
+        jinja2 templates, and which are undeclared
+        '''
```
christinaholtNOAA, line 308: It might be helpful to add some information about the arguments, including hints that the last three are lists.

src/uwtools/config.py
```diff
@@ -301,6 +301,30 @@ def update_values(self, new_dict, dict_to_update=None):
...
+            if isinstance(val, dict):
+                set_var.append(key)
+                self.iterate_values(val, set_var, jinja2_var, undeclared_var)
+            elif isinstance(val, list):
+                set_var.append(key)
+                for item in val:
+                    self.iterate_values(item, set_var, jinja2_var, undeclared_var)
+            elif "{{" in str(val) or "{%" in str(val):
+                jinja2_var.append(key)
+            elif val == "" or not val:
```
christinaholtNOAA, line 320: Is this one meant to pick up `val = 0` or `val = False`?

src/uwtools/set_config.py
```diff
@@ -120,6 +119,24 @@ def create_config_obj(argv):
...
+        log.info('Filled template variables:')
+        for var in set_var:
+            log.info(var)
+        log.info('Variables left as jinja2 templates:')
+        for var in jinja2_var:
+            log.info(var)
+        log.info('Values still needed:')
+        for var in undeclared_var:
+            log.info(var)
+        return
```
christinaholtNOAA, line 138: For these section headers, could we change them to be a little clearer and more readable? Perhaps...

- Keys that are complete
- Keys that have unfilled jinja2 templates
- Keys that are set to empty

Also, if there's a way to highlight the key lists in the output it would be good. Either indent the list, or have a horizontal line after the header, maybe?

tests/test_set_config.py
```diff
@@ -306,3 +306,88 @@ def test_show_format():
...
+    args = ['-i', input_file, '--values_needed']
+
+    # Capture stdout for values_needed output
+    outstring = io.StringIO()
+    with redirect_stdout(outstring):
+        set_config.create_config_obj(args)
+    result = outstring.getvalue()
+    outcome=\
+    """Filled template variables:
+FV3GFS
```
christinaholtNOAA, line 324: It seems like this one might show up as a section heading to tell the user which section to find the variable:

```
Variables left as jinja2 templates:
FV3GFS.nomads.protocol
```

This might help the user if we had other entries in this file like this:
```
RAP:
  nomads:
    url: '{{ some template }}'     
```

christinaholtNOAA (changes requested):
This looks really clean! Nice.

I left a couple of comments about minor change, and one sort of medium-ish lift. Let me know if that ask isn't clear.

src/uwtools/config.py
j-derrico: if not val picks up the empty variables in all three file types (e.g. toppings =), when we did further testing just now we realized we didn't need the empty string

tests/test_set_config.py
j-derrico: We have a fix for this that still iterates through everything once (and keeps track of the variable path as it goes), but the linter is flagging it for too-many-locals (16/15) and too-many-arguments(6/5).  The alternatives seem to be iterating through the whole file again, or creating a list of tuples which would be bad for readability

src/uwtools/set_config.py
```diff
@@ -120,6 +119,26 @@ def create_config_obj(argv):
 
         config_obj.update_values(user_config_obj)
 
+    if user_args.values_needed:
+        config_dict = config_obj.data
+        set_var = []
+        jinja2_var = []
+        empty_var = []
+        config.Config.iterate_values(config_obj, config_dict, set_var, jinja2_var, empty_var)
```
christinaholtNOAA, line 127: Oh, I missed this call in the last review...
I think this call should be `config_obj.iterate_values(set_var, jinja2_var, empty_var)`.

src/uwtools/config.py
christinaholtNOAA: Suggest an edit:
```
        Recursively parse which keys in the object are complete (set_var), which keys have 
        unfilled jinja templates (jinja2_var), and which keys are set to empty (empty_var).  
```

src/uwtools/config.py
```diff
@@ -301,6 +301,32 @@ def update_values(self, new_dict, dict_to_update=None):
             else:
                 dict_to_update[key] = new_val
 
+    def iterate_values(self, config_dict, set_var, jinja2_var, empty_var):
```
christinaholtNOAA, line 304: After noticing the call to this method below, I realized the interface may need some cleaning up. Since its a class method, you won't need to pass a `config_dict`...that information is stored in `self.data`.

So the method interface becomes:

```
    def iterate_values(self, set_var, jinja2_var, empty_var):

        for key, val in self.data.items():
              ...
```

src/uwtools/config.py
```diff
@@ -301,6 +301,32 @@ def update_values(self, new_dict, dict_to_update=None):
...
+                set_var.append(f'    {key}')
+                self.iterate_values(val, set_var, jinja2_var, empty_var)
+            elif isinstance(val, list):
+                set_var.append(f'    {key}')
+                for item in val:
+                    self.iterate_values(item, set_var, jinja2_var, empty_var)
+            elif "{{" in str(val) or "{%" in str(val):
+                jinja2_var.append(f'    {key}')
+            # elif val == "" or not val:
+            elif not val:
```
christinaholtNOAA, line 322: I like the simplification here, but it will still pick up `False` and `0`. Those are legit values. Perhaps an empty string is important to distinguish. To pick up `None`, which is also important, consider using `elif val is not None`.

tests/test_set_config.py
christinaholtNOAA: Hmmm. I was thinking it might be simpler than that. 

If you add an argument to the method like `parent=""`, then you could update that as you recurse through (using the syntax I suggested in the review)....

```
def iterate_values(self, set_var, jinja2_var, empty_var, parent=""):

        for key, val in self.data.items():
            if isinstance(val, dict):
                set_var.append(f'    {parent}{key}')
                self.iterate_values(val, set_var, jinja2_var, empty_var, parent=f"{parent}{key}.")

```

src/uwtools/config.py
j-derrico: Thank you! Passing config_dict is how we did it when it was still in set_config.py and I just carried it over when we moved it into the class instead of actually using the class.  This will look much better

src/uwtools/config.py
j-derrico: I think this is what we were trying to get at when we were asking about the difference between no value provided and None/Null values

src/uwtools/config.py
christinaholtNOAA: Oh, shoot! Yep, totally missed that part. I'm sorry. For a recursive method, I typically add an optional argument for a dictionary, then it can recurse where needed and pass the nested dict.

Leave it as-is. It's totally good.

src/uwtools/config.py
christinaholtNOAA: Ahhh. Right. So I think checking explicitly for an empty string or None would be the way to go.

src/uwtools/set_config.py
christinaholtNOAA: Disregard this one after the above conversation!

christinaholtNOAA (approved):
LGTM.

WeirAE (approved):
Nice loops, clean & clear, all tests pass. We may have to come in later and simplify the number of test files we're using but that's a separate issue. LGTM

PR #176, opened 2023-03-17
UW-208: Test 2

WeirAE:
**Description**

Given: the user provides a debug flag to any logger-enabled tool
When: a logger decorator function has been applied to a called method
Then: information about the caller, local variables, and expected output will be printed to the screen and log file

Proof-of-concept base test: test_templater.py now calls back the logger object created by templater and writes to it, and also calls a simple decorator

**Type of change**

**How Has This Been Tested?**

- [ ] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -19,6 +19,7 @@ import f90nml
 import yaml
 
 from uwtools.j2template import J2Template
+from uwtools.logger import Logger
 
 class Config(collections.UserDict):
 
@@ -73,13 +74,14 @@ class Config(collections.UserDict):
         self.config_path = config_path
 
         #initialize logger with parent inheritance
-        py_caller = getmodule(stack()[1][0])
-        if py_caller.__name__ ==  __name__:
-            name = __name__
+        if __name__ == '__main__':
+            _name = __name__
         else:
-            name = f"{py_caller.__name__}.{__name__}"
-        log = logging.getLogger(name)
-        log.debug("Beginning logging with %s",name)
+            py_caller = getmodule(stack()[1][0])
+            _name = f"{py_caller.__name__}.{__name__}"
+        self.log = Logger(name=_name).get_logger()
+        message = "existing" if _name in logging.root.manager.loggerDict else "new"
+        self.log.debug("Beginning %s logging with %s" % (message,_name))
 
     def __repr__(self):
         ''' This method will return configure contents'''
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -80,8 +80,9 @@ class Config(collections.UserDict):
             py_caller = getmodule(stack()[1][0])
             _name = f"{py_caller.__name__}.{__name__}"
         self.log = Logger(name=_name).get_logger()
+        #pylint: disable=no-member, consider-using-f-string
         message = "existing" if _name in logging.root.manager.loggerDict else "new"
-        self.log.debug("Beginning %s logging with %s" % (message,_name))
+        self.log.debug("Beginning {message} logging with {name}", message=message, name=_name)
 
     def __repr__(self):
         ''' This method will return configure contents'''
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -11,6 +11,8 @@ import copy
 import json
 import os
 import re
+import logging
+from inspect import getmodule, stack
 
 import jinja2
 import f90nml
@@ -69,6 +71,15 @@ class Config(collections.UserDict):
         super().__init__()
 
         self.config_path = config_path
+
+        #initialize logger with parent inheritance
+        py_caller = getmodule(stack()[1][0])
+        if py_caller.__name__ ==  __name__:
+            name = __name__
+        else:
+            name = f"{py_caller.__name__}.{__name__}"
+        log = logging.getLogger(name)
+        log.debug("Beginning logging with %s",name)
 
     def __repr__(self):
         ''' This method will return configure contents'''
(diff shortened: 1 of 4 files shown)


tests/test_templater.py
```diff
@@ -291,3 +293,46 @@ def test_set_template_verbosity():
...
+
+        print("Test templater with logger decorator")
+        # Capture verbose stdout
+        outstring = io.StringIO()
+        with redirect_stdout(outstring):
+            templater.set_template(args)
+            print("Test calling templater logging object")
+            #loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
+            #print(f"Current loggers are: {loggers}")
+            log = logging.getLogger('uwtools.templater')
```
WeirAE, line 326: @christinaholtNOAA I intended this to establish what we talked about on Friday as a check-in, then complete the implementation in a proper full PR.

It turns out this is all that is required to access the logging object, there's no need to try to force it to pass. Once its instantiated, the root manager can access it from the name. I added `name=__name__` to the initial Logger() call which will ensure the logger created by a method now carries that method's name and can be accessed as such, rather than leaving all of them as root. The commented out section was my verification of registered logger objects and the debug call proves that the called log is the existing one rather than a new one.

src/uwtools/templater.py
```diff
@@ -26,7 +26,14 @@ def path_if_file_exists(arg):
         raise argparse.ArgumentTypeError(msg)
     return arg
 
-def parse_args(argv):
+def details():
+    ''' Placeholder for a wrapper'''
+    # writing custom logs specific to function, outside of log decorator, if needed
+    # provide caller, local variables and expected output to the log
+
+    print("Call Logger decorator to print 'hello world'")
```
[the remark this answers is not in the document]
WeirAE: AH sorry, this time I piggybacked on this as the function to call within the decorator to demonstrate the decoration and print. Probably unnecessary, honestly

WeirAE:
@christinaholtNOAA Before moving this to the full PR, I've implemented the details, I'd like to see what you think of this setup first. Detail information used in the start of test_templater has been moved to the "decorator" to be universal, as that's the information we really want to print out in this task I think.

src/uwtools/templater.py
```diff
@@ -104,16 +108,15 @@ def set_template(argv):
...
+#{('-' * 70)}
+#{('-' * 70)}""")
+#    for name, val in user_args.__dict__.items():
+#        if name not in ["config"]:
+#            log.info("{name:>15s}: {val}".format(name=name, val=val))
+#    log.info(f"""{('-' * 70)}
+#{('-' * 70)}""")
 
 
     if user_args.config_file:
```
christinaholtNOAA, line 122: Instead of adding a new function to the logger here, I'd like to see something like passing the templater logger name to the YAMLConfig class here, and then maybe in the YAMLConfig __init__ , you could get the logger, and report that it's creating a config object and list the args that were passed.

christinaholtNOAA (changes requested):

src/uwtools/templater.py
WeirAE: This is really why I kept this on the draft - I felt that the log block here from templater.py covered what we needed, but I was wary of changing that part and making a new function. I think to pass the name variable, the right way would be with a decorator, otherwise I think I'd have to add another arg to all of the configs

christinaholtNOAA: Just the base config....

src/uwtools/config.py
```diff
@@ -70,6 +72,15 @@ def __init__(self, config_path=None):
 
         self.config_path = config_path
 
+        #initialize logger with parent inheritance
+        py_caller = getmodule(stack()[1][0])
+        if py_caller.__name__ ==  __name__:
+            name = __name__
+        else:
+            name = f"{py_caller.__name__}.{__name__}"
+        log = logging.getLogger(name)
```
christinaholtNOAA, line 81: Should be `self.log` so we can use it throughout.

Maybe also `_name`. (Per Brian in person)

src/uwtools/templater.py
```diff
@@ -86,15 +88,25 @@ def set_template(argv):
     '''Main section for rendering and writing a template file'''
     user_args = parse_args(argv)
 
+    #initialize logger with parent inheritance
+    py_caller = getmodule(stack()[1][0])
+    if py_caller.__name__ ==  __name__:
+        name = __name__
+    else:
+        name = f"{py_caller.__name__}.{__name__}"
```
christinaholtNOAA, line 96: Could this be just `__name__` if the caller is `__main__`?

tests/test_templater.py
```diff
@@ -291,3 +293,62 @@ def test_set_template_verbosity():
...
+             ]
+
+        print("Test templater with logger decorator")
+        # Capture verbose stdout
+        outstring = io.StringIO()
+        with redirect_stdout(outstring):
+            templater.set_template(args)
+            print("Test calling templater logging object")
+            #Note: all instantiated loggers are stored in logging.root.manager.loggerDict
+            log = logging.getLogger('test_templater.uwtools.templater')
```
christinaholtNOAA, line 346: Should `logging` here be the UW Logger... maybe like this:

```
log = Logger(name='test_templater.uwtools.templater').get_logger()
```

There's a chance that Logger needs to be modified to make sure that if a logger object is already created with the same name it just returns that object that exists....
In __init__ for UW Logger, add a debug statement for whether it matches an existing object or is creating a new one (whether its in `logging.root.manager.loggerDict`).

christinaholtNOAA (commented):
This is looking really good!

tests/test_templater.py
WeirAE: Turns out there's an open pylint issue on this, just have to disable it

WeirAE:
Closing for full PR

Issue #165, opened 2023-03-06
UW-208


WeirAE:
**Description**
As a user, I'd like to see uniform logging for the ins/outs of a whole variety of functions/methods
(https://jira-epic.woc.noaa.gov/browse/UW-208)

**Requirements**
Logger.py should be updated with a decorator function to be called with debug to provide additional information

**Acceptance Criteria (Definition of Done)**
Given: the user provides a debug flag to any logger-enabled tool
When: a logger decorator function has been applied to a called method
Then: information about the caller, local variables, and expected output will be printed to the screen and log file

**Additional context**
This ticket is more about being able to wrap calls to any existing methods

WeirAE:
Completed by #201

PR #188, opened 2023-03-22
UW-208: As a user, I'd like to see uniform logging for the ins/outs of a whole variety of functions/methods

WeirAE:
**Description**

Given: the user provides a debug flag to any logger-enabled tool
When: a logger decorator function has been applied to a called method
Then: information about the caller, local variables, and expected output will be printed to the screen and log file
Completes issue #165 
Previous PRs #174 , #176 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**


tests/test_templater.py
```diff
@@ -291,3 +292,26 @@ def test_set_template_verbosity():
...
+    args = [
+        '-i', input_file,
+        '--dry_run',
+        '-v'
+        ]
+
+    print("Test templater with logger decorator")
+    templater.set_template(args)
+    print("Test calling templater logging object")
+    log = Logger(name='test_templater.uwtools.templater').get_logger()
```
WeirAE, line 312: Currently some issue with this implementation, testing now

src/uwtools/config.py
```diff
@@ -70,6 +73,17 @@ def __init__(self, config_path=None):
...
+        #initialize logger with parent inheritance
+        if __name__ == '__main__':
+            _name = __name__
+        else:
+            py_caller = getmodule(stack()[1][0])
+            _name = f"{py_caller.__name__}.{__name__}"
+        self.log = Logger(name=_name).get_logger()
+        #pylint: disable=no-member, consider-using-f-string
+        message = "existing" if _name in logging.root.manager.loggerDict else "new"
+        self.log.debug("Beginning {message} logging with {name}", message=message, name=_name)
```
WeirAE, line 85: Should use a test of this in test_config.py, may be an additional issue to resolve after this

tests/test_templater.py
WeirAE: I suspect the issue is that logger.py is overwriting the logger object after instantiating it. I would recommend changing back to logging.getLogger() for now and opening an issue to add the "if _name in logging.root.manager.loggerDict" check to logger to bypass the level and handlers if the object already exists

src/uwtools/config.py
WeirAE: Added issue #189

tests/test_templater.py
WeirAE: Reverted and created issue #190

src/uwtools/templater.py
```diff
@@ -86,15 +88,25 @@ def set_template(argv):
     '''Main section for rendering and writing a template file'''
     user_args = parse_args(argv)
 
+    #initialize logger with parent inheritance
+    if __name__ == '__main__':
+        name = __name__
+    else:
+        py_caller = getmodule(stack()[1][0])
+        name = f"{py_caller.__name__}.{__name__}"
```
christinaholtNOAA, line 96: Same here as above.

src/uwtools/config.py
```diff
@@ -70,6 +73,17 @@ def __init__(self, config_path=None):
...
 
+        #initialize logger with parent inheritance
+        if __name__ == '__main__':
+            _name = __name__
+        else:
+            py_caller = getmodule(stack()[1][0])
+            _name = f"{py_caller.__name__}.{__name__}"
+        self.log = Logger(name=_name).get_logger()
+        #pylint: disable=no-member, consider-using-f-string
+        message = "existing" if _name in logging.root.manager.loggerDict else "new"
```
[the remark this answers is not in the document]
WeirAE: AH. yep. Needs to go before the logger!

src/uwtools/config.py
```diff
@@ -70,6 +73,17 @@ def __init__(self, config_path=None):
 
         self.config_path = config_path
 
+        #initialize logger with parent inheritance
+        if __name__ == '__main__':
+            _name = __name__
```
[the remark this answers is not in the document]
WeirAE: Aha, ok - I think I lost track of the logic here, I was focusing on the class conditional example rather than cutting down to what we really meant to do, thanks

src/uwtools/config.py
```diff
@@ -70,6 +73,17 @@ def __init__(self, config_path=None):
...
+        else:
+            py_caller = getmodule(stack()[1][0])
+            _name = f"{py_caller.__name__}.{__name__}"
+        self.log = Logger(name=_name).get_logger()
```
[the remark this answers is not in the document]
WeirAE: Hmm, so my logic here is to make the logger a child of the parent function's logger if available. It maintains the description of the history in the log by name stamp and inherits the parent's attributes while linking as well, though it does get somewhat convoluted. I'm not sure if getmodule can pull the full tree of modules to join() them?
It would be simpler to just find the parent and reference its logger object if we don't care about tracking the called modules explicitly

christinaholtNOAA:
I just wanted to come back to say now that I have started to get a better picture of why both a Logger object and the logging.getlogger will potentially need to be used in conjunction. I apologize for my reluctance around this topic so far.

src/uwtools/config.py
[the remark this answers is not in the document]
WeirAE: Fixed to check directly against the logger directory

tests/test_templater.py
```diff
@@ -291,3 +292,26 @@ def test_set_template_verbosity():
     outcome = ''
     # Check that only the correct messages were logged
     assert result == outcome
+
+def test_log_passing():
+    """Unit test for checking application of the logger object passing"""
+
+    input_file = os.path.join(uwtools_file_base, "fixtures/nml.IN")
+    logfile = os.path.join(os.path.dirname(templater.__file__), "templater.log")
+    ref = "Only the templater logger would see this"
```
christinaholtNOAA, line 301: What happens here if we want to run two templater tests simultaneously?

christinaholtNOAA (changes requested):
The code looks good, I think. The test might cause us some problems. I don't know that it's safe to run multiple instances simultaneously given that everybody is using the same log file.

tests/test_templater.py
WeirAE: Good point, I pulled that from my verbosity test, which should also be changed to use `tmp_path`. As-is, if two tests run simultaneously, they'll write to the same log file, which shouldn't cause an error despite it, but is bad manners at least. However, looking back through, templater and set_config both also hard code a fixed file, which will lead to the same problem if two users run from the same install or, more likely, the stack. I'll add an issue to address that

WeirAE: Removed the logfile logic and replaced with stdout

src/uwtools/config.py
```diff
@@ -70,6 +72,17 @@ def __init__(self, config_path=None):
...
+        #initialize logger with parent inheritance
+        #pylint: disable=no-member
+        py_caller = getmodule(stack()[1][0])
+        if py_caller.__name__ in logging.root.manager.loggerDict:
+            _name = f"{py_caller.__name__}.{__name__}"
+        else:
+            _name = __name__
+
+        self.log = logging.getLogger(_name)
+        self.log.debug("Beginning logging with {name}", name=_name)
```
christinaholtNOAA, line 84: Would we not want the test to make sure that this line ends up in the log?

src/uwtools/config.py
```diff
@@ -70,6 +72,17 @@ def __init__(self, config_path=None):
 
         self.config_path = config_path
 
+        #initialize logger with parent inheritance
+        #pylint: disable=no-member
+        py_caller = getmodule(stack()[1][0])
+        if py_caller.__name__ in logging.root.manager.loggerDict:
```
christinaholtNOAA, line 78: You have named this thing already in templater. To keep a strict contract, I think you need to just pass that name into this object and use that. Otherwise, we might start seeing some funky behavior if a caller decided to use a different naming convention.

src/uwtools/config.py
WeirAE: Hm. I may have reached the stage in revisions where I'm outsmarting myself....

src/uwtools/config.py
WeirAE: Thats fair, thanks! I was being overly committed to tier inheritance when I should just let the log entries track the modules.

WeirAE:
Hm, tried to manually update to fix the conflicts but its still upset, I'll work on rebasing it

WeirAE:
Good gracious that took some hammering, hopefully sorted now

PR #192, opened 2023-03-23
UW-273: Enable all config file types with templater.

christinaholtNOAA:
**Description**

This should fix an issue found in acceptance testing that doesn't allow an INI file to configure a template. Opens it up to a namelist file also working...technically, but why??

Also added tests and removed the old template.py that is deprecated.

**Type of change**

**How Has This Been Tested?**

**Checklist**


src/uwtools/template.py
christinaholtNOAA, line 1: This file is deprecated. Removing it and its associated tests in test_template.py.

src/uwtools/templater.py
```diff
@@ -117,7 +136,8 @@ def set_template(argv):
 
 
     if user_args.config_file:
-        cfg = config.YAMLConfig(user_args.config_file)
+        config_type = get_file_type(user_args.config_file)
+        cfg = getattr(config, f"{config_type}Config")(user_args.config_file)
```
christinaholtNOAA, line 140: @WeirAE's logic for creating the right config object. :)

WeirAE (approved):
Ah, I had a note to look at template.py, thanks for concatenating those - LGTM!

venitahagerty:
I don't know if this is a problem, and I'm sorry that I didn't notice this during the first round of testing, but the output of templater.py does not have a newline at the end of file, while the jinja template does. This is true both when using a yaml config file and an ini config file.

I also tried using ini config files, one (.ini) without a section header, and one (.sh) with a section header. I did not get errors (progress!) but the fields in the output file were blank. The output from the yaml config file had the fields filled in.

christinaholtNOAA:
@venitahagerty Would you mind sharing you test parameters and I can add it as a functional test here so that we can have fewer issues with those problems in the future?

venitahagerty:
I'm doing my hera testing in /scratch1/BMC/zrtrr/uw as role.rtrr.
I copied a file regional.nml.IN from the weather model with 10 fields to fill in, and ran it through atparse_to_jinja2.py
I created a yaml file, two ini config files, and a shell file that set environment variables for the 10 fields
Do you want the commands I'm using to run templater.py?

venitahagerty:
This may be the testing for the second PR.
I ran templater.py with no config file and the 10 fields filled in on the command line (i.e. CCPP_SUITE='FV3_GFS_v17_coupled_p8' EXTERNAL_IC=.true.)
The output file was identical to using a yaml config file.
I ran templater.py with no config file and no environment vars set. No error, no output file.
I sourced a shell to set the env vars, and echoed a couple of them to check that they were set. 
I ran templater.py with no config file. No error, no output file.

christinaholtNOAA:
Thanks, @venitahagerty !!!

christinaholtNOAA:
@venitahagerty I think I've addressed all the _gotchas_ you found for us. Let me know if there are others.

venitahagerty:
I cloned the latest code and re-ran the tests. All of the tests worked: yaml config file, ini config file with no section header, using command line key-value pairs, and sourcing a script that exports environment variables. Output files do not have a newline at end of file, but that is ok.

venitahagerty (dismissed):
LGTM

Issue #179, opened 2023-03-17
UW-229 Add conversion flag to config_parser.


elcarpenter:
**Description**

Add additional CLI args that make the config parser more user friendly. Please reference the [Toolbox MVP](https://confluence-epic.woc.noaa.gov/display/UW/Toolbox+MVP) for specifics.
The "conversion" flag (please choose an appropriate flag name) is meant to allow a user the flexibility to provide say a yaml base config and write out a f90 namelist.
Each of the base, config, and output file types could technically be different file types (Jeez, I hope that folks don't ACTUALLY do THAT regularly). The most common use case might be that we have a base file in f90nml format, the user config in YAML, and the output needs to be either yaml or f90nml.
Tests already exist for these conversions in test_config:: test_transform_config. Reference those tests for how this is implemented.

**Acceptance Criteria (Definition of Done)**

GIVEN: A user wants to provide different types of files for input and output 
WHEN: The file is provided via the existing flags and an output type is also provided 
THEN: A conversion to the final config type is performed before writing the file (or printing to log/stdout in dry-run mode)

PR #194, opened 2023-03-23
UW-229 Draft: Adding conversion flags

elcarpenter:
**Description**

This is a second check-in to make sure Janet and I are on the right track. Here we've removed the original logic and reworked it with how we understood the requests. Want confirmation before moving too much more forward. 

For issue [#179](https://github.com/ufs-community/workflow-tools/issues/179). 
Add additional CLI args that make the config parser more user friendly. Please reference the [Toolbox MVP](https://confluence-epic.woc.noaa.gov/display/UW/Toolbox+MVP) for specifics.
The "conversion" flag (please choose an appropriate flag name) is meant to allow a user the flexibility to provide say a yaml base config and write out a f90 namelist.
Each of the base, config, and output file types could technically be different file types (Jeez, I hope that folks don't ACTUALLY do THAT regularly). The most common use case might be that we have a base file in f90nml format, the user config in YAML, and the output needs to be either yaml or f90nml.
Tests already exist for these conversions in test_config:: test_transform_config. Reference those tests for how this is implemented.

**Type of change**

**How Has This Been Tested?**

**Checklist**

### DIFF
--- a/src/uwtools/set_config.py
+++ b/src/uwtools/set_config.py
@@ -74,17 +74,17 @@ def parse_args(argv):
     )
 
     parser.add_argument(
-        '--convert_input_file',
+        '--input_file_type',
         help='If provided, will convert provided input file to provided file type. Accepts YAML, bash/ini or namelist',
     )
 
     parser.add_argument(
-        '--convert_config_file',
+        '--config_file_type',
         help='If provided, will convert provided config file to provided file type. Accepts YAML, bash/ini or namelist',
     )
 
     parser.add_argument(
-        '--convert_output_file',
+        '--output_file_type',
         help='If provided, will convert provided output file to provided file type. Accepts YAML, bash/ini or namelist',
     )
     return parser.parse_args(argv)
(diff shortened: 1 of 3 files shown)


src/uwtools/set_config.py
```diff
@@ -72,6 +72,21 @@ def parse_args(argv):
         action='store_true',
         help='If provided, print a list of required configuration settings to stdout',
     )
+
+    parser.add_argument(
+        '--convert_input_file',
```
christinaholtNOAA, line 77: Totally a nit-pick...could these be named as "input_file_type"? Then later, the decision point could be to use the flag if the flag is provided or to try to infer it from the file name suffix if it isn't provided.

christinaholtNOAA (commented):
I think that taking more of a declarative approach to the user interface might be nice. What I mean is that it's nice to be able to provide any file name I want for a given config file (user, base, output), and can then explicitly set the filetype to override our expected conventions. 

The "conversion" part then depends only on the logic that happens when the input flag is different from the output flag. And the validation is that each file, no matter the type, gives you a dictionary of the same depth.

PR #195, opened 2023-03-27
Set a more appropriate configuration order.

christinaholtNOAA:
**Description**

The dereferencing was done too early in the previous implementation. We want the ability to dereference _after_ we load the user config so the config can potentially fill in missing values.

Adds the logic from dereferencing until no changes are made to a method `dereference_all` that can be used by the caller when the time is necessary.

Also adds more information about what templates are left in place so it's easier to identify what user-defined variables should be provided.

This behavior was identified as unfavorable during acceptance testing.

**Type of change**

- [X] Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

**Checklist**


tests/test_config.py
```diff
@@ -274,6 +274,7 @@ def test_dereference():
 
     test_yaml = os.path.join(uwtools_file_base,pathlib.Path("fixtures/gfs.yaml"))
     cfg = config.YAMLConfig(test_yaml)
+    cfg.dereference_all()
```
WeirAE, line 277: Should we really be calling dereference_all() outside of config.py?

christinaholtNOAA: What are some reasons we may not want to?

WeirAE: Hmm, I feel design-wise it should be an internal utility rather than a user command, so run automatically or with a user flag at the end of YAMLConfig()._load(). Might just be a decorated/undecorated function aesthetic preference of mine?

christinaholtNOAA: I think that public methods that are used by the caller are also an important piece to ensure that there is enough external control over the object. Here, for example, dereferencing is an important step to happen _after_ the user's additional config settings have been applied. Do it before, and we don't have any way to update some of those fields, or reference the ones a user must include.

If we need control over when something happens in the process, we need a common, public interface for that. For example, if I wanted to update the object after reading in a series of config files in a specified order, I'd need to dereference after each of the updates. That kind of activity cannot _only_ be included in the building of the object initially.

I agree that there should be an internal utility for doing the job of dereferencing, and we do have that as part of the design. This method, however, calls that internal method as many times as is needed to ensure all settings have been dereferenced that can be with the given information...from the base config, and from the user.

WeirAE: Fair, and appreciated! No other concerns, LGTM

WeirAE (approved):
As above, LGTM

PR #197, opened 2023-03-29
UW-229: Add conversion flag to config_parser.

elcarpenter:
**Description**

[Fixes Issue #179](https://github.com/ufs-community/workflow-tools/issues/179)

Add additional CLI args that make the config parser more user friendly. Please reference the [Toolbox MVP](https://confluence-epic.woc.noaa.gov/display/UW/Toolbox+MVP) for specifics.
The "conversion" flag (please choose an appropriate flag name) is meant to allow a user the flexibility to provide say a yaml base config and write out a f90 namelist.
Each of the base, config, and output file types could technically be different file types (Jeez, I hope that folks don't ACTUALLY do THAT regularly). The most common use case might be that we have a base file in f90nml format, the user config in YAML, and the output needs to be either yaml or f90nml.
Tests already exist for these conversions in test_config:: test_transform_config. Reference those tests for how this is implemented.

**Type of change**

- [x] This change requires a documentation update

**How Has This Been Tested?**

**Checklist**


tests/test_set_config.py
```diff
@@ -405,3 +405,123 @@ def test_values_needed_f90nml():
...
+            set_config.create_config_obj(args)
+        result = outstring.getvalue()
+
+        expected = config.YAMLConfig(input_file)
+        expected_file = f'{tmp_dir}/test.yaml'
+        expected.dump_file(expected_file)
+
+        assert result.rstrip('\n') == str(expected)
+
+def test_output_file_conversion():
```
j-derrico, line 428: def test_output_file_conversion(): is where we're seeing the MissingSectionHeader error

tests/fixtures/simple2_yaml.cfg
```diff
@@ -0,0 +1,8 @@
+scheduler:
+    jobname:
+        - extra_stuff: 12345
+        - account: user_account
```
christinaholtNOAA, line 4: This is interesting....I wasn't sure exactly what pyyaml would do with this structure, but it seems it loads as a list of dicts. Does that count as a new level of depth in the depth algorithm?

tests/test_set_config.py
```diff
@@ -399,3 +399,102 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
     salad.appetizer
 """
     assert result == outcome
+
+def test_cfg_to_yaml_conversion(): #pylint: disable=unused-variable
+    ''' Test that .a cfg file can be used to create a yaml object.'''
```
christinaholtNOAA, line 404:
```suggestion
    ''' Test that a .cfg file can be used to create a yaml object.'''
```

tests/test_set_config.py
```diff
@@ -399,3 +399,102 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+        outstring = io.StringIO()
+        with redirect_stdout(outstring):
+            set_config.create_config_obj(args)
+        result = outstring.getvalue()
+
+        expected = config.YAMLConfig(input_file)
+        expected_file = f'{tmp_dir}/test.yaml'
+        expected.dump_file(expected_file)
+
+        assert result.rstrip('\n') == str(expected)
```
christinaholtNOAA, line 420: I think from our templater experience, we want to make sure our files do end with a new line. 

I think it makes sense to strip this on here to check for content independently of the new line.

Would you mind adding just one additional assert in here that checks that the expected file ends with a new line? I think that would be sufficient to check the YAML dump_file method.

tests/test_set_config.py
```diff
@@ -399,3 +399,102 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+        out_file = f'{tmp_dir}/test_ouput.cfg'
+        args = ['-i', input_file, '-o', out_file, '--output_file_type', '.nml']
+
+        set_config.create_config_obj(args)
+
+        expected = config.F90Config(input_file)
+        expected_file = f'{tmp_dir}/expected_nml.nml'
+        expected.dump_file(expected_file)
+
+        assert compare_files(expected_file, out_file)
```
christinaholtNOAA, line 437: Another similar check that we get a new line here would be helpful.

tests/test_set_config.py
```diff
@@ -399,3 +399,102 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+
+        # test --input_file_type
+        input_file = os.path.join(uwtools_file_base, pathlib.Path("fixtures/simple2_nml.cfg"))
+        args = ['-i', input_file, '--input_file_type', ".pdf"]
+
+        outstring = io.StringIO()
+        with redirect_stdout(outstring):
+            set_config.create_config_obj(args)
+        result = outstring.getvalue()
+        outcome = "Set config failure: bad file type\n"
```
christinaholtNOAA, line 472: There are some examples elsewhere to ensure that a specific exception is raised. That might be useful with the comment I left above.

christinaholtNOAA (commented):
This is looking really nice. I have just a couple of items below that jumped out at me.

tests/test_set_config.py
```diff
@@ -399,3 +398,99 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+        # test --ouput_file_type
+        input_file = os.path.join(uwtools_file_base, pathlib.Path("fixtures/srw_example.yaml"))
+        out_file = f'{tmp_dir}/test_outfile_conversion.yaml'
+        args = ['-i', input_file, '-o', out_file, '--output_file_type', ".nml"]
+
+        with pytest.raises(ValueError) as error:
+            set_config.create_config_obj(args)
+
+            expected = "Set config failure: output object not compatible with input object\n"
+            actual = str(error.value)
```
j-derrico, line 495: This is the only place where we're getting the context manager error

tests/test_set_config.py
```diff
@@ -217,13 +217,12 @@ def test_incompatible_file_type(): #pylint: disable=unused-variable
...
-    with redirect_stdout(outstring):
+    with pytest.raises(ValueError) as error:
         set_config.create_config_obj(args)
-    result = outstring.getvalue()
-    outcome = "Set config failure: bad file type\n"
 
-    assert result == outcome
+        expected = "Set config failure: bad file type\n"
+        actual = str(error.value)
+        assert actual == expected
```
christinaholtNOAA, line 225: Here, I think just calling set_config.create_config_obj() with bad args would be the only thing needed for the context manager to catch the expected Exception.

tests/test_set_config.py
```diff
@@ -217,13 +217,8 @@ def test_incompatible_file_type(): #pylint: disable=unused-variable
...
     args = ['-i', input_file]
 
-    outstring = io.StringIO()
-    with redirect_stdout(outstring):
+    with pytest.raises(ValueError):
         set_config.create_config_obj(args)
-    result = outstring.getvalue()
-    outcome = "Set config failure: bad file type\n"
-
-    assert result == outcome
```
christinaholtNOAA, line 226: I really appreciate this magic simplification we get with pytest. This line diff is a nice illustration of what it buys us.

tests/test_set_config.py
```diff
@@ -399,3 +394,106 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+        # test --input_file_type
+        input_file = os.path.join(uwtools_file_base, pathlib.Path("fixtures/simple2_nml.cfg"))
+        args = ['-i', input_file, '--input_file_type', ".pdf"]
+
+        with pytest.raises(ValueError) as error:
+            set_config.create_config_obj(args)
+
+            expected = "Set config failure: input base file not compatible\n"
+            actual = str(error.value)
+            assert actual == expected
```
christinaholtNOAA, line 475: I'm not sure these three lines will be reachable because the ValueError should be raised and handled in line 471. You may be able to delete them. Same below.

christinaholtNOAA (dismissed):
This looks great. Thanks for addressing my earlier comments. Just a couple of spots that could use cleanup, but I'm good with the functionality.

elcarpenter:
Looks like the changes from 197 are causing one of the tests to fail. The pytest error is showing that when a key's value is a list, the out_file and expected file within test_cfg_to_yaml_conversion() are being interpreted differently.

j-derrico:
> Looks like the changes from 197 are causing one of the tests to fail. The pytest error is showing that when a key's value is a list, the out_file and expected file within test_cfg_to_yaml_conversion() are being interpreted differently.

@christinaholtNOAA we've figured out that the changes to YAMLconfig in config.py in #197 are what's causing the config obj to return python lists instead of yaml lists, specifically that replacing the self.dereference() means that the keys that are lists aren't being dereferenced, so they're not converted to yaml

tests/test_set_config.py
```diff
@@ -401,3 +396,94 @@ def test_values_needed_f90nml(): #pylint: disable=unused-variable
...
+    input_file = os.path.join(uwtools_file_base, pathlib.Path("fixtures/srw_example_yaml.cfg"))
+
+    with tempfile.TemporaryDirectory(dir='.') as tmp_dir:
+
+        out_file = f'{tmp_dir}/test_ouput.yaml'
+        args = ['-i', input_file, '-o', out_file, '--input_file_type', '.yaml']
+
+        set_config.create_config_obj(args)
+
+        expected = config.YAMLConfig(input_file)
```
christinaholtNOAA, line 411: @j-derrico Nice find. The call to set_config is doing the dereferencing for you now. I think adding `expected.dereference_all()` here could do the trick.

christinaholtNOAA (approved):

tests/test_set_config.py
elcarpenter: It looks like it's the out_file that's used in set_config.create-config_obj(args) that isn't being dereferenced. (The naming order for expected/rendered were accidentally flipped in compare_files().)

christinaholtNOAA (approved):
LGTM! Go for it!

PR #203, opened 2023-04-03
UW-208: Uniform logging for the ins/outs of functions/methods.

christinaholtNOAA:
**Description**

The intent is to wrap calls to any existing methods with a logging capability that will give the user debug-level information about the caller, the local defined variables, and expected output.

**Type of change**

**How Has This Been Tested?**

**Checklist**

Co-authored-by: @WeirAE

### DIFF
--- a/src/uwtools/logger.py
+++ b/src/uwtools/logger.py
@@ -2,10 +2,11 @@
 Logger
 """
 
-import sys
-from pathlib import Path
-from typing import Union, List
+import functools
 import logging
+from pathlib import Path
+import sys
+from typing import Union, List
 
 
 class ColoredFormatter(logging.Formatter):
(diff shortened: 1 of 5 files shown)


src/uwtools/logger.py
```diff
@@ -146,8 +147,40 @@ def add_file_handler(cls, logfile_path: Union[str, Path],
         if not logfile_path.parent.is_dir():
             logfile_path.mkdir(parents=True, exist_ok=True)
 
-        handler = logging.FileHandler(str(logfile_path))
+        handler = logging.FileHandler(str(logfile_path), mode="w")
```
christinaholtNOAA, line 150: This is a change in behavior to the log. It will overwrite previous logs, instead of append! Please let me know if this is undesirable.

tests/test_templater.py
```diff
@@ -86,7 +86,8 @@ def test_set_template_dryrun(): #pylint: disable=unused-variable
         templater.set_template(args)
     result = outstring.getvalue()
 
-    assert result == outcome
+    for outcome_line in outcome.split('\n'):
+        assert outcome_line in result
```
christinaholtNOAA, line 90: This change in how the tests are validated allows us to keep adding log lines to output without updating verbatim the output in the test. 

Let me know if you think this requirement is too loose.

tests/test_templater.py
```diff
@@ -287,11 +290,13 @@ def test_set_template_verbosity(): #pylint: disable=unused-variable
   how_many = 22
   dressing = 'balsamic'
 /
+J2Template._load_file INPUT Args: \n{input_file}
```
christinaholtNOAA, line 293: Because who doesn't like a good f-string when the linter allows?

src/uwtools/logger.py
WeirAE: So, I had been looking at this - the log gets kind of unhelpfully cumbersome otherwise. Should we do appending UID or datetime to the log name now or hold it for later with the issue?

tests/test_templater.py
WeirAE: Its not a strict match check so it should be a-ok, and we do that for similar reasons in other places.
Hm, now that you mention it, we should be checking specifically for the log.info() messages from --dry_run rather than the block text. Maybe that should be part of a later general refactoring pass?

src/uwtools/logger.py
christinaholtNOAA: I think probably we can hold off on that until a little later. We may definitely want that change, though.

tests/test_templater.py
christinaholtNOAA: From one of our sessions this week, it may be that we want to do a group walk through of the tests in general to make sure we've tested everything appropriately, and to organize the tests more effectively.

WeirAE (approved):
Thanks for straightening me out and making it better too - LGTM

venitahagerty (approved):
Ran templater.py with -v option. Extra color-coded debug lines were written while running. Wrote out log (and gave location). Ran templater.py again, and log was indeed overwritten and not appended.

LGTM

PR #206, opened 2023-04-07
UW 230: Adding comparison flag to set_config.py

christinaholtNOAA:
**Description**
Adds a comparison flag to set_config. Also addresses recent change to use logging instead of printing. Modifications were needed to the tests to capture log output instead of stdout.

**Type of change**

**How Has This Been Tested?**

**Checklist**


tests/test_set_config.py
```diff
@@ -492,3 +494,41 @@ def test_erroneous_conversion_flags(): #pylint: disable=unused-variable
...
+    result = outstring.getvalue()
+
+    # Make sure the tool output contains all the lines above
+    for expected_line in expected.split('\n'):
+        assert expected_line.strip() in result
+
+    # Make sure it doesn't include any additional significant diffs
+    # A very rough estimate is that there is a word/colon set followed
+    # by a -/+ set
+    pattern = re.compile(r'(\w):\s+(\w+):\s+-\s+(\w+)\s+\+\s+(\w+)')
```
WeirAE, line 531: Oof, no way to avoid regex?

WeirAE (commented):
So much cleaner this way!

tests/test_set_config.py
venitahagerty: I thought I was the only one who didn't like regex :). If it can't easily be done another way, maybe an example or two to make it easier to decode?

WeirAE: @christinaholtNOAA What was the error or issue that needed a second assert with regex?

venitahagerty:
Ran set_config.py with --compare option on hera. Compared a .nml file with a .yaml file, and also .yaml with .yaml. Comparisons were accurate, even if second file had key-value pairs in a different order.

The one odd thing is that variables in upper case were changed to lower case in the comparison output. Would users find that confusing?

tests/test_set_config.py
WeirAE: Thanks for sorting us out in the stand-up, there is a time and place for regex and validating logs is one that we've been avoiding but needed eventually

WeirAE (approved):
With the regex agreed, LGTM

tests/test_set_config.py
christinaholtNOAA: I added a comment here that is hopefully a bit clearer.

venitahagerty (approved):
LGTM

PR #209, opened 2023-04-14
moved user-facing files to top level (scripts)

j-derrico:
**Description**
Addresses issue [#210](https://github.com/ufs-community/workflow-tools/issues/210)

Moved `set_config.py`, `templater.py`, and `atparse_to_jinja2.py`  to scripts directory in top level of `workflow-tools`.  

**Type of change**

**How Has This Been Tested?**

**Checklist**


scripts/__init__.py
```diff
@@ -0,0 +1,8 @@
+# pylint: disable=unused-variable
```
christinaholtNOAA, line 1: There's a high probability this file could just be an empty file and we don't have to manage any of its contents.

christinaholtNOAA (dismissed):
Thanks for this move! I have one very minor suggestion/comment below.

Also, would you consider adding the necessary changes to the docs for this structure change? I'd be cool with a follow up Issue/PR, but I don't know if that's more/less trouble than modifying it here, with this code.

scripts/__init__.py
j-derrico: Since I'm already working on updating the docs with the changes to templater and the new flags I think it makes more sense to either roll these changes into that PR or do them in a follow up after that's done (which will hopefully be Monday)

WeirAE (changes requested):
Looks good! I think I agree with Christina that we should edit the docs here while we're at it

christinaholtNOAA (approved):
This looks good. I'm happy with the follow-up PR for docs.

venitahagerty (approved):
Cloned fork on hera. Did one run each of templater.py and set_config.py from scripts directory. Ran as expected.

Issue #212, opened 2023-04-18
UW-284


WeirAE:
**Description**
This scaffolding will be a best attempt at capturing the methods, inputs, and outputs of the Class and it's methods.
Make this designed for FV3 weather model – uncoupled. 
 
**Requirements**
We need to capture this type of workflow:

The user provides a YAML
Defaults will be filled in from some default YAML. User settings will override where provided.
The driver will stage fix files, create a namelist and a model_config (and any others)
The driver will run the forecast executable as defined by the machine it's being run on.

**Acceptance Criteria (Definition of Done)**
Best effort in the classes methods needed to identify what the end product looks like, but with blank methods.

**Additional context**
For example:

def stage_fix_files():
    ''' Holds the knowledge for how to modify a list of fix files and stages them in the working directory. Likely gets all its info from config_obj. Calls data mover tool (could be python copy). Fix files usually are specific to a given named grid and resolution.''' 

def create_namelist():
    ''' collects all the user intputs required to create a namelist file, calling the existing namelist config tools. '''

WeirAE:
Completed by #214

PR #214, opened 2023-04-18
UW-284: Create the scaffolding for the forecast driver

WeirAE:
**Description**

We need to capture this type of workflow:

The user provides a YAML
Defaults will be filled in from some default YAML. User settings will override where provided.
The driver will stage fix files, create a namelist and a model_config (and any others)
The driver will run the forecast executable as defined by the machine it's being run on.

Completes issue #212 

**Type of change**

**How Has This Been Tested?**

**Checklist**


src/uwtools/forecast_driver.py
```diff
@@ -0,0 +1,61 @@
...
+'''
+
+import argparse
+import inspect
+import logging
+import os
+import sys
+
+from uwtools.j2template import J2Template
+from uwtools import config
```
christinaholtNOAA, line 18: For scaffolding, it makes sense to leave out the imports except those defined as used/needed following the YAGNI principle.

src/uwtools/forecast_driver.py
```diff
@@ -0,0 +1,61 @@
...
+
+def parse_args(argv):
+
+    '''
+    Function maintains the arguments accepted by this script. Please see
+    Python's argparse documentation for more information about settings
+     of each argument.
+    '''
+
+def create_model_config():
```
christinaholtNOAA, line 28: This and all the following functions should probably go in a Forecast class. Recall, that we have a description of a component driver design [here](https://github.com/ufs-community/workflow-tools/wiki/Migrating-Production-Workflows-to-the-Unified-Workflow-using-the-Strangler-Fig-Pattern#component-drivers).

Again, we'll likely want a command line interface that handles user input and such, but remember that here, the goal is capturing the methods, inputs, and outputs of the _Class_ and its methods. The command line interface isn't quite as big a deal right now, as it will likely consist of only a couple of meaningful calls like:

```
def main():
    args = parse_args()
    forecast = Forecast(args)
    forecast.run()
```

src/uwtools/forecast_driver.py
```diff
@@ -0,0 +1,196 @@
...
+        '''
+        # Parse user arguments
+        user_args = parse_args(argv)
+
+        # Set up logging
+        name = f"{inspect.stack()[0][3]}"
+        log = cli_helpers.setup_logging(user_args, log_name=name)
+
+    @classmethod
+    def requirements(cls):
```
christinaholtNOAA, line 131: Why choose classmethod here instead of a regular method?

christinaholtNOAA (commented):
A couple suggests and a question below....

It's looking pretty good. :)

src/uwtools/forecast_driver.py
```diff
@@ -0,0 +1,196 @@
...
+        action='store_true',
+        help='If provided, print no logging messages',
+        )
+    parser.add_argument(
+        '-l', '--log_file',
+        help='Optional path to a specified log file',
+        default=os.path.join(os.path.dirname(__file__), "set_config.log")
+        )
+    return parser.parse_args(argv)
+class Forecast:
```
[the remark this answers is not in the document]
WeirAE: Oh, ok! I was looking at the Luigi source for reference and thats how they were set up, but I had been reluctant to break it up

src/uwtools/forecast_driver.py
```diff
@@ -0,0 +1,196 @@
...
+    stage_output_files()
+        Holds the knowledge for how to modify a list of output files and
+        stages them in the working directory.
+
+    job_card()
+        Turns the resources config object into a batch card for the 
+        configured Task. Can be called from the command line on the front end 
+        node to show the user what the job card would have looked like.
+
+    run()
```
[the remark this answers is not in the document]
WeirAE: Ah! I had been kind of trying to shove it into init for some reason, thats what was bothering me about it, thanks!

src/uwtools/forecast_driver.py
WeirAE: Hmm, I had been considering them usable externally, but I did change --dry_run to do all of that itself, so its probably not needed now!

WeirAE:
@christinaholtNOAA getting some weird linter errors for micromamba, not sure what's different here that's failing?

src/uwtools/drivers/__init__.py
```diff
@@ -0,0 +1,8 @@
+# pylint: disable=unused-variable
+''' The Unified Workflow Tools package. '''
+
+__docformat__ = "NumPy/SciPy"
+
+import os
+
+uwtools_directory = os.path.dirname(__file__)
```
christinaholtNOAA, line 8: We can probably just leave this file empty.

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,237 @@
...
+        '''
+        Initialize the Forecast driver.
+
+        '''
+        # Parse user arguments
+        user_args = parse_args(argv)
+
+        # Set up logging
+        name = f"{inspect.stack()[0][3]}"
+        log = cli_helpers.setup_logging(user_args, log_name=name)
```
christinaholtNOAA, line 130: So far, we've been creating our logger in the caller, and getting it in the init method of the class that uses it.

christinaholtNOAA (commented):
The test is failing because:

```
FAIL Required test coverage of 90% not reached. Total coverage: 86.36%
```

Also I saw a "resolved" on the request to put the command line interface in the scripts directory, and the associated class in the drivers directory. The change for that didn't show up on GitHub if you'd already made that mod.

src/uwtools/drivers/forecast.py
WeirAE: OHHH wait, yes, my mistake, as its in the base class but its not abstract, so actual creation should go to the concrete class, right?

WeirAE: Moved logging to main()

src/uwtools/forecast_driver.py
WeirAE: Still fixing main() unfortunately!

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,247 @@
...
+'''
+
+import abc
+import argparse
+import inspect
+import os
+
+from uwtools.utils import cli_helpers
+
+def parse_args(argv):
```
christinaholtNOAA: This function belongs in a totally separate file under scripts, along with a main function, and any of its imports above.

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,247 @@
...
+        help='If provided, print no logging messages',
+        )
+    parser.add_argument(
+        '-l', '--log_file',
+        help='Optional path to a specified log file',
+        default=os.path.join(os.path.dirname(__file__), "forecast.log")
+        )
+    return parser.parse_args(argv)
+
+class Forecast:
```
christinaholtNOAA: The base class should likely be `Driver` or something similar. It likely deserves its own file because we'll have MANY drivers.

It's subclasses would then be created like:

```
class Forecast(Driver):
```

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,247 @@
...
+    resources()
+        Determine necessary task objects and fill in resource requirements of each.
+        Returns a Config object containing the HPC resources needed.
+
+    validate()
+        Validates the objects generated by the driver from the provided
+        config and platform files. If the --dry_run flag is provided, complete all
+        stages through validation and print results without running the forecast.
+
+    main()
```
christinaholtNOAA, line 94: main() belongs in the script mentioned above.

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,250 @@
...
+        )
+    parser.add_argument(
+        '-l', '--log_file',
+        help='Optional path to a specified log file',
+        default=os.path.join(os.path.dirname(__file__), "forecast.log")
+        )
+    return parser.parse_args(argv)
+
+class Forecast: # pragma: no cover
+    #remove pragma when completed
```
christinaholtNOAA, line 66: This should likely be the base class, named something like `Driver`, and included in its own file since there will be MANY drivers of various types and complexity.

The subclasses, defined in their own files, would inherit these base methods and interfaces like this:

```
class FV3Forecast(Driver):
```

For my comments below, I'll assume this class has become the base class.

I'm choosing FV3Forecast as the subclass name because there _could_ be different forecast models, and we wouldn't want to assume this is the only one, or that others would work with this subclass.

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,250 @@
...
+
+import abc
+import argparse
+import inspect
+import os
+
+from uwtools.utils import cli_helpers
+
+def parse_args(argv):# pragma: no cover
+    #remove pragma when completed
```
christinaholtNOAA, line 23: This, along with its imports above, belong in a separate script under the `scripts/` directory at that top level, and not as part of the uwtools package.

src/uwtools/drivers/forecast.py
```diff
@@ -0,0 +1,250 @@
...
+
+        ''' Turns the resources config object into a batch card for the 
+        configured Task. Can be called from the command line on the front end 
+        node to show the user what the job card would have looked like.'''
+
+    @abc.abstractmethod
+    def run(self):
+
+        ''' Runs the forecast executable with the namelist file and staged
+        input files. This will both build the executable and run it. '''
```
christinaholtNOAA, line 198: The `@abc.abstractmethod` decorated methods here don't seem to be defined as I would have expected.

The ones that ALL drivers should conform to the same interface should be abstract methods, while the ones that are specific to a given component should be defined in that component's sub class, and may not necessarily have a strict interface. Those that have the same implementation in all Drivers can be defined fully in the base class (`run()` and `job_card()` _might_ be examples of those, but perhaps we don't want to assume that here).

Of the ones you've included, the ones that belong to all drivers include:

```
requirements
resources
validate
job_card
run
```

This set seems to be missing the `output()` method mentioned in the [docs](https://github.com/ufs-community/workflow-tools/wiki/Migrating-Production-Workflows-to-the-Unified-Workflow-using-the-Strangler-Fig-Pattern#component-drivers).

All the others should likely go in the subclass `FV3Forecast(Driver)` definition.

christinaholtNOAA (commented):
There are still some structural elements that should be addressed below before I go too much further with the doc strings.

src/uwtools/drivers/forecast.py
WeirAE: So I kept `output()` as `stage_output_files()`, but I can change that. I think the methods otherwise are an issue due to the comments above, just a fundamental misunderstanding of the intended structure. Definitely overly simplistic and missing the necessary levels - thanks for the clarification, working on it now!

src/uwtools/drivers/forecast.py
WeirAE: Ah, so, my intended structure was just starting with the driver directory:
|Forecast() [base class]
|-->SRWForecast() [sub-class by app]
|---->Create_Model_config() [build config by inputs (FV3/HAFS/etc)]

To clarify, I think what you're looking for is to move that one stage higher and implement more of the [directory tree](https://unified-workflow.readthedocs.io/en/latest/Contributors_Guide/directory_tree.html):
scripts/
|create_experiment.py [includes import]
|-->parse_args()
|-->main()
src/uwtools/drivers/
|driver.py
|-->Driver() [base class by app, starting with SRW]
src/uwtools/runners/
|Forecast.py [top level by process, may have to split later by app]
|-->FV3Forecast(Driver) [sub-class by model, starting with SRW FV3]

christinaholtNOAA: Almost. This is what I think I was looking for:

```
scripts/
  run_forecast.py
  --> parse_args()
  --> main()
src/uwtools/drivers/
  driver.py
  --> Driver()
  forecast.py
  --> FV3Forecast(Driver)
  
```

I am aware that I'm making a fairly nuanced distinction between a runner and a driver in this particular example, and could be talked out of it. ;)

What I mean is in the example above, I'm giving a user a script that says it's going to run a forecast -- run_forecast.py. It's a pretty descriptive, human-readable name for the task that it's meant to do. But then when we include it in the classes to actually setup and run that component in the uwtools package, I'm calling those drivers. That's more like a standard [software driver](https://learn.microsoft.com/en-us/windows-hardware/drivers/gettingstarted/what-is-a-driver-) that provides the application the interface to control a lower-level piece of software (the fortran executable).

src/uwtools/drivers/forecast.py
christinaholtNOAA: The names of the two methods to me have significantly different meanings. The `stage_output_files()` method is a verb, and indicates that it's going to do some action, hopefully putting some files in a location. Whereas `output()` is a noun that says it will likely give me some _thing_, hopefully a data structure or object containing the expected "output" of the driver (maybe it's an action, files, a boolean value, or nothing at all).

src/uwtools/drivers/forecast.py
WeirAE: Aha, ok - also fair! I was being less nuanced and just bluntly shoved the naming conventions in the wiki examples, so I'm actually happier with this version 😃

src/uwtools/drivers/forecast.py
WeirAE: Ah, its more of a 'emphasis of priority' difference then! I was emphasizing the staging more than the return, but you're right that the actual output is more important than the directories and should be the focus - gotcha!

christinaholtNOAA (commented):
I think this is looking pretty darn good! :)

Let's see what @venitahagerty thinks, if she has a bit of time to take a look.

venitahagerty:
I read the files and the documentation. I'm seeing the pieces that I expect to see. Having just run the SRW app, I'm thinking of some of those arguments, which will probably be in the user's YAML file. NCO (production) vs. community, and the compiler (i.e. Intel vs GNU). LGTM

WeirAE:
It was hard to not get carried away looking at the upcoming arguments!
Previous conversations resolved, previous updates merged, converted to full PR - thanks everyone!

j-derrico (approved):
LGTM!

PR #234, opened 2023-05-03
UW-12 Create correct directory structure

venitahagerty:
**Description**

As specified by Jira Issue UW-12 and github Issue 233

Each configuration of the UFS requires a specific run directory structure. For each model configuration the tool needs to read in the portion of the model schema that defines the run directory structure for that specific model configuration and create the corresponding run directory tree. The tool needs to confirm that the run directory structure has been created successfully or to exit with a clear and informative error code and message.

**Type of change**

**How Has This Been Tested?**

Has been tested manually at the command line.
Still needs pytests to be created

**Checklist**

### DIFF
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -114,6 +114,11 @@ class FV3Forecast(Driver): # pragma: no cover
         try:
             os.makedirs(run_directory + "/INPUT")
             os.makedirs(run_directory + "/RESTART")
+            # Verify creation of new directory with subdirectories
+            if not os.path.isdir(run_directory + "/RESTART"):
+                logging.critical("New run directories not created")
+                sys.exit(1)                
+            logging.info("Directories created")
         except (RuntimeError, FileExistsError):
             raise RuntimeError("Could not create new run directories")
 
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -98,8 +98,9 @@ class FV3Forecast(Driver): # pragma: no cover
         try:
             if exist_act == "delete" and os.path.isdir(run_directory):
                 shutil.rmtree(run_directory)
-        except (RuntimeError):
-            raise RuntimeError("Could not delete old run directory")
+        except (RuntimeError, FileExistsError) as del_error:
+            raise RuntimeError("Could not delete " +
+                               "old run directory") from del_error
 
         # Try to rename existing run directory if caller chooses Rename
         try:
@@ -107,8 +108,9 @@ class FV3Forecast(Driver): # pragma: no cover
                 now = datetime.now()
                 save_dir = run_directory + now.strftime("_%Y%m%d_%H%M%S")
                 shutil.move(run_directory, save_dir)
-        except (RuntimeError):
-            raise RuntimeError("Could not rename old run directory")
+        except (RuntimeError, FileExistsError) as rename_error:
+            raise RuntimeError("Could not rename " +
+                               "old run directory") from rename_error
 
         # Create new run directory with two required subdirectories
         try:
@@ -117,10 +119,11 @@ class FV3Forecast(Driver): # pragma: no cover
             # Verify creation of new directory with subdirectories
             if not os.path.isdir(run_directory + "/RESTART"):
                 logging.critical("New run directories not created")
-                sys.exit(1)                
+                sys.exit(1)
             logging.info("Directories created")
-        except (RuntimeError, FileExistsError):
-            raise RuntimeError("Could not create new run directories")
+        except (RuntimeError, FileExistsError) as create_error:
+            raise RuntimeError("Could not create " +
+                               "new run directories") from create_error
 
     def output(self):
         ''' Create list of SRW output files and stage them in the working
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -106,8 +106,7 @@ class FV3Forecast(Driver): # pragma: no cover
         try:
             if exist_act == "rename" and os.path.isdir(run_directory):
                 now = datetime.now()
-                save_dir = os.path.join(run_directory,
-                                        now.strftime("_%Y%m%d_%H%M%S"))
+                save_dir = run_directory + now.strftime("_%Y%m%d_%H%M%S")
                 shutil.move(run_directory, save_dir)
         except (RuntimeError, FileExistsError) as rename_error:
             raise RuntimeError("Could not rename " +
@@ -115,10 +114,10 @@ class FV3Forecast(Driver): # pragma: no cover
 
         # Create new run directory with two required subdirectories
         try:
-            os.makedirs(os.path.join(run_directory, "/INPUT"))
-            os.makedirs(os.path.join(run_directory, "/RESTART"))
+            os.makedirs(os.path.join(run_directory, "INPUT"))
+            os.makedirs(os.path.join(run_directory, "RESTART"))
             # Verify creation of new directory with subdirectories
-            if not os.path.isdir(os.path.join(run_directory, "/RESTART")):
+            if not os.path.isdir(os.path.join(run_directory, "RESTART")):
                 logging.critical("New run directories not created")
                 sys.exit(1)
             logging.info("New run directories created")
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -106,15 +106,15 @@ class FV3Forecast(Driver): # pragma: no cover
             if exist_act == "rename" and os.path.isdir(run_directory):
                 now = datetime.now()
                 save_dir = run_directory + now.strftime("_%Y%m%d_%H%M%S")
-                shutil.copytree(run_directory, save_dir)
+                shutil.move(run_directory, save_dir)
         except (RuntimeError):
             raise RuntimeError("Could not rename old run directory")
 
         # Create new run directory with two required subdirectories
         try:
-            os.makedirs(run_directory + "/INPUT", exist_ok=True)
-            os.makedirs(run_directory + "/RESTART", exist_ok=True)
-        except (RuntimeError):
+            os.makedirs(run_directory + "/INPUT")
+            os.makedirs(run_directory + "/RESTART")
+        except (RuntimeError, FileExistsError):
             raise RuntimeError("Could not create new run directories")
 
     def output(self):
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -79,7 +79,7 @@ class FV3Forecast(Driver): # pragma: no cover
 
         Args:
            run_directory: path of desired run directory
-           exist_act: - could be any of [ 'delete', 'rename', 'quit' ]
+           exist_act: - could be any of 'delete', 'rename', 'quit'
                       - how program should act if run directory exists
                       - default is to delete old run directory
            Returns: None
@@ -112,8 +112,8 @@ class FV3Forecast(Driver): # pragma: no cover
 
         # Create new run directory with two required subdirectories
         try:
-            os.makedirs(run_directory + "/INPUT")
-            os.makedirs(run_directory + "/RESTART")
+            os.makedirs(run_directory + "/INPUT", exist_ok=True)
+            os.makedirs(run_directory + "/RESTART", exist_ok=True)
         except (RuntimeError):
             raise RuntimeError("Could not create new run directories")
 
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -89,7 +89,7 @@ class FV3Forecast(Driver): # pragma: no cover
             raise ValueError("Bad argument to create_directory_structure")
 
         # Exit program with error if caller chooses to quit
-        if exist_act == "quit":
+        if exist_act == "quit" and os.path.isdir(run_directory):
             logging.critical("User chose quit option when creating directory")
             sys.exit(1)
 
@@ -98,17 +98,13 @@ class FV3Forecast(Driver): # pragma: no cover
 
         # Create new run directory with two required subdirectories
         try:
-            # Create and verify new directory with subdirectories
-            os.makedirs(os.path.join(run_directory, "INPUT"))
-            if not os.path.isdir(os.path.join(run_directory, "INPUT")):
-                msg = f"Directory {run_directory} with INPUT not created"
-                logging.critical(msg)
-                sys.exit(1)
-            os.makedirs(os.path.join(run_directory, "RESTART"))
-            if not os.path.isdir(os.path.join(run_directory, "RESTART")):
-                msg = f"Directory {run_directory} with RESTART not created"
-                logging.critical(msg)
-                sys.exit(1)
+            for subdir in ("INPUT", "RESTART"):
+                # Create and verify new directory with subdirectories
+                os.makedirs(os.path.join(run_directory, subdir))
+                if not os.path.isdir(os.path.join(run_directory, subdir)):
+                    msg = f"Directory {run_directory} with {subdir} not created"
+                    logging.critical(msg)
+                    raise RuntimeError(msg)
             msg = f"Directory {run_directory} created with subdirectories"
             logging.info(msg)
         except (RuntimeError, FileExistsError) as create_error:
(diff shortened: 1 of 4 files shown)


venitahagerty:
May need improvements to logging.

src/uwtools/drivers/forecast.py
```diff
@@ -6,9 +6,18 @@
...
+from datetime import datetime
+import shutil
+
 from .driver import Driver
 
-class SRWFV3Forecast(Driver): # pragma: no cover
+logging.getLogger(__name__)
+
+
+class FV3Forecast(Driver): # pragma: no cover
```
WeirAE, line 20: Strictly, the forecast.py was setup to include forecasts from all apps, so I put the app name within the class - if we remove this, we'll have to change /scripts/run_forecast.py's autocomplete. I called it 'FV3' as thats how SRW calls it, though MRW just says 'ATM', so some degree of input parsing will be necessary

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +50,87 @@ def validate(self):
...
         usually are specific to a given named grid and resolution. '''
 
     def create_namelist(self):
-        ''' Collects all the user inputs required to create a namelist 
+        ''' Collects all the user inputs required to create a namelist
         file, calling the existing namelist config tools. For SRW,
         this will take the fix files and model_config previously created to
         to generate config.yaml'''
 
+    def create_directory_structure(self, run_directory, exist_act="delete"):
```
WeirAE, line 74: At some point, there was a 'stage input files' but that got lost along the way it seems - thanks for catching that and adding this!

WeirAE (commented):
Sorry for many trailing spaces - Thanks for cleaning them up!

src/uwtools/drivers/forecast.py
venitahagerty: I should have checked to see where the class was used. @christinaholtNOAA requested this change, so hopefully she'll have a chance soon to give guidance on resolution.

christinaholtNOAA: My hope is that all the Apps will be able to configure the same FV3Forecast object with different parameters. We do want to differentiate between FV3 versus a different forecast model. And we may want to grow the inheritance tree in the future to include a sub-class for a coupled FV3, but I don't think we want to divide this up based on existing Apps. 

@venitahagerty Would you mind changing the instance definition in run_forecast.py to change the naming convention.

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +50,87 @@ def validate(self):
...
+
+        # Try to rename existing run directory if caller chooses Rename
+        try:
+            if exist_act == "rename" and os.path.isdir(run_directory):
+                now = datetime.now()
+                save_dir = run_directory + now.strftime("_%Y%m%d_%H%M%S")
+                shutil.move(run_directory, save_dir)
+        except (RuntimeError, FileExistsError) as rename_error:
+            raise RuntimeError("Could not rename " +
+                               "old run directory") from rename_error
```
christinaholtNOAA, line 113: I'm curious if it make sense to pull out the above logic into a helper method (on the base class) or even a util function. This seems like a task we might want to do in a variety of places throughout the workflow and other run tasks.

That could break up the single responsibility of this method to only create the structure. Then the helper function could be responsible for checking what to do in certain cases.

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +50,87 @@ def validate(self):
...
+
+        # Create new run directory with two required subdirectories
+        try:
+            os.makedirs(run_directory + "/INPUT")
```
christinaholtNOAA, line 117: We can achieve more platform independence if we try to use `os.path.join()` when we build up paths. It is a bit annoying that it's so much more syntax, though.

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +50,87 @@ def validate(self):
...
+            os.makedirs(run_directory + "/RESTART")
+            # Verify creation of new directory with subdirectories
+            if not os.path.isdir(run_directory + "/RESTART"):
+                logging.critical("New run directories not created")
+                sys.exit(1)
```
christinaholtNOAA, line 122: I'm curious if there is a situation where `os.makedirs()` would _not_ have raised its own exception,  but we still wouldn't have this path.

christinaholtNOAA (commented):
@venitahagerty This was fast! I left a couple of structural comments/suggestions below. I haven't had a chance to run it or consider logging options just yet.

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +50,87 @@ def validate(self):
...
 
+    def create_directory_structure(self, run_directory, exist_act="delete"):
+        ''' Collects the name of the desired run directory, and has an
+        optional flag for what to do if the run directory specified already
+        exists. Creates the run directory and adds subdirectories
+        INPUT and RESTART. Verifies creation of all directories.
+
+        Args:
+           run_directory: path of desired run directory
+           exist_act: - could be any of 'delete', 'rename', 'quit'
```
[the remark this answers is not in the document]
venitahagerty: I was trying to make it easier to pick out the different options, but it could be confusing. I'll change the comments.

src/uwtools/drivers/forecast.py
venitahagerty: The Jira issues says: The tool needs to confirm that the run directory structure has been created successfully... I'm wondering about permissions, or disk full, or some such.

src/uwtools/drivers/forecast.py
venitahagerty: Changed the three instances

src/uwtools/drivers/forecast.py
venitahagerty: I've been thinking about what that might look like. Perhaps:

if exist_act == "delete":
    delete_directory(run_directory)
if exist_act == "rename":
    rename_directory(run_directory)

I could pass in a suffix to rename_directory, or we could always use datetime.now(). Is that what you have in mind?

src/uwtools/drivers/forecast.py
christinaholtNOAA: These different labels define compile-time options for preparing the FV3 forecast executable. Here, we want to make a class that will run all of them. There will be additional settings per UFS App (SRW, MRW, HAFS) that go along with the choice here, but in general ATM is the "simple" one, and then all other compile-time options add on features needed for coupling or additional modeling components for chemistry, waves, ocean, etc.

src/uwtools/drivers/forecast.py
christinaholtNOAA: I _think_ so...I'd say that the example you have here could certainly go into the helper function I was suggesting. Something like this (untested code) perhaps:
```
def handle_existing(run_directory, exist_act="delete"):
    
      if exist_act == "delete":
          shutil.rmtree(run_directory)
      if exist_act == "rename":
          rename_directory(run_directory)

def rename_directory():
    ...

```

And then in `create_direcctory_structure`:

```
def create_directory_structure(self, run_directory, exist_act="delete"):

    handle_existing(run_directory, exist_act)
    ...
```

src/uwtools/drivers/forecast.py
christinaholtNOAA: Gotcha. We definitely need to make sure that we've done what we expected. I'm just questioning whether Python will reliably handle that for us, or do we need to explicitly check.

If we do think it's a good idea to specifically check, that's cool. But we do want to make sure we check all of the subdirs.

src/uwtools/drivers/forecast.py
venitahagerty: Will handling an existing directory, for which the caller specifies delete or rename, be that common?

src/uwtools/drivers/forecast.py
venitahagerty: Two identical commands. My thought was that if the second subdirectory was created, the first one (as well as the containing directory) were as well.

The commands are in a try/except block, so theoretically if the except block is not fired, things are okay. Do you, @christinaholtNOAA, product owner, feel that that is enough to meet the confirmation specified in the Jira issue?

venitahagerty: Decision in standup - do both verifications

src/uwtools/drivers/forecast.py
venitahagerty: Decision in standup - this is common, put it in utils, but not in cli

src/uwtools/drivers/forecast.py
```diff
@@ -85,45 +84,36 @@ def create_directory_structure(self, run_directory, exist_act="delete"):
...
 
-        # Caller can only provide correct argument
+        # Caller should only provide correct argument
         if exist_act not in ["delete", "rename", "quit"]:
             raise ValueError("Bad argument to create_directory_structure")
 
         # Exit program with error if caller chooses to quit
         if exist_act == "quit":
-            logging.critical("User chose quit option")
+            logging.critical("User chose quit option when creating directory")
```
christinaholtNOAA, line 93: Here, the script should only quit if the `run_directory` exists.

src/uwtools/drivers/forecast.py
```diff
@@ -85,45 +84,36 @@ def create_directory_structure(self, run_directory, exist_act="delete"):
...
+            if not os.path.isdir(os.path.join(run_directory, "INPUT")):
+                msg = f"Directory {run_directory} with INPUT not created"
+                logging.critical(msg)
+                sys.exit(1)
             os.makedirs(os.path.join(run_directory, "RESTART"))
-            # Verify creation of new directory with subdirectories
             if not os.path.isdir(os.path.join(run_directory, "RESTART")):
-                logging.critical("New run directories not created")
+                msg = f"Directory {run_directory} with RESTART not created"
+                logging.critical(msg)
```
christinaholtNOAA, line 110: This section could use a loop for reduced duplication:

```
for subdir in ("INPUT", "RESTART"):
    os.makedirs(os.path.join(run_directory, subdir))
    if not os.path.isdir(os.path.join(run_directory, subdir)):
        msg = f"Directory {run_directory} with {subdir} not created"
        logging.critical(msg)
        raise RuntimeError(msg)
```

Since it's in the `try` block, raising the error here will get handled in the except.

christinaholtNOAA (commented):
Before I skip out for the day...just a couple comments below. Also, I didn't see the additional file_helpers script come it (if it exists at this point).

christinaholtNOAA (commented):
I think the code looks great! It might be helpful to have tests for `create_directory_structure` that test behavior when `quit` and `rename` options, and to make sure that when a file already exists and `delete` is chosen it does the right thing.

src/uwtools/drivers/forecast.py
venitahagerty: Put in utils/file_helpers.py

src/uwtools/drivers/forecast.py
venitahagerty: Changed to a loop

tests/drivers/test_forecast.py
```diff
@@ -0,0 +1,52 @@
+#pylint: disable=invalid-name, missing-module-docstring, missing-function-docstring
```
christinaholtNOAA, line 1: I think it's good to remove the first line of the pylint disablers. Only the `missing-module-docstring` seems to apply and that may be one we want to fix by adding a docstring at the header of the file instead of ignoring.

tests/drivers/test_forecast.py
```diff
@@ -0,0 +1,52 @@
...
+
+from uwtools.drivers.driver import Driver
+from uwtools.drivers.forecast import FV3Forecast
+
+
+def test_create_directory_structure():
+    """Tests create_directory_structure method given a directory."""
+    forecast_obj = FV3Forecast(Driver)
+    uwtools_file_base = os.path.join(os.path.dirname(__file__))
+    run_directory = os.path.join(uwtools_file_base, "expt_dir")
```
christinaholtNOAA, line 17: Does it make sense to use a named temporary directory for this path so that it's less likely to have collisions with other tests?

It also means we wouldn't have to remove anything that may already be existing, as in the next two lines.

src/uwtools/drivers/forecast.py
```diff
@@ -40,33 +49,74 @@ def validate(self):
...
         usually are specific to a given named grid and resolution. '''
 
     def create_namelist(self):
-        ''' Collects all the user inputs required to create a namelist 
+        ''' Collects all the user inputs required to create a namelist
         file, calling the existing namelist config tools. For SRW,
         this will take the fix files and model_config previously created to
         to generate config.yaml'''
 
+    def create_directory_structure(self, run_directory, exist_act="delete"):
```
christinaholtNOAA, line 73: Sorry I haven't caught this question sooner. Does it make sense for run_directory to be an argument to this method, or for it to be an object attribute?

venitahagerty: @christinaholtNOAA UW-12 says, "The location of that directory will be communicated to the script by the caller." Are there OO rules of thumb that suggest when a piece of information should be an attribute vs an argument? Is this something that an earlier method will have found in a schema file and want to squirrel away into an attribute?

src/uwtools/drivers/forecast.py
venitahagerty: Also, file_helpers.py is in uwtools/utils

src/uwtools/drivers/forecast.py
christinaholtNOAA: In general, I think about the OO principle such that an object is a set of data with a group of actions associated with it. If a piece of data is inherently tied to the object then it might be helpful to associate that data to the specific object. So, in this case, if I decided to create a whole set of forecast objects for an experiment (one every 6 hours, maybe), then each of those objects could tell me what location on disk it was configured to run.

I could see how we might just have a configuration object associated with each Forecast object and then query tht to get the same information. I have no rule of thumb to decide between those two choices.

Now that I've written it out like this...let's keep it as-is. If we choose to make it an attribute later, along with any of the other configuration settings required by its configuration object, we can refactor.

tests/drivers/test_forecast.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+from uwtools.drivers.driver import Driver
+from uwtools.drivers.forecast import FV3Forecast
+
+
+def test_create_directory_structure():
+    """Tests create_directory_structure method given a directory."""
+    forecast_obj = FV3Forecast(Driver)
+    temp_file_base = tempfile.gettempdir()
```
christinaholtNOAA, line 18: Eeek! I was not at all clear in my last suggestion. I apologize for that. What I meant to suggest was creating a temporary directory for use here with`tempfile.mkdtemp`. This utility gives you the ability to create a local space on the filesystem of your choice that has a unique name is automatically cleaned up when used in conjunction with a context manager.

For example, if you do something like this example, it wouldn't have any collisions with other tests doing something similar that are also named "expt_dir".

```
with tempfile.mkdtemp(dir="./", prefix="expt_dir") as run_directory:
    forecast_obj.create_directory_structure(run_directory, "delete")
    ...
```

The whole set of tests would likely need to be moved inside the scope of the context manager, though.

tests/drivers/test_forecast.py
```diff
@@ -0,0 +1,45 @@
...
+        # Test create_directory_structure when run directory does exist
+        forecast_obj.create_directory_structure(run_directory, "rename")
+        copy_directory = glob.glob(run_directory + "_*")[0]
+        assert os.path.isdir(os.path.join(copy_directory, "RESTART"))
+
+        # Test create_directory_structure when run directory does exist
+        with pytest.raises(SystemExit) as pytest_wrapped_e:
+            forecast_obj.create_directory_structure(run_directory, "quit")
+        assert pytest_wrapped_e.type == SystemExit
+        assert pytest_wrapped_e.value.code == 1
```
christinaholtNOAA, line 45: We might want to just do a quick cleanup down here of all the `copy_directory` paths that might have been created above.

christinaholtNOAA (approved):
I think this looks really good. I realized the "rename" test might leave some residual test directories. If we can clean those up, I think that will do the trick here.

christinaholtNOAA:
I think the "with" block will only clean up the directory with the name assigned to the original. Since one of the tests moves that to a new name, I'm not sure.

What are you seeing on your local disk? If there aren't relics, then I think it would be fine to leave it as-is.

Since there is no "dir" argument to `TemporaryDirectory`, these leftovers would likely show up in the `/tmp` directory under the user where you ran the test suite.

venitahagerty:
@christinaholtNOAA I rummaged through the "tmp" directory to look for leftover files. There were indeed leftovers from rename, but not from the other actions. I added code to clean up the copied directory from rename.

christinaholtNOAA (approved):
LGTM!

PR #237, opened 2023-05-11
UW-255: Changes needed for SRW fill_jinja_template replacement.

christinaholtNOAA:
**Description**

While integrating uwtools into SRW, a few changes were needed. 
- The arguments when generating the templates should point to default values, so I removed those.
- The name of the called script is now included in the logged information since it's helpful in the output in the log for SRW.

In addition, running the tests on Hera highlighted that we don't yet have jsonschema installed/supported there, so we should skip that test when the package isn't available.

**Type of change**

**How Has This Been Tested?**

I ran the linter and pytests manually on hera.

**Checklist**

### DIFF
--- a/src/uwtools/j2template.py
+++ b/src/uwtools/j2template.py
@@ -123,8 +123,7 @@ class J2Template(): #pylint: disable=unused-variable
         Jinja2 Template object
         '''
 
-        self._j2env=Environment(loader=FileSystemLoader(searchpath='/'),
-                               trim_blocks=True,lstrip_blocks=True)
+        self._j2env=Environment(loader=FileSystemLoader(searchpath='/'))
         register_filters(self._j2env)
         return self._j2env.get_template(template_path)
 
--- a/.pylintrc
+++ b/.pylintrc
@@ -374,7 +374,7 @@ ignore-docstrings=yes
 ignore-imports=no
 
 # Minimum lines number of a similarity.
-min-similarity-lines=4
+min-similarity-lines=15
 
 
 [IMPORTS]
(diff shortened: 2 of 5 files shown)


WeirAE (commented):
Tests clear, good change, just checking one thing - is the log entry in SRW not capturing the script name because of our log name change or is the SRW script not capturing the child call? We may want to change other logging to ensure the entries are still showing the correct script

src/uwtools/j2template.py
venitahagerty, line 1: I'm unclear what changes happen when these two options (trim_blocks and lstrip_blocks) are not set to true, and if that would change anything in our testing.

christinaholtNOAA:
@WeirAE SRW doesn't really have consistent logging throughout. There is a logger at the top level of the generate workflow script, but it isn't named or passed to any of the calls to other functions. There are no classes currently, so we're in a bit of a gray area. 

In this way, we can have the output that is sent to either a basic bash log, or a python log show up with some additional information. For example, this function is called separate three times for three separate config files from the bash run script for the forecast model.

src/uwtools/j2template.py
WeirAE: trim_blocks removes the newline after a new block and lstrip_blocks strips leading spaces - I tested these a little and it didn't seem to affect the results and the tests seem to pass at least here...perhaps we should add another test to ensure this is being handled as expected?

christinaholtNOAA: Removing these lines allows us to keep the same output structure as the template file we provide as input.

We likely want to open this method up enough that the user could potentially pass any additional native flags they'd like, but for now, I think going with the defaults is likely the short-term path forward.

WeirAE (approved):
Should be fine for now, we can revisit taking over logging for other apps later. LGTM

venitahagerty (approved):
LGTM

christinaholtNOAA:
@venitahagerty Did I address your concerns? Could I do more?

venitahagerty:
I'm good

Issue #242, opened 2023-06-02
UW-263


WeirAE:
**Description**
As a user, I need the implementation of a strangler pattern wrapper

**Requirements**
shall provide a new file to the repo: experiment_manager.py. There may be additional files that hold scaffolding that will be added
shall include the scaffolding of the work of subsequent tasks
it need NOT provide any real functionality for this deliverable
shall include a mock functional test
host team meeting to group peer review

**Acceptance Criteria (Definition of Done)**

experiment_manager.py -> manager.load_config() -> facade.py -> (new_way)?(new_class.new_way):(old_class.old_way)
does these things:
load config
validate config
create experiment directory
create workflow manager files
link fix files

WeirAE:
Completed by #243

PR #243, opened 2023-06-02
UW-263: As a user, I need the implementation of a strangler pattern wrapper

WeirAE:
**Description**

Shall provide a new file to the repo: experiment_manager.py. There may be additional files that hold scaffolding that will be added. Shall include the scaffolding of the work of subsequent tasks. Shall include a mock functional test.

Completes issue #242 

**Type of change**

**How Has This Been Tested?**

- [ ] Tests on XYZ OS/HPC/CSP

**Checklist**


WeirAE:
Still adding the mock functional test!

src/uwtools/drivers/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+
+        '''
+
+    def create_experiment(self):
+        '''
+        Create the experiment directory and manager files.
+
+        '''
+
+    def create_directory_structure(self, run_directory, exist_act="delete"):
```
christinaholtNOAA, line 50: This method, as implemented, is specific to the forecast model. Please remove it from the experiment scaffolding.

christinaholtNOAA (commented):
I have a couple of surface level comments below. I think perhaps we should schedule the in-person code review next to discuss design and structure a bit more (part of the original Jira ticket)

src/uwtools/drivers/experiment.py
```diff
@@ -0,0 +1,101 @@
...
+
+        '''
+        Initialize the facade driver.
+
+        '''
+        super().__init__(argv)
+
+    def load_config(self):
+        '''
+        Load the configuration file and validate it.
```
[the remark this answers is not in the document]
WeirAE: My thought was to have the load always validate the config while allowing the validation to be called separately, but thats fair

scripts/experiment_manager.py
```diff
@@ -0,0 +1,74 @@
...
+import inspect
+import os
+import sys
+
+from uwtools.drivers import experiment
+from uwtools.utils import cli_helpers
+
+
+def parse_args(argv):# pragma: no cover
+    #remove pragma when completed
```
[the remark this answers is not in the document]
WeirAE: That removes it from coverage checks, its really good for scaffoldings like this, I used it on the previous one

src/uwtools/drivers/experiment.py
WeirAE: I had included it as I liked the validation checks, but I will remove it from here

scripts/experiment_manager.py
```diff
@@ -0,0 +1,74 @@
...
+        '--forecast_app',
+        help='If provided, will define the app to be run.\
+            Currently accepts SRW, MRW, or HAFS',
+        choices=["SRW", "MRW", "HAFS"],
+    )
+    parser.add_argument(
+        '-d', '--dry_run',
+        action='store_true',
+        help='If provided, validate configuration but do not run the forecast',
+    )
```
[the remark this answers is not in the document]
WeirAE: I like having dry run options - for this tool, my intended use would be to complete the read, validation and directory checks without creating the config

WeirAE:
Not quite sure what the issue is? When I run `coverage report` locally I get 96% - is it perhaps not accepting my `#pragma: no cover` flag after all?

WeirAE:
Turns out pytest-cov doesn't follow the same rules as coverage.py does unless you use a .coveragerc file (and even then only sort-of)

scripts/experiment_manager.py
```diff
@@ -0,0 +1,69 @@
...
+    Defines the user interface for the experiment manager. Parses arguments
+    provided by the user and passes to the facade to be run.'''
+    user_args = parse_args(argv)
+
+    # Set up logging
+    name = f"{inspect.stack()[0][3]}"
+    log = cli_helpers.setup_logging(user_args, log_name=name)
+
+    experiment_class = getattr(experiment, f"{user_args.forecast_app}Experiment")
+    managed_experiment = experiment_class(user_args.config_file, log_name=name)
```
maddenp-noaa, line 63: The signature of e.g. `SWRExperiment`'s constructor is `__init__(self, argv)`. Is that one of the classes that we'd be instantiating here? If so, `argv` doesn't seem like the right variable name on the constructor side, as these are already-parsed arguments (`argv` indicates raw, unparsed args); and it doesn't accept a second argument. The linter can't help us here because of the dynamic class lookup we're doing; however, when we add `mypy` type annotations (I'm working on a demo now), we could have something like in this `testing.py`:
```
from typing import Type

class Base:
    def __init__(self, n: int) -> None:
        self.n = n

class Sub(Base):
    pass

the_class: Type[Base] = globals()["Sub"]  # similar to getattr()
s = the_class(88, "foo")
```
and
```
% mypy testing.py 
testing.py:13: error: Too many arguments for "Base"  [call-arg]
Found 1 error in 1 file (checked 1 source file)
```
`pylint` and `mypy` both have to do static analysis, and `pylint` has no way of knowing what class we're going to end up instantiating. But via type annotations, `mypy` knows that we're going to have a subclass of `Base`, which has exactly one argument.

src/uwtools/drivers/facade.py
```diff
@@ -0,0 +1,90 @@
...
+to do the work. It will also be responsible for setting up the 
+working directory and moving files around. 
+
+See UW documentation for more information:
+https://github.com/ufs-community/workflow-tools/wiki/Implementation-of-a-Service-Oriented-Architecture-Fa%C3%A7ade
+'''
+
+import abc
+
+class Facade: # pragma: no cover
```
maddenp-noaa, line 17: I also wonder if `Facade` might be too generic a name. I don't have a better suggestion at the moment, but maybe one will emerge as we flesh this out over time.

maddenp-noaa (changes requested):

src/uwtools/apps/srw.py
```diff
@@ -0,0 +1,75 @@
...
+logging.getLogger(__name__)
+
+
+class SRW210(Facade): # pragma: no cover
+    #remove pragma when completed
+
+    '''
+        Concrete class to handle UFS Short Range Weather app forecasts.
+    '''
+    def __init__(self, argv):
```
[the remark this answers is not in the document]
christinaholtNOAA: I believe it will be an important and useful refactor and new requirement. Since we have several other instances that will require significant changes, I'd prefer not to blow @WeirAE's work out of scope and include that requirement on this work. 

The only caveat being that if we don't get the work done on this PR in a timeline that the refactor beats it. I certainly hope that is not the case given we haven't actually planned that work yet!

src/uwtools/apps/srw.py
```diff
@@ -0,0 +1,75 @@
...
+        '''
+        Load the configuration file.
+
+        '''
+        file_type = cli_helpers.get_file_type(config_file)
+        if file_type == 'INI':
+            with open('config.yaml', 'w', encoding="utf-8") as file_name:
+                ## Note: this is a temporary path until parsing the SRW directory is implemented
+                subprocess.call(["python", "config_utils.py", "-c", config_file, "-t",
+                             "$PWD/config_defaults.yaml", "-o", "yaml"], stdout=file_name)
```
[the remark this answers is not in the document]
WeirAE: 1) Yes, this is the current SRW config method. Strictly including it explicitly here is out of scope, but I found it helpful for reference.
2) True! It may just still be an old habit to overly describe placeholders
3) Actually, both are new to me, so any more clear techniques are appreciated!

src/uwtools/apps/srw.py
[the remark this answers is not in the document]
WeirAE: Totally agreed!

src/uwtools/apps/srw.py
```diff
@@ -0,0 +1,75 @@
...
+import subprocess
+
+from uwtools.utils import file_helpers
+from uwtools.utils import cli_helpers
+from ..drivers.facade import Facade
+
+logging.getLogger(__name__)
+
+
+class SRW210(Facade): # pragma: no cover
```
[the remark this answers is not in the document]
WeirAE: I found myself at a bit of a loss for how to track multiple possible versions after you mentioned it, so I created app top levels with release version named classes - SRW release version 2.1.0 here. Class naming conventions hampering that, as well as my lack of a clear comment explaining it!

src/uwtools/apps/srw.py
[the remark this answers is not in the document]
maddenp-noaa: I mostly like `subprocess.run()` and `.check_output()` because you can just write out the command you're executing instead of breaking it down into an array, which is, for me, harder to read/understand.

src/uwtools/drivers/facade.py
WeirAE: Extremely generic - I'm not a fan of patterns-named-after-patterns, but for lack of a better way, sometimes it reluctantly has to be shaped like it is I guess

scripts/experiment_manager.py
WeirAE: Oh thats thoughtful - git really needs a 'lightbulb' react! I'm still learning so thats great, thanks

src/uwtools/apps/uw.py
```diff
@@ -0,0 +1,66 @@
...
+from ..drivers.facade import Facade
+
+logging.getLogger(__name__)
+
+
+class UW(Facade): # pragma: no cover
+    #remove pragma when completed
+
+    '''
+        Concrete class to handle UFS Short Range Weather app forecasts.
```
christinaholtNOAA, line 26: If this is truly the UW-enabled SRW Facade we will want a clearer name on this class, too. We might have other Facades for other parts of the system. We might have other Apps that need slightly different Facades. Big question for the group: Do we expect that this single class will handle all UFS Apps transitions appropriately?

christinaholtNOAA (changes requested):
@WeirAE I think this has come a long way. I have included some questions in the code, but also wanted to circle back to the reduced-scope option that we'd discussed in the in-person session you hosted for us.

It's likely that the first implementation of this code would be to put a facade on exactly what is currently happening now to configure an experiment for SRW (or the other Apps). We don't even necessarily need the granularity in the first stage of loading a config file and validating it, etc. I do think it's great to see how that would work when we do make it to that stage. 

I think it could also be useful for grasping the concept with what currently exists to see how the facade could run SRW's `generate_FV3LAM_wflow.py` vs a no-op for the UW alternative. Would you please still consider squeezing that switch in as we'd discussed?

scripts/experiment_manager.py
```diff
@@ -0,0 +1,69 @@
...
+     of each argument.
+    '''
+    parser = argparse.ArgumentParser(
+       description='Set config with user-defined settings.'
+    )
+    parser.add_argument(
+        '--forecast_app',
+        help='If provided, will define the app to be run.\
+            Currently accepts SRW, MRW, or HAFS',
+        choices=["SRW", "MRW", "HAFS"],
```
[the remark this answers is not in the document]
WeirAE: I've been struggling with that as well - right now its a placeholder and calling the top app level which I like, but the version number is hardcoded below. Optional version number argument maybe?

src/uwtools/drivers/experiment.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(argv)
+
+    def load_config(self, config_file):
+        '''
+        Load the configuration file.
+
+        '''
+        modname = "uwtools.apps.uw.UW" if len(sys.argv) > 1 else "uwtools.apps.srw.SRW210"
+        load_config = getattr(import_module(modname), "load_config")
+        self.load_config(config_file)
```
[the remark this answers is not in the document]
WeirAE: Even if it doesn't, its kind of a mess - I think I got my load_configs in my load_configs!

src/uwtools/apps/uw.py
WeirAE: The scope management got me a little off here to be honest. I liked the idea of being able to define an app-named class with available functions to be added and called as necessary, but that will likely get unwieldly fast if we add to it. Perhaps for now I version it as 'UWforSRW'?

src/uwtools/drivers/experiment.py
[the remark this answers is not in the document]
WeirAE: Even if it doesn't, its kind of a mess - I think I got my load_configs in my load_configs!

scripts/experiment_manager.py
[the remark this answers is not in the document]
WeirAE: Sorry, scrambled in my head a bit - I think here should be opaque to the downstream user, just calling the app to the facade. Going back to earlier or alternate original versions should be a manual override of the strangler by a power user would be my thought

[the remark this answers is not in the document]
maddenp-noaa: I'd vote for hardcoding a (correct w.r.t. to the actually available classes) list now and replacing it with a dynamically-populated list (definitely nice to have!) in a later ticket.

scripts/experiment_manager.py
maddenp-noaa: Just to be clear, my only ask on this topic is that we either remove the `log_name=name` argument from the caller (here), or add a receiving argument to `Facade` and its subclasses. The rest is ideas for future work.

scripts/experiment_manager.py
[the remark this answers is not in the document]
WeirAE: Sorry, scrambled in my head a bit - I think here should be opaque to the downstream user, just calling the app to the facade. Going back to earlier or alternate original versions should be a manual override of the strangler by a power user would be my thought
(vscode keeps breaking my replies!)

src/uwtools/apps/srw.py
```diff
@@ -0,0 +1,77 @@
...
+import os
+import sys
+import shutil
+import subprocess
+
+from uwtools.utils import file_helpers
+from uwtools.utils import cli_helpers
+from ..drivers.facade import Facade
+
+logging.getLogger(__name__)
```
maddenp-noaa, line 18: I don't think this line or the two copies in other classes actually do anything for us: They return a reference to the current logger, but we don't save it. The docs (https://docs.python.org/3/library/logging.html#logging.getLogger) don't mention any side effects, so I think these lines are no-ops. I'd be happy to be mistaken.

WeirAE: Hmm - PR this came from is #188 There are a few other places where we instantiate the logger object with a name based on the first creating class, then calling that object back as needed to prevent too many objects floating and if we need its attributes. We have refactored this a bit though, I'll have to think on this again in the morning to make sure its still doing what we intend to

src/uwtools/apps/srw.py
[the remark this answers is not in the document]
WeirAE: "invalid name 'f'" c'mon pylint, sheesh

scripts/experiment_manager.py
WeirAE: Gotcha - just removed log_name=name for now to leave the rest for the future

src/uwtools/apps/srw.py
```diff
@@ -0,0 +1,80 @@
...
+        '''
+
+    def create_experiment(self):
+        '''
+        Generate the regional workflow.
+        This sets up the workflow based on config.yaml, links fix files, creates input.nml and FV3LAM_wflow.xml.
+        '''
+        with open('config.yaml', 'w', encoding="utf-8") as file:
+            # Note: this is a temporary path until parsing the SRW directory is implemented
+            subprocess.run("python generate_FV3LAM_wflow.py", capture_output=True, check=False, shell=True, stdout=file)
```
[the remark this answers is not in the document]
WeirAE: AH, I left that in as part of the copy over from load_config(), thanks for catching me!

scripts/experiment_manager.py
```diff
@@ -0,0 +1,78 @@
...
+    user_args = parse_args(argv)
+
+    # Set up logging
+    name = f"{inspect.stack()[0][3]}"
+    log = cli_helpers.setup_logging(user_args, log_name=name)
+
+    experiment_class = getattr(experiment, f"{user_args.forecast_app}Experiment")
+    managed_experiment = experiment_class(user_args.config_file)
+
+    managed_experiment.load_config()
```
christinaholtNOAA, line 67: It's unclear to me still that we'd need this here. I think that with a big switch for `create_experiment()` we don't necessarily need to call this directly from the top level. I might be missing something here, though.

If we do need to keep it, we probably want to make sure that it provides the arguments required by the `load_config()` method interface.

src/uwtools/drivers/experiment.py
```diff
@@ -0,0 +1,74 @@
...
+        Validate the configuration file.
+
+        '''
+
+    def create_experiment(self):
+        '''
+        Create the experiment directory.
+
+        '''
+        modname = "uwtools.apps.srw.SRW210"
```
christinaholtNOAA, line 57: Should there also be a similar switch here as in `load_config()` so that the caller has some control over whether to do the old thing or the new thing?

christinaholtNOAA (changes requested):
I think we're getting real close. Just a couple more questions below.

src/uwtools/drivers/experiment.py
WeirAE: Oh sorry, I may have misunderstood your previous comment on this - we don't actually have a full UW handling of everything that generate_FV3LAM_wflow.py does so I left that switch disabled pending that

scripts/experiment_manager.py
WeirAE: Ah fair! You're right, one switch to rule them all as it were, totally agreed.

WeirAE: OK, finished my coffee, now I can read - yes,  I was calling manually originally just to not also catch methods that hadn't been completed, but thats actually unnecessary. Fixing this now, thanks!

src/uwtools/drivers/experiment.py
WeirAE: moved the switch to _init_ and clarified the override here

christinaholtNOAA (commented):
Thanks for the quick update @WeirAE. LGTM.

Just needs to make the linter happy.

christinaholtNOAA (approved):
LGTM.

maddenp-noaa (approved):

PR #245, opened 2023-06-02
UW-271 Implement -q/-v args as mutually exclusive

MYWeidner:
**Description**
Made -q and -v mutually exclusive in set_config.py and templater.py. Excluded dry run from this change as dry run with -q or -v flags is appropriate. Created tests for this change in test_set_config.py and test_templater.py. No changes were made to at parse_to_jinja2.py as -q and -v flags are not available in this file.

Completed for issue [162](https://github.com/ufs-community/workflow-tools/issues/162#issue-1601622865) 

**Type of change**

**How Has This Been Tested?**

**Checklist**


tests/test_templater.py
```diff
@@ -157,8 +157,7 @@ def test_set_template_no_config_suffix_fails(): #pylint: disable=unused-variable
         args = [
             '-i', input_file,
             '-c', tmp_file.name,
-            '-d',
-            '-q',
+            '-d'
```
christinaholtNOAA, line 160: Super tiny suggestion here -- a Python life hack, really. When managing lists over multiple lines, it's often helpful to include the comma after the last item for ease of supportability. If someone (or some tool) came in to rearrange the order or the list, or add another item to the list, there'd be no need to remember to add the extra comma here. Things move around in future iterations much more fluidly with that pesky extra comma.

tests/test_templater.py
```diff
@@ -327,3 +325,13 @@ def test_set_template_verbosity(): #pylint: disable=unused-variable
     outcome = ''
     # Check that only the correct messages were logged
     assert result == outcome
+
+def test_mutually_exclusive_args(): #pylint: disable=unused-variable
+    ''' Test that -q, -v, -d args are mutually exclusive'''
```
christinaholtNOAA, line 330: Here, you're suggesting that `-q`, `-v`, and `-d` should be mutually exclusive, but you only included `-q` and `-v` here in the test. They were also the only two included in the `group` addition in the `scripts` files.

Should the comment just include `-q` and `-v`?

christinaholtNOAA (commented):
Mayra, this is a really nice (and quick) first addition! I have a couple of small suggestions for changes below, otherwise it looks great.

WeirAE:
Well, dry_run was just in my original comment, but since it only limits the file not any output, probably just remove it from the comment

christinaholtNOAA:
Do either of you happen to know if you get output if you choose `--dry-run` and `-q`? If so, leaving it as-is should be fine.

I think `--dry-run` should be able to print to screen/log no matter what, so we might consider making those mutually exclusive if the user might get baffled about something like "I added the dry-run flag but it isn't telling me anything".

WeirAE:
AFAIK, overlapping exclusive groups is not especially pleasant. Could add a message "Quiet dry-run: this is the last message you will see - are you sure this is what you want to do?" in the meantime?

tests/test_templater.py
MYWeidner: Good to know. I didn't realize we could have a comma at the end of a list.

tests/test_templater.py
MYWeidner: Comment updated in test. -q and -v should be mutually exclusive.

venitahagerty:
I don't know if I'm being confused by multiple edits, but it looks like -d is still being included in the "mutually exclusive" comments.

MYWeidner:
I added an error for when users use -d and -q, but now I'm failing tests/test_templater.py::test_set_template_verbosity - TypeError: expected str, bytes or os.PathLike object, not NoneType. I can't figure out how to test -q without dry run flag. Does anyone have any ideas?

PR #253, opened 2023-06-12
UW-254: Replace set_namelist in SRW with set_config

christinaholtNOAA:
**Description**

A handful of minor changes that made the integration into SRW go a bit smoother.

**Type of change**

**How Has This Been Tested?**

Standard pytests and pylint. Also integrated in SRW running its fundamental WE2E tests.

**Checklist**


.pylintrc
```diff
@@ -60,7 +60,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use "--disable=all --enable=classes
 # --disable=W".
-disable=
+disable=consider-using-f-string
```
christinaholtNOAA, line 63: This should get us away from this back and forth issue linter issues when we don't put an f-string in the call to the log.

scripts/set_config.py
```diff
@@ -105,14 +104,19 @@ def parse_args(argv):
         )
     return parser.parse_args(argv)
 
-def create_config_obj(argv, log=None):
+def create_config_file(argv, config_dict=None, log_name=None):
```
christinaholtNOAA, line 107: When we call this function from a python script, its name is a bit misleading. It is creating a file, not an object. It's also nice to be able to pass a dict when we don't have a file written out to disk.

I plan to open an issue about how this should be refactored for easier use as a function outside the CLI script.

scripts/set_config.py
```diff
@@ -212,10 +219,7 @@ def create_config_obj(argv, log=None):
 
 if __name__ == '__main__':
 
-    cli_args = parse_args(sys.argv[1:])
-    LOG_NAME = "set_config"
-    cli_log = cli_helpers.setup_logging(cli_args, log_name=LOG_NAME)
     try:
-        create_config_obj(sys.argv[1:], cli_log)
+        create_config_file(sys.argv[1:])
```
christinaholtNOAA, line 223: The log handler was already doing it's thing inside `create_config_file`, so I removed this duplicate work.

scripts/templater.py
```diff
@@ -107,14 +107,11 @@ def set_template(argv):
...
-    for name, val in user_args.__dict__.items():
-        if name not in ["config"]:
-            log.info("{name:>15s}: {val}".format(name=name, val=val))
-    log.info(f"""{('-' * 70)}""")
-    log.info(f"""{('-' * 70)}""")
+    cli_helpers.log_input(
+        log=log,
+        script_name=name,
+        user_args=user_args,
+        )
```
christinaholtNOAA, line 114: This move to utils will give us consistent output from all our tools.

src/uwtools/config.py
```diff
@@ -331,14 +331,19 @@ def update_values(self, new_dict, dict_to_update=None):
         '''
 
         if dict_to_update is None:
-            dict_to_update = self.data
+            dict_to_update = self.data or {}
```
christinaholtNOAA, line 334: Apparently `self.data` is `None` before we initialize it. 

I'll be opening an Issue about the need for any testing at all of this method.

src/uwtools/config.py
```diff
@@ -331,14 +331,19 @@ def update_values(self, new_dict, dict_to_update=None):
...
+            return
 
         for key, new_val in new_dict.items():
             if isinstance(new_val, dict):
                 if isinstance(dict_to_update.get(key), dict):
                     self.update_values(new_val, dict_to_update[key])
                 else:
                     dict_to_update[key] = new_val
+            elif new_val is None:
+                dict_to_update.pop(key, None)
```
christinaholtNOAA, line 346: The standard approach in SRW is to use this mechanism to remove items that may have been set in the base file.

tests/test_set_config.py
```diff
@@ -241,10 +241,10 @@ def test_set_config_dry_run(): #pylint: disable=unused-variable
 
     outstring = io.StringIO()
     with redirect_stdout(outstring):
-        set_config.create_config_obj(args)
+        set_config.create_config_file(args)
     result = outstring.getvalue()
 
-    assert result.rstrip('\n') == expected_content.rstrip('\n')
+    assert expected_content.rstrip('\n') in result.rstrip('\n')
```
christinaholtNOAA, line 247: The adoption of the same logging as was in templater.py meant that none of the strings matched exactly anymore. This is just making sure that the expected string is at least part of the output.

christinaholtNOAA (commented):
A few clarifying (hopefully) comments.

venitahagerty (dismissed):
LGTM

WeirAE (approved):
Lots of good refactoring! LGTM

scripts/set_config.py
maddenp-noaa: It sounds like this function is not even necessarily creating a config file. Is it possible that it might do nothing at all, e.g. if the configuration doesn't call for an output file and no `config_dict` is provided? If so, then I think this is still misnamed.

The length of this function suggests to me that it's trying to do too much, which makes naming it hard (not to mention testing it, or writing more than a not-very-informative one-liner docstring). I think that would all improve were it broken up. That's probably not something for this PR, but I think we should consider it for future work. I'd be happy to contribute.

scripts/set_config.py
```diff
@@ -113,14 +112,19 @@ def parse_args(argv):
...
-def create_config_obj(argv, log=None):
+def create_config_file(argv, config_dict=None, log_name=None):
     '''Main section for processing config file'''
 
     user_args = parse_args(argv)
 
-    if log is None:
-        name = f"{inspect.stack()[0][3]}"
-        log = cli_helpers.setup_logging(user_args, log_name=name)
+    name = log_name or f"{inspect.stack()[0][3]}"
```
maddenp-noaa, line 120: Out of curiosity, given that I haven't seen these tools running yet: Is `inspect.stack()[0][3]` the name of this function? And is that being incorporated into log messages that users see? If so, is that maybe unnecessary detail for a user-facing CLI tool, exposing internals that they shouldn't need to know about? I can see that it could be valuable for diagnostics for developers; if so, could it maybe be enabled by some sort of debug switch? There are also other tracing tools developers can use that might be even more helpful.

scripts/set_config.py
```diff
@@ -151,6 +155,9 @@ def create_config_obj(argv, log=None):
 
         config_obj.update_values(user_config_obj)
 
+    if config_dict is not None:
+        config_obj.update_values(config_dict)
```
maddenp-noaa, line 159: Tying back to my comment about naming and this function possibly being responsible for too many things, mutating a `Config` object as a side effect might not belong in a function called `create_config_file()`. A refactor to a number of simpler functions would probably help here.

src/uwtools/config.py
maddenp-noaa: Can you share what error you encountered that led to this change? I can't see how `self.data` could be `None`. `UserDict` initializes it to `{}`, so something must be changing it to `None`, and that seems like violation of the invariant that `self.data` should be a `dict`, per the `UserData` documentation:
> The instance’s contents are kept in a regular dictionary, which is accessible via the `data` attribute of `UserDict` instances.
If you get the stacktrace of the failure, it might tell us where the invalid `None` value is originating.

I'll take this opportunity to sound the first note of a drum I'll be beating in the future: We should incorporate `mypy` typing into this codebase. In addition to providing insanely useful documentation to anyone reading the code, it also catches cases where unexpected `None` values arise and force you to deal with them. I can't overstate how valuable I've found it both in catching and avoiding bugs, but also in making it clear when my code is poorly designed: If it's hard to write a correct (and useful!) type for a function, argument, variable, etc., it's probably too complicated. I'll be very happy to help do the work!

@fgabelmannjr I heard you mention that you were more steeped in the JS ecosystem than Python, so I wonder if you've also used TypeScript and what you think of it.

scripts/set_config.py
christinaholtNOAA: Just to be clear, @maddenp-noaa are you okay with me leaving the `create_config_file`  name here?

When it's incorporated into the SRW workflow, creating a config file is exclusively what it is supposed to do. The additional options and perks from the CLI are primarily for added user benefit outside the workflow.

src/uwtools/config.py
christinaholtNOAA: @maddenp-noaa Good call all around here to sound drums. I'm looking back at the offender of this, and it's definitely a caller breach of contract...check out what dumb thing I did in the UW integration.

```
    # Load pre-defined settings for the physics suite
    suite_config = uw_config.YAMLConfig(
        config_path=FV3_NML_YAML_CONFIG_FP,
        )
    # Update the config object to include only the chosen suite data
    suite_config.data = suite_config.data.get(CCPP_PHYS_SUITE)

    # Combine the information from the settings dict and the default
    # suite, where items defined in settings take priority.
    suite_config.update_values(settings)
```

Now, the question is... given that this dumb move is totally allowed, how do we protect against it? I mean...should I have really done it this way? No, I had some icky feelings that it was bad when I did it. I mean, first pass it that I change the SRW integration piece to make sure it am setting the object to an actual dict with:

```
suite_config.data = suite_config.data.get(CCPP_PHYS_SUITE, {})
```

Is this the place where a setter comes into play? If so, would you mind guiding that process a bit more? I've never actually (knew that I) needed one before. Also, I'm not sure what that looks like given that `data` is a built-in public, default method and I want to be careful about overwriting native features when setting `data`.

scripts/set_config.py
maddenp-noaa: @christinaholtNOAA My complaint is that, unless I'm mistaken, this function can be called in such a way that it does nothing at all. That shouldn't be possible. It should either create a config file, or populate an in-memory config object. I think the issue stems from that fact that this function is doing too much. Better, IMO, would be to break this up into _at least_ two functions -- one that constructs an in-memory config object and returns it (a cleaner pattern than taking un- or semi-initialized objects as arguments and then mutating, a Python antipattern I believe appears elsewhere in this codebase), and another that takes such an object and writes it to a file. Callers can then call the function with the desired behavior, rather than implying it via a combination of set/unset function arguments. A third function could combine the two behaviors -- call the cfg object creator, then pass the cfg object to the writer -- as a convenience. I think there may be a couple more functions that could usefully be factored out of this big one, too.

I won't hold up this PR for this, but I think we should address this, as this is a crucial bit of logic, and currently too complex due to bypassing single-responsibility principle. It should be possible to simplify the tests, too, by refactoring this.

scripts/set_config.py
maddenp-noaa: As I've been refactoring on the `condev` branch, I can answer my own question: That `stack()` call indeed returns the name of the function. I think we can vastly simplify how logging is being done in `uwtools` -- it's currently complicated, with logging being set up in many places -- sometimes more than once is a given call path! -- and I'm not seeing what benefit the complexity is providing. At some point maybe we can sync up on requirements / desired behavior and I can see whether the much simpler pattern I'm used to employing would meet those.

christinaholtNOAA: It is the name of the function. TIL that it's not a list that you actually get back from `inspect.stack()`, but a list of `FrameInfo` objects which are named tuples (since Python 3.5) and now class instances backward compatible with the named tuples (since Python 3.11). 

We can at least use their names here for clarity instead of their index, which is nice and self-documenting.

I am 100% on board to learn about and implement better logging tools. Using this approach was the solution to DRYing the output to include the function name.

May I propose the use of the `FrameInfo` attributes here for a clearer current DRY version and leaving the broader logging discussion for a later ticket?

scripts/set_config.py
christinaholtNOAA: Let's definitely have some open discussions about how to consider refactoring this tool and our others. Many of them follow a similar pattern.

src/uwtools/config.py
maddenp-noaa: @christinaholtNOAA Probably disappointingly, my answer has to be "It depends". I mean, there are lots of interesting things we could do to make things safer and enforce desirable constraints/invariants, but today I lack the background to know what those constraints/invariants would be. As a first guess, we might not want to inherit from `UserDict` but, instead, design a class with a _subset_ of `dict` semantics but that enforces behavior we want to rely on. Rather than having a public `data` member, it might manage a private `dict` and force callers to modify it via method calls (including e.g. `__setattr__()` to support `dict` `a[b] = c` setter semantics), which would never allow wholesale replacement of the private `dict` like we saw here -- the private `dict` would _always_ be a `dict`.

But I'd probably benefit from a 15-minute discussion about what that class should be/do, looking ahead and making sure nobody sees any obvious deficiencies before we make it. And I don't think it would be hard at all, by the way -- probably just tweaks to what we've already go. But it'd be worth trying to poke holes in the idea.

scripts/set_config.py
maddenp-noaa: I still (lol, looking forward to the upcoming demo!) haven't seen these tools used in practice, but if we're exposing internals of our code (e.g. function names) in user output, I think that would be worth reconsidering. I know that's useful for developers, but there are other/better tools -- debuggers, tracers, profilers, etc. -- for developers to glean internals insight. I suspect that users would value clean, reliable, concise output. For example, if we did internal refactoring that changed function names, which then changed the format of log output, I don't think users necessarily need to be exposed to that. But is that what we actually have now, function names in log output that users see? And, if not, what benefit do we get from involving the function name in logging?

If there's a chance that we could revisit logging requirements and then potentially greatly simplify what we're doing, I'd say don't spend a minute changing the current code.

christinaholtNOAA: I maybe missed a piece of this bigger picture before. We should indeed have this discussion (and demo!) in real life. Here, I think we are naming the logger itself so that we can make sure we retrieve the correct one (in the case where there are multiple instantiated). We may or may not actually be using it in the output information.

maddenp-noaa: > make sure we retrieve the correct one
The question I'll ask later, in a more free-form, non-PR discussion: Do we need multiple loggers? (Maybe we do! But I never have. Maybe I've just gotten lucky so far.) I might make the outrageous claim, later, that we should have a single logger, and that we should log to `stdout` and let users redirect however they like. I may well be missing the needs of some use case.

christinaholtNOAA: We should probably consider a few scenarios about how the Apps have been handling loggers. We can certainly propose an alternative, if needed, and see how workflow-tools can fit in there appropriately.

maddenp-noaa:
> @maddenp-noaa Given this morning's discussion, does it make sense to also close this PR?

I think so. I think we could revisit this a little later, get the benefits this was meant to provide, but also potentially address some of the questions that were raised here. I'd be happy to work any of the related tickets.

Issue #256, opened 2023-06-14
UW-266 Create validation tool for a machine file in SRW.


venitahagerty:
**Description**
Create a command line tool in the scripts directory that takes a YAML or JSON schema file and a config file of a supported type and validates the config file against the schema file using the package jsonschema. This will eventually (not in this PR) expand to validate config files in SRW, HAFS, and Global.

**Requirements**
Create a schema that describes the machine files for SRW (ufs-community/ufs-srweather-app/ush/machine/*.yaml) and uses as a reference ufs-community/ufs-srweather-app/ush/valid_param_vals.yml. This will eventually be complete and structured in a logical way, but not necessarily for this PR. This schema will be able to be used on any of the Tier 1 systems.

Create a command line tool in the scripts directory that validates config files against the schema file

**Acceptance Criteria (Definition of Done)**
This validation tool will generates useful user feedback for the machine files for SRW (excluding the Rocoto section).

venitahagerty:
Closed by PR 258

PR #258, opened 2023-06-26
UW-266 Create validation tool for a machine file in SRW

venitahagerty:
**Description**

Create a command line tool in the scripts directory that takes a YAML or JSON schema file and a config file of a supported type and validates the config file against the schema file using the package jsonschema. This will eventually (not in this PR) expand to validate config files in SRW, HAFS, and Global.

Jira issue UW-266
Github issue #256

**Type of change**

**How Has This Been Tested?**

- [X] pytests in GitHub actions.
- [X] Tests on Mac OS

**Checklist**

### DIFF
--- a/schema/workflow.jsonschema
+++ b/schema/workflow.jsonschema
@@ -78,6 +78,15 @@
                 "format": "uri"
                 }
             }
+        },
+        "task_get_da_obs": {
+            "description": "task for data assimilation",
+            "type": "object",
+            "properties": {
+                "OBS_SUFFIX": {
+                    "type": "string"
+                    }
+                }
+            }
         }
-    }
 }
--- a/schema/workflow.jsonschema
+++ b/schema/workflow.jsonschema
@@ -21,7 +21,48 @@
             "CCPA_OBS_DIR": {
                 "type": "string",
                 "format": "uri"
-                }
+                },
+              "MRMS_OBS_DIR": {
+                  "type": "string",
+                  "format": "uri"
+                  },
+              "NDAS_OBS_DIR": {
+                  "type": "string",
+                  "format": "uri"
+                  },
+              "METPLUS_PATH": {
+                  "type": "string",
+                  "format": "uri"
+                  },
+              "MET_BIN_EXEC": {
+                  "type": "string"
+                  },
+              "MET_INSTALL_DIR": {
+                  "type": "string",
+                  "format": "uri"
+                  },
+              "DOMAIN_PREGEN_BASEDIR": {
+                  "type": "string",
+                  "format": "uri"
+                  },
+              "PARTITION_DEFAULT": {
+                  "type": "string"
+                  },
+              "QUEUE_DEFAULT": {
+                  "type": "string"
+                  },
+              "PARTITION_FCST": {
+                  "type": "string"
+                  },
+              "QUEUE_FCST": {
+                  "type": "string"
+                  },
+              "PARTITION_HPSS": {
+                  "type": "string"
+                  },
+              "QUEUE_HPSS": {
+                  "type": "string"
+                  }
             }
         },
     "cpl_aqm_parm": {
@@ -29,6 +70,10 @@
         "type": "object",
         "properties": {
             "AQM_CONFIG_DIR": {
+                "type": "string",
+                "format": "uri"
+                },
+            "AQM_BIO_DIR": {
                 "type": "string",
                 "format": "uri"
                 }
--- a/schema/workflow.jsonschema
+++ b/schema/workflow.jsonschema
@@ -1,0 +1,38 @@
+{
+"title": "workflow config",
+"description": "This document is to validate config files from SRW, HAFS, Global",
+"type": "object",
+"properties": {
+    "platform": {
+        "description": "attributes of the platform",
+        "type": "object",
+        "properties": {
+            "WORKFLOW_MANAGER": {
+                "type": "string",
+                "enum": ["rocoto", "none"]
+                },
+            "NCORES_PER_NODE": {
+                "type": "number"
+                },
+            "SCHED": {
+                "type": "string",
+                "enum": ["slurm", "pbspro", "lsf", "lsfcray", "none"]
+                },
+            "CCPA_OBS_DIR": {
+                "type": "string",
+                "format": "uri"
+                }
+            }
+        },
+    "cpl_aqm_parm": {
+        "description": "attributes of coupled air quality",
+        "type": "object",
+        "properties": {
+            "AQM_CONFIG_DIR": {
+                "type": "string",
+                "format": "uri"
+                }
+            }
+        }
+    }
+}
--- a/scripts/validate_config.py
+++ b/scripts/validate_config.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+#pylint: disable=too-many-branches, too-many-statements, too-many-locals
+
+'''
+This utility validates a config file using a validation schema.
+'''
+import argparse
+import inspect
+import os
+import sys
+import json
+import jsonschema
+import yaml
+
+from uwtools import config
+from uwtools import exceptions
+from uwtools.utils import cli_helpers
+
+
+def parse_args(argv):
+
+    '''
+    Function maintains the arguments accepted by this script. Please see
+    Python's argparse documentation for more information about settings of each
+    argument.
+    '''
+
+    parser = argparse.ArgumentParser(
+       description='Validate config with user-defined settings.'
+    )
+
+    group = parser.add_mutually_exclusive_group()
+
+    parser.add_argument(
+        '-s', '--validation_schema',
+        help='Path to a validation schema.',
+        required=True,
+        type=cli_helpers.path_if_file_exists,
+    )
+
+    parser.add_argument(
+        '-c', '--config_file',
+        help='Path to configuration file. Accepts YAML, bash/ini or namelist',
+        required=True,
+        type=cli_helpers.path_if_file_exists,
+    )
+
+    parser.add_argument(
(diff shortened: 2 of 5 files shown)


tests/test_validate_tool.py
```diff
@@ -0,0 +1,52 @@
...
+import json
+import jsonschema
+import pytest
+
+from scripts import validate_config
+
+uwtools_file_base = os.path.join(os.path.dirname(__file__))
+
+
+def test_config_no_errors(): #pylint: disable=unused-variable
```
WeirAE, line 19: Do we need a test that reads a json as well?

WeirAE (commented):
Looks great so far, just one question on the tests and the usual git sorting

schema/workflow.jsonschema
```diff
@@ -0,0 +1,92 @@
+{
```
christinaholtNOAA, line 1: I think keeping this in JSON is consistent with a concern that @maddenp-noaa brought up last week. 👍

scripts/validate_config.py
```diff
@@ -0,0 +1,136 @@
...
+    # Create a list of fields that could contain a file or path
+    path_list = []
+    for field in schema["properties"]:
+        for value in schema["properties"][field]["properties"]:
+            if "format" in schema["properties"][field]["properties"][value]:
+                path_list.append(value)
+
+    # Check for existence of those files or paths
+    for field in config_obj.data:
+        for value in config_obj.data[field]:
```
christinaholtNOAA, line 115: These two lines could clean up to:
```
for field, value in config_obj.data.items():
   ...
```

tests/fixtures/bad.yaml
```diff
@@ -0,0 +1,10 @@
+platform:
+  WORKFLOW_MANAGER: 20
+  NCORES_PER_NODE: forty
+  SCHED: slurm
+  CCPA_OBS_DIR: /scratch1/NCEPDEV/nems/role.epic/UFS_SRW_data/develop/obs_data/ccpa/proc
+  MRMS_OBS_DIR: /scratch1/NCEPDEV/nems/role.epic/UFS_SRW_data/develop/obs_data/mrms/proc
```
christinaholtNOAA, line 6: I'd suggest using `scratchy` here and below so that we could make sure this is bad on all our platforms. This might exist on Hera, for example.

christinaholtNOAA (dismissed):
LGTM!

I left a few small comments below as we'd discussed in person.

schema/workflow.jsonschema
maddenp-noaa: Not for this PR, but we might like to adopt a tool like `jq` to format any JSON documents we're maintaining, for consistency and to avoid any formatting nitpicking that might otherwise arise. (Running JSON through `jq` is also a good sanity check for basic correctness.)

scripts/validate_config.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Load the json validation schema
+    with open(user_args.validation_schema, 'r',
+              encoding="utf-8") as schema_file:
+        schema = json.load(schema_file)
+
+    schema_error = 0
+
+    # Validate the config file against the schema file
+    validator = jsonschema.Draft7Validator(schema)
```
maddenp-noaa, line 96: I was playing with `jsonschema` briefly last week to learn a little about it and found that I had to use one of these validator subclasses (like `Draft7Validator` as you have here), but I didn't understand which class to use for what purpose. Did you get any understanding of that that you can share either here, or in a code comment in case others might wonder later?

schema/workflow.jsonschema
venitahagerty: Excellent!

tests/fixtures/bad.yaml
maddenp-noaa: @christinaholtNOAA What is `scratchy`?

scripts/validate_config.py
```diff
@@ -0,0 +1,136 @@
...
+            if value in path_list:
+                if not os.path.exists(config_obj.data[field][value]):
+                    schema_error += 1
+                    logging.error('%s has Invalid Path %s',
+                                  value, config_obj.data[field][value])
+
+    if schema_error > 0:
+        sys.exit(f'This configuration file has {str(schema_error)} errors')
+    else:
+        sys.exit(0)
```
maddenp-noaa, line 125: I think the `str()` wrapper is unnecessary (f-strings will do the `int` -> `str` conversion).

I think it would be better if this function returned a value and left the exiting to the caller, as its stated responsibility is to validate a config, not to terminate the program. IMO it should only validate the config and return a `bool` indicating whether or not it was valid. It should not set up logging: The caller -- or some ancestor -- should do that.

It might be useful to even split this function into two: One that validates the config vs the schema, and one that validates the path values. I can imagine that other correctness checks might be needed later, and the obvious path given the current code will be to add more complexity to this function. Replacing this function with two smaller, single-responsibility functions would suggest a cleaner path for maintainers / extenders to follow.

But I know that @venitahagerty needs to be done with this, so this is probably a suggestion for later work.

maddenp-noaa (dismissed):

scripts/validate_config.py
venitahagerty: The "field, value" code returns the key and the value, which is another dictionary. The second level dictionary also needs to be iterated through. In the original code, the first for loop is iterating through the YAML sections, and the second for loop is iterating through each field in a section looking for paths and files.

tests/fixtures/bad.yaml
venitahagerty: The name "scratchy" is a planned typo - the name of a disk which does not exist on any of the HPC platforms.

maddenp-noaa: Oh, `scratchy` instead of e.g. `scratch1`. I like it.

scripts/validate_config.py
venitahagerty: The python program validate_config.py is a command line script that will be part of a workflow.  It needs to provide a return code that can be tested to see whether the workflow should terminate or continue.

I'll remove the `str()` wrapper.

tests/test_validate_tool.py
venitahagerty: The validation schema will be in json, and the config file will be in YAML.

scripts/validate_config.py
maddenp-noaa: I should probably add that I do realize that this code is following existing idioms, so my comments are really about the idioms, not about this specific instance.

tests/test_validate_tool.py
WeirAE: Ah, sorry, I mean a YAML schema, but we should probably just update the ticket to only support a json schema

scripts/validate_config.py
maddenp-noaa: Understood, but how about e.g.
```
try:
    if not validate_config(sys.argv[1:], cli_log):
        sys.exit(1)
...
```
? Then CLI interface -- getting/parsing command-line arguments, setting exit status -- lives in the outermost scope. I'm not suggesting that the _script_ shouldn't be terminating the program (it must!) but that a function responsible for validating a config shouldn't.

scripts/validate_config.py
maddenp-noaa: I know what @christinaholtNOAA meant, but see that @venitahagerty is right. Since the outermost keys are not actually needed (except to get their values), an option might be
```
for section in config_obj.data.values():
    for key, value in section.items():
        ...
```
e.g.
```
>>> s = """
... fruit-colors:
...   apple: red
...   banana: yellow
...   pear: green
... pet-names:
...   dog: spot
...   parrot: polly
... """
>>> d = yaml.load(s, Loader=yaml.Loader)
>>> for section in d.values():
...     for key, value in section.items():
...         if value == "yellow":
...             print(f"{key} is yellow")
... 
banana is yellow
```
I'm not sure if `section`, `key`, and `value` are good names here, or if I've got the structure of the input data right, but this potentially eliminates some repetitive `[]` `dict` lookups.

This brings up a question for me though: Does our input data currently have only two levels of nesting, and is that something we think we can maintain in the future?

scripts/validate_config.py
venitahagerty: I see. Exit not in validate_config(), but in main. I think your idea of returning a boolean and testing for it in main is good.

maddenp-noaa: Just wanted to make sure I wasn't missing anything for when I bring it up again with the people who are still working on this codebase after this week :)

scripts/validate_config.py
venitahagerty: Christina and I discussed that, and she said there would only be 2 levels of nesting.

tests/test_validate_tool.py
venitahagerty: I just added a comment to the ticket that the validation schema will be in JSON.

scripts/validate_config.py
christinaholtNOAA: Ahh, yes. For these machine files, we do have a limitation that this is as deep as it can go. I think we will want to delve into recursion at some point since that this limited depth is only a requirement for SR , but it is likely out of scope of this ticket. We have a lot of additional schema validator work on the backlog already.


Also, I totally missed the boat on the nested loops above. Thanks to both of you for the clarification.

christinaholtNOAA (approved):
Still looks good.

scripts/validate_config.py
maddenp-noaa: I do think using `.values()` and `.items()` is cleaner and that we should update this code later to use `section` and `value` in place of `config_obj.data[field]` and (2 copies of) `config_obj.data[field][value]`. Later...

Issue #261, opened 2023-07-06
UW-8 Add a forecast method that stages files from known locations on disk.


NaureenBharwaniNOAA:
**Description**

The script needs to stage a list of files that are configuration specific (i.e., based on model resolution, regional vs global, ICS/LBCS model, etc).

This method should know what to do with static data files and cycle-specific data files. 

This method will likely need to be generic since most UFS components follow this "stage static and cycle-specific data" pattern. Perhaps the forecast-specific part is concerned with which information may need to be formatted – for example, forecast lead time, resolution, etc.

**Describe alternatives you've considered**

**Acceptance Criteria (Definition of Done)**

Scenario: I receive a set of parameters and need to make sure a set of files appears in the run directory

GIVEN: All files are assumed to be available on local disk. The default file names are set for required files.
WHEN: A properly structured configuration object is available to the forecast class
THEN: It will stage a set of files necessary to run the forecast as part of its run method.
 
GIVEN: The method is run
WHEN: A file is staged successfully
THEN: A log info message on where the files came from and where they were put is available.

GIVEN: A required file is unavailable at it configured location
WHEN: All files have attempted
THEN: The a status report of available/unavailable files will be logged, and the method will raise an exception.

**Additional context**

christinaholtNOAA:
Fixed by PR #262

PR #262, opened 2023-07-06
UW-8 Add a forecast method that stages files from known locations on disk. 

elcarpenter:
**Description**
The script needs to stage a list of files that are configuration specific (i.e., based on model resolution, regional vs global, ICS/LBCS model, etc).
This method should know what to do with static data files and cycle-specific data files.
This method will likely need to be generic since most UFS components follow this "stage static and cycle-specific data" pattern. Perhaps the forecast-specific part is concerned with which information may need to be formatted – for example, forecast lead time, resolution, etc.

For issue [#261](https://github.com/ufs-community/workflow-tools/issues/261).

**Current issues** 
Tests are passing right now, but they also pass when line 186 is commented out: `forecast_obj.stage_fix_files(run_directory, test_cfg)`. We are stuck on how to get this working and where exactly the issue lies. Currently we are only trying to test `stage_fix_files()` and have not yet done any edge case testing. We would appreciate any advice or guidance. 

**Type of change**

**How Has This Been Tested?**

**Checklist**


tests/drivers/test_forecast.py
```diff
@@ -165,3 +166,27 @@ def test_create_directory_structure():
...
+        files_to_stage = test_cfg.get('static', {})
+
+        for target, file_path in files_to_stage.items():
+            pathlib.Path(file_path).touch()
+
+        forecast_obj.stage_fix_files(run_directory, test_cfg)
+
+        for target, file_path in files_to_stage.items():
+            assert os.path.isfile(file_path)
+            assert os.path.isfile(target)
```
christinaholtNOAA, line 190: Is this line checking the current directory for these links instead of the `run_directory`? Maybe the links exist there from an earlier iteration on your system?

WeirAE (approved):
Ok, nice fix! Make sure there's a second reviewer just in case there's a path validation issue I missed, but otherwise LGTM!

src/uwtools/drivers/forecast.py
```diff
@@ -72,12 +74,23 @@ def create_model_config(self, config_obj, outconfig_file, base_file=None):
...
         ''' Holds the knowledge for how to modify a list of fix files and
-        stages them in the working directory. Likely gets all its info from
-        config_obj. Calls data mover tool (could be python copy). Fix files
-        usually are specific to a given named grid and resolution. '''
+        stages them in the working directory. Calls data mover tool (could
+        be python copy). Fix files usually are specific to a given named
+        grid and resolution. '''
+
+        files_to_stage = config_obj.get('static', {})
+        for target, file_path in files_to_stage.items():
```
maddenp-noaa, line 85: Would `source` and `destination` (or the more concise `src` and `dst`) be applicable here, and possibly better describe the values? It might help to distinguish full paths vs just filenames with variable names including the idiomatic `path` and `fn` identifiers; I'm guessing that `config_obj["static"]` maps filenames to pull paths -- is that right? If so, something like
```
for dst_fn, src_path in files_to_stage.items():
    if os.path.isfile(src_path):
        dst_path = os.path.join(run_directory, dst_fn)
        shutil.coypyfile(src_path, dst_path)
        ...
```
tell the story more clearly? (`file_path` got the `path` part right, but there are two files here, so the `file` part is not very descriptive.)

Note that Python itself prefers and uses `src` and `dst`:
```
>>> help(shutil.copyfile)
Help on function copyfile in module shutil:

copyfile(src, dst, *, follow_symlinks=True)
    Copy data from src to dst in the most efficient way possible.
```

src/uwtools/drivers/forecast.py
```diff
@@ -72,12 +74,23 @@ def create_model_config(self, config_obj, outconfig_file, base_file=None):
...
+            if os.path.isfile(file_path):
+                file = os.path.join(run_directory, target)
+                shutil.copyfile(file_path, file)
+                msg = f"File {file_path} staged in working directory at {target}"
```
maddenp-noaa, line 89: Here, too, is it a "working" directory, or a "run" directory. IMO we should use consistent terminology.

src/uwtools/drivers/forecast.py
```diff
@@ -72,12 +74,23 @@ def create_model_config(self, config_obj, outconfig_file, base_file=None):
...
+                logging.info(msg)
+            else:
+                msg = f"File path {file_path} not found"
+                raise RuntimeError(msg)
```
maddenp-noaa, line 93: What does `shutil.copyfile()` do if the source file doesn't exist? Are we really improving on the default behavior that Python would provide if we just boldly invoked `copyfile()` without checking the source file first.

What if `run_directory()` doesn't exist?

src/uwtools/drivers/forecast.py
```diff
@@ -72,12 +74,23 @@ def create_model_config(self, config_obj, outconfig_file, base_file=None):
         msg = f"Config file {outconfig_file} created"
         logging.info(msg)
 
-    def stage_fix_files(self):
+    def stage_fix_files(self, run_directory, config_obj):
```
maddenp-noaa, line 77: Instead of passing in the whole `config_obj`, then pulling the `static` `dict` that we need out of it, could we require the caller to pass in the `dict` that we need? If the caller _has_ the big config object, **they** could pass in `config_obj.get("static", {})`. Or, if they have a mapping of source and destination files they want copied -- wherever they got it -- the could pass that in.

Why should this function know that some dictionary has a `static` key? I think we should start moving away from operating on these large, complex `dict` objects and use specific bits of them as needed.

This should make testing easier, and means that we could write a type-annotated function signature for this function like
```
def stage_files(self, run_directory: str, file_map: Dict[str, str]) -> None:
```
That is, a function that takes a `str` run directory, and a `dict` mapping source to destination (or is it the other way around?) files, where both the keys and values are `str`s.

I left `fix` out of the function name in my example due to my `fix` vs `static` question, below, but also because: Does what this function is doing have anything to do with the files being static? That strikes me as an irrelevant detail here; it might be useful to have in a comment at the call site, like
```
# Stage static files into the run directory:
fv3obj.stage_files(run_directory=rundir, file_map=config.obj.get("static", {}))
```
Could we potentially also use this function to also stage cycle-specific files in the run directory?

tests/drivers/test_forecast.py
```diff
@@ -165,3 +166,29 @@ def test_create_directory_structure():
...
+        test_yaml = os.path.join(uwtools_file_base,pathlib.Path("../fixtures/expt_dir.yaml"))
+        test_cfg = config.YAMLConfig(test_yaml)
+        forecast_obj = FV3Forecast()
+
+        forecast_obj.create_directory_structure(run_directory)
+
+        files_to_stage = test_cfg.get('static', {})
+
+        for target, file_path in files_to_stage.items():
+            pathlib.Path(file_path).touch()
```
maddenp-noaa, line 185: I didn't know about `Path.touch()`! Thanks, I need this in a bunch of places!

tests/drivers/test_forecast.py
```diff
@@ -165,3 +166,29 @@ def test_create_directory_structure():
             forecast_obj.create_directory_structure(run_directory, "quit")
         assert pytest_wrapped_e.type == SystemExit
         assert pytest_wrapped_e.value.code == 1
+
+def test_create_forecast_obj():
```
maddenp-noaa, line 170: Nice test.

tests/drivers/test_forecast.py
```diff
@@ -165,3 +166,29 @@ def test_create_directory_structure():
...
+            pathlib.Path(file_path).touch()
+
+        forecast_obj.stage_fix_files(run_directory, test_cfg)
+
+        for target, file_path in files_to_stage.items():
+            assert os.path.isfile(file_path)
+            file = os.path.join(run_directory, target)
+            assert os.path.isfile(file)
+            assert filecmp.cmp(file_path, file)
+            os.remove(file_path)
```
maddenp-noaa, line 194: I'd suggest adopting `src` / `dst` and `fn` / `path` terminology here, too. Differentiating between filenames, paths, and actual file objects (like what `open()` returns) is a perennial issue in Python codes. Adopting a convention to make it clear which kind of thing you're dealing with is helpful. (Type annotations help distinguish file objects from the other two, but filenames and paths are both `str` objects, so it doesn't help there.)

maddenp-noaa (changes requested):
Hey @elcarpenter -- Some of the suggestions I've made are truly specific to this PR, but some are more general and I was probably going to make them on the next PR I saw, which just happens to be this one (lucky you). I hope the difference is obvious. I'd be grateful for discussion on any of them, but especially hope that others may comment on the more general ones if they have opinions.

src/uwtools/drivers/forecast.py
```diff
@@ -72,12 +74,23 @@ def create_model_config(self, config_obj, outconfig_file, base_file=None):
...
 
         ''' Holds the knowledge for how to modify a list of fix files and
-        stages them in the working directory. Likely gets all its info from
-        config_obj. Calls data mover tool (could be python copy). Fix files
-        usually are specific to a given named grid and resolution. '''
+        stages them in the working directory. Calls data mover tool (could
+        be python copy). Fix files usually are specific to a given named
+        grid and resolution. '''
+
+        files_to_stage = config_obj.get('static', {})
```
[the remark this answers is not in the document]
elcarpenter: The ticket description seems to imply "static" is standard within UFS, so I will stick with that terminology.

src/uwtools/drivers/forecast.py
elcarpenter: I added a check for `run_directory`, but I'm not sure what other checks would be necessary for checking the source file beyond line 89 `if os.path.isfile(src_path):`. Was there something specific you had in mind?

src/uwtools/drivers/forecast.py
elcarpenter: In conversations with @christinaholtNOAA yesterday, we decided not to include cycle-specific files for now. I do agree, though, that this could be done within the same function.

elcarpenter: I attempted to add the type-annotated function signature, but it kept failing the linter, so I reverted the changes.

src/uwtools/drivers/forecast.py
maddenp-noaa: Here's what Python does on its own when the source file does not exist:
```
>>> shutil.copyfile("/does/not/exist", "/cannot/write/here")
  File "/home/maddenp/sw/conda/lib/python3.10/shutil.py", line 254, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '/does/not/exist'
```
I think a `FileNotFoundError` better describes what went wrong better than the generic `RuntimeError` we'd be raising. So is it worth interfering with Python's default behavior? I don't think so.

Going further, other things can go wrong, like e.g. a. non-existent destination directory, or a permissions problem, e.g.
```
>>> shutil.copyfile("/etc/passwd", "/passwd")
  File "/home/maddenp/sw/conda/lib/python3.10/shutil.py", line 256, in copyfile
    with open(dst, 'wb') as fdst:
PermissionError: [Errno 13] Permission denied: '/passwd'
```
Here, too, Python does a great job, which I'm not sure we could improve on. And why manually handle just one of (at least) three obvious potential failure cases? So I would suggest just letting Python do its thing here and removing the check.

An added benefit: We _should_ also have a unit test that exercises the failure case and the `else` clause, but if we remove the check, there's nothing to test!

src/uwtools/drivers/forecast.py
maddenp-noaa: Thanks for trying the type signature. I'm planning to give a short tutorial on type annotation sometime in the near future, which I hope will help going forward, but here, specifically, we'd need to `from typing import Dict`, which is what I guess `pylint` complained about. (Quick type geekery: We'd have to annotate the current big config object as `Dict[Any, Any]` (also importing `Any` from `typing`), i.e. `config_obj: Dict[Any, Any]`, which is the same as just `config_obj: dict`, which is saying close to, though not quite, nothing. At least we'd know it's a dictionary, but we'd know nothing about what to expect as content.)

But we'd also need to change type of the argument being passed in, extracting the file mapping from the larger `dict` at the call site. Though I think that's worth doing now, to set a precedent that we can build on, I won't insist. But I think we should talk about whether that's the direction we should go. As a general design principle, I'll assert that functions should not be given more to work with than what they need. And they shouldn't need to have knowledge of things they're not responsible for. The first principle means limits the harm they can do: There's no reason this function should have access to other parts of the config `dict`, but now it could access, and even change, _everything_. The second principle insulates the code from changes it shouldn't have to care about. If we decided to change the key `static` to something else, this code would have to be updated to track that change.

Do we disagree about those ideas? If not, then the question isn't if we should head in that direction, but when. I'm fine with that being "later", but the longer we wait the more debt we accrue.

maddenp-noaa: Roger on not dealing with cycle files for now. We could generalize this function's name later if we found it useful for other purposes. As long as we address the `fix` vs `static` question, I'm happy with a specific (and accurate) function name.

src/uwtools/drivers/forecast.py
maddenp-noaa: Oh, I see now that you added a directory check, too -- I'm sorry, I should not have asked a leading question but rather just given a straightforward suggestion to remove the checks. Apologies for wasting your time if it ends up going that way...

maddenp-noaa (approved):
Nice work, thanks for addressing all those suggestions.

elcarpenter:
@maddenp-noaa I fixed one last linter issue, but it should be good now if you're willing to approve again.

maddenp-noaa (approved):

tests/drivers/test_forecast.py
christinaholtNOAA: Same here. Nice. :)

PR #265, opened 2023-07-18
UW-9

NaureenBharwaniNOAA:
**Description**
Different platforms support different parallel run commands – srun, mpirun, etc. These may or may not need to know additional parameters like total number of processes, etc. The run script should have zero decision points about named platform dependencies.
This will come in as a (likely generic) method that given the correct config settings for the platform and task, will know what run command to perform for the given task.
The forecast method will know that this will need to be a parallel executable, and will know which config params to communicate for the platform and resource settings.

**Type of change**

**How Has This Been Tested?**
Passes all development-shell `make format && make test` code-quality checks, including (updated) unit tests.

**Checklist**

Co-authored-by: Emily Carpenter `<emily.l.carpenter@noaa.gov>`


src/uwtools/drivers/forecast.py
```diff
@@ -162,12 +162,28 @@ def job_card(self):
...
+        node_arg: Optional[str] = None,
+        pe_member01: Optional[int] = None,
+        num_of_procs: Optional[int] = None,
+        oversubscribe: Optional[str] = None,
+        procs_per_nodes: Optional[str] = None,
+        ppn: Optional[int] = None,
+        cpu_bind: Optional[str] = None,
+        core: Optional[str] = None,
+        depth: Optional[str] = None,
+        omp_num_threads: Optional[int] = None,
```
maddenp-noaa, line 179: Since these arguments (all but `self` and `exec_name`) are all optional, can we sort them since their order is irrelevant?

src/uwtools/drivers/forecast.py
```diff
@@ -162,12 +162,28 @@ def job_card(self):
...
+    ) -> str:
+        """
+        Builds the forecast executable with the user provided run commands.
+
+        This will build the executable.
```
maddenp-noaa, line 184: I think that "build an executable" will be thought by many developers to mean "compile source code into an executable program" and will be surprising to see in this context. I think something like `Constructs a command to be used to run the forecast executable` might describe this better.

src/uwtools/drivers/forecast.py
```diff
@@ -162,12 +162,28 @@ def job_card(self):
...
+        """
+        return (" ".join(str(i) if i is not None else "" for i in vars().values())).strip()
```
maddenp-noaa: Nice use of `vars()` here!

The `else ""` part will leave anomalous spaces in the constructed command, as `""` will still be in the sequence of `str(i)` values. For example:
```
>>> def foo(a, b=None, c=None):
...   return (" ".join(str(i) if i is not None else "" for i in vars().values())).strip()
... 
>>> foo("hello", b="world")
'hello world'
>>> foo("hello", c="world")
'hello  world'
```
Note the _two_ spaces between `hello` and `world` in the second call's return value, since `b` was converted from `None` to `""` and is still there. Better, IMO, we could filter out the `None` values completely:
```
>>> def foo(a, b=None, c=None):
...   return (" ".join(str(i) for i in filter(None, vars().values()))).strip()
... 
>>> foo("hello", b="world")
'hello world'
>>> foo("hello", c="world")
'hello world'
```
(Note that this use of `None` with `filter` is a weird edge case that's discussed in the Python docs. If you're not familiar with it, it probably isn't doing what you think it's doing, so check that out if `filter` surprises you later.)

maddenp-noaa: I wonder if `str(i).strip()` might be nice instead of the trailing `.strip()` -- if all the individual components are first stripped of leading/trailing whitespace, does that take care of the final string, too? If so, that would also clean up any unintentional leading/trailing whitespace that users might pass as part of values.

maddenp-noaa: A quick search didn't turn up for me any docs saying that `vars()` *must* be ordered the same way as the arguments are given in the function, but it seems to be the case and is probably reliable.

Note that `vars()` will also pick up any locally-defined variables:
```
>>> def foo(a, b=None, c=None):
...   d = 88
...   return (" ".join(str(i).strip() for i in filter(None, vars().values())))
... 
>>> foo("hello", c="world")
'hello world 88'
```
So beware, but I don't think we need to change anything now.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -167,6 +167,22 @@ def test_create_namelist_without_base_file(create_namelist_assets):
...
+    hera_expected = "srun --export=ALL"
+    assert hera_expected == FV3Forecast.run("srun", "--export=ALL")
+
+    cheyenne_expected = "mpirun -np 4"
+    assert cheyenne_expected == FV3Forecast.run("mpirun", "-np", 4)
+
+    wcoss2_expected = "mpiexec -n 4 -ppn 8 --cpu-bind core -depth 2"
+    assert wcoss2_expected == FV3Forecast.run(
+        "mpiexec", "-n", 4, "-ppn", 8, "--cpu-bind", "core", "-depth", 2
+    )
```
maddenp-noaa, line 183: Do these tests work? `run()` isn't a static or class method, so I don't see how it can be called without first constructing an `FV3Forecast` instance. Maybe I'm missing something...

src/uwtools/drivers/forecast.py
elcarpenterNOAA: Is the order of the arguments always irrelevant in the run command? We weren't sure since all the examples from the various platforms all seemed to follow the same order, which is why we have ordered them in this way here.

maddenp-noaa: I'm fairly certain that any `-` or `--` switches are independent of order. Maybe the executable name needs to come last, but I'm not sure even that's the case. Definitely `mpirun` / `mpiexec` / `srun` has to come first.

Speaking of the executable name, did we decide to delegate that to the caller to tack on to the end of the command? I think so...

src/uwtools/drivers/forecast.py
NaureenBharwaniNOAA: We were able to update how we were contructing the command and were able to remove the else part that left the anomalous spaces and the strip method. It should look clearer now

NaureenBharwaniNOAA: We removed strip altogether and cleaned up the leading/trailing whitespaces.

maddenp-noaa: Looks good, thanks.

src/uwtools/drivers/forecast.py
NaureenBharwaniNOAA, line 179: @christinaholtNOAA How would you like us to downsize these arguments? Here we were considering combining `node_arg: Optional[str] = None` and `pe_member01: Optional[int] = None`  as a dictionary to limit the number of arguments. Same can be done for `procs_per_nodes: Optional[str] = None` and `ppn: Optional[int] = None`, and others as well.

christinaholtNOAA: I don't think we'll have cases where the order matters outside the command name (mpirun / mpiexec / srun) first and the executable last. Both should be required arguments. I think it only makes sense to have this method construct the string with the executable name included.

christinaholtNOAA: I think it might make sense to allows all the optional flags as keyword arguments. Perhaps it also makes sense to allow for a `*args` argument for any flags that might need to be added that will not have a value. Perhaps that boils down to something like this:

```
    def run(
        self,
        run_cmd: str,
        exec_name: str,
        *args,
        **kwargs,
        ) -> str:

```

Then, the return string might be built from any number of generic arguments provided by the caller.

src/uwtools/drivers/forecast.py
maddenp-noaa: That comment was just musing about possible issues that don't seem to currently be in play, so I'll resolve this conversation to keep this PR tidy.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -167,6 +168,22 @@ def test_create_namelist_without_base_file(create_namelist_assets):
...
+    hera_expected = "srun --export=ALL test_exec.py"
+    assert hera_expected == FV3Forecast().run("srun", "test_exec.py", "--export=ALL")
+
+    cheyenne_expected = "mpirun -np 4 test_exec.py"
+    assert cheyenne_expected == FV3Forecast().run("mpirun", "test_exec.py", "-np", 4)
+
+    wcoss2_expected = "mpiexec -n 4 -ppn 8 --cpu-bind core -depth 2 test_exec.py"
+    assert wcoss2_expected == FV3Forecast().run(
+        "mpiexec", "test_exec.py", "-n", 4, "-ppn", 8, "--cpu-bind", "core", "-depth", 2
+    )
```
NaureenBharwaniNOAA, line 184: After working through implementing our tests, one question that came up was how would address if the user did not provide the executable name. In the case of our tests it would just accept the next arg as the placeholder for the exec_name and would falsely pass the test cases.

src/uwtools/drivers/forecast.py
```diff
@@ -162,12 +162,19 @@ def job_card(self):
...
+        exec_name: str,
+        *args,
+        **kwargs,
+    ) -> str:
+        """
+        Constructs a command to be used to run the forecast executable.
+        """
+        args_str = " ".join(str(i) for i in args if i is not None)
+        kwargs_str = " ".join(str(i) for i in kwargs if i is not None)
+        return f"{run_cmd} {args_str} {kwargs_str}{exec_name}"
```
maddenp-noaa, line 177: I've got some misgivings about this -- not at all due to the current PR or implementation and certainly not due to the developers, but more generally. Two minor and more maybe more major concern:
1. This `run()` method isn't specified in the `Driver` parent class. It's _mentioned_ there, but not defined as an abstract method that must be implemented in subclasses. Unless there's a use case where we have a `Driver` that cannot be `run()`, I think it should be mandated by `Driver`.
2. The current implementation doesn't actually "run" anything; it only constructs a command that could be run. Should we maybe leave `run()` un-implemented here and create a helper method like `run_cmd()` or something that that contains the logic here, and that could then be _called by_ `run()` or by any other method that wants the run command (e.g. to place into a batch script)?
3. The hard one, and an honest question: What utility are we providing here? The caller is going to have to have such significant knowledge of the run command to be constructed -- every single character of it, in fact -- that they might as well construct the command themselves. Only slightly exaggerating, why don't they just pass us `mpirun -np 256 ufs_weather_model` and we pass it right back to them? It seems we're jumping through unnecessary hoops, as all we're doing is string concatenation. Now, if we insisted on keyword-only arguments and forced the caller to structure their call that way, by naming each argument, that would at least enforce some rigor by making the caller fill in a kind of template with some constraints. This is a hard problem because it leads to the question of how `uwtools` can add value by adding structure, in the face of so much variability in run environments. But I think it's worth thinking/talking about because it's going to come up repeatedly in other `Driver` implementations.

christinaholtNOAA: 1. Yes.
2. Yes.
3. I think this is the part that got lost somewhere in the conversation of whose responsibility it is to take the larger config object and gather the necessary information to build this command. This method doesn't offer too much utility if we're putting that on the caller, but if the method has access to say `self.config_obj` and knows where to find that information instead of it being passed in, then we start to see some the original intended utility. Maybe that's not even necessary given how we can use templates in config files to reference other pieces of information.

We could potentially have a platform define a template for this section as:

```
platform:
  run_cmd_fcst: 'mpirun -np {run_fcst.pe_member01}'
```

\*Just an example.

maddenp-noaa: On 3., it's a great point that knowledge about how to spelunk through a big config dict and extract the required information would create value and justify the existence of an otherwise almost-no-op method like this. But I'm still uneasy about that because

a) I suspect that this will end up coupling multiple methods/functions to the exact structure of that config dict, such that changes to the structure of the config dict (which I fear are inevitable, though I could be wrong) will require changing code in many places, leading to brittleness. On the other hand, extracting values from the config dict in one place and then passing explicit arguments means there's a single piece of code that would have to be updated to deal with config-structure/content changes; and
b) It's not possible to say anything other than `cfgobj: dict` about such a big config object, which doesn't make typechecking useful or give someone looking at a method/function any indication of what they're dealing with. Of course, if it's the responsibility of the method accepting such a config dict to be _the single authority_ on how to access certain sections/keys of that object, then it's fair that someone implementing/maintaining that method has to deal with that ambiguity, because somebody has to! But I'm worried about duplication of that knowledge across different pieces of code. Using discrete method arguments, while admittedly delaying the inevitable fact that _some code_ has to know how to pull info from the big config dict, leaves open the possibility that that can be done in one, authoritative place. This feels right to me, but it's difficult to justify. My gut feeling is that a big config dict is a bad interface that will lead to complexity, confusion, and britleness, and that we should extract more targeted, specific pieces of data as early in the call graph as possible. Intuition isn't rigor, but it's the best argument I've got at the moment, and could be wrong.

That said...

- I agree that having the config provide a template for the MPI executable could be a good way forward. That's not far from my joke about forcing the user to give us a string that we'll just pass back to them, except that we'd add value by e.g. interpolating in the process count, which must vary.
- Another thing we could do, here or elsewhere (or not), would be to define our own containers -- e.g. either `dataclasses` (https://docs.python.org/3/library/dataclasses.html) or `namedtuple` objects -- to encapsulate bits of data from a big config obj for specific purposes. Then there could be a single place in `uwtools` where such a container instances would be created and populated from info in the big config dict (and possibly computed values). Then, the (explicitly typed) container instance would be passed on into the call graph. I haven't seen a full YAML-originated config dict in the wild yet so can't offer an illustration (so this might not be a useful idea), but it could provide a layer of insulation between the world of YAML configs and more specific Python data structures used internally. For IDE users, being able to type `x.` and seeing a list of attributes on instance `x` is useful, and which using data classes or named tuples would support, while you're on your own when unpacking a `dict` with `.[]` lookups. Good API documentation could also be generated for such container objects. I think the only alternative for understanding a big config dict would be the JSON Schema doc, which might not be great for humans to digest.

I'm just thinking out loud on the last part about `dataclasses` / `namedtuple` containers.

maddenp-noaa: Everything I said above could be premature optimization. Maybe the MVP way is to just pass `cfgobj: dict` and refactor when we learn more. I think we just need to take care to fall back and redesign as soon as we start feeling pain / knowing more and not wait until it's so painful to refactor that we avoid doing it.

christinaholtNOAA: I think these thoughts are great to have documented! It may make sense for us to capture them in some Confluence documentation so we have a shot at finding them again. :)

One thing that we might be forgetting here is that type-checking is not the only trick we have up our sleeves. We are also working on the schema validation tools for exactly this purpose. Calling that to make sure that the config object meets the required structure and content has been a part of this design I haven't been overly vocal about lately. I'm hoping that thing will help us keep a strict, well-defined interface for these config objects.

We might look into how a tool like AWS's Parallel Cluster handles their configuration layer, which is also all in YAML and nicely defined to do A LOT of stuff. @fgabelmannjr has referred to this thing as a rules engine in the past, but I think perhaps we can keep our rules in the JSON Schema and use that tool to do the same thing.

src/uwtools/tests/drivers/test_forecast.py
NaureenBharwaniNOAA: I updated the tests so now they should be constructing an instance of `FV3Forecast` similar to other methods within the class.

src/uwtools/drivers/forecast.py
```diff
@@ -162,11 +162,19 @@ def job_card(self):
         Turns the resources config object into a batch card for the configured Task.
         """
 
+    def run_cmd(self, run_cmd: str, *args, exec_name: str, **kwargs) -> str:
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: It seems that there is a census that removing **kwargs would be the best option due to the unpredictably of the args needing 1 or 2 dashes. 

To my understanding, I think it would be best to remove **kwargs to not overcomplicate what arguments need to be passed in with keywords and just leave it at *args. This way the only keyword argument we have is exec_name.

Speaking of the only keyword argument being exec_name, I realize that the other required argument is run_cmd which currently if a user did not supply the run_cmd the tests would still pass. Just confirming that we have something more like:
`FV3Forecast().run_cmd("--export=ALL" run_cmd="mpiexec", exec_name="test_exec.py")`
Adding run_cmd towards the end due to the keywords positional requirement. Then the concatenation of the run command will look like `mpiexec --export=ALL test_exec.py`

If everyone is ok with removing **kwargs I will push my new changes and get some more feedback on these updates.

christinaholtNOAA (approved):
LGTM!

WeirAE (approved):
Second! LGTM

PR #266, opened 2023-07-19
Remove realpath, pin condev version in README.md

maddenp-noaa:
Two trivial updates:

1. Replace calls to `realpath`, which I have discovered is not available by default under OSX (but is under GNU/Linux) with more portable calls. This is in response to a couple OSX users on our team experiencing difficulties, which they solved by doing `homebrew install coreutils`, an easy workaround it would nevertheless be better to avoid having to rely on.
2. Pin `condev` (once in an executable-code context, once in a documentation context) to a specific version known to be compatible with the current `uwtools` codebase, to insulate the latter from unexpected upstream changes.

I'll happily accept a quick LGTM review from anyone.


WeirAE (approved):
Good catch, LGTM

Issue #267, opened 2023-07-20
UW-268: create validation tool for the drivers' required configuration parameters


WeirAE:
**Description**
There is a link between the global environment variables set at runtime in the Rocoto XML job definition and the j-job (that i runs) that is not validated until runtime (using set -eux in bash). Create a tool that helps with this in the new way/

**Requirements**
Tool to be included will interrogate a driver to determine whether required sections and settings are correct/consistent
ex: needs to know node size, queues, length, grid, suites, locations
Mock with reference schema

**Acceptance Criteria (Definition of Done)**
Tool exists in workflow-tools (need not have a user interface)
Tests included as needed
sample config and schema have been included in the repo

WeirAE:
closed by #268

PR #268, opened 2023-07-20
UW-268: create validation tool for the drivers' required configuration parameters

WeirAE:
**Description**

given: a model_config file for an experiment
when: the driver is interrogated for consistency with the validation tool
then: the tool returns results of required sections and settings

Jira issue UW-268
Completes issue #267 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**


WeirAE:
Opened as draft while fussing with a test reference schema that is reasonably bounded but also reflects the expected behavior sufficiently

src/uwtools/drivers/experiment.py
```diff
@@ -36,6 +36,9 @@ def validate_config(self, config_file):
         """
         Validate the configuration file.
         """
+        config_validation = getattr(import_module(self.modname), "validate_config")  # pragma: no cover
```
maddenp-noaa, line 39: I saw that CI failed (https://github.com/ufs-community/workflow-tools/actions/runs/5616367277/job/15218559714?pr=268) due to this line being too long. `make format` will take care of that for you, along with most other formatting concerns. Always a useful thing to run (and any potential changes it makes committed) before a `git push`. My habit is to do `make format && make test` before a push (and pretty often anyway).

WeirAE: Thanks - I 100% forgot to do that, trying to see if I can add to my VS Code tutorial with support for this but its a little fussy.

WeirAE:
Several issues have been identified and are in progress!
1) import_module fix: import_module("uwtools.apps.uw").UWforSRW
2) validation focus move to forecast.py (but will keep here)
3) additional logging necessary and in progress
4) validation moving to parent Driver from its subclass
5) Additional constructors to be updated in a later PR

maddenp-noaa:
@WeirAE I'm looking at the error you're getting. I have an idea about what's going on, more soon...

maddenp-noaa:
I checked out your branch, created a development shell, and was able to reproduce your problem. Here's a diff that fixes the problem -- discussion below.
```
(DEV-uwtools) ~/git/uwtools % git diff -U0
diff --git a/src/uwtools/drivers/experiment.py b/src/uwtools/drivers/experiment.py
index 0435340..621cab7 100644
--- a/src/uwtools/drivers/experiment.py
+++ b/src/uwtools/drivers/experiment.py
@@ -6 +5,0 @@ import sys
-from importlib import import_module
@@ -7,0 +7,2 @@ from importlib import import_module
+from uwtools.apps.srw import SRW210
+from uwtools.apps.uw import UWforSRW
@@ -20,9 +21 @@ class SRWExperiment(Facade):
-        self.modname = (
-            "uwtools.apps.uw.UWforSRW" if len(sys.argv) > 1 else "uwtools.apps.srw.SRW210"
-        )
-        # Note: for alternate versions, manually set the modname to the appropriate app version
-        # see the src/uwtools/apps/ directory for options
-        # The following steps parse the modname to call later funtions appropriately
-        _func = self.modname.rsplit(".", maxsplit=1)[-1]
-        _module = self.modname.replace(f".{_func}", "")
-        self.module = getattr(import_module(_module), _func)
+        self.srwobj = UWforSRW() if len(sys.argv) > 1 else SRW210()
@@ -35,3 +28 @@ class SRWExperiment(Facade):
-
-        import_config_file = getattr(self.module, "load_config")  # pragma: no cover
-        import_config_file(config_file)
+        self.srwobj.load_config(config_file)
@@ -43,2 +34 @@ class SRWExperiment(Facade):
-        config_validation = getattr(self.module, "validate_config")
-        return config_validation(config_file)
+        return self.srwobj.validate_config(config_file)
@@ -51,3 +41 @@ class SRWExperiment(Facade):
-        # Note: UW version coming soon, replace module with self.modname when available
-        workflow_generator = getattr(import_module("uwtools.apps.srw.SRW210"), "create_experiment")
-        workflow_generator()
+        self.srwobj.create_experiment()
```
First, instead of dynamically importing the SRW class we need, we statically import it in the imports section, then create an instance of the appropriate class and assign it to `self.srwobj`. This exposes more analysis-time information to `pylint` and `mypy`, which promotes errors to the linting or typechecking phases -- rather than only seeing the failures at run-time, via the unit tests.

Then, we can make instance-method calls on `self.srwobj` directly, again without needing to do any more dynamic imports or lookups.

Can you give that a try and see how it goes?

maddenp-noaa:
I know we talked about using `getattr()` to select specific behaviors, and that may still be useful in the future, but it looks possibly premature until we know that we've got more complicated needs. At the moment, a simpler solution seems to work well, and eliminates some of the indirection that you were rightly concerned about.

WeirAE:
Significant changes added from offline feedback, summarizing here:

- `Facade` changes will require followup discussions on design and will be removed from this PR
- New schema needed for subclasses of `Driver`
- `Driver` constructor needs to be configured and validated
- `Driver` needs an abstract `schema()` `@property`
- `Driver.validate()` needs to be instance, not static
- Update `Driver` test to create a subclass and test schema

Last commit reverted `Facade` and its subclasses and tests, while adding the requested structure. The `Driver` test needs to be fixed and the `FV3Workflow.jsonschema` needs to be further updated, but progress! 
Tremendous thanks to @christinaholtNOAA and @maddenp-noaa for their feedback, support and patience!

src/uwtools/tests/drivers/test_experiment.py
```diff
@@ -4,12 +4,13 @@
...
 import pytest
-from pytest import fixture
 
 from uwtools.drivers import experiment
+from uwtools.logger import Logger
+from uwtools.tests.support import fixture_path
 
 
-@fixture
+@pytest.fixture
```
WeirAE, line 13: some leftovers here, will revert as well

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -16,6 +17,17 @@
...
 
+@pytest.mark.parametrize("vals", [("good", True), ("bad", False)])
+def test_validity(vals):
+    """
+    Tests validation on initialization when a config file is given.
+    """
+    cfgtype, boolval = vals
+    config = FV3Forecast(config_file=fixture_path(f"schema_test_{cfgtype}.yaml"))
+
+    assert config.validate() is boolval
```
maddenp-noaa, line 28: If we implement a test of `Driver.validate()` as discussed above, via a small subclass defined in `test_driver.py`, then that should be a sufficient test of `validate()` and we shouldn't need to test it in subclasses, as it's inherited as-is.

At the moment, I think the only new code to test in `FV3Forecast` is its `schema()` method. Since we intend to make the `config_file` argument to the constructor mandatory, it would probably be best to not rely on the current optional status of that argument; instead we could mock `Driver.validate()` to return `True`, then instantiate an `FV3Forecast` object passing whatever value we like as `config_file`, then assert that the value of its `schema` attribute is actually a file -- I think that'd be a decent test of `schema()`.

maddenp-noaa (changes requested):
I hope this isn't premature -- you might still be looking at the optional argument / static methods idea. But I had some time and the inclination to read this over, so here's some early feedback.

src/uwtools/drivers/forecast.py
```diff
@@ -20,10 +21,23 @@ class FV3Forecast(Driver):
...
+        super().__init__(config_file)
+        self._schema = ""
+
+    @property
+    def schema(self) -> str:
+        """
+        The schema to validate the config file against.
+        """
+        with resources.as_file(resources.files("uwtools.resources")) as path:
+            self._schema = (path / "workflow.jsonschema").as_posix()
```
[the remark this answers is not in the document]
WeirAE: Yep - I had swapped it over in testing for troubleshooting and never swapped it back it seems

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,106 @@
...
+            "description": "task for data assimilation",
+            "type": "object",
+            "properties": {
+                "OBS_SUFFIX": {
+                    "type": "string"
+                    }
+                }
+            }
+        }
+}
```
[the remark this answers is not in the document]
WeirAE: Yep, definitely a placeholder pending additional work to nail it down, thanks for the formatter reference!

src/uwtools/drivers/forecast.py
```diff
@@ -20,10 +21,23 @@ class FV3Forecast(Driver):
...
 
-    def __init__(self):
+    def __init__(
+        self, config_file: Optional[str] = None
+    ):  # pylint: disable=useless-super-delegation
         """
         Initialize the Forecast driver.
         """
+        super().__init__(config_file)
+        self._schema = ""
```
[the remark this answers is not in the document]
WeirAE: I had added the internal underscore due some recursion errors in testing, but it should work in this version

src/uwtools/drivers/forecast.py
```diff
@@ -20,10 +21,23 @@ class FV3Forecast(Driver):
     FV3 (ATM-only) mode.
     """
 
-    def __init__(self):
+    def __init__(
+        self, config_file: Optional[str] = None
+    ):  # pylint: disable=useless-super-delegation
```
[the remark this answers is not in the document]
WeirAE: Hmm, maybe I have some linter issues in my environment, I'm happy to get rid of it!

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
maddenp-noaa: To be clear, I'm suggesting that we may not need an instance variable at all, with or without an underscore. I think the `schema()` `@property` method might be all we need.

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
WeirAE: Weird, still getting the same error in my environment, I'll see if I can track that down

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
WeirAE: When I remove it, I get `Property "schema" defined in "FV3Forecast" is read-only`, I think that's where I had previously also used the `@schema.setter` to deal with it as well. Is there a clean setting that lets me bypass the `self._schema` workaround?

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
maddenp-noaa: Do you see the error when you run `make test`, or just in your editor/IDE?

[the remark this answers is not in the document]
WeirAE: yep, on `make test` - I'm hammering away at the forecast test but I'll get back to tracking it down in a bit

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
maddenp-noaa: Working from your latest pushed code, here's what works for me:
```
(DEV-uwtools) ~/git/uwtools % git diff -U0
diff --git a/src/uwtools/drivers/forecast.py b/src/uwtools/drivers/forecast.py
index 21997c2..3ee17cd 100644
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -10 +10 @@ from importlib import resources
-from typing import Dict, Optional
+from typing import Dict
@@ -24,9 +23,0 @@ class FV3Forecast(Driver):
-    def __init__(
-        self, config_file: Optional[str] = None
-    ):  # pylint: disable=useless-super-delegation
-        """
-        Initialize the Forecast driver.
-        """
-        super().__init__(config_file)
-        self._schema = ""
-
@@ -39,2 +30 @@ class FV3Forecast(Driver):
-            self._schema = (path / "workflow.jsonschema").as_posix()
-        return self._schema
+            return (path / "workflow.jsonschema").as_posix()
```
That is: Have `schema()` return its value without assigning it to anything, then get rid of the now-unnecessary `__init__()` method.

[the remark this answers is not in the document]
WeirAE: Ohhhhhhh  I don't have to override `__init__` at all to pass `@property`, thats excellent!

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -3,9 +3,31 @@
...
 
 def test_import():
     # Test stub asserting that module was imported.
     assert dir(driver)
+
+
+@pytest.mark.parametrize("cls", Driver.__subclasses__())
+@pytest.mark.parametrize("vals", [("good", True), ("bad", False)])
+def test_instance(cls, vals):
+    # test instantiation of abstract class
```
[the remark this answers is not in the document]
WeirAE: So, I believe I've gotten the structure correct, but now I'm getting `redefining name 'create_config' from outer scope` error and its not creating the temp config values correctly. I've pushed those changes for following, although its not complete and differences not yet resolved

src/uwtools/tests/drivers/test_forecast.py
WeirAE: I think I got the hang of this one!

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -3,9 +3,75 @@
...
+    """
+
+    @property
+    def schema(self) -> str:
+        return create_schema
+
+    def requirements(self):
+        """
+        requirements.
+        """
```
maddenp-noaa, line 38: A common Python idiom for do-nothing functions/methods is to just use the statement `pass` as the body, e.g.
```
def requirements(self):
    pass
```
Slightly more concise, FWIW, if you like it. Thought I'd share.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -3,9 +3,75 @@
...
+@fixture
+def create_schema(tmp_path):
+    schema_file = tmp_path / "schema_file.jsonschema"
+    with open(schema_file, "w", encoding="utf-8") as f:
+        print("undefined_filter: '{{ 34 | not_a_filter }}'", file=f)
+    return schema_file
+
+
+@pytest.mark.usefixtures("create_schema")
+class Testinstance(Driver):
```
maddenp-noaa, line 26: nitpick: I find the word `instance` kind of odd in the name of a _class_ -- kind of a contradiction of terms. How about something like `TestDriver`, as it's a kind of `Driver`?

WeirAE: Nitpick away, its great for me to be better in my work!

src/uwtools/tests/drivers/test_driver.py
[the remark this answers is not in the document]
maddenp-noaa: `redefining ... from outer scope` is expected and should be disabled at the top of the file. Browse around the other test modules and you'll see this warning being disabled in many of them. `pylint` doesn't like to see a module-level name (the `create_config()` function in this case) shadowed/hidden by an argument to a function. But it's fine for us in this case.

I'll pull your code and have a look at the tests.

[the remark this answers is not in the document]
maddenp-noaa: In case it's helpful to anyone, I think I can motivate `pylint`'s concern a little more. Consider
```
>>> def foo() -> str:
...     return "foo"
... 
>>> def bar(foo) -> str:
...     return f"You gave me {foo}"
... 
>>> 
```
`pylint` would complain about the redefinition of `foo`. In the module scope `foo` is a function. But inside `bar()`, `foo` is an argument, whatever the caller passes in. The concern is that `bar()` would no longer be able to call `foo()` because, in its scope, `foo` has been redefined; `foo()` is now inaccessible to `bar()`.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -3,9 +3,75 @@
...
+
+@fixture
+def create_config(tmp_path):
+    config_file_good = tmp_path / "config_file_good.yaml"
+    config_file_bad = tmp_path / "config_file_bad.yaml"
+
+    with open(config_file_good, "w", encoding="utf-8") as f:
+        print("undefined_filter: '{{ 34 | not_a_filter }}'", file=f)
+    with open(config_file_bad, "w", encoding="utf-8") as f:
+        print("undefined_filter: '{{ 34 | not_a_filter }}'", file=f)
```
maddenp-noaa, line 64: If my eyes don't deceive me, both files will have identical content. Is that intended?

WeirAE: yep, just a copied placeholder until the functionality is fixed

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -3,9 +3,75 @@
...
+    return str(config_file_good), str(config_file_bad)
+
+
+@pytest.mark.parametrize("vals", [("good", True), ("bad", False)])
+def test_validation(vals, create_config):
+    # test concrete classes will validate correctly
+    config_file_good, config_file_bad = create_config  # pylint: disable=unused-variable
+    cfgtype, boolval = vals
 
+    instance = Testinstance(config_file=f"config_file_{cfgtype}.yaml")
```
maddenp-noaa, line 75: I think you want to pass either `config_file_good` or `config_file_bad` to the constructor. What's being passed here is just a filename without a path, while the config files that have been created by the fixture are in some temp directory, so you have to pass those values as-is.

maddenp-noaa:
```
@pytest.mark.parametrize("vals", [("good", True), ("bad", False)])
```
could be simplified to
```
@pytest.mark.parametrize("valid", [True, False])
```
and then something like
```
instance = TestDriver(config_file=config_file_good if valid else config_file_bad)
```
should work.

WeirAE: Yep, and it looks better anyway - as usual, fix one bug, i++ bugs. Next it was upset at the non-jsonschema-formatted examples, so I gave them a placeholder. But now it turns out there is a peculiarity - it turns out that if you use `@pytest.mark.usefixtures` to bypass the input specification, it passes it as a function and is very stubborn to complete

maddenp-noaa: [test_driver.txt](https://github.com/ufs-community/workflow-tools/files/12247003/test_driver.txt)

Riffing on what you had, I came up with the attached. A few notes:

1. I wasn't familiar with [`usefixtures`](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures) but I think I see how that's supposed to be used. In the example, notice that `cleandir()` doesn't `return` a value -- it just does some setup work and then calls `yield`, which passes control back to the caller so that it can execute in a context/environment that is set up in some useful way. When the caller passes control back, the fixture picks up at the statement following the `yield`, where it can do whatever cleanup is needed. But the fixture doesn't supply an actual value for the test to use, which I think we need here.
2. So I simplified `TestDriver` slightly (and renamed it to `ConcreteDriver` because TIL `pylint` tries to treat `Test`-prefixed classes as containers for tests -- see [here](https://docs.pytest.org/en/latest/explanation/goodpractices.html#conventions-for-python-test-discovery)), giving it a trivial `schema()` implementation that gets patched later, in the test function.
3. I simplified the two `@fixture`s to just return strings, and let the test method create the needed temporary files. This eliminates the need to use the `tmp_path` fixture in more than one place, or to create both good and bad config files. Not a big deal, but a little more concise.
4. The JSON doc had a syntax error (a trailing comma where JSON doesn't allow one), which I discovered when running it through `jq` to get a neatly-formatted version. Fixed.
5. In the test function, I mocked the `ConcreteDriver.schema` attribute, making it a simple `str` value -- the path to the temporary schema file.

As simple as the `configs()` and `schema()` `@fixture`s are, their values probably could have just been inlined in the `test_validation()` function, but I can imagine that they might grow more complicated later, and keeping them in the separate fixtures keeps the test-function logic uncluttered. Also, these fixtures might be used by tests to come later.

Please let me know what you think.

WeirAE: AH, fantastically cleaned up! Skips over the limitations of `usefixtures` and the test class - Keep It Simple, and all of that. Thank you so much for everything!

WeirAE:
Currently pending conflict resolution in the update, working on it, will re-request review once resolved

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -3,17 +3,32 @@
...
+from uwtools.drivers.driver import Driver
 from uwtools.drivers.forecast import FV3Forecast
 from uwtools.tests.support import compare_files, fixture_path
 
 
+def test_schema():
+    """
+    Tests validation on initialization when a config file is given.
+    """
+    config_file = fixture_path("fruit_config_similar.yaml")
```
maddenp-noaa, line 23: nitpicks:
1. In re: the docstring, this test doesn't really test validation, since we are mocking `validate()` to return `True`; rather, it's testing that the `schema` property is defined and has a value that is actually a file.
2. Since we mock `validate()`, the `config_file` value will never be used in the test, and I think it might be a bit misleading to define this as a fixture file -- it might give the impression that this actually has meaning here. Changing `FV3Forecast(config_file)` to something like `FV3Forecast(config_file="/not/used")` might make that clear.

maddenp-noaa (changes requested):
Looking good, just a couple of nitpicks...

src/uwtools/tests/drivers/test_forecast.py
WeirAE: 1. Yep, completely missed updating this, thanks for keeping me honest!
2. Ah, I've been needlessly wary of dummy values, good point! The declaration of `config_file` on line 23 is unnecessary as well

maddenp-noaa (approved):
Looks good. Nice work, and thanks for your persistence iterating on all the ideas!

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,106 @@
+{
+  "title": "FV3 Forecast config",
+  "description": "This document is to validate forecast config files from SRW, HAFS, Global",
+  "type": "object",
+  "properties": {
+    "platform": {
+      "description": "attributes of the platform",
+      "type": "object",
+      "properties": {
```
christinaholtNOAA, line 9: I think probably here we only want to start with the settings that we require in the workflow-tools driver. In the tests we currently have, we expect `static` and `cycledep` sections. 

Field table updates may also be handled here, perhaps we'd want to enforce something like a `field_table` section, and then we can't say too much about which sections are there, but each of those sections require entries like this:
```
  longname: specific humidity
  units: kg/kg
  profile_type: 
    name: fixed
    surface_value: 2.0
```

We might also be interested in `run_cmd`, `exec_name`, and optional kwargs for the run command.

Many of the settings included below are for SRW, specifically and may or may not relate directly to the forecast task. I think it makes sense to build out this schema as we build out the driver to make sure we have only the information we need.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -3,17 +3,31 @@
...
 from uwtools.tests.support import compare_files, fixture_path
 
 
+def test_schema():
+    """
+    Tests that the schema is properly defined with a file value.
+    """
+
+    with patch.object(Driver, "validate", return_value=True):
+        forecast = FV3Forecast(config_file="/not/used")
```
christinaholtNOAA, line 25: Are we testing what happens when `config_file = None`?

christinaholtNOAA (commented):
@WeirAE Nice work. 

I left a few comments below.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,106 @@
+{
+  "title": "FV3 Forecast config",
+  "description": "This document is to validate forecast config files from SRW, HAFS, Global",
```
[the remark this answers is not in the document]
WeirAE: AH, caught me again, thank you!

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: Oh good point, thanks! I was kind of going off of the comments but still left a few things in extra. I'll scrub some of that out and add a few

src/uwtools/tests/drivers/test_forecast.py
WeirAE: Ok, I thiiiink I got it set

src/uwtools/drivers/driver.py
```diff
@@ -53,10 +57,32 @@ class Driver(ABC):
     run and will appropriately parse subsequent stages of the workflow.
     """
 
-    def __init__(self):
+    def __init__(self, config_file: Optional[str] = None):
```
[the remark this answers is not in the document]
maddenp-noaa: +1 to alphabetical order as a general rule. Sometime's it's useful to have different blocks, e.g. `# Public methods` and `# Private methods` or `# Abstract methods` and `# Concrete methods`, within which the contents are sorted, and it's nice if those are marked with comments to that the next contributor knows where to put new additions.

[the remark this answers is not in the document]
maddenp-noaa: It's also worth noting that, while Python is very flexible about ordering, there are some cases where something _must_ appear earlier in a module and cannot be sorted. For example, base classes must be defined before their subclasses.

src/uwtools/drivers/driver.py
```diff
@@ -53,10 +58,36 @@ class Driver(ABC):
...
+        log = Logger() if log is None else log
+        return (
+            config_validator.config_is_valid(
+                config_file=self.config_file,
+                validation_schema=self.schema,
+                log=log,
+            )
+            if self.config_file and self.schema
+            else sys.exit(1)
+        )
```
maddenp-noaa, line 83: I'm still catching up on my emails and haven't seen yet how this is going to be tested, but in case @WeirAE is working on that now, I'd like to propose that, instead of doing this check and potentially exiting here, we should just make `validate()` -> `_validate()` (private). The constructor already makes sure that this method is only called when `config_file` is provided, so we shouldn't **be** here if it wasn't. Also, a method on an object should not take down the entire Python application unless there's a very serious issue!

maddenp-noaa: In general, it's also good design to avoid side effects (e.g. killing Python) wherever possible. From best to worst error-handling behavor:
1. A pure function that simply returns a value that can be evaluated by the caller, which can then decide what to do.
2. Raise an exception, which can at least be caught by higher-level code and handled if possible. If no code catches the exception, _then_ the application will die.
3. Killing the application.

src/uwtools/tests/drivers/test_forecast.py
maddenp-noaa: I don't think we should spend much effort on testing `config_file == None`, because we want to get rid of that possibility ASAP. We agreed that we want `config_file` to be mandatory, and will create static methods to provide any functionality that would previously have been obtained by creating a no-config `Driver` object and then calling instance methods on it. There are just some existing use cases for the no-config objects that @WeirAE looked at and (presumably) decided would be better addressed outside this PR.

src/uwtools/drivers/driver.py
WeirAE: Sorry for the extra email, turns out I missed a save there! I do need to rework the `Driver` and test functions now, since we can't query `instance.validate()` anymore so it'll have to return a message. The full sys.exit(1) I actually pulled from the behavior of `validate_config.py`, so that may need to be updated as well

maddenp-noaa: Yes, I assumed there were other places where we're not doing this as well as we could that we should improve, but I figured we could at least not add another. Thanks for addressing it.

In a well-designed code, all functions/methods would just propagate values (or, worst case, exceptions) back up the call stack -- logging their complaints, of course, because they've got the local knowledge of what happened -- until the outermost CLI level, which would then call `exit()` with either `0` or `1` to denote success or failure. (Or other non-`0` values if it's really useful to distinguish different failure reasons numerically -- this was old-school Unix behavior, but a lot of modern software sticks with `0` and `1`.)

For motivation, imagine that you're running different activities inside your program concurrently, in threads or futures. If one thread calls `exit()`, the entire Python interpreter dies, and all the other threads die in the middle of whatever they were doing, which may leave a mess. But if a thread with an issue just exits nicely with a return value (or even an exception), the main thread that spawned all the other worker threads can detect that and decide what to do. It might wait until all the other threads return on their own, letting them complete their work, clean up after themselves, etc.; or it might decide that the situation is so dire that it has to kill all those threads mid-work anyway. But that decision should be made at a fairly high level. This also makes all those functions/methods easier to test, because they don't have to test for the "else exit(1)` branches!

src/uwtools/tests/drivers/test_forecast.py
WeirAE: Hmm, well, I did manage to set something up regardless - let me know what you think

src/uwtools/drivers/driver.py
WeirAE: Replaced this behavior with logging like a sensible person! The CLI should be making decisions about exiting with that information

src/uwtools/tests/drivers/test_forecast.py
maddenp-noaa: That looks good. It'll be easy to remove when we make `config_file` mandatory, and locks down good behavior in the meantime.

src/uwtools/drivers/driver.py
maddenp-noaa: Looks good!

maddenp-noaa (approved):

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+          "type": "string",
+          "enum": [
+            "slurm",
+            "pbspro",
+            "lsf",
+            "lsfcray",
+            "none"
+          ]
+        },
+        "CCPA_OBS_DIR": {
```
christinaholtNOAA, line 30: None of these directories (here to line 56) are relevant to this class.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+        },
+        "MET_INSTALL_DIR": {
+          "type": "string",
+          "format": "uri"
+        },
+        "DOMAIN_PREGEN_BASEDIR": {
+          "type": "string",
+          "format": "uri"
+        },
+        "PARTITION_DEFAULT": {
```
christinaholtNOAA, line 57: Partitions and queues should be specified, as we'd discussed before in the arguments to the batch card or run command creator. These are all bad constructs invented by SRW that we shouldn't be enforcing.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+        },
+        "PARTITION_HPSS": {
+          "type": "string"
+        },
+        "QUEUE_HPSS": {
+          "type": "string"
+        }
+      }
+    },
+    "parameters": {
```
christinaholtNOAA, line 77: I'm not sure what these two parameters are. They are both listed as `uri` but `length` and `grid` as properties don't map to anything I'm particularly aware of.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+        },
+        "GRID": {
+          "type": "string",
+          "format": "uri"
+        }
+      }
+    },
+    "run_cmd": {
+      "description": "parameters for the run command",
+      "type": "object",
```
christinaholtNOAA, line 93: Does it make sense for us to use all caps here, or match the required variable names in the FV3Forecast method?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+      "properties": {
+        "RUN_CMD": {
+          "type": "string"
+        },
+        "EXEC_NAME": {
+          "type": "string"
+        },
+        "KWARGS": {
+          "type": "string"
```
christinaholtNOAA, line 102: Is kwargs a string?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
+{
+  "title": "FV3 Forecast config",
+  "description": "This document is to validate user-defined FV3 forecast config files",
+  "type": "object",
+  "properties": {
+    "platform": {
```
christinaholtNOAA, line 6: Now that I'm taking another look at this, does it make sense for the FV3Forecast class to be validating information about the platform?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+        },
+        "EXEC_NAME": {
+          "type": "string"
+        },
+        "KWARGS": {
+          "type": "string"
+        }
+      }
+    },
+    "locations": {
```
christinaholtNOAA, line 106: This one is for my own edification, but if we put in a section like this, does jsonschema require the input YAML to be organized like this, too? Would we need to organize the YAML like this:

```
locations:
  static:
  cycledep:
```

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -0,0 +1,167 @@
...
+        }
+      }
+    },
+    "field_table": {
+      "description": "attributes of a field table",
+      "type": "object",
+      "properties": {
+        "LONGNAME": {
+          "type": "string",
+          "enum": [
```
christinaholtNOAA, line 126: We probably don't need enum on the longname.

src/uwtools/resources/workflow.jsonschema
```diff
@@ -1,92 +1,101 @@
 {
```
christinaholtNOAA, line 1: Let's not do anything with this file at all for now since it's out of scope for the changes to the Forecast validation component. Instead, we'll keep a pretty clean history that it came in as part of the first validation tool. We should circle back around and clean it up with what we require from a workflow-level validation schema.

I am seeing that the changes are mostly whitespace and a couple of list formatting changes. I'm not super familiar with JSON formatting, but a quick search indicates that the 4 space indentation is standard. Does it make sense for us to also keep that standard?

christinaholtNOAA (commented):
I'm happy with the code at this point. Thanks @WeirAE.

I do still have concerns about the schema, though. I've listed several of those below.

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: ah uri is a slip up from me, they should be number

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: I'm still learning jsonschema - is it not implicit dict?

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: Thanks for the catch

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: I believe so? Thats what I was trying for at least

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: Could hard-specify it as `array`?

src/uwtools/resources/workflow.jsonschema
maddenp-noaa: There is a variety of opinion about 2 vs 4. We can either accept the `jq` default of 2, or use `--indent 4` and standardize on 4. I don't have an opinion, except that we should use automated formatting so that everyone follows the same rules and doesn't have to do manual formatting. The one fact I can offer is that `--indent 4` makes for a longer command. @christinaholtNOAA if you'd like to make an executive decision we should go with that.

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: And the Jira issue requested "ex: needs to know node size, queues, length, grid, suites, locations", so I included the forecast length and grid here

src/uwtools/resources/workflow.jsonschema
maddenp-noaa: @WeirAE In case we don't hear back on this, it might be worth re-doing the `jq` formatting with `--indent 4`. If we decide to stick with 4-space indents for JSON, I'll make sure it gets in the Contributor Guide at some point.

maddenp-noaa (approved):

NaureenBharwaniNOAA (approved):
Looked through the final comments and they all seemed to be resolved. Code looked good to me!

Issue #275, opened 2023-08-10
UW-10 As a model user, I want to be able to submit a forecast job


WeirAE:
**Description**
The forecast.py class needs to be extended to complete the forecast.run() class and any necessary additiona functions. It will have to be runnable as an interactive job or on a node, and be able to run as a script or part of a user batch script. 

**Requirements**
The tools should handle submission of a job by either running a script as a part of a user-supplied batch script. For example, I can call the python scripts inside an SBATCH-ready script that the user sets up, or it can create the sbatch job for you and run only the fortran executable when you call the python script from the front end. As a first pass, the tooling to handle a single batch system is fine...preferably Slurm.

**Acceptance Criteria (Definition of Done)**
GIVEN: The user chooses to run the Python scripts in an interactive job
WHEN: The tools run the forecast executable
THEN: The executable is run with platform-specific mpi command.

GIVEN: The user choose to run the Python scripts on a front-end node
WHEN: The tools run the forecast executable
THEN: The tools submit a batch job for the Slurm batch manager.

WeirAE:
Completed by #276

PR #276, opened 2023-08-10
Uw 10

WeirAE:
**Description**

GIVEN: The user chooses to run the Python scripts in an interactive job
WHEN: The tools run the forecast executable
THEN: The executable is run with platform-specific mpi command.

GIVEN: The user choose to run the Python scripts on a front-end node
WHEN: The tools run the forecast executable
THEN: The tools submit a batch job for the Slurm batch manager.

Jira issue UW-10
Fixes issue #275 

**Type of change**

**How Has This Been Tested?**

- [ ] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -109,7 +109,7 @@ class Driver(ABC):
         """
 
     @abstractmethod
-    def job_card(self):
+    def job_card(self, job_resources):
         """
         Turns the resources config object into a batch card for the configured Task.
 
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -5,11 +5,13 @@ This file contains the forecast drivers for a variety of apps and physics suites
 import logging
 import os
 import shutil
+import subprocess
 import sys
 from importlib import resources
 from typing import Dict
 
 from uwtools import config
+from uwtools.scheduler import JobScheduler
 from uwtools.drivers.driver import Driver
 from uwtools.utils import file_helpers
 
@@ -147,10 +149,13 @@ class FV3Forecast(Driver):
         Create list of SRW output files and stage them in the working directory.
         """
 
-    def job_card(self):
+    def job_card(self, resources: str):
         """
         Turns the resources config object into a batch card for the configured Task.
+        This function expects self.resources to be called first.
         """
+        return JobScheduler.get_scheduler(resources).job_card
+        
 
     def run_cmd(self, *args, run_cmd: str, exec_name: str) -> str:
         """
@@ -165,6 +170,32 @@ class FV3Forecast(Driver):
 
         This will take in the executable built in run_cmd and then run it.
         """
+        # read in the config file
+        experiment_config = self.create_model_config(self.config_file)
+        experiment_resources = {}
+
+        # prepare directories
+        run_directory = experiment_config["parameters"]["run_directory"]
+        self.create_directory_structure(run_directory)
+
+        static_files = experiment_config["locations"]["static"] 
+        self.stage_files(run_directory, static_files, link_files=False)
+        cycledep_files = experiment_config["locations"]["cycledep"] 
+        self.stage_files(run_directory, cycledep_files, link_files=True)
+
+        # set up the job
+        for key in experiment_config["resources"]:
+            experiment_resources[key] = experiment_config["resources"][key]
+        job_card = self.job_card(experiment_resources)
+        args = self.run_cmd(experiment_config["run_cmd"])
+
+        # run the job
+        subprocess.run(
+                    f"{args} {job_card}",
+                    capture_output=True,
+                    check=False,
+                    shell=True,
+                )
 
     def stage_files(
         self, run_directory: str, files_to_stage: Dict[str, str], link_files: bool = False
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -11,8 +11,8 @@ from importlib import resources
 from typing import Dict
 
 from uwtools import config
-from uwtools.scheduler import JobScheduler
 from uwtools.drivers.driver import Driver
+from uwtools.scheduler import JobScheduler
 from uwtools.utils import file_helpers
 
 
@@ -149,13 +149,12 @@ class FV3Forecast(Driver):
         Create list of SRW output files and stage them in the working directory.
         """
 
-    def job_card(self, resources: str):
+    def job_card(self, job_resources):  # pragma: no cover
         """
         Turns the resources config object into a batch card for the configured Task.
-        This function expects self.resources to be called first.
+
         """
-        return JobScheduler.get_scheduler(resources).job_card
-        
+        return JobScheduler.get_scheduler(job_resources).job_card
 
     def run_cmd(self, *args, run_cmd: str, exec_name: str) -> str:
         """
@@ -164,38 +163,44 @@ class FV3Forecast(Driver):
         args_str = " ".join(str(arg) for arg in args)
         return f"{run_cmd} {args_str} {exec_name}"
 
-    def run(self):
+    def run(self):  # pragma: no cover
         """
         Runs the forecast executable with the namelist file and staged input files.
 
         This will take in the executable built in run_cmd and then run it.
         """
         # read in the config file
-        experiment_config = self.create_model_config(self.config_file)
+        #self.create_model_config(config_file=self.config_file)
+        experiment_config = config.YAMLConfig(self.config_file)
         experiment_resources = {}
 
         # prepare directories
         run_directory = experiment_config["parameters"]["run_directory"]
         self.create_directory_structure(run_directory)
 
-        static_files = experiment_config["locations"]["static"] 
+        static_files = experiment_config["locations"]["static"]
         self.stage_files(run_directory, static_files, link_files=False)
-        cycledep_files = experiment_config["locations"]["cycledep"] 
+        cycledep_files = experiment_config["locations"]["cycledep"]
         self.stage_files(run_directory, cycledep_files, link_files=True)
 
         # set up the job
         for key in experiment_config["resources"]:
             experiment_resources[key] = experiment_config["resources"][key]
         job_card = self.job_card(experiment_resources)
-        args = self.run_cmd(experiment_config["run_cmd"])
+        args = str(experiment_config["run_cmd"]["kwargs"])
+        run_command = self.run_cmd(
+            args,
+            run_cmd=experiment_config["run_cmd"]["run_cmd"],
+            exec_name=experiment_config["run_cmd"]["exec_name"]
+        )
 
         # run the job
         subprocess.run(
-                    f"{args} {job_card}",
-                    capture_output=True,
-                    check=False,
-                    shell=True,
-                )
+            f"{run_command} {job_card}",
+            capture_output=True,
+            check=False,
+            shell=True,
+        )
 
     def stage_files(
         self, run_directory: str, files_to_stage: Dict[str, str], link_files: bool = False
--- a/src/uwtools/resources/FV3Forecast.jsonschema
+++ b/src/uwtools/resources/FV3Forecast.jsonschema
@@ -33,6 +33,10 @@
             "description": "parameters of the forecast",
             "type": "object",
             "properties": {
+                "RUN_DIRECTORY": {
+                    "type": "string",
+                    "format": "number"
+                },
                 "LENGTH": {
                     "type": "string",
                     "format": "number"
--- a/src/uwtools/tests/drivers/test_driver.py
+++ b/src/uwtools/tests/drivers/test_driver.py
@@ -29,7 +29,7 @@ class ConcreteDriver(Driver):
     def output(self):
         pass
 
-    def job_card(self):
+    def job_card(self, job_resources):
         pass
 
 


src/uwtools/drivers/forecast.py
```diff
@@ -147,10 +149,12 @@ def output(self):
         Create list of SRW output files and stage them in the working directory.
         """
 
-    def job_card(self):
+    def job_card(self, job_resources):  # pragma: no cover
```
NaureenBharwaniNOAA, line 152: This may be premature, but after looking at Paul's PR #278. I'm wondering how those changes will affect this ticket specifically around the language of job card vs. batch script, and the movement of the code.

WeirAE: Definitely not premature! Pending that resolution it may have to change

maddenp-noaa: It might make for a messy merge. I will happily volunteer a follow-on PR targeting _this branch_ (and its PR) that resolves any conflicts and preserves the in-progress work on this branch.

WeirAE: Haha oh I've been keeping an eye on that one, don't worry!

WeirAE: Well, the answer is "batch script"! On that update and will wrangle the subsequent resolutions :)

src/uwtools/cli/run_forecast.py
```diff
@@ -22,9 +23,20 @@ def main() -> None:
...
+            config_file=args.config_file,
+            dry_run=args.dry_run,
+            forecast_app=args.forecast_app,
+            forecast_model=args.forecast_model,
+            log_file=args.log_file,
+            machine=args.machine,
+            quiet=args.quiet,
+            verbose=args.verbose,
+        )
+    except UWConfigError as e:
```
maddenp-noaa, line 38: I wonder if it might be time to build up a bit more of an exception hierarchy. For example, is the _only_ reason `run()` might fail a configuration error? If not, we should probably start raising (and handling) other exception types. For example, we could now (or soon) define `UWError(Exception)` as a base class for _all_ UW issues, redefine `UWConfigError(UWError)`, and catch `UWError` here, in case we are failing due to a UW issue other than a config issue. Is there an immediate use case for a new type of exception relating to _running_ and not just _configuring_?

WeirAE: Ohh, that is a good point - we have taken over several other major features and we might as well finish taking over the exception handling entirely.

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +131,64 @@ def resources(self) -> None:
...
+        dry_run: bool = False,
+        log_file: Optional[str] = None,
+        machine: Optional[str] = None,
+        quiet: bool = False,
+        verbose: bool = False,
+    ) -> None:  # pragma: no cover
         """
         Runs FV3.
         """
+        # set up logging
```
maddenp-noaa, line 148: For consistency with the rest of the codebase, please use full-sentence English comments with initial capitals and punctuation throughout, e.g.
```
# Set up logging.
```

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +131,64 @@ def resources(self) -> None:
...
+            log.info(f"Configuration: {forecast_app} {forecast_model} {machine}")
+            log.info(f"Run command: {run_command} {batch_script}")
+            return
+        # run the job
+        subprocess.run(
+            f"{run_command} {batch_script}",
+            capture_output=True,
+            check=False,
+            shell=True,
+        )
```
maddenp-noaa, line 191: I think that, if `run_command` fails in the subshell, `subprocess.run()` will raise an exception that -- since we've got `capture_output=True` -- we'll want to catch and handle by iterating over the captured output and logging what was captured. Otherwise, `uw` will terminate, but the user will not see any complaints that the program being run might have made and will possibly be mystified.

Well, I think that would be the _better_ way to do it, but I think the options are:

1. Remove `capture_output=True`, which will result in the output being printed directly to the appropriate stream (`stdout` and/or `stderr`, depending on how well-behaved the program we're running is); and just let whatever exception is raised terminate `uw`.
2. The better option IMO: Assign the return value of `subprocess.run()` to a variable, then extract the captured output from that variable and log the output, which we might want to do even if there is no error -- if the program being run produces information, it might be good to log it for the user to see. Then execute that inside a `try` block, handle whatever exception is raised (can't remember, but the `subprocess` docs should say), and then raise a (new!) subclass of `UWError` -- maybe `UWRuntimeError`? -- for code higher up the call stack to handle.

I can also think of two ways to deal with the fact that programs being run in a subshell may or may not use both `stdout` and `stderr`:

1. Set `stderr=STDOUT` as described in the [subprocess](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) docs, which will merge `stderr` into `stdout` so that there will be only one set of output to iterate over and log, which will look more like what a user would see in a terminal, where `stdout` and `stderr` messages are (normally, barring stream redirection in the shell) naturally interspersed.
2. Log e.g. `Captured output:`, iterate over and log `stdout` messages, log `Captured errors:`, iterate over and log `stderr` messages. This might be more intelligible to the user, as they won't see error and non-error output interspersed.

It's anybody's guess as to whether any given program will use `stdout` and/or `stderr` "correctly". Many NWP applications ignore this, but I think we should be ready for anything.

It might be worth checking to see whether there are other `subprocess.run()` calls that could benefit from the same treatment and, if so, factor this more robust handling out to a helper method that could be called here and from those other sites.

src/uwtools/tests/cli/test_run_forecast.py
```diff
@@ -39,7 +41,21 @@ def test_main(files):
     with patch.object(run_forecast, "parse_args", return_value=args):
         with patch.object(run_forecast.forecast, "FV3Forecast") as fv3fcst:
             run_forecast.main()
-            fv3fcst.assert_called_once_with(cfgfile, machinefile, log_name="run-forecast")
+            fv3fcst.assert_called_once_with(
```
WeirAE, line 44: The current main bug seems to be the syntax here is not actually successfully calling `FV3Forecast`. I've tried rewriting it more like the `test_main()` in `test_set_config()` but with no success. I believe `run_forecast.main()` is set up ok, so any observations on this are appreciated

src/uwtools/drivers/forecast.py
WeirAE: Ah thats very fair, and an extension of your other comment. In my local branch I've placed in the quick solution of removing `capture_output=True` and replacing it with `stderr=subprocess.STDOUT` so there's a fix in place while I sort out the failing tests, but I'm hoping I can slip in adding more to `UWError`!

src/uwtools/tests/cli/test_run_forecast.py
maddenp-noaa: Try `assert fv3fcst.__init__.called_once_with(...)`? It's the constructor of the class that gets called when `FV3Forecast` (or a mock of it) gets instantiated.

WeirAE: Interesting - I think I might still be misunderstanding the way python handles the constructor vs the class, thats a good one!

..Ah, the type checker rapped my knuckles on that, "error: Accessing "__init__" on an instance is unsound"

maddenp-noaa: If this simulation is realistic, what you had first looks like it should work:
```
% cat foo.py
class C:
    pass
```
Then in the Python REPL:
```
>>> from unittest.mock import patch
>>> import foo
>>> with patch.object(foo, "C") as C:
...     C(a=1, b=2)
...     C.assert_called_once_with(a=1, b=2)
... 
<MagicMock name='C()' id='281472952832592'>
```
I didn't need to say `C.__init__` was called, just `C`. I'm going to check out your branch and see if I can reproduce whatever error you're getting.

src/uwtools/drivers/driver.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, config_file: str, log: Optional[Logger] = None):
     # Public methods
 
     @abstractmethod
-    def batch_script(self) -> None:
+    def batch_script(self, job_resources):
```
christinaholtNOAA, line 28: We seem to have lost the return type here.

src/uwtools/drivers/driver.py
```diff
@@ -49,7 +49,17 @@ def resources(self) -> None:
         """
 
     @abstractmethod
-    def run(self) -> None:
+    def run(
+        self,
+        config_file: str,
```
christinaholtNOAA, line 54: Since `config_file` is a required object attribute, we shouldn't need it as an argument here. 

`forecast_app` is not necessary at all given that an abstract Driver will not need any connection to an app to be able to be useful. Presumably there'd be some information provided elsewhere (maybe in the config file) to determine if this thing is running as part of an app so that other appropriate assumptions can be made.

I think the forecast model will need to be specified before we get to this point so that we can decide which class to invoke an object from. 

And `machine` should be information included in the high-level info contained in the config file. 

I think that leaves us with `dry_run`, `log_file`, `quiet`, and `verbose`, and I have other concerns there. Should those also be handled at the object attribute level? Also, do we need both `quiet` and `verbose` flags, or a single flag that handles verbosity in general with various degrees?

I did a LOT of weeding there, but perhaps the signature for `run()` should stay with just `self`.

src/uwtools/drivers/forecast.py
```diff
@@ -21,10 +25,11 @@ class FV3Forecast(Driver):
 
     # Public methods
 
-    def batch_script(self) -> None:
+    def batch_script(self, job_resources):  # pragma: no cover
```
christinaholtNOAA, line 28: Should we lose the output type here?

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +131,64 @@ def resources(self) -> None:
...
+            log_file=log_file or "/dev/stdout",
+            log_name=name,
+            quiet=quiet,
+            verbose=verbose,
+        )
+        # read in the config file
+        # self.create_model_config(config_file=config_file)
+        experiment_config = config.YAMLConfig(config_file)
+        # prepare directories
+        run_directory = experiment_config["parameters"]["run_directory"]
```
christinaholtNOAA, line 160: I'd expect that the section of the config file we'd want these settings to be a part of would be named something similar to `forecast`. Perhaps it's more specific with `FV3Forecast`?

src/uwtools/tests/cli/test_run_forecast.py
maddenp-noaa: I think I see what's going on. In `uwtools.cli.run_forecast`, you have
```
    forecast_class = getattr(forecast, f"{args.forecast_model}Forecast")
    try:
        forecast_class.run(
            config_file=args.config_file,
            dry_run=args.dry_run,
            forecast_app=args.forecast_app,
            forecast_model=args.forecast_model,
            log_file=args.log_file,
            machine=args.machine,
            quiet=args.quiet,
            verbose=args.verbose,
        )
```
The line `forecast_class = getattr(forecast, f"{args.forecast_model}Forecast")` gives you a reference to the `FV3Forecast` class, on which you then make the call `forecast_class.run(...)`. This would work if `run()` were a static method. But it's an instance method, so you first need to instantiate an `FV3Forecast` object and _then_ call the `run()` method on that instance, like
```
    forecast_class = getattr(forecast, f"{args.forecast_model}Forecast")
    forecast_obj = forecast_class()
    try:
        forecast_obj.run(...)
```
This would have been apparent if you'd tried running `run_forecast.main()` outside of a unit test with mocking going on, because Python would have complained about `run()` not being something you can call on the `FV3Forecast` **class**. (And `pylint` and probably `mypy` would have complained if we weren't using `getattr()` to obtain a reference to the `FV3Forecast` class -- if we literally had `FV3Forecast` in `run_forecast.main()`, the static-analysis tools could have done their job, but the dynamic `getattr()` lookup hides that from them.)

But mocking makes the impossible possible, in a rare, unhelpful way in this case. I was already writing up what follows to share, but it just became especially pertinent, so here goes:

--

I should probably share something that I lost a lot of time by not understanding when I started working with `unittest.mock`: When you have a `Mock` object and reference a member, either as a property (`mymock.property`) or a method (`mymock.method()`), `Mock` _creates those members_ dynamically as you reference them, and to any depth of dot-chaining. They just spring into existence, e.g.
```
>>> from unittest.mock import Mock
>>> m = Mock()
>>> type(m.property)
<class 'unittest.mock.Mock'>
>>> type(m.n.o.func)
<class 'unittest.mock.Mock'>
>>> m.n.o.func(foo=88, bar=99)
<Mock name='mock.n.o.func()' id='281472987072064'>
>>> m.n.o.func.assert_called_once_with(foo=88, bar=99)
>>> 
```
I didn't have to do anything special to create `n` as a member of `m`, `o` as a member of `n`, etc.

--

So, this is why `forecast_class.run()` doesn't crash via the unit test: `forecast_class` is a `Mock` (probably `MagicMock`, but the difference isn't important here) object, so a `.run` member is automatically created upon reference. `Mock` objects are callable, so it's fine that that member is called as a function/method. But the _constructor_ of `FV3Forecast` is never called because we never instantiate an object.

Upon reflection, the line `forecast_class.run(` should have told us everything: An `FV3Forecast` **class** doesn't have a `run()` (static) method to call; but a `FV3Forecast` **instance** has a `run()` (instance) method to call.

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +131,64 @@ def resources(self) -> None:
...
+        self.create_directory_structure(run_directory)
+
+        static_files = experiment_config["locations"]["static"]
```
christinaholtNOAA, line 163: Again, perhaps the configuration section under `FV3Forecast` (name pending) looks something like:

```
FV3Forecast:
  run_directory:
  static:
  cycledep:
  exec_name:
platform:
  mpi_cmd:
  
```

Please also disregard my lack of ordering there.

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +131,64 @@ def resources(self) -> None:
...
+        )
+
+        if dry_run:
+            # Apply switch to allow user to view the run command of config
+            # This will not run the job.
+            log.info(f"Configuration: {forecast_app} {forecast_model} {machine}")
+            log.info(f"Run command: {run_command} {batch_script}")
+            return
+        # run the job
+        subprocess.run(
```
christinaholtNOAA, line 186: I think right around here is where the script should make a decision about whether to run the mpi command or submit a batch script that runs the mpi command.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -33,6 +33,10 @@
             "description": "parameters of the forecast",
             "type": "object",
             "properties": {
+                "RUN_DIRECTORY": {
+                    "type": "string",
+                    "format": "number"
```
christinaholtNOAA, line 38: What is a number format here?

christinaholtNOAA (changes requested):

src/uwtools/drivers/driver.py
WeirAE: Ah, interesting - so I had passed all args mirroring set_config() but several of the args we had set up originally are redundant since they should be in the config file. I had left them in as a thought to be a fail-safe to check the arg against the config to ensure user certainty, but I wasn't particularly convinced by that. I think you're probably right, keep it simple, prune the redundant parts and let the validator do any work that needs to be done - thanks!

src/uwtools/drivers/driver.py
WeirAE: There had been an issue in a previous iteration that was failing on not returning correctly, but the proper version should just have a non-None type, thanks

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: A copy/paste mistake!

src/uwtools/tests/cli/test_run_forecast.py
WeirAE: Hooray, for lack of an insightful lightbulb emoji! `Mock` is still new to me and I'm a little clumsy with constructors, so that was what I had a feeling the issue was, but its a more fundamental 'gotcha' that is hard to parse and practically impossible for usual patchwork solutions to fix. Thanks, I'm going to have to really reread this last part to internalize it!

src/uwtools/drivers/driver.py
maddenp-noaa: We haven't had the full conversation yet about what to do with logging, but absent use cases I haven't imagined, I would hope that `log_file`, `quiet`, and `verbose` would all be handled far above this class, when configuring the Python base logger for the entire `uw` invocation. I think that would be handled at the cli level -- currently in each of the modules under `uwtools.cli`, soon in the single modal-cli module. (We may still need to accept an override logger from API users: I would hope to receive that via to-be-defined API functions and use it to reconfigure the Python base logger, so that no downstream logging calls need to change.)

I'm not sure about `dry_run`, but my gut feeling is that that should be set in the constructor, too.

src/uwtools/tests/cli/test_run_forecast.py
maddenp-noaa: Assuming what I wrote is correct, I found this out by inserting a `breakpoint()` in the right place, dropping into the Python debugger, and stepping through the code, introspecting variables as I went. Maybe a quick tutorial on `pdb` would be useful for our team. It's a really powerful (and surprisingly simple) way to delve into weird problems.

src/uwtools/drivers/forecast.py
maddenp-noaa: Maybe
```
forecast:
    model: FV3
    <etc>
```
? This avoids leaking Python-code names into the DSL. We can map to the correct class with e.g.
```
forecast_class = {"FV3": FV3Forecast, "MPAS": MPASForecast}[config["forecast"]["model"]
```
(I think we're doing that already, just constructing the class name with an f-string or similar.)

It terms of schema, I think it would be nice to say "There's an optional top-level block called `forecast` and, if it exists, it may have a `model` value in the enum `[FV3, MPAS]`, and also ..." vs having to say "There's an optional top-level block called `FV3Forecast` and another optional top-level block called `MPASForecast` and they are mutually exclusive, and their subtrees look like ..." -- with the names of Python classes also appearing in YAML.

Unless we can describe all the pertinent forecast parameters for both (or more) models with a consistent set of key-value pairs, we might have to overs-specify the parameters under `forecast` and check them in code post schema validation, unless JSON Schema permits saying things like "parameter `x` only goes with `model: FV3`, and parameter `y` only goes with `model: MPAS`. I don't know yet. I think it'd be worth trying to distill the common parameters and then complicating however we must from there.

Failing any/all of that, I'd suggest that we at least choose and stick to a convention for key strings, e.g. "lower-case with hyphens". (Underscores are ok, too, but I've found value in using hyphens to remind everyone that these things are _not anything except YAML config keys_ Specifically, they're _not_ Python variables! For the same reason, I suggest using e.g. `true` instead of `True` in YAML: As a reminder that these are YAML values, not Python values -- as otherwise people supply `None` instead of `null` as a YAML value and get `"None": str` in Python instead of `None: NoneType`. Schema checking will help, but conventions help, too. And we can map from our DSL keys to whatever we need for various uses.)

I'm not suggesting any restriction on _value_ case, punctuation, whitespace, etc. -- whatever's needed is fine. Just on keys.

src/uwtools/drivers/forecast.py
maddenp-noaa: I agree, but what would that decision be based on? Should that be handled at the cli level, e.g. `uw forecast run ...` vs `uw forecast submit ...`? Or do we already know the user's intent some other way?

src/uwtools/drivers/forecast.py
WeirAE: I've gone through some of the SRW YAML files and here's my proposed edit, with some similar naming and structure to the others:
```
 platform:
    RUN_ENVIR: community
    MACHINE: hera
    MPI_CMD: srun
    ACCOUNT: any_acct
forecast:
    APP: SRW
    EXTRN_MDL_NAME: FV3GFS
    STATIC:
    CYCLEDEP:
    VERBOSE: false
```
This should formally pull the args out to the config file as well

src/uwtools/drivers/forecast.py
WeirAE: My preference is at the cli level, though I can set some structure here

src/uwtools/drivers/driver.py
WeirAE: Working on moving anything not covered in the config file to the constructor, which I think is just `dry_run` and `log_file` now!

src/uwtools/tests/cli/test_run_forecast.py
WeirAE: Yep, going to make sure to use breakpoints now, that'll help a lot. I had gotten tied up blaming the mock patching for the failure to make the instance right when it was just the darn `getattr()` again. It worked correctly this way, although there was some fussing in passing the args correctly. However, with moving all of those args out it should lessen the headache by far

src/uwtools/cli/run_forecast.py
```diff
@@ -18,13 +19,17 @@ def main() -> None:
     Parses arguments provided by the user and passes to the Forecast driver class to be run.
     """
     args = parse_args(sys.argv[1:])
-    name = "run-forecast"
-    cli_helpers.setup_logging(
+    name = "validate-config"
```
maddenp-noaa, line 22: Should this still be `run-forecast`?

src/uwtools/drivers/driver.py
```diff
@@ -7,25 +7,30 @@
...
 
 
 class Driver(ABC):
     """
     An abstract class representing drivers for various NWP tools.
     """
 
-    def __init__(self, config_file: str, log: Optional[Logger] = None):
+    def __init__(
+        self, config_file: str, log: Optional[Logger] = None, dry_run: Optional[bool] = False
```
maddenp-noaa, line 19: Unless we need to be able to explicitly pass in a `None` value for `dry_run`, which doesn't immediately make sense to me, `dry_run: Optional[bool] = False` should just be `dry_run: bool = False`, which means that the argument is optional but will be `False` if not supplied. `Optional[<type>]` is shorthand for `Union[<type>, None]`, meaning that a value is expected, but that value may be `None`.

If I'm not mistaken here and this can be simplified, I think the line
```
self._dry_run = dry_run if dry_run is not None else False
```
can be just
```
self.dry_run = dry_run
```
since the `dry_run` argument will definitely have a value of either `True` or `False`.

src/uwtools/drivers/forecast.py
```diff
@@ -21,10 +26,11 @@ class FV3Forecast(Driver):
 
     # Public methods
 
-    def batch_script(self) -> None:
+    def batch_script(self, job_resources) -> BatchScript:  # pragma: no cover
```
maddenp-noaa, line 29: Don't forget to type-annotate `job_resources`.

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +132,60 @@ def resources(self) -> None:
...
+            log = cli_helpers.setup_logging(
+                log_file="/dev/stdout",
+                log_name=name,
+                quiet=False,
+                verbose=False,
+            )
+        # Read in the config file.
+        experiment_config = config.YAMLConfig(self._config_file)
+
+        # define forecast app and model
```
maddenp-noaa, line 151: nitpick: Capitalize & punctuate.

src/uwtools/tests/cli/test_run_forecast.py
```diff
@@ -3,6 +3,7 @@
 Tests for the run-forecast CLI.
 """
 
+# from itertools import chain
```
maddenp-noaa, line 6: nitpick: Don't forget to remove dead code.

src/uwtools/drivers/forecast.py
WeirAE: pick away!

src/uwtools/drivers/driver.py
WeirAE: OK, good catch - I have a bad habit of over-stating things to a fault, I'll clean it up, thanks!

src/uwtools/cli/run_forecast.py
WeirAE: caught my copy-paste rushing errors, thanks

src/uwtools/drivers/driver.py
WeirAE: hm - I think the issue is that the constructor is taking the args, so I think I should change this to see if `Forecast` is `called_once_with` the `args` and that `FV3Forecast` is just `called_once`

src/uwtools/cli/run_forecast.py
maddenp-noaa: I take it that a promising opportunity to diversify the UW exceptions did not present itself. Please feel free to resolve this comment if that's the case.

WeirAE: I have an opportunity still to get it in, I'll see if I can

maddenp-noaa: Only if it looks genuinely useful!

src/uwtools/cli/run_forecast.py
```diff
@@ -60,11 +70,23 @@ def parse_args(args: List[str]) -> Namespace:
         metavar="FILE",
     )
     optional.add_argument(
-        "-m",
-        "--machine",
-        help="Path to YAML platform-definition file",
+        "--forecast-app",
+        choices={"HAFS", "MRW", "SRW"},
+        help="The app to be run",
+    )
```
maddenp-noaa, line 76: Did we decide that the "app" concept wasn't applicable at the level of just running a forecast? If so, I think this argument could be removed.

src/uwtools/drivers/driver.py
```diff
@@ -7,25 +7,35 @@
...
     An abstract class representing drivers for various NWP tools.
     """
 
-    def __init__(self, config_file: str, log: Optional[Logger] = None):
+    def __init__(
+        self,
+        config_file: str,
+        log: Optional[Logger] = None,
+        dry_run: bool = False,
+        outfile: Optional[str] = None,
```
maddenp-noaa, line 23: Is `outfile` a concept that would apply to **all** drivers? My guess is "no". If we need this option for the `FV3Forecast` subclass, it could move to that constructor of that class. Thoughts?

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +133,71 @@ def resources(self) -> None:
         ???
         """
 
-    def run(self):
+    def run(self, log: Optional[Logger] = None) -> None:  # pragma: no cover
```
maddenp-noaa, line 136: Please remove `# pragma no cover`.

src/uwtools/drivers/driver.py
WeirAE: Since I'm using 'outfile' here as the output batch script file name, it would apply to all drivers - I guess the better question is whether its an appropriate name for the sake of menu symmetry or if it should be a bool flag or something like "batch_file_name"?

src/uwtools/cli/run_forecast.py
WeirAE: I had maintained this along with the model call due to having to parse the correct class to run, even though its not currently looking at the app for versioning. I could remove it at this stage for us to add back if we need it later?

src/uwtools/drivers/forecast.py
WeirAE: Hopefully handled better in the updated YAML now!

src/uwtools/drivers/driver.py
WeirAE: Fixed!

src/uwtools/cli/run_forecast.py
maddenp-noaa: If we're not using it, YAGNI is probably the right philosophy.

src/uwtools/cli/run_forecast.py
```diff
@@ -60,11 +70,23 @@ def parse_args(args: List[str]) -> Namespace:
...
+    optional.add_argument(
+        "--forecast-model",
+        choices={"CCPP", "CICE", "CMEPS", "FV3", "MOM6"},
+        help="The experiment to be run",
+    )
```
maddenp-noaa, line 81: Is this actually a _required_ argument? Can we do without it?

Also, I think it would be better not to add support for choices we don't have, i.e. all the models _except_ for `FV3`. We can easily add those later when we have code support for them. If we release this, users get the impression (from the `--help` switch) that these are supported.

That said, it's probably not worth doing much more work in this module, as it will mostly be OBE by the modal-CLI work anyway. Just getting the programmatic interface working, and unit tests validating it, should be enough for this ticket.

src/uwtools/drivers/driver.py
maddenp-noaa: Do we (currently) need this value -- the path to the batch script to write -- anywhere but in the `run()` method? If no, could it (or, like you say, a more explicitly-named e.g. `batch_script_name`) be an argument to that method instead? Then one might, elsewhere, do something like
```
fv3fcst = FV3Forecast(...)
fv3fcst.run(..., batch_script_name="whatever")
```
Is it accurate that we should only need `batch_script_name` instead of `batch_script_path` because we'll be writing it to a run directory whose path is known by other means? Does the filename itself even need to be configurable? Could/should we have a standard name for a forecast batch script?

src/uwtools/drivers/forecast.py
```diff
@@ -126,10 +133,71 @@ def resources(self) -> None:
...
+        if log is None:
+            name = f"{inspect.stack()[0][3]}"
+            log = cli_helpers.setup_logging(
+                log_file="/dev/stdout",
+                log_name=name,
+                quiet=False,
+                verbose=False,
+            )
+        # Read in the config file.
+        experiment_config = config.YAMLConfig(self._config_file)
```
maddenp-noaa, line 150: I'm confused by `experiment_config` because we've got an "experiment" concept, but here we're in a "forecast" context. Do we need a clarification of concepts, or is this ok?

src/uwtools/cli/run_forecast.py
WeirAE: I can depreciate the unsupported models for this ticket :)

src/uwtools/drivers/forecast.py
WeirAE: nope, you're right, I missed that

src/uwtools/drivers/driver.py
WeirAE: True! I can move that over to run() specifically. For the name, hmm, 'can' and 'should' give different answers here - I'm inclined to give it just a name and pull `run_directory` from the YAML for the write?

src/uwtools/cli/run_forecast.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=duplicate-code
+# pylint: disable=duplicate-code, unused-import
```
maddenp-noaa, line 1: What's the unused import? I haven't found a reason yet to use this particular disable directive instead of removing the actually unused import. Maybe TIL something?

src/uwtools/drivers/driver.py
maddenp-noaa: I like that. If we find a use case where users need to use a different name, we could field a feature request.

src/uwtools/cli/run_forecast.py
WeirAE: Ohh - ok, this was a weird one, might be an issue of the implementation. I'm catching `UWError` on line 36 instead of `UWConfigError` like previously. However, I was mocking `UWConfigError` in `test_run_forecast()` to test catching the base from the parent, but it was erroring on not having `UWConfigError`. If I include it without disabling the import, it doesn't count it below and errors. Strictly, its a test remnant for an issue I hadn't resolved to where I was happy with it

src/uwtools/cli/run_forecast.py
```diff
@@ -60,11 +70,18 @@ def parse_args(args: List[str]) -> Namespace:
...
-        "-m",
-        "--machine",
-        help="Path to YAML platform-definition file",
+        "--forecast-model",
+        choices={"FV3"},
+        help="The experiment to be run",
+    )
+    optional.add_argument(
+        "-o",
+        "--outfile",
```
christinaholtNOAA, line 79: I might suggest that this should be `batch_script` instead of `outfile`. The latter is quite ambiguous when we're talking about running a whole task. It could refer to the model output, a log of some sort, etc.

src/uwtools/drivers/driver.py
```diff
@@ -43,9 +55,9 @@ def requirements(self) -> None:
         """
 
     @abstractmethod
-    def resources(self) -> None:
+    def resources(self, platform: dict) -> Mapping:
         """
-        ???
+        Set up the formatting for the scheduler batch script.
```
christinaholtNOAA, line 60: Perhaps leave this one as-is with `???` so that we can recognize it as something we need to handle explicitly later when we move beyond the MVP?

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
         ???
         """
 
-    def resources(self) -> None:
+    def resources(self, platform: dict) -> Mapping:
         """
-        ???
+        Set up the formatting for the scheduler batch script.
```
christinaholtNOAA, line 134: See the comment above concerning the interface.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
         """
-
-    def run(self):
+        # Add required fields to platform.
+        # Currently supporting only slurm and none.
+        slurmqueues = ["hera", "jet", "orion", "stampede"]
```
christinaholtNOAA, line 138: None of this information belongs in the driver. We haven't talked about this explicitly, but we will need some utility that transforms platform information into Driver-usable information without any additional processing here.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+            "nodes": 1,
+            "queue": platform["queue"],
+            "scheduler": platform["scheduler"],
+            "tasks_per_node": 1,
+            "walltime": "00:01:00",
+        }
+
+    def run(self, log: Optional[Logger] = None) -> None:
         """
         Runs FV3.
```
christinaholtNOAA, line 155:
```suggestion
        Runs FV3 either as a subprocess or by submitting a batch script.
```

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+        if log is None:
+            name = f"{inspect.stack()[0][3]}"
+            log = cli_helpers.setup_logging(
+                log_file="/dev/stdout",
+                log_name=name,
+                quiet=False,
+                verbose=False,
+            )
+        # Read in the config file.
+        forecast_config = config.YAMLConfig(self._config_file)
```
christinaholtNOAA, line 167: Do we want to consider loading this file in `__init__()`? Sorry if I missed this discussion already.

Also, I'd suggest that we load it in like this:

```
forecast_config = config.YAMLConfig(self._config_file).get('forecast')
```

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+
+        # Define forecast model.
+        forecast_model = forecast_config["forecast"]["EXTRN_MDL_NAME"]
+        machine = forecast_config["platform"]["machine"]
```
christinaholtNOAA, line 171: No machine information like this should live in the Driver.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+
+        # Prepare directories.
+        run_directory = forecast_config["forecast"]["FCSTDIR"]
```
christinaholtNOAA, line 174: We likely want each driver to have the same entry that is something like `run_directory` or `work_dir`. I'm open to others, but it's already in its own namespace, so let's make it consistent and generic.

We also likely want to pare down the scope of what the driver needs to just be in the `forecast` section just after it loads an experiment YAML. That would mean we don't have to repeat asking for the `forecast` section request when we look for config settings the driver needs.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+        self.create_directory_structure(run_directory, "delete")
+
+        static_files = forecast_config["forecast"]["STATIC"]
+        self.stage_files(run_directory, static_files, link_files=False)
+        cycledep_files = forecast_config["forecast"]["CYCLEDEP"]
+        self.stage_files(run_directory, cycledep_files, link_files=True)
```
christinaholtNOAA, line 180: It's likely safer to link the static files, and copy the cycledep ones by default.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+
+        # Create the job script.
+        platform = {key: forecast_config["platform"][key] for key in forecast_config["platform"]}
```
christinaholtNOAA, line 183: I think I'm missing something here. Why not:

```
platform = forecast_config.get('platform')
```

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+        platform_resources = self.resources(platform)
+        batch_script = self.batch_script(platform_resources)
+        args = "--export=ALL -n"
```
christinaholtNOAA, line 186: This args statement may not be true on all platforms.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+        run_command = self.run_cmd(
+            args,
+            run_cmd=forecast_config["forecast"]["run_cmd"],
+            exec_name=forecast_config["forecast"]["exec_name"],
```
christinaholtNOAA, line 190: These entries don't exist in the YAML provided with this PR.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +129,92 @@ def requirements(self) -> None:
...
+            log.info(f"Configuration: {forecast_model} {machine}")
+            log.info(f"Run command: {run_command} {batch_script}")
+            return
+
+        # Run the job.
+        if self._outfile is not None:
+            outpath = run_directory.join(self._outfile)
+            with open(outpath, "w", encoding="utf-8") as f:
+                subprocess.run(
+                    f"{run_command} {batch_script}",
```
christinaholtNOAA, line 205: This subprocess should run `f"sbatch {batch_script}"`. The batch command should come from a property in scheduler, which I don't think currently exists. 

@maddenp-noaa Any ideas on where to get this piece of info?

src/uwtools/exceptions.py
```diff
@@ -2,8 +2,16 @@
 Custom exceptions for the UWTools package.
 """
 
+from abc import ABC
 
-class UWConfigError(Exception):
+
+class UWError(ABC, Exception):
```
christinaholtNOAA, line 8: Why do we need this to be an abstract class?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -3,116 +3,71 @@
     "description": "This document is to validate user-defined FV3 forecast config files",
     "type": "object",
     "properties": {
-        "resources": {
+        "platform": {
```
christinaholtNOAA, line 6: The platform section should be included in a workflow-level section, not in a driver-level section. Each driver can then use the single platform section as a single source of information.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -3,116 +3,71 @@
...
             "description": "Expected resources for the given platform",
             "type": "object",
             "properties": {
-                "WORKFLOW_MANAGER": {
+                "RUN_ENVIR": {
```
christinaholtNOAA, line 10: This entry is not needed at all.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -3,116 +3,71 @@
...
                 }
             }
         },
-        "parameters": {
+    "forecast": {
             "description": "parameters of the forecast",
             "type": "object",
             "properties": {
-                "LENGTH": {
+                "APP": {
```
christinaholtNOAA, line 42: App information is not needed for the forecast to run.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -3,116 +3,71 @@
...
                     "type": "string",
-                    "format": "number"
                 },
-                "GRID": {
+                "EXTRN_MDL_NAME": {
```
christinaholtNOAA, line 45: This one is not needed.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -3,18 +3,42 @@
...
+
+
+def test_batch_script(slurm_props):
+    expected = """
+#SBATCH --account=account_name
+#SBATCH --nodes=1
+#SBATCH --ntasks-per-node=1
+#SBATCH --qos=batch
+#SBATCH --time=00:01:00
+""".strip()
```
christinaholtNOAA, line 37: I'd hope to see that this batch script would have the mpirun command in it.

christinaholtNOAA (changes requested):
I've left quite a few comments below with expectations that this PR will go in before @maddenp-noaa's current open one. If that does not end up being the case, I might have some additional feedback. 

It would also be good to get the tests passing, too.

src/uwtools/tests/drivers/test_forecast.py
WeirAE: At the moment, this is just checking the output of `scheduler.batch_script()` as used in `forecast()` and the mpirun command is handled by `run_cmd`... Actually now that you mention it, this brings up a question - `scheduler.batch_script()` just generates the batch system block, not the actual full .sh script with `!/bin/bash` and the run command. Is this something we should handle here in `forecast()`, or as part of `scheduler.batch_script()`?

WeirAE: I'm having trouble finding one of the completed EX scripts as recheck reference in the AWS bucket that we have

src/uwtools/resources/FV3Forecast.jsonschema
WeirAE: We had previously intended to move the model name to the YAML, do we want to revert that back to the CLI?

src/uwtools/drivers/forecast.py
WeirAE: OK, I was wondering about that. We'll have to talk about whether to include that as part of this PR or I'll have to put it in the YAML/schema to get the proper run command

src/uwtools/cli/run_forecast.py
WeirAE: OK cool, we were discussing naming conventions and this was my placeholder, but that works

src/uwtools/drivers/forecast.py
maddenp-noaa: OMG I didn't know you could make suggestions like that. Brilliant.

WeirAE: oh wow, nice! However, it hid this whole block of suggestions from me and I only now saw it so thats a quirk to be aware of 0.o

src/uwtools/drivers/forecast.py
maddenp-noaa: @christinaholtNOAA Why did you choose `.get("forecast")` here instead of `["forecast"]`? If we get a `None` result, the next executable statement will fail (no?) but in a less obvious way than if this one had failed. I'm thinking that we should be able to expect correct input (it should at least have been validated against a schema) at this spot in the code, and so should be able to confidently use `[]` lookups wherever non-optional values are expected.

src/uwtools/exceptions.py
WeirAE: This was the offline discussion suggestion that was included here - to make a parent general UWError class, of which ConfigErrors are one type, with additional ones to be included as needed

src/uwtools/drivers/forecast.py
WeirAE: ok, we might want to recheck how stage_files() is set up here and in test_forecast() too then

src/uwtools/drivers/forecast.py
maddenp-noaa: Probably a discussion for another place/time, but practice for a long time has been to export NOTHING from the front-end / head nodes to the batch system, but rather to explicitly set any environment variables needed in scripts. The reason is that this avoids the possibility that the same operation mysteriously fails for one user that succeeds for another, or even fails/succeeds at different times, based on whatever might be in the user's shell at the time. Much better IMO is to export nothing, start in a pristine environment on the batch system, and explicitly arrange for precisely the required environment. I've wasted a lot of time helping people troubleshoot nonsense arising from this IMO anti-pattern of exporting shell environments to the batch system. For your consideration, and maybe future discussion.

src/uwtools/drivers/forecast.py
maddenp-noaa: There are `Slurm` and `PBS` classes in `scheduler.py`: These facts about the respective batch managers could be class attributes or static/class methods (the latter would allow us to declare them abstract in the `JobScheduler` base class to ensure that they are implemented in subclasses); or maybe (abstract in `JobScheduler`, concrete in `Slurm` / `PBS`) instance methods since IIRC we can get an instance in `FV3Forecast` (it looks like we already get, and then discard, and instance in the statement `return JobScheduler.get_scheduler(job_resources).batch_script`, so we could obtain it, cache it, and call it for any facts about the batch manager we with e.g. a method on `FV3Forecast` like
```
from functools import cached_property
...
@cached_property
def batch_manager(self):
    return JobScheduler.get_scheduler(self.config[...])
```
That assumes that, as @christinaholtNOAA suggested, `self.config` is defined in the constructor, by instantiating a `YAMLConfig` object.

(It may not be much of / any optimization to cache this value, but it also shouldn't be _necessary_ to obtain it more than once, I think.)

Then, elsewhere in the class, code could call methods on this object like `self.batch_manager.batch_script` (already exists) or `self.batch_manager.submit_command` (doesn't exist yet).

src/uwtools/drivers/driver.py
```diff
@@ -3,29 +3,41 @@
...
+        self._dry_run = dry_run
         self._config_file = config_file
+        self._outfile = outfile
         self._validate()
 
     # Public methods
 
     @abstractmethod
-    def batch_script(self) -> None:
+    def batch_script(self, job_resources: Mapping) -> BatchScript:
```
maddenp-noaa, line 40: The `job_resources` argument is called, if I'm not mistaken, `platform_resources` at the call site. Any reason for two names for (I think) the same concept?

src/uwtools/drivers/forecast.py
maddenp-noaa: Upon reflection, we probably don't want both "batch manager" and "job scheduler" terminology and should pick one or the other. "Since "job scheduler" is already in the codebase, maybe we should stick with that.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Would we have access to this fact from the CLI, e.g. because we'd run `uw forecast run --model=fv3` or similar? Certainly by the time we've instantiated an `FV3Forecast` object we know that we're running FV3, but would we want that information to come from CLI or YAML? (Not both, I'd probably argue.)

src/uwtools/tests/drivers/test_forecast.py
maddenp-noaa: I think I'd opt to have `JobScheduler.batch_script()` include the initial `#!/bin/bash` line (because it will always be needed) and a trailing newline after the `#SBATCH` block; and then, in `FV3Forecast`, append the return value of `self.run_cmd()` to actually run FV3.

src/uwtools/drivers/forecast.py
christinaholtNOAA: Shoot. I was being sloppy with the `get("forecast")` instead of `["forecast"]`. I'm sorry for the confusion. I think we can rely on it to be there at this point in the code. 

The part that I was getting at is that we want to go ahead and load the _contents_ of the forecast section into the `forecast_config` variable and remove that top level so that we don't have to search as deep in all the other references.

src/uwtools/drivers/forecast.py
christinaholtNOAA: I think the `stage_files` method doesn't care which way it goes, so any combination would be fine. This request is because static files are typically staged on disk and don't often move or change (although it can happen), while many times your cycle-dependent files will be scrubbed during the course of your experiment. It's safer to keep those big files with your experiment.

src/uwtools/drivers/forecast.py
christinaholtNOAA: Just to clarify, as above, this would also be fine syntactically:

```
platform = forecast_config["platform"]
```

I wasn't sure why dictionary comprehension was used.

src/uwtools/drivers/forecast.py
maddenp-noaa: I'm not suggesting a change now, but just throwing this out for maybe future consideration: Making a hardlink when possible (that is, when source and destination files are on the same filesystem) would be an ideal first thing to try, falling back to a copy if that fails: A hardlink uses the same underlying blocks on disk, and therefore 1. is instantaneous and 2. uses no extra disk space. If the "source" is removed, the "destination" remains. Hardlinks will fail either if the source is on a different filesystem, or if the user doesn't have write access to the source. But for cases where those requirements are met, it can save a lot of time and space. For example, if a user has their own collection of files and is running many experiments against them, hardlinks can avoid eating in to their quota.

src/uwtools/exceptions.py
christinaholtNOAA: I think that inheritance alone gives you that for these simple Exception classes and you wouldn't need it to also be abstract.

src/uwtools/resources/FV3Forecast.jsonschema
christinaholtNOAA: Oh. I see. In SRW `EXTRN_MDL_NAME` is the variable that defines which system to get initial and boundary conditions from. We do not want to use this variable if what we are after is `model` to mean that we are looking at fv3, mpas, etc. Maybe let's go with `model` as in @maddenp-noaa's example. It can live here as we'd discussed previously.

src/uwtools/drivers/forecast.py
WeirAE: Hm, probably a remnant of an earlier version to be honest

src/uwtools/drivers/forecast.py
WeirAE: This is pulled from the SRW most common srun arg, so this may be more complicated to change there

src/uwtools/exceptions.py
WeirAE: @maddenp-noaa If this works for what you were thinking, I'll revert back no prob

maddenp-noaa: I think it comes down to the question: Do we want to be able to raise a `UWError`, or force ourselves to raise something more specific than that. I think I'd err on the side of removing `ABC` from the parent classes and accepting that we might raise a fairly generic `UWError if a more-specific exception isn't obvious.

The other half of the original question was: Do we now have a use case for a new _more-specific_ error? Are we raising `UWConfigError` where the issue isn't actually a *configuration* error? If so, we could raise the more generic `UWError` instead, or define a new subclass to describe the actual problem. I'm not saying there IS an opportunity for change here, just suggesting that it's something to keep in mind.

But I'm fine with what's here (probably minus the `ABC`), as it at least lays the groundwork for more specific error types.

src/uwtools/drivers/forecast.py
WeirAE: From today, moving this to "--export=None"

WeirAE:
Many changes implemented - some weird formatting changes though: I've got a lot of single lines, and some issues where the `make formatter` is reversing the import order in `test_logger` causing the make test to fail.
Still in progress: Fixing test mocking, updating `JobScheduler.batch_script()`, tweaking the YAML/JSONSCHEMA, and fixing file staging

maddenp-noaa:
@WeirAE If import order causes tests to pass/fail, something else is wrong. I'd be happy to take a look if you can point me to the code in question, especially if I can fetch your code and reproduce the failure myself.

WeirAE:
> 
@maddenp-noaa make format && make test is causing the following:
Module uwtools.tests.test_logging
uwtools/tests/test_logging.py:11:0: C0411: third party import "from pytest import raises" should be placed before "import uwtools.logging" (wrong-import-order)

maddenp-noaa:
@WeirAE Is that code pushed? I just did a `git fetch` on your fork, checked out the `uw-10` branch, and `make format && make lint` pass for me. Can you push if your remote is not up-to-date?

src/uwtools/drivers/forecast.py
WeirAE: I've moved both to link for now after today's discussion

src/uwtools/drivers/forecast.py
WeirAE: Hm - for defining in the constructor, I've moved  `self.config_data = config.YAMLConfig(self._config_file)` into `Driver.__init__` which makes it more broadly accessible. This does annoyingly require me to rework all the tests that had been getting away with `fcstobj = FV3Forecast(config_file="/not/used")` - working on it!

src/uwtools/cli/set_config.py
```diff
@@ -7,11 +7,12 @@
 from argparse import ArgumentError, ArgumentParser, HelpFormatter, Namespace
 from typing import List
 
-from uwtools import config
 from uwtools.exceptions import UWConfigError
 from uwtools.logging import setup_logging
 from uwtools.utils import cli_helpers
 
+from uwtools import config
```
maddenp-noaa, line 14: I get a lot of changes when I run `make format`. Could you run it in your clone and commit the changes? A `make format && make test` passes all tests for me, but reformats the imports in a bunch of files. For example, here, this import should be in a block with the other `uwtools` imports. But `make format` should take care of that.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +138,75 @@ def requirements(self) -> None:
...
+        if self._dry_run:
+            # Apply switch to allow user to view the run command of config.
+            # This will not run the job.
+            logging.info("Batch Script:")
+            logging.info(batch_script)
+            return
+
+        # Run the job.
+        if self._batch_script is not None:  # pragma: no cover
+            outpath = run_directory.join(self._batch_script)
```
maddenp-noaa, line 193: What is `outpath` set to here? If I try this in the REPL I get
```
>>> "/some/dir".join("a-file")
'a/some/dir-/some/dirf/some/diri/some/dirl/some/dire'
```
Unless I'm missing something, it seems like this would have caused problems in a unit test -- but maybe the `# pragma: no cover` is masking that, so it would be good to remove that.

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +138,75 @@ def requirements(self) -> None:
...
+            with open(outpath, "w+", encoding="utf-8") as file_:
+                print(batch_script, file=file_)
+                batch_command = JobScheduler.get_scheduler(platform_resources).submit_command
+            subprocess.run(
+                f"{batch_command} {outpath}",
+                stderr=subprocess.STDOUT,
+                check=False,
+                shell=True,
+            )
+        else:  # pragma: no cover
```
maddenp-noaa, line 203: Please remove `# pragma: no cover`.

src/uwtools/exceptions.py
```diff
@@ -3,7 +3,13 @@
 """
 
 
-class UWConfigError(Exception):
+class UWError(Exception):
+    """
+    An abstract class representing generic exceptions for UWTools.
```
maddenp-noaa, line 8: We can remove `abstract` from the docstring as that's no longer accurate.

src/uwtools/scheduler.py
```diff
@@ -151,6 +152,13 @@ def batch_script(self) -> BatchScript:
 
         return BatchScript(sorted(processed))
 
+    @property
+    def submit_command(self) -> str:
+        """
+        Returns the command for running a batch script.
+        """
+        return self.command
```
maddenp-noaa, line 160: `command` is probably too generic: There are other batch-related commands, like ones to interrogate job status (which we may need at some point), one to cancel jobs (may or may not ever need), etc. So this should probably be `self.submit_command`.

src/uwtools/tests/cli/test_run_forecast.py
```diff
@@ -18,25 +17,36 @@
...
 def files(tmp_path):
     cfgfile = tmp_path / "cfg.yaml"
-    machinefile = tmp_path / "machine.yaml"
-    for path in cfgfile, machinefile:
-        path.touch()
-    return str(cfgfile), str(machinefile)
+    logfile = tmp_path / "log"
+    for fn in cfgfile, logfile:
+        with open(fn, "w", encoding="utf-8"):
+            pass
```
maddenp-noaa, line 23: Since these files are `Patch` objects, these two lines could be replaced with the more concise `fn.touch()`.

maddenp-noaa (changes requested):

src/uwtools/cli/set_config.py
WeirAE: Aha - that's part of that issue I've been having where the `make format` keeps trying to break `test_logger`, there's clearly something wrong in my format definition

src/uwtools/drivers/forecast.py
WeirAE: Interesting - thanks for catching that. I was thinking about how to catch the true run in a unit test without it actually running, which is why I had been leaving it out

src/uwtools/scheduler.py
WeirAE: Yeah, this is what I was pondering in our meeting yesterday, but `submit_command` is better than `batch_command` or anything like that I was focusing on

src/uwtools/tests/cli/test_run_forecast.py
WeirAE: Ah, thats a nice clean up to get in here too!

src/uwtools/cli/set_config.py
WeirAE: I have `isort 5.12.0` and `black 23.3.0` so I'm not sure why its scrambling on me

src/uwtools/drivers/forecast.py
maddenp-noaa: There are a couple small gaps in my calendar today -- send me an invitation if you'd like to pair on this.

src/uwtools/cli/set_config.py
maddenp-noaa: I have the same versions in my environment. We could look at this in a pairing session if you don't figure out what's up.

WeirAE: Fixed - just had to clean something weird out of git and remake it

src/uwtools/drivers/forecast.py
WeirAE: replaced with `Path(run_directory) / self._batch_script`

maddenp-noaa (approved):
Thank you for your tireless effort and openness to feedback!

NaureenBharwaniNOAA (approved):
LGTM!

src/uwtools/drivers/forecast.py
```diff
@@ -121,15 +138,75 @@ def requirements(self) -> None:
         ???
         """
 
-    def resources(self) -> None:
+    def resources(self, platform: dict) -> Mapping:
         """
         ???
         """
```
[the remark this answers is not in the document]
WeirAE: Thanks! There was a previous comment that requested reverting to an empty string, but I for one do prefer putting something there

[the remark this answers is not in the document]
WeirAE: Ah, that had been on `Driver` whew lost track there, thanks for noting this one!

christinaholtNOAA (approved):
LGTM!

PR #277, opened 2023-08-11
UW-259 As a user, I need to handle Rocoto XML config files

NaureenBharwaniNOAA:
**Description**
The Config Parser tool in the UW Framework must be able to support parsing a structured YAML file and output a Rocoto XML File.

NOTE: SRW already has an XML Generator tool that uses a variety of methods in the framework of SRW. The current XML file is created through a very generic JINJA 2 template. Christina can share a recorded demo of the SRW tool with individuals on the team.

**Type of change**

**How Has This Been Tested?**

**Checklist**


src/uwtools/resources/rendered_xml.xml
maddenp-noaa, line 1: I think we could improve the name of this file. The `_xml` part is repetition given the `.xml` extension, and the content is not yet rendered. Maybe something like `rocoto.jinja2`, to clarify that it's not yet XML at all but, in fact, Jinja2 syntax (mostly)?

src/uwtools/tests/fixtures/hello_workflow.yaml
```diff
@@ -0,0 +1,39 @@
...
+      var:
+        member: foo bar baz
+      task_hello_#member#:
+        attrs:
+          cycledefs: howdy
+          maxtries: "1"
+        account: "&ACCOUNT;"
+        command: "echo hello #member#"
+        nodes: 1:ppn=1
+        walltime: 00:01:00
```
NaureenBharwaniNOAA, line 34: It seems after running _add_jobname on my side that an extra line appears in the actual generated from the function, compare to the created hello_workflow.xml. Do we need to add a jobname here? @christinaholtNOAA

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
+"""
+Support for writing a Rocoto XML using J2Templating
+"""
+
+from uwtools.j2template import J2Template
+from uwtools.config import YAMLConfig
+
+# private functions
```
maddenp-noaa, line 8: `# Private functions`, and `# Main function` to `# Public functions` down below, for consistency with other modules, please.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str) -> None:
+    """
+    Main entry point.
+
+    :param input_yaml: Yaml file provided by the user.
+    :param input_template: Path to input template file.
+    :param rendered_output: Path to directory to write rendered XML file.
+    """
+    # make an instance of the YAMLConfig class
+    # pass the input yaml to return a dict
```
maddenp-noaa, line 38:
```
# Comments should be sentences with an initial capital letters and punctuation.
```

src/uwtools/tests/fixtures/hello_workflow.yaml
```diff
@@ -0,0 +1,39 @@
...
+        dependency:
+          taskdep:
+            attrs:
+              task: hello
+
```
maddenp-noaa, line 39: Please remove trailing newline.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
+"""
+Support for writing a Rocoto XML using J2Templating
```
maddenp-noaa, line 2: The use of `J2Template` is an implementation detail that isn't useful information for the person who'd be reading the docstring, which may be translated into HTML documentation. Maybe
```
Support for creating Rocoto XML workflow documents
```
?

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+
+# private functions
+def _add_jobname(tasks: dict) -> None:
+    """
+    Add the jobname entry for all the tasks in the workflow.
+
+    :param tasks: Dict of tasks in workflow.
+    """
+    if not isinstance(tasks, dict):
+        return
```
maddenp-noaa, line 16: These two lines are not needed because the `tasks` argument is type-annotated and we use typechecking. In an API method exposed to outside users we might want to do such a check.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+    for task, task_settings in tasks.items():
+        task_type = task.split("_", maxsplit=1)[0]
+        if task_type == "task":
+            # Use the provided attribute if it is present, otherwise use
+            # the name in the key
+            tasks[task]["jobname"] = \
+                task_settings.get("attrs", {}).get("name") or \
+                task.split("_", maxsplit=1)[1]
```
maddenp-noaa, line 24: The expression `task.split("_", maxsplit=1)` is repeated a few lines up. Could we assign this value to a variable and then reference that variable in both places?

I realize that this code was copied from SRW, but now's our best chance to improve it a bit.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+        elif task_type == "metatask":
+            _add_jobname(task_settings)
+
+# Main function
+def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str) -> None:
+    """
+    Main entry point.
+
+    :param input_yaml: Yaml file provided by the user.
```
maddenp-noaa, line 33: I'd recommend
```
:param input_yaml: Path to YAML input file.
```
because 1. The other two path argument descriptions all start (helpfully) with "Path to", 2. `YAML` is an acronym and so should be in all-caps, and 3. "the user" is not a useful concept here -- there may be no user, for example via a call from a unit test or some other programmatic use case.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+    :param input_template: Path to input template file.
+    :param rendered_output: Path to directory to write rendered XML file.
+    """
+    # make an instance of the YAMLConfig class
+    # pass the input yaml to return a dict
+    config_obj = YAMLConfig(input_yaml)["rocoto"]
```
maddenp-noaa, line 39: We (@NaureenBharwaniNOAA and I) added this `["rocoto"]` lookup to make the tests pass, but probably this doesn't belong here. The unit test could do this initial extraction of the required sub-`dict`, or the test-fixture file could omit the top-level `rocoto` key. Thoughts @christinaholtNOAA ?

src/uwtools/tests/test_rocoto.py
```diff
@@ -0,0 +1,67 @@
...
+        input_template=resc / "rocoto.jinja2"
+    output = tmp_path / "rendered.xml"
+    rocoto.write_rocoto_xml(input_yaml=input_yaml, input_template=str(input_template), rendered_output=str(output))
+
+    expected = support.fixture_path("hello_workflow.xml")
+    assert True == support.compare_files(expected, output)
+    #with open(expected, "r", encoding="utf-8") as f:
+    #    with open(tmp_path / output, "r", encoding="utf-8") as f2:
+    #        assert f.read() == f2.read()
+
```
maddenp-noaa, line 67: Please remove dead code and trailing newline.

maddenp-noaa (changes requested):
A few initial comments. I may have more once these are addressed.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I agree with this. It makes it simpler when take a quick glance at the code.

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,43 @@
...
+
+# Private functions
+def _add_jobname(tasks: dict) -> None:
+    """
+    Add the jobname entry for all the tasks in the workflow.
+
+    :param tasks: Dict of tasks in workflow.
+    """
+    for task, task_settings in tasks.items():
+        task_split = task.split("_", maxsplit=1)
```
maddenp-noaa, line 17: Here's an idea/technique for you: We can "destructure" the collection that we get back from `.split()` and assign the result to variables. Usually in Python we'd destructure into a number of variables matching the size of the collection, e.g.
```
a, b, c, d = [1, 2, 3, 4]
>>> a, b, c, d = [1, 2, 3, 4]
>>> a
1
>>> d
4
```
But we can also destructure into a smaller number, assigning the remainder to another collection (`list` in this case):
```
>>> a, b, *rest = [1, 2, 3, 4]
>>> a
1
>>> b
2
```
`rest` is a traditional name for "the rest of" something, but in case the linter doesn't like it because `rest` is then unused, you can use `*_` instead: `_` is a traditional name of an unused variable, and `pylint` knows that and will not complain that you do not subsequently use it.

So, if we can think of good names for the `[0]` and `[1]` elements that we're pulling out of `task` after splitting it, we could `foo, bar, *_ = task.split("_", maxsplit=1)` and then just refer to -- hopefully meaningful -- names `foo` and `bar` later without using indexes. `foo` -> `task_type` and `bar` -> `task_name` maybe?

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,43 @@
...
+        task_type = task_split[0]
+        if task_type == "task":
+            # Use the provided attribute if it is present, otherwise use.
+            # The name in the key.
```
maddenp-noaa, line 21: Looks like maybe these two comments should be a single sentence instead of two.

src/uwtools/tests/test_rocoto.py
```diff
@@ -0,0 +1,72 @@
...
+  task_hola_#mem#:
+    command: echo hello 
+    jobname: hola_#mem#
+  metatask_hey_#mem#:
+    var:
+      day: mon tues
+    task_hey_#mem#_#day#:
+      command: echo hello 
+      jobname: hey_#mem#_#day#
+"""
```
maddenp-noaa, line 36: Honest question: Is this block of YAML more than we actually need to test the `add_jobname()` functionality? It's not a problem to use incomplete, unrealistic bits of data in tests; in fact, having extraneous data that is not actually used by the code-under-test can distract from understanding the test. I'd probably recommend removing unneeded data here if it's not needed for the test.

src/uwtools/tests/test_rocoto.py
```diff
@@ -0,0 +1,72 @@
...
+    input_yaml = support.fixture_path("hello_workflow.yaml")
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        input_template = resc / "rocoto.jinja2"
+    output = tmp_path / "rendered.xml"
+    rocoto.write_rocoto_xml(
+        input_yaml=input_yaml, input_template=str(input_template), rendered_output=str(output)
+    )
+
+    expected = support.fixture_path("hello_workflow.xml")
+    assert True == support.compare_files(output, expected)
```
maddenp-noaa, line 72: `assert support.compare_files(output, expected)` should be sufficient: `assert` is only looking for a truthy value as its argument, and `compare_files()` returns a `bool`, so that should do. (Similarly, for negative tests, `assert not x` would be more idiomatic than `assert x == False` or `assert x is False`, `assert False == x`, etc.)

src/uwtools/tests/fixtures/hello_workflow.yaml
```diff
@@ -0,0 +1,38 @@
...
+    FOO: test.log
+  log: /some/path/to/&FOO;
+  tasks:
+    task_hello:
+      attrs:
+        cycledefs: howdy
+        maxtries: "2"
+      account: "&ACCOUNT;"
+      command: "echo hello $person"
+      jobname: hello
```
NaureenBharwaniNOAA, line 19: Taking a look at this again, is `jobname` suppose to be added here in the yaml? I believe we are adding jobname using the `_add_jobname` function and the yaml shouldn't be providing this? If that's the case then I'm not sure `_add_jobname `is correctly iterating through the config_obj passed from write_xml_rocoto. If jobname is suppose to be included here then should we be including it below for `task_hello_#member#:` as well? @christinaholtNOAA

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: I think making testing the functionality of the `_add_jobname` can be made smaller such as:

`def test_add_jobname():
    expected = yaml.safe_load(
        """
task_hello:
  command: echo hello 
  jobname: hello
metatask_howdy:
  var:
    mem: 1 2 3
  task_howdy_#mem#:
    command: echo hello 
    jobname: howdy_#mem#` 

Thoughts? @christinaholtNOAA and @maddenp-noaa

src/uwtools/tests/test_rocoto.py
```diff
@@ -0,0 +1,73 @@
...
+    command: echo hello
+  metatask_hey_#mem#:
+    var:
+      day: mon tues
+    task_hey_#mem#_#day#:
+      command: echo hello
+"""
+    )
+
+    # pylint: disable=protected-access
```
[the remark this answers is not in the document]
maddenp-noaa: That's the right fix, but since accessing protected (private) members is commonplace in unit tests, this should go at the top of the test module so that it applied throughout, to avoid needing to repeat it. You'll see that a lot of other test modules have this `pylint` annotation.

src/uwtools/rocoto.py
christinaholtNOAA: Yes, this is likely the caller's responsibility.

src/uwtools/tests/test_rocoto.py
christinaholtNOAA: Here's the rub. Each of these `task_` sections needs to be a dictionary on input to `_add_jobname`, otherwise the logic breaks down. We can certainly remove the `var:` entries under metatasks without a problem.

I'm also curious if it makes sense to require the input type to be a dict of dicts...

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I attempted to implement this solution and found that at certain parts specifically when `task_split = ['var']` this becomes an issue because not enough values are supplied to unpack `(expected 2, got 1)`. I used `rest` to help alleviate the issue and then indexed into the list from there. Let me know if this is what you were looking for

src/uwtools/rocoto.py
maddenp-noaa: @NaureenBharwaniNOAA Given that, I think we should remove the `["rocoto"]` bit here and employ one of the two ideas above to have the caller pass in a `dict` without that top level. Please feel free to resolve this conversation when you've made a fix.

src/uwtools/tests/fixtures/hello_workflow.yaml
christinaholtNOAA: This entry likely should not go here.

src/uwtools/tests/test_rocoto.py
christinaholtNOAA: I missed the last part that Naureen asked. Do we want to simplify the input dict, or make sure that several tasks and nested metatasks are tested, or is the abridged version above enough?

src/uwtools/rocoto.py
maddenp-noaa: The destructuring I described wouldn't work if there are sometimes less than 2 elements in `task.split("_", maxsplit=1)`. I see now that that can happen in a `metatask_*` subtree. It seems to me that the names `task` and `task_settings could be generalized, as it seems that something other than "tasks" can be assigned to these. Maybe something like
```
for element, subtree in tree.items():
    element_parts = element.split("_", maxsplit=1)
    element_type = element_parts[0]
    if element_type == "task":
        task_name = element_parts[1]
        element["jobname"] = subtree.get("attrs", {}).get("name") or task_name
    elif element_type == "metatask":
        _add_jobname(subtree)
```
? (That also renames `tasks` to `tree`, as the tree isn't necessarily composed only of task entires, apparently.)

src/uwtools/tests/test_rocoto.py
maddenp-noaa: I think we need at least one task and one metatask, as otherwise the recursive step won't be tested. I don't think we necessarily need more than one of each, though.

NaureenBharwaniNOAA: I simplified the expected output and task example down further, removing `var` altogether.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: That makes sense to me. Using element and subtree will provide a better understanding since it can be misleading that not every item fits into `tasks` category. I'll update this shortly.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: It might be good to keep one non-task element in the tree, even something like `foo: bar`, to demonstrate that nothing goes wrong if a non-task element is present. Assuming that that's something we should expect to happen in practice. @christinaholtNOAA ?

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I think this also puts into question if we should be renaming the input parameter from `tasks` to something more general such as `tree`? Same can be said for the docstring on this function. Thoughts? @maddenp-noaa

src/uwtools/tests/test_rocoto.py
```diff
@@ -0,0 +1,54 @@
...
+Tests for uwtools.rocoto module.
+"""
+from importlib import resources
+
+import yaml
+
+from uwtools import rocoto
+from uwtools.tests import support
+
+# pylint: disable=protected-access
```
maddenp-noaa, line 12: Please combine with line l:
```
# pylint: disable=missing-function-docstring,protected-accesss
```

src/uwtools/rocoto.py
maddenp-noaa: I agree, and tried to say so perhaps a bit vaguely, with
> (That also renames `tasks` to `tree`, as the tree isn't necessarily composed only of task entires, apparently.)
I'm glad that we both thought of the same thing!

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: Just for the future in case someone sees this, it took me a second to realize there was an extra s on the end of access.
`# pylint: disable=missing-function-docstring,protected-access`

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,44 @@
...
+from uwtools.j2template import J2Template
+
+# Private functions
+
+
+def _add_jobname(tree: dict) -> None:
+    """
+    Add the jobname entry for all the tasks in the workflow.
+
+    :param tree: Dict of tree  in workflow.
```
maddenp-noaa, line 15: Alternative phrasing for your consideration:

1. `Add a "jobname" attribute to each "task" element in the given config tree`.

To me, "the jobname" could mean that there's a single value instead of different values for each task; and "for all the tasks" is different than "to each task", in that the latter means we're taking action on multiple "things", rather than maybe just creating a single thing that is somehow _about_ all those things. Does that make sense?

2. `:param tree: A config tree containing "task" elements.`

By looking at the function signature, we can see that `tree` is a `dict`, so there's no need to repeat that. "workflow" is not a concept otherwise present in this function, so it might not be useful to introduce it here; instead this param tag uses the terms `task` and `element`, which also appear as concepts in the code.

What do you think? Do those make sense and would they provide any more clarity to a reader not necessarily familiar with this module and function?

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# Public functions
+def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str) -> None:
+    """
+    Main entry point.
+
+    :param input_yaml: Path to YAML input file.
+    :param input_template: Path to input template file.
+    :param rendered_output: Path to directory to write rendered XML file.
```
maddenp-noaa, line 35: Is `rendered_output` a directory, or a file?

NaureenBharwaniNOAA: `rendered_output` should be a path to write the rendered XML file. I'll up the `param` for this

src/uwtools/rocoto.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    # Make an instance of the YAMLConfig class.
+    # Pass the input yaml to return a dict.
```
maddenp-noaa, line 38: `yaml` -> `YAML`, as it's an acronym.

Also, the return type of `YAMLConfig` if a `YAMLConfig` object which -- as the typechecker made so painfully clear -- is not a `dict`. In fact, since it's pretty clear that the line just below is a call to an object constructor, it shouldn't be necessary to mention, in a comment, a return type: The return type is the class whose constructor is being called.

All considered, I don't think these two lines of comment are very useful, as the following line of code says essentially just that.

maddenp-noaa (approved):
Thanks for patiently entertaining (and implementing, where appropriate) the long string of comments. This looks great.

christinaholtNOAA (approved):
I agree.

PR #278, opened 2023-08-11
mandatory-driver-config

maddenp-noaa:
**Description**

Don't Panic.

There's a big diff, but the vast majority is due to code moving in its module, due to organizing methods into public and private sections and lexicographically ordering them within their sections. I've made comments on actual, interesting code changes.

- As discussed in a previous PR where it was deemed out-of-scope, make the `config_file` argument to `Driver` and its subclasses mandatory instead of optional.
- Rework `Driver` (and subclass) methods either as users of the `config_file` value provided via the constructor, or as static methods.
- Mark `Driver` (and subclass) methods as either public or private, group and sort.
- Update unit tests accordingly.
- Update docstrings, leaving some to-be-written when their methods are implemented.
- Replace "job card" terminology with "batch script".

**How Has This Been Tested?**

Passes dev-shell `make test` and GitHub Actions.

**Checklist**


src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
-        else:
-            self.log.info("No config file provided, available functions are limited.")
+        self._config_file = config_file
+        self._validate()
 
-    def _validate(self) -> bool:
+    # Public methods
+
+    @abstractmethod
+    def batch_script(self) -> None:
```
maddenp-noaa, line 28: Prefer "batch script" terminology to "job card". For example, compare the results returned from Google for query `"slurm" "batch script"` vs `"slurm" "job card"` (or replace `"slurm"` with `"pbs"`). While "job card" is not _totally_ unheard of, it seems clear to me that "batch script" is the much more standard terminology.

src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
-        The schema to validate the config file against.
+        ???
         """
 
     @abstractmethod
-    def requirements(self):
+    def requirements(self) -> None:
         """
-        Recursively parse config and platform files to determine and fill in any dependencies.
+        ???
```
maddenp-noaa, line 42: I wasn't able to make sense of some of the existing docstrings where there's no implementation yet, so I blanked these out to be completed later. In general, docstrings should describe what a method does at a high level, without delving into implementation specifics. For example, a user running `help(fcstobj.requirements)` wouldn't need to know that recursion is being used -- they need to know what the method needs as input, what it returns, and what side effects it might have; they don't need to know _how_ the work gets done. `#` comments in code are the preferred place for algorithmic details.

I also added `-> None` return-type annotations, though these will probably need to be updated when the methods are actually implemented.

src/uwtools/drivers/forecast.py
```diff
@@ -16,74 +16,55 @@
...
 
-    def requirements(self):
+    def batch_script(self) -> None:
         """
-        Recursively parse config and platform files to determine and fill in any dependencies.
+        Write to disk, for submission to the batch system, a script to run FV3.
         """
 
-    def resources(self):
+    @staticmethod
```
maddenp-noaa, line 29: One of several methods that were reformulated as static methods, so that they can be called without first instantiating an `FV3Forecast` object.

src/uwtools/drivers/forecast.py
```diff
@@ -185,3 +166,23 @@ def stage_files(
                 shutil.copyfile(src_path, dst_path)
             msg = f"File {src_path} staged in run directory at {dst_fn}"
             logging.info(msg)
+
+    # Private methods
+
+    def _create_model_config(self, base_file: str, outconfig_file: str) -> None:
```
maddenp-noaa, line 172: It may turn out that this method should be public, but starting private and upgrading to public is the conservative choice.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -30,37 +30,32 @@ def test_schema():
...
-    Tests create_directory_structure method given a directory.
+    Tests _create_directory_structure method given a directory.
     """
 
     rundir = tmp_path / "rundir"
-    forecast_obj = FV3Forecast()
 
     # Test delete behavior when run directory does not exist.
-    forecast_obj.create_directory_structure(rundir, "delete")
+    FV3Forecast.create_directory_structure(rundir, "delete")
```
maddenp-noaa, line 58: Instead of creating a no-config `FV3Forecast` object, simply test the now-static method.

src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
         """
-        Initialize the Forecast driver.
+        Initialize the driver.
         """
         self.log = log if log is not None else Logger()
-        if config_file is not None:
-            self.config_file = config_file
-            self._validate()
-        else:
-            self.log.info("No config file provided, available functions are limited.")
```
maddenp-noaa, line 69: No special treatment needed now that `config_file` is mandatory.

src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
+        Run the NWP tool.
         """
-        Holds the knowledge for how to modify a list of output files and stages them in the working
-        directory.
 
-        Output files usually are specific to a given app.
+    @abstractmethod
+    def run_cmd(self, *args, run_cmd: str, exec_name: str) -> str:
+        """
+        The command-line command to run the NWP tool.
```
maddenp-noaa, line 60: I added `run()` and `run_cmd()` to the set of abstract methods. These were already implemented for the `FV3Forecast` driver, but would we have a use case of a `Driver` subclass that would **not** need these methods?

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -186,36 +180,36 @@ def test_forecast_run_cmd():
...
-        "-ppn",
-        8,
-        "--cpu-bind",
-        "core",
-        "-depth",
-        2,
-        run_cmd="mpiexec",
-        exec_name="test_exec.py",
-    )
+    with patch.object(FV3Forecast, "_validate", return_value=True):
```
maddenp-noaa, line 183: Now that `config_file` is mandatory, its value will automatically be validated against its schema by the constructor of `Driver` and its subclasses. To avoid the necessity of providing test files that can actually be schema-validated, mock the `_validate()` method so that it always returns `True` and does no actual validation.(We know that `_validate()` works correctly because it has its own unit tests -- testing `_validate()` is not the point of **this** test, so we can just neutralize it with a mock.)

src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
-    This base class provides the interface to methods used to read in a user-provided YAML
-    configuration file, convert it to the required config file, fix files, and namelist for a
-    forecast. Subsequent methods will be used to stage the input files and run the forecast.
-
-    Attributes
-    ----------
-    config_path : Path
-        The file path to the configuration file to be parsed.
-
-    Methods
```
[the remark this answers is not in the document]
WeirAE: Yeah, at the time this was sort of a pattern of previous scaffoldings but agreed its redundant now

src/uwtools/drivers/driver.py
WeirAE: Interestingly, both are used within SRW - it looks like "batch script" specifically refers to `rt.sh`,  which leads to somewhat mixed usage like `sbatch job_card`. Here job_card parses files like `fv3_slurm.IN_gaea` which is the sort of sbatch file we're trying to generate here. Looks like we're already at one of those conflicting terms things!

src/uwtools/drivers/driver.py
WeirAE: Thats fair - I had put text here to effectively propose the planned structure as part of the facade, though its not needed now, not to mention depreciated

src/uwtools/drivers/driver.py
WeirAE: I went back to the original PR #214 and I think the intent was to include them as abstract methods but to not assume  they could be defined fully in the base class. Which is to say, I missed making sure they were in the base!

NaureenBharwaniNOAA (commented):
I appreciate the description on this PR. It made understanding the diffs and code movement simple to follow. Great additional comments as well on code changes. If we go with batch script over job card, this PR should be all set to approve.

src/uwtools/drivers/driver.py
WeirAE: So it doesn't get lost in the mix, let me @christinaholtNOAA on this one for an executive decision

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -186,36 +180,36 @@ def test_forecast_run_cmd():
...
+            2,
+            run_cmd="mpiexec",
+            exec_name="test_exec.py",
+        )
 
 
 @pytest.mark.parametrize("section", ["static", "cycledep"])
 @pytest.mark.parametrize("link_files", [True, False])
-def test_stage_files(tmp_path, section, link_files):
+def test__stage_files(tmp_path, section, link_files):
```
christinaholtNOAA, line 210: It doesn't seem that `stage_files()` became private after all.

src/uwtools/drivers/driver.py
christinaholtNOAA: Let's choose "batch script" as the modern-era, single term we use to describe this thing.

christinaholtNOAA (dismissed):
I'm happy with the use of "batch script" instead of "job card". 

I found only a minor thing that we might change in the tests for consistency, but it looks good to me, otherwise.

src/uwtools/drivers/driver.py
```diff
@@ -16,103 +11,70 @@
...
-            log=self.log,
-        )
 
-    @property
     @abstractmethod
-    def schema(self):
+    def output(self) -> None:
         """
-        The schema to validate the config file against.
+        ???
```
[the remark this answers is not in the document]
maddenp-noaa: Very interesting! Maybe we can hit, in our ongoing design discussion, so ideas about actually getting some of that Luigi-style dependency-tracking in place here.

src/uwtools/tests/drivers/test_forecast.py
maddenp-noaa: Fixing.

christinaholtNOAA (approved):

PR #279, opened 2023-08-14
Validation improvements

maddenp-noaa:
**Description**

- Factored the two types of config check -- conformance to schema, and existence of filesystem paths -- into separate functions, each respecting Single Responsibility Principle.
- Concentrated all IO (except for `.exists()` calls on `Path` objects to check if they exist) in function `config_is_valid()`, for easier unit testing (i.e. all checks of log messages are in tests of `config_is_valid()`).
- Fixed a bug in the old `config_is_valid()` whereby the `for value in config_obj.data[field]` loop would iterate over a `str` value of `config_obj.data[field]`, such that e.g. the string `asdf` would yield `value` values of `a`, then `s`, then `d`, then `f`. The bug was not apparent because these single-letter strings were never also variables marked as `uri` (that is, paths) in the schema. But they could be in the future. The improved code takes a different approach: It traverses the config in lock step with a traversal of the schema, which should (if the schema is complete) have the same general shape. For each value in the config, check its corresponding schema entry and, if it has `uri` format, interpret it as a filesystem path and then verify that the named path exists. This also permits checks of paths at _any_ level in the config, not just at the top. (Long term, I'm not sure any of this is sound: It may be that a path that doesn't exist during one `uw` invocation may exist later, and may not be immediately _needed_. So it's not clear to me that just mentioning a file that does not exist (yet) should be a fatal error.)
- Expanded unit tests for functions in module `config_validator`.
- Removed several unit-test fixture files that were not in use by any tests.

**How Has This Been Tested?**

Passed `make test` in dev shell and GitHub Actions.

**Checklist**

- [ ] My changes need updates to the documentation.  I have made corresponding changes to the documentation **NA**
- [] Any dependent changes have been merged and published **NA**


src/uwtools/config_validator.py
```diff
@@ -2,61 +2,70 @@
...
-        for value in schema["properties"][field]["properties"]:
-            # the json pair "format": "uri" labels a path or file
-            if "format" in schema["properties"][field]["properties"][value]:
-                path_list.append(value)
+# Private
 
-    # Check for existence of those files or paths.
 
-    for field in config_obj.data:
-        for value in config_obj.data[field]:
```
maddenp-noaa, line 53: Here's where a `str` value of `config_obj.data[field]` would result in iteration over the individual letters of the string.

src/uwtools/tests/cli/test_validate_config.py
```diff
@@ -21,7 +21,7 @@
 def files(tmp_path):
     cfgfile = tmp_path / "cfg.yaml"
     logfile = tmp_path / "log"
-    schemafile = fixture_path("schema_test_good.yaml")
+    schemafile = fixture_path("/dev/null")
```
maddenp-noaa, line 24: `/dev/null` always exists, and since the CLI tests only need to confirm existence, and file will do. I think it's misleading to supply a real file with a real schema that is never actually used.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -87,10 +87,6 @@ def test_validation(capsys, configs, schema, tmp_path, valid):
...
         print(schema, file=f)
-    with patch.object(ConcreteDriver, "schema", new=schema_file):
+    with patch.object(ConcreteDriver, "schema_file", new=schema_file):
         ConcreteDriver(config_file=config_file)
-        assert (
-            "error(s)" not in capsys.readouterr().err
-            if valid
-            else "error(s)" in capsys.readouterr().err
-        )
+        assert ("errors" in capsys.readouterr().err) is False if valid else True
```
WeirAE, line 92: Nice clean version, appreciated!

WeirAE (dismissed):
Much cleaner, learned a lot following through this!

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -207,7 +207,7 @@ def test_forecast_run_cmd():
 
 @pytest.mark.parametrize("section", ["static", "cycledep"])
 @pytest.mark.parametrize("link_files", [True, False])
-def test_stage_files(tmp_path, section, link_files):
+def test__stage_files(tmp_path, section, link_files):
```
christinaholtNOAA, line 210: This is still a public method in the forecast driver as far as I can tell.

src/uwtools/config_validator.py
```diff
@@ -2,61 +2,70 @@
...
-    # Validate the config file against the schema file.
-
-    schema_error = 0
-    validator = jsonschema.Draft7Validator(schema)
-
-    # Print out each schema error.
-
-    errors = validator.iter_errors(config_obj.data)
+    # Load the config and schema.
+    yaml_config = YAMLConfig(config_file, log_name=log.name)
```
christinaholtNOAA, line 22: Ultimately, we would want to be able to validate any type of config we support (INI and namelists, too). I definitely think this name is fine matching the current capability, but we will likely want to rename it when we actually add that feature.

christinaholtNOAA (commented):
This looks good to me. I left a broad question and a tiny item below.

src/uwtools/tests/drivers/test_forecast.py
maddenp-noaa: Will fix, thanks!

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -207,7 +207,7 @@ def test_forecast_run_cmd():
 
 @pytest.mark.parametrize("section", ["static", "cycledep"])
 @pytest.mark.parametrize("link_files", [True, False])
-def test__stage_files(tmp_path, section, link_files):
+def test_stage_files(tmp_path, section, link_files):
     """
     Tests that _stage_files() is copying or linking files from static or cycledep sections of the
```
christinaholtNOAA, line 212: I'm so sorry to nitpick, but it's here, too in the comment.

maddenp-noaa: Fixed, thank you for noticing this. I've simplified the comment to _not_ repeat the name of the function-under-test.

christinaholtNOAA (approved):
LGTM. Thanks for allowing the nitpick. :)

Issue #272, opened 2023-08-03
UW-269 Extend set_config.py tool to "export" variables.


NaureenBharwaniNOAA:
**Description**

We will need to interface between a new YAML-configured App, and the j-jobs/ex-scripts (collectively referred to as run scripts) currently written mostly in bash. We will need to be able run a command line tool in bash that "exports" the variables in a given section of our YAML or INI file.

These sections will need to be "flat" and we will want to ensure we have namespace protection. Run scripts will need access to variables from multiple sections of an experiment config file, but it is an NCO restriction that we shouldn't export all variables for an experiment at run time. That's a good practice in general to ensure namespace protection.

So, here we will want a tool that exports all the key/value pairs in given sections to the shell in which it is invoked. An interface that looks something like is sufficient.

```export_variables.py -c /path/to/experiment.yaml --sections run_fcst workflow user```

What it does, then is loads the config, finds those sections, and ensures that the variables are available in the shell.

This need not necessarily be an extension of the set_config tool. 

**Requirements**

**_Acceptance Criteria_**
* Introduces the functionality to export variables from a config file to the existing shell.
* Must support INI and YAML input, at minimum.
* Shall provide tests that ensure environment variables are set in the shell that this script is called from.
* Shall include any relevant unit tests.

**_Contractual Requirements_**

4.5.6. Support the design, development and deployment of a Unified Workflow (UWF) for the UFS, associated tools, and documentation for multiple cloud and on-premise HPC environments, as well as delivering the UWF integration plan.

Due - Support activities, existing Confluence page and user manual updates reflecting community needs, as approved by EPT. - Online and MS Word Report via email

PRS:

* An integration plan to integrate the UWF toolkit into UFS Apps and configurations, including but not limited to, SRW, HAFS, global workflow, and land DA. Key stakeholder inputs shall be captured.
* Support scheduling, handling and managing a database for dataflow, configuration of applications, tasks and functional tools. Provides an overarching framework for applications to be configured and executed.
* Up-to-date documentation.
* Online documentation linked to EPIC and UFS document databases.

christinaholtNOAA:
Fixed by PR #280

PR #280, opened 2023-08-14
UW-269 Extend set_config.py tool to "export" variables

elcarpenterNOAA:
**Description**

For Issue [#272](https://github.com/ufs-community/workflow-tools/issues/272)

We will need to be able run a command line tool in bash that "exports" the variables in a given section of our YAML or INI file.

It was decided that the cli code was out of scope for this ticket and instead part of the modal-cli ticket planned for the next sprint. 

**Type of change**

**How Has This Been Tested?**

**Checklist**


src/uwtools/config.py
```diff
@@ -752,3 +753,17 @@ def create_config_obj(
                     raise ValueError(err_msg)
                 # Dump to file:
                 dump_method(path=outfile, cfg=config_obj)
+
+
+def export_variables(config_dict: dict, section_path: list, log: Logger) -> None:
+    """
+    Allows users to export sections of the config file as environment variables.
```
maddenp-noaa, line 760: I think using the word "export" here could be misleading, as "to export an environment variable" has a specific conventional meaning, and that's not what's happening here. I think we could just state explicitly what we _are_ doing, e.g.
> Descend into the config via the given section keys, then print the contents of the located subtree as key=value pairs, one per line.
Or something like that.

src/uwtools/config.py
```diff
@@ -752,3 +753,17 @@ def create_config_obj(
...
+    """
+    for key in section_path:
+        config_dict = config_dict[key]
+    for key, value in config_dict.items():
+        if type(value) not in (bool, str, int, float):
```
maddenp-noaa, line 765: nitpick: `(bool, float, int, str)` -- might as well order them.

src/uwtools/config.py
```diff
@@ -752,3 +753,17 @@ def create_config_obj(
...
+            log.error(f"Non-scalar variable {key} was provided")
```
maddenp-noaa, line 766: I think `variable` -> `value`. We're talking about keys and values in the context of a dictionary, rather than "variables". This also agrees better with the text of the exception we raise below.

src/uwtools/config.py
```diff
@@ -752,3 +753,17 @@ def create_config_obj(
...
                 # Dump to file:
                 dump_method(path=outfile, cfg=config_obj)
+
+
+def export_variables(config_dict: dict, section_path: list, log: Logger) -> None:
+    """
+    Allows users to export sections of the config file as environment variables.
+    """
+    for key in section_path:
+        config_dict = config_dict[key]
```
maddenp-noaa, line 763: It might be worth a check here that we've landed on the kind of subtree we expect, namely, one whose value is also a `dict`. For example, this is good:
```
>>> config = {"a": {"b": {"c": {"x": 88, "y": 99}}}}
>>> config = config["a"]["b"]["c"]  # simulate descent via keys a, b, and c
>>> for key, value in config.items():
...   print(key, value)
... 
x 88
y 99
```
but this is bad
```
>>> config = {"a": {"b": {"c": "foo"}}}
>>> config = config["a"]["b"]["c"]  # simulate descent via keys a, b, and c
>>> for key, value in config.items():
...   print(key, value)
... 
AttributeError: 'str' object has no attribute 'items'
```
I've asked in other PRs whether it's worth interfering with Python's normal error handling / exception raising. Here it might be worth a check and exiting with an informative error, as it might not be obvious to the user what has gone wrong.

maddenp-noaa (changes requested):

src/uwtools/config.py
elcarpenterNOAA: Good call. This wasn't covered in my initial unit tests.

src/uwtools/config.py
```diff
@@ -752,3 +753,17 @@ def create_config_obj(
                     raise ValueError(err_msg)
                 # Dump to file:
                 dump_method(path=outfile, cfg=config_obj)
+
+
+def export_variables(config_dict: dict, section_path: list, log: Logger) -> None:
```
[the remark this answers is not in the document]
elcarpenterNOAA: If `section_path` is being typed as `List[str]`, what is the reasoning for leaving `config` as `dict` and not `Dict[str, Any]`? Simply because it's a way to say "we know it may not _necessarily_ be turtles all the down"?

src/uwtools/config.py
```diff
@@ -752,3 +753,22 @@ def create_config_obj(
...
+
+
+def print_config_section(config: dict, section_path: List[str], log: Logger) -> None:
+    """
+    Descends into the config via the given section keys, then prints the contents of the located
+    subtree as key=value pairs, one per line.
+    """
+    for key in section_path:
+        if isinstance(config[key], dict):
+            config = config[key]
```
christinaholtNOAA, line 765: What happens here if `key` is not in `config`? Should we handle that explicitly, too?

src/uwtools/config.py
```diff
@@ -752,3 +753,22 @@ def create_config_obj(
...
+        else:
+            log.error(f"{key} type must be a dictionary")
+            raise UWConfigError("Key type must be a dictionary")
+    for key, value in config.items():
+        if type(value) not in (bool, float, int, str):
```
christinaholtNOAA, line 770: Should we handle a list as a value?

src/uwtools/config.py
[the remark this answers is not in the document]
maddenp-noaa: Yes, I feel like it might give a false sense of assurance. I was just wondering if there's not a way to express what we really want somehow and found this: https://github.com/python/mypy/issues/731. I haven't had time to look into / test it yet, though. It might be the way to go, but possible out-of-scope for this PR unless you're feeling adventurous.

christinaholtNOAA (commented):
This is looking really nice. :)

src/uwtools/tests/test_config.py
```diff
@@ -810,3 +811,46 @@ def test_YAMLConfig__load_unexpected_error(tmp_path):
...
+side=False
+servings=0\n"""
+    assert actual == expected
+
+
+def test_print_config_section_ini_missing_section():
+    config_obj = config.INIConfig(fixture_path("simple3.ini"))
+    section = ["sandwich"]
+    with raises(KeyError):
+        config.print_config_section(config_obj.data, section, log=Logger())
```
maddenp-noaa, line 856: Nice tests!

A couple small ideas:

1. In the second code box in [Assertions about expected exceptions](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions) the `pytest` docs show how to assign the caught exception to a variable and then make an assertion about its text content. Since you give useful messages in the `UWConfigError` exceptions you raise, you could potentially strengthen for `_for_nonscalar()` and `_not_dict()` tests to look for e.g. `"must be scalar"` and `"must be a dictionary"`, respectively. That would rule out any chance that the test was catching the _other_ exception.

2. I think you might be able to set `expected` values like e.g.
```
expected = """
name=fixed
surface_value=0.0
""".lstrip()
```
if that seems any better to you. That makes the `expected` value maybe a little easier to read and should do the right thing with the newlines. I think... I do not feel strongly about this.

3. Consider sorting the `_ini*()` functions above the `_yaml*()` ones?

src/uwtools/config.py
```diff
@@ -752,3 +753,22 @@ def create_config_obj(
...
+    """
+    Descends into the config via the given section keys, then prints the contents of the located
+    subtree as key=value pairs, one per line.
+    """
+    for key in section_path:
+        if isinstance(config[key], dict):
+            config = config[key]
+        else:
+            log.error(f"{key} type must be a dictionary")
+            raise UWConfigError("Key type must be a dictionary")
```
maddenp-noaa, line 768: `Key` -> `Value`, since we're testing the type of `config[key]`, which is the value?

src/uwtools/config.py
maddenp-noaa: @christinaholtNOAA Honest question: Would it make sense to export a Python list as the value to assign to a shell environment variable?

src/uwtools/config.py
maddenp-noaa: Would something like
```
keys = []
for key in section_path:
    keys.append(key)
    current_path = " -> ".join(keys)
    try:
        subconfig = config[key]
    except KeyError:
        msg = f"Bad config path: {current_path}"
        log.error(msg)
        raise UWConfigError(msg)
    if isinstance(subconfig, dict):
        config = subconfig
    else:
        msg = f"Value at {current_path} must be a dictionary"
        log.error(msg)
        raise UWConfigErrror(msg)
```
give the error check @christinaholtNOAA suggests, plus, usefully, the actual path from the top of the config down to the place we found an error? Maybe something similar below in the value-type check. (Also `msg` / `log` / `raise` starts to look like a pattern...)

src/uwtools/config.py
christinaholtNOAA: We have use cases like this in SRW, but it's not immediately clear to me that we _need_ those situations. We might want to consider them in the context of this being a tool we'd use in the interim of having a YAML configuration layer but with some of the drivers still in bash.

One particular case in SRW is in [fixed_files_mapping.yaml](https://github.com/ufs-community/ufs-srweather-app/blob/develop/parm/fixed_files_mapping.yaml), which sets `FV3_NML_VARNAME_TO_SFC_CLIMO_FIELD_MAPPING` to a list that looks sort of like the dictionary we set up for `stage_files`. I'm not sure how many other instances exist.

maddenp-noaa: Given something like https://github.com/ufs-community/ufs-srweather-app/blob/develop/parm/fixed_files_mapping.yaml#L158-L169 (if I'm looking at the right thing), what would you expect a bash `export` statement to look like?

christinaholtNOAA: I think we'd need to end up with a bash iterable so that it might be something like this:

```
 FV3_NML_VARNAME_TO_SFC_CLIMO_FIELD_MAPPING=(
  "FNALBC  | snowfree_albedo" \
  "FNALBC2 | facsf" \
 )
```

Perhaps that doesn't work out well as an iterable, though. Maybe the true iterables won't be able to have spaces in them? So like in lines 86-119?

src/uwtools/config.py
```diff
@@ -752,3 +753,39 @@ def create_config_obj(
...
+                log.error(f"Non-scalar value {key} was provided")
+                raise UWConfigError("Section values provided must be scalar values")
+        for key, value in config.items():
+            print(f"{key}={value}")
+    else:
+        for value in config:
+            if type(value) not in (bool, float, int, str):
+                log.error(f"Non-scalar value {keys[-1]} was provided")
+                raise UWConfigError("Section values provided must be scalar values")
+        print(f"{keys[-1]}={config}")
```
maddenp-noaa, line 791: Since you mentioned the growing function length in standup, I can offer that you could reduce 10 lines with 4 by defining a helper function that does the `msg` / `log` / `raise` business and calling it 4 times, i.e.
```
some_good_name(f"Non-scalar value {key} was provided")
```
in this function would do the `log` and `raise` calls. That would also mean making the `log` and `raise` messages identical, but that might be a good thing.

Two other minor suggestions:
1. Since there's an outer loop (line 764) using variable `key`, it would be safer not to reuse/redefine that variable inside, in the nested loops. Perhaps use `k` and `v` for keys and values, given that those loops/scopes are so small? I probably missed a chance to give better advice in this regard in some previous comment...
2. If `key` maintains its identity throughout the loop, it should be possible (and IMO) better to replace `keys[-1]` with just `key`, which should have the same value throughout the loop.

maddenp-noaa: And if you feel like you could go full leave-the-campsite-better on this module, you could find other places where a new `some_good_name()` logging / exception-raising function could be used and update them to simplify and make code more consistent. You may not find any opportunities, but it might be worth a look.

elcarpenterNOAA: Regarding point 2, when I initially tried `key` instead of `keys[-1]`, I got linter complaints stating "using possibly undefined loop variable 'key'". Is there a way around this (outside of disabling it)?

maddenp-noaa: In re: item 2, sorry, you're right -- we're outside of the first `for` loop by then. The linter isn't lying. So both my items 1 and 2 are invalid, please disregard!

My (maybe?) last suggest would be to factor out the `if type(value) not in (bool, float, int, str)` check & error-generating code to a helper function to avoid the repetition of identical logic.

elcarpenterNOAA: What is the rule of thumb for creating helper functions? If the code is written twice, might as well have one?

maddenp-noaa: A matter of taste/intuition, I'm afraid.

I wouldn't factor `x = x + 1` and `y = y + 1` out as calls to `def add_one(n: int) -> int`. But in this case we've got that tuple of acceptable types that seems to me to be a concept worth naming, though we could also just factor that out as a local variable in the function like `scalar_types = (bool, float, int, str)` (or if `list` is acceptable, maybe `acceptable_types = (bool, float, int, list, str)`); but then the fact that we want to log a message (possibly/probably the identical message) and raise an exception all starts to look like a "chunkier" concept that might deserve its own identity. If we found that the tuple of types wasn't the same in both cases, I might not opt to factor the whole 3-4 lines out and pass the tuple into that function, but would probably at least call a single factored-out `msg` / `log` / `raise` function.

Well, one rule of thumb might be: "If I have to change one line of code for <reasons>, do I have to go change another identical line and keep them in sync?" If so, probably you want to DRY. So far, I think that's the case here.

maddenp-noaa: What about this way to re-org the code you've already got:
```
output_lines = []
for item in config:
    value = config[item] if isinstance(config, dict) else item
    if type(value) not in (bool, float, int, str):
        log_and_error(f"Non-scalar value {value} found at {current_path}")
    else:
        output_lines.append(f"{key}={value}") if isinstance(config, dict) else f"{keys[-1]}={config}"
print("\n".join(output_lines))
```
I didn't try it in a Python REPL and might have missed something, but is that useful?

elcarpenterNOAA: Yes, thank you, that is useful. I've implemented it with some added constraints for the printing.

maddenp-noaa: From looking at your implementation I think I realize now that we can't just build up `output_lines` no matter what, because if `config` is a `list` then `.append()`ing `f"{keys[-1]}={config}"` repeatedly isn't what we want. Have I got that right? Would it work if we did `output_lines = set()` instead, and did `output_lines.add()` instead of `output_lines.append()`? That should nullify the effect of adding `f"{keys[-1]}={config}"` multiple times and would squeeze some conditional logic out. (In your long-ago Slack question asking for recommendations for learning resources, I suggested the "Confident Ruby" video. What I'm suggesting here is in line with those ideas: Instead of using conditionals to check for and handle various cases, try to arrange for it to be safe to do _one thing_ in all cases. I'm not sure what I'm suggesting here will work, but that's the idea.)

We could also declutter slightly by taking advantage of the fact that `log_and_error()` does not return by changing e.g.
```
if type(subconfig) in (dict, list):
    config = subconfig
else:
    log_and_error(f"Value at {current_path} must be a dictionary or list", log)
```
to
```
if not type(subconfig) in (dict, list):
    log_and_error(f"Value at {current_path} must be a dictionary or list", log)
config = subconfig
```
We can never reach `config = subconfig` if `log_and_error()` is called.

I'm just nitpicking and sharing ideas now, since we're waiting for @christinaholtNOAA to comment in the discussion about list handling, above :)

maddenp-noaa: Another technique share, whether or not you want to use it:

Python's [ternary operator](https://en.wikipedia.org/wiki/Ternary_conditional_operator) can go anywhere another expression can go, so we could also
```
print("\n".join(output_lines) if isinstance(config, dict) else f"{keys[-1]}={config}")
```
if the idea I posted above about using a `set` instead of a `list` for `output_lines` doesn't pan out.

maddenp-noaa: If you are able to / want to adopt the `set` idea, `print("\n".join(sorted(output_lines)))` might be good to ensure that the order is the same run-to-run, as `set`s are unordered.

elcarpenterNOAA: The issue with the original re-org you provided was that because `print("\n".join(output_lines))` was simply the last line, it printed even when `config` was a list and `output_lines` was empty, so it would print an extra `\n`.

maddenp-noaa: I don't immediately see why `output_lines` would be empty, unless the `.append()` (or `.add()` with a `set`) is being guarded with `if isinstance(config, dict)`, which I'm hoping we don't need to do if we're using a `set`. I mean `config` is a container, either a `dict` or ` list`, so the block under `for item in config` should execute at least once and -- unless we call `log_and_error()` and bail out -- `.append()` or `.add()` should be called at least once. Not so?

elcarpenterNOAA: Sorry about that - I didn't see the second and third response yet when I commented. That was in reference to your comment [here](https://github.com/ufs-community/workflow-tools/pull/280#discussion_r1296368363) with the re-organization when `output_lines` was still a list. 

This morning I've converted `output_lines` to a set and sorted the output. The code itself is now under 20 lines as well.

src/uwtools/config.py
```diff
@@ -752,3 +753,37 @@ def create_config_obj(
...
+        keys.append(section)
+        current_path = " -> ".join(keys)
+        try:
+            subconfig = config[section]
+        except KeyError:
+            log_and_error(f"Bad config path: {current_path}", log)
+        if not type(subconfig) in (dict, list):
+            log_and_error(f"Value at {current_path} must be a dictionary or list", log)
+        config = subconfig
+    output_lines = set()
```
maddenp-noaa, line 782: I checked out your branch and did a few tests to see if any of the remaining conditional logic could be removed, but I don't think so. Given that we do need to treat `dict` and `list` objects differently (e.g. I think we should, as you are doing, sort `dict` contents so that order is predictable; but we cannot, as you are correctly _not_ doing, sort `list` contents, as `list` order should be preserved), we can't just add everything to a `set` and treat it one way; so I think we should go back to `output_lines = []` and `output_lines.append()` since we're not actually relying on the special properties of the `set` and its presence might be misleading.

Sorry for all the back and forth. I've learned more about this logic by seeing what you did and did not find that you could implement from the ideas that I shared. Thanks.

src/uwtools/config.py
maddenp-noaa: Doing it this way will have edge cases / failure modes that may lead to behavior ranging from mystifying to dangerous depending on user's inputs and how they're using the outputs. If you're ok with that, here's a diff of @elcarpenterNOAA 's current code that does, I think, what you're suggesting:
```
(DEV-uwtools) ~/git/uwtools % git diff -U0
diff --git a/src/uwtools/config.py b/src/uwtools/config.py
index 6ab54d3..71f923f 100644
--- a/src/uwtools/config.py
+++ b/src/uwtools/config.py
@@ -782,0 +783 @@ def print_config_section(config: dict, section_path: List[str], log: Logger) ->
+    isdict = isinstance(config, dict)
@@ -784 +785 @@ def print_config_section(config: dict, section_path: List[str], log: Logger) ->
-        value = config[item] if isinstance(config, dict) else item
+        value = config[item] if isdict else item
@@ -787 +788 @@ def print_config_section(config: dict, section_path: List[str], log: Logger) ->
-        if isinstance(config, dict):
+        if isdict:
@@ -789 +790 @@ def print_config_section(config: dict, section_path: List[str], log: Logger) ->
-    print("\n".join(sorted(output_lines)) if isinstance(config, dict) else f"{keys[-1]}={config}")
+    print("\n".join(sorted(output_lines)) if isdict else "%s=%s" % (keys[-1], " ".join(config)))
```
I added an `isdict` variable to DRY out multiple `isinstance()` calls.

In a REPL:
```
>>> d = {"x": ["a", "b", "c"]}
>>> config.print_config_section(d, ["x"], Logger())
x=a b c
```

src/uwtools/config.py
```diff
@@ -752,3 +753,37 @@ def create_config_obj(
...
+    subtree as key=value pairs, one per line.
+    """
+    keys = []
+    for section in section_path:
+        keys.append(section)
+        current_path = " -> ".join(keys)
+        try:
+            subconfig = config[section]
+        except KeyError:
+            log_and_error(f"Bad config path: {current_path}", log)
```
christinaholtNOAA, line 778: So, I broke this on my first attempt in a weird way. I got a `TypeError` here by accidentally passing in a string instead of a dict and it took me a while to figure out where I'd gone wrong. This is a more general question, but does it make sense to make sure that our caller is obeying the contract and actually passing in a dict for config? It looked like this:

```
  File "/Users/christina.holt/Work/workflow-tools/test_export.py", line 21, in <module>
    config.print_config_section(list_example, section, log=Logger())
  File "/Users/christina.holt/Work/workflow-tools/src/uwtools/config.py", line 776, in print_config_section
    subconfig = config[section]
                ~~~~~~^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

This also got me thinking to check what happens with my first instinct...passing in `section_path` as a string instead of a list since I only wanted one. I got essentially the same TypeError. In the case the caller provided a string, does it make sense to assume the user wanted only one section and make it a list?

maddenp-noaa: The typechecker will enforce correct argument types for internal code, so we should not be doing manual checks of argument types in library modules like this one. If you add new/test code is to `src/uwtools` source files, `make typecheck` should catch and report this kind of error. If you're interacting with the code in a REPL, you're a dev power-user and on your own, but you can e.g. type `help(uwtools.config.print_config_path)` in the REPL to see the argument names, types, docstring, etc. FWIW, I did exactly the same thing: Passed in a string value for `section_path` instead of a list.

When we have API code, we might want to to extra work there in case users are not doing their own typechecking, because they probably aren't.

maddenp-noaa: I missed the last bit of the question: I think a `str` -> `List[str]` conversion could be appropriate in an API function. For the CLI case, which is all we have now, the CLI code will certainly pass in a `List[str]` object. So I don't think we should be massaging values here.

src/uwtools/config.py
maddenp-noaa: Not totally sure I understand the question, but in a (working) bash script, the author would already have had to deal with all of the quoting issues to represent the values they want in a form that could be iterated over, possibly including adding parens. There are certainly a lot of mystifying rules/techniques that they would have to deal with, of course, but they would have had to sort all of that out to get to working code. If their values contain spaces, or special characters, or whatever, they would have had to deal with it, and they'd know their specific intent / use case.

The mystery here would come from the fact that a perfectly legit YAML list (for example, one where the values contained spaces), to be used for purposes unknown, would produce something in bash that wouldn't reflect what the YAML said. Like
```
% export 'x="a apple" "b banana"'
% for e in ${x[*]}; do echo $e; done
"a
apple"
"b
banana"
```
Unexpected interpretation of that YAML data.

Something like this might be better:
```
% export x=( "a apple" "b banana" )
% for e in "${x[@]}"; do echo "$e"; done
a apple
b banana
```
Or, bringing `python` into the mix:
```
% cat foo.py 
print('x=( "a apple" "b banana" )')
% eval export $(python foo.py)
% for e in "${x[@]}"; do echo "$e"; done
a apple
b banana
```
I haven't been able to make it work without `eval`, though, and `eval` reasonably makes people nervous. But I don't have a better idea.

maddenp-noaa: Well, I do have a better (I think) idea: Use a structured format for structured data, and a tool to destructure it, e.g. JSON + `jq`. We're essentially trying to generate bash code in Python and have bash interpret it, which is tricky, but what we'll have to do if we need a pure-bash solution with list-ish values.

src/uwtools/config.py
elcarpenterNOAA: I'm enjoying the back and the forth, so no worries. `output_lines` has been made into a list again. I'm not sure where to move from here beyond the possibility of implementing your idea of using an `isdict` variable above.

src/uwtools/config.py
maddenp-noaa: Everything related to this conversation looks good to me now. We just need to finalize treatment of lists in the other conversation.

src/uwtools/config.py
christinaholtNOAA: Let's go ahead and leave lists out for now. It seems like we're getting into space that is too big a hoop to jump through and we could likely find better ways for the bash scripts to handle passing this information around. Perhaps we can take a harder look at SRW and see if it's better to propose a change on that side.

src/uwtools/config.py
christinaholtNOAA: I'm fine with not hand-holding our power users. Just thought I'd bring it up since we were already doing some error-catching here. But, yeah, RTFM is a perfectly fine way to go. :)

src/uwtools/config.py
```diff
@@ -752,3 +753,35 @@ def create_config_obj(
                     raise ValueError(err_msg)
                 # Dump to file:
                 dump_method(path=outfile, cfg=config_obj)
+
+
+def log_and_error(msg: str, log: Logger) -> None:
```
maddenp-noaa, line 758: Probably this should be marked private, as `_log_and_error()`, since we don't intend for it to be called from outside this module.

src/uwtools/config.py
```diff
@@ -752,3 +753,35 @@ def create_config_obj(
...
+            subconfig = config[section]
+        except KeyError:
+            log_and_error(f"Bad config path: {current_path}", log)
+        if not isinstance(subconfig, dict):
+            log_and_error(f"Value at {current_path} must be a dictionary", log)
+        config = subconfig
+    output_lines = []
+    for key, value in config.items():
+        if type(value) not in (bool, float, int, list, str):
+            log_and_error(f"Non-dictionary value {value} found at {current_path}", log)
```
maddenp-noaa: Maybe I've lost the plot, but I think that by the time we reach this `for` loop, we know that `config` is a `dict`, and now we want to iterate over its key-value pairs, ensure that all values are scalars, add them to a list of key/val pairs to output if so, abort if not. If that's right, 1. Should we remove `list` from the tuple of acceptable (i.e. scalar) types, and 2. Complain about a `Non-scalar` value instead of a `Non-dictionary` value?

maddenp-noaa: Please ignore, I commented on old code. Resolving...

elcarpenterNOAA: That was me who lost the plot. I already corrected it in the last push.

src/uwtools/config.py
elcarpenterNOAA: Should the last three functions in `config.py` be re-ordered to have `# Private functions` and `# Public functions` with their respective functions sorted underneath them?

src/uwtools/tests/test_config.py
```diff
@@ -810,3 +811,68 @@ def test_YAMLConfig__load_unexpected_error(tmp_path):
...
+
+def test_print_config_section_yaml_not_dict():
+    config_obj = config.YAMLConfig(fixture_path("FV3_GFS_v16.yaml"))
+    section = ["sgs_tke", "units"]
+    with raises(UWConfigError) as e:
+        config.print_config_section(config_obj.data, section, log=Logger())
+    assert "must be a dictionary" in str(e.value)
+
+
+def test_log_and_error():
```
maddenp-noaa, line 875: If you make `log_and_error()` private, it would be helpful for consistency to add the extra underscore here, too: `def test__log_and_error()`.

src/uwtools/config.py
maddenp-noaa: I like that 👍

src/uwtools/tests/test_config.py
elcarpenterNOAA: The linter is responding with a warning: `W0212: Access to a protected member _log_and_error of a client class (protected-access)`. What is the best way to fix this?

maddenp-noaa: Some of the other unit-test modules suppress this warning -- you can crib from one of those.

src/uwtools/config.py
```diff
@@ -752,3 +767,27 @@ def create_config_obj(
                     raise ValueError(err_msg)
                 # Dump to file:
                 dump_method(path=outfile, cfg=config_obj)
+
+
+def print_config_section(config: dict, section_path: List[str], log: Logger) -> None:
```
christinaholtNOAA, line 772: I am totally going to leave this up to you at this point...I've waited so late in the game, but is `section_path` a bit misleading? We often see `path` used in context with a filesystem path. 

Does it make sense to use some other word to describe this concept? Hierarchy? Breadcrumbs? Chain? 

Maybe none of those are any clearer, though.....hmmm

christinaholtNOAA (approved):
@elcarpenterNOAA This looks great! 

Please take my only comment below with a grain of salt. I'm 💯 fine with whatever you decide to do there (ignoring me included)!

maddenp-noaa (approved):

src/uwtools/config.py
elcarpenterNOAA: In searches, "key path" was the term I found most often for this concept, so I'll convert it to that. I'm not sure '_hierarchy` or `_chain` would be clearer.

christinaholtNOAA: Perfect. Leave it as-is. :)

maddenp-noaa (approved):

PR #283, opened 2023-08-21
pre-uw-343-logging-and-streams

maddenp-noaa:
**Description**

As outlined in our modal-CLI design discussion and as a prelude to the implementation in UW-343, this PR establishes the following invariants:

1. The `stdout` stream is preserved for data exchange, so that e.g. IO can stream between `uw` and other programs (including other invocations of `uw`, e.g. a `uw config translate` command translating an atparse-formatted input to Jinja2-formatted output streaming into a `uw config render` command to render that Jinja2 template, without the need for intermediate files).
2. The `stderr` stream is used for all other console IO, either from `print` statements before logging is configured, or via `logging` calls.

This is done by simplifying `uwtools` logging by having the CLI entry-point modules (to be replaced in UW-343 with a single module) configure the Python root logger -- logging either to `sys.stderr` for normal logging, or to `os.devnull` in quiet mode -- and having all code call `logging.info()`, `logging.error()`, logging.debug()`, etc. Logging is configured immediately after the command-line arguments are parsed and the quiet/verbose settings are known. Errors encountered _before_ logging is configured are printed with `print(..., file=sys.stderr)` to keep them on the `stderr` stream.

When we design the programmatic API for `uwtools` we can include mechanisms for non-CLI users to override the log configuration with a custom logger, which can be used to reconfigure the Python root logger, which should require no change to code already logging to that logger.

This PR has another long diff, but much of it is either repetitive patterns or code removal. I have annotated it for a top-down read.

**How Has This Been Tested?**

Passes `make test` in local dev shell + GitHub Actions.

**Checklist**

- [ ] **NA** My changes need updates to the documentation.  I have made corresponding changes to the documentation


src/uwtools/cli/atparse_to_jinja2.py
```diff
@@ -15,6 +16,7 @@ def main() -> None:
     Main entry point.
     """
     args = parse_args(sys.argv[1:])
+    setup_logging(quiet=args.quiet, verbose=args.verbose)
```
maddenp-noaa, line 19: Here's the first example of logging setup immediately following command-line argument parsing. Others follow.

src/uwtools/cli/atparse_to_jinja2.py
```diff
@@ -53,8 +55,23 @@ def parse_args(args: List[str]) -> Namespace:
...
+        "--quiet",
+        action="store_true",
+        help="Print no logging messages",
+    )
+    optional.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="Print all logging messages.",
+    )
```
maddenp-noaa, line 69: In UW-343, a **lot** of duplication will be removed by coalescing the CLI logic into a single module. For now, to keep things consistent on `develop`, add the `--quiet` and `--verbose` flags to the `arparse-to-jinja2` program so that logging can be correctly configured.

src/uwtools/cli/experiment_manager.py
```diff
@@ -50,13 +52,6 @@ def parse_args(args: List[str]) -> Namespace:
         type=cli_helpers.path_if_it_exists,
     )
     optional = parser.add_argument_group("optional arguments")
-    optional.add_argument(
-        "-l",
-        "--log-file",
-        default="/dev/null",
-        help="Path to a file to log to",
-        metavar="FILE",
-    )
```
maddenp-noaa, line 59: Few tools that I'm aware of offer an option to redirect non-data (i.e. for humans) output.

- Like the version of `uw` in this PR, `conda create` offers `--quiet` and `--verbose` options, but no redirection option.
- `npm` has a `--verbose` switch and, at least in some modes, `--silent`, but no redirection option that I can find.
- `git` has `--quiet` and `--verbose` but no redirection.

This follows Unix tradition of keeping messages and data on separate streams and providing tools to redirect from the console if desired. For example, with `git` the error output from specifying a bad mode can be directed to a file, including to `/dev/null` to suppress it:
```
% git asdf # normal behavior
git: 'asdf' is not a git command. See 'git --help'.

The most similar command is
	add
```
```
% git asdf 2>err # send to file 'err'
% cat err
git: 'asdf' is not a git command. See 'git --help'.

The most similar command is
	add
```
```
% git asdf 2>/dev/null # suppress stderr
% 
```
So, barring unforeseen circumstances, I'd like to suggest that YAGNI. It will be easy to add this option via the modal CLI in one place for all modes, if we find we need it.
 
(Full disclosure: `curl`, like the proposed `uw`, streams data to `stdout` by default (and offers a `-o` switch to direct data to a file, as I believe `uw` should, too), as well as `--silent` and `--verbose` switches. It does also have a `--stderr` option to redirect its own human-readable output without using shell redirection.)

src/uwtools/cli/experiment_manager.py
```diff
@@ -71,6 +66,6 @@ def parse_args(args: List[str]) -> Namespace:
     )
     parsed = parser.parse_args(args)
     if parsed.quiet and parsed.verbose:
-        print("Options --dry-run and --outfile may not be used together", file=sys.stderr)
+        print("Options --quiet and --verbose may not be used together", file=sys.stderr)
```
maddenp-noaa, line 69: bugfix

src/uwtools/cli/run_forecast.py
```diff
@@ -18,12 +19,9 @@ def main() -> None:
     Parses arguments provided by the user and passes to the Forecast driver class to be run.
     """
     args = parse_args(sys.argv[1:])
-    name = "run-forecast"
-    cli_helpers.setup_logging(
-        log_file=args.log_file, log_name=name, quiet=args.quiet, verbose=args.verbose
-    )
+    setup_logging(quiet=args.quiet, verbose=args.verbose)
     forecast_class = getattr(forecast, "%sForecast" % args.forecast_model)
-    experiment = forecast_class(args.config_file, args.machine, log_name=name)
```
maddenp-noaa, line 26: Here and in much of what follows the diffs involve simplification and removal of unneeded code.

src/uwtools/cli/atparse_to_jinja2.py
```diff
@@ -53,8 +55,23 @@ def parse_args(args: List[str]) -> Namespace:
...
     parsed = parser.parse_args(args)
     if not parsed.dry_run and not parsed.outfile:
         print("Specify either --dry-run or --outfile", file=sys.stderr)
         sys.exit(1)
+    if parsed.quiet and parsed.verbose:
+        print("Options --quiet and --verbose may not be used together", file=sys.stderr)
```
maddenp-noaa, line 75: Note here and below, in CLI code, that `print(..., file=sys.stderr)` is used to emit messages to `stdout` since logging has not been set up yet, and cannot be until argument parsing is complete and the status of the `--quiet` and `--verbose` switches is known.

src/uwtools/utils/cli_helpers.py
```diff
@@ -51,35 +50,7 @@ def path_if_it_exists(path: str) -> str:
     p = Path(path)
     if not p.exists():
         msg = f"{path} does not exist"
-        logging.critical(msg)
```
maddenp-noaa, line 54: `logging` cannot be used here because this function is called from `parse_args()` functions, so logging has not been set up yet. Use `print(..., file=sys.stderr)` instead.

src/uwtools/cli/templater.py
```diff
@@ -119,10 +111,12 @@ def parse_args(args: List[str]) -> Namespace:
     parsed = parser.parse_args(args)
     if not parsed.outfile and not parsed.dry_run and not parsed.values_needed:
         msg = "You need outfile, dry_run, or values_needed to continue."
-        raise ArgumentError(None, msg)
+        print(msg, file=sys.stderr)
+        sys.exit(1)
```
maddenp-noaa, line 115: It's useful to raise exceptions deep in library code so that they can be caught and dealt with gracefully by higher-level code, but this is CLI code and there's nothing higher that can catch that exception. An end user probably won't have much use for a stack trace -- which in any case will be very shallow and not give much context -- here. The problem pertains to the command they just typed not being quite correct, so we can just tell them about that like humans.

src/uwtools/logging.py
```diff
@@ -0,0 +1,26 @@
...
+    if quiet and verbose:
+        print("Specify at most one of 'quiet' and 'verbose'", file=sys.stderr)
+        sys.exit(1)
+    kwargs: dict = {
+        "datefmt": "%Y-%m-%dT%H:%M:%S",
+        "format": "[%(asctime)s] %(levelname)8s %(message)s",
+        "level": logging.DEBUG if verbose else logging.INFO,
+        **({"filename": os.devnull} if quiet else {}),
+    }
+    logging.basicConfig(**kwargs)
```
maddenp-noaa, line 26: Here's how this looks in practice:
```
>>> from uwtools.logging import setup_logging
>>> setup_logging()
>>> import logging
>>> logging.info("Some info")
[2023-08-22T00:41:14]     INFO Some info
>>> logging.critical("Oh no!")
[2023-08-22T00:41:23] CRITICAL Oh no!
```
If the `filename` argument to `basicConfig()` is not set, Python's root logger defaults to logging to `stderr`, which is what we want.

src/uwtools/tests/cli/test_atparse_to_jinja2.py
```diff
@@ -26,6 +26,22 @@ def test_main_bad_args(sw, capsys):
...
+        ns(d="--dry-run", i="--input-template", q="--quiet", v="--verbose"),
+    ],
+)
+def test_main_bad_args_1(capsys, sw):
+    argv = ["test", sw.i, fixture_path("ww3_multi.inp.IN"), sw.d, sw.q, sw.v]
+    with patch.object(atparse_to_jinja2.sys, "argv", argv):
+        with raises(SystemExit) as e:
+            atparse_to_jinja2.main()
+        assert e.value.code == 1
+        assert "Options --quiet and --verbose may not be used together" in capsys.readouterr().err
```
maddenp-noaa, line 42: Added a test to deal with the new `--quiet` and `--verbose` flags of the `atparse-to-jinja2` program. A shame since it will disappear in UW-343, but this keeps everything running smoothly in the meantime.

src/uwtools/tests/fixtures/schema_test_good.yaml
```diff
@@ -0,0 +1,7 @@
+platform:
+  WORKFLOW_MANAGER: rocoto
+  NCORES_PER_NODE: 40
+  SCHED: slurm
+
+task_get_da_obs:
+  OBS_SUFFIX: grib2
```
maddenp-noaa, line 7: I prematurely removed this file in my previous PR. Restoring.

src/uwtools/tests/test_logging.py
```diff
@@ -0,0 +1,52 @@
...
+# it, since the roog logger will not be what one would expect.
+
+
+def test_setup_logging():
+    with patch.object(logging, "basicConfig") as basicConfig:
+        uwtools.logging.setup_logging()
+        basicConfig.assert_called_once_with(
+            datefmt=ANY,
+            format=ANY,
+            level=logging.INFO,
```
maddenp-noaa, line 25: Use the `ANY` value for `datefmt` and `format` to avoid getting to specifc about these, as their values are not really what we're testing for here. (We're testing that `filename=` was not specified, since we're not in `--quiet` mode; and that the level is correct/default.)

src/uwtools/cli/atparse_to_jinja2.py
christinaholtNOAA: I'm a bit confused here. Would these messages not be emitted to `stderr`? Is that a typo or black magic?

src/uwtools/cli/templater.py
christinaholtNOAA: This is a great rule of thumb! I'm curious if you think it might be a nice addition to the _Coding Advice_ document we have floating around.

src/uwtools/logging.py
christinaholtNOAA: Just to make sure I'm reading the correctly...we lose the coloring scheme that was set up in the class that was removed, is that right?

christinaholtNOAA (approved):
LGTM!

src/uwtools/cli/atparse_to_jinja2.py
maddenp-noaa: Yes, to `stderr`. Is it the `file=` part that is surprising, i.e. `stderr` is not a file?

src/uwtools/logging.py
maddenp-noaa: Yes, but we could add that, or something similar, back later if we find that we need it. It seemed to me maybe premature at this point. I'm happy to commit to adding color-coding if/when we want it -- including now.

IMO colors can be nice in human-facing tools, but unwelcome in automated contexts where the output might be captured/parsed, as the ANSI color codes can complicate e.g. regex processing. For just about any tool with colorized output, animated progress bars, etc., there are tickets from disgruntled users asking for ways to disable it. So it should at least be behavior that can be disabled -- which I believe was indeed already the case.

src/uwtools/cli/experiment_manager.py
WeirAE: At the time I was thinking that outputting nothing with `--dry-run` but requesting output with `--outfile` would be a user error. Since then we've moved to using the output file for more internal things, so replacing it with the standard q/v is the way  to go

PR #285, opened 2023-08-28
UW-272 Develop unit tests that verify and validate that the Rocoto XML generator outputs a valid Rocoto XML

NaureenBharwaniNOAA:
**Description**

- One of the most difficult/complex workflow end-to-end tests to get "right" has been the MET_ensemble_verification. It creates multiple nested metatasks, and uses many of the supported dependency flags. It most fully utilizes the functionalities we need to support for Rocoto XML generation.
- Perhaps we can ensure that we match the SRW expected file in one of our tests. We do NOT want to be running this WE2E test as part of our suite.
- Re-scoping this work for now to have a test read in a Rocoto XML with lxml and tell us if it is valid or not. We will have both a valid and an invalid one.
- The only thing needed to satisfy that goal would be a unit test that does that comparison and returns a true/false.

**Type of change**

**How Has This Been Tested?**

Passes all development-shell `make format && make test` code-quality checks, including (updated) unit tests.

**Checklist**

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -18,6 +18,7 @@ requirements:
     - python >=3.8,<3.12
     - python {{ release }}
     - pyyaml 6.0.*
+    - lxml 4.9.*
 test:
   requires:
     - black 23.3.*
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -15,10 +15,10 @@ requirements:
     - f90nml 1.4.*
     - jinja2 3.0.*
     - jsonschema 4.17.*
+    - lxml 4.9.*
     - python >=3.8,<3.12
     - python {{ release }}
     - pyyaml 6.0.*
-    - lxml 4.9.*
 test:
   requires:
     - black 23.3.*
--- a/src/uwtools/resources/schema_with_metatasks.rng
+++ b/src/uwtools/resources/schema_with_metatasks.rng
@@ -1,3 +1,4 @@
+<!-- Copied from the Rocoto repository at release version 1.3.5.-->
 <grammar xmlns="http://relaxng.org/ns/structure/1.0" datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
                                                                       
   <start>
--- a/src/uwtools/resources/schema_with_metatasks.rng
+++ b/src/uwtools/resources/schema_with_metatasks.rng
@@ -1,0 +1,622 @@
+<grammar xmlns="http://relaxng.org/ns/structure/1.0" datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
+                                                                      
+  <start>
+
+    <element name="workflow">
+
+      <!-- Realtime workflow attribute is required-->
+      <attribute name="realtime">
+        <choice>
+          <value>f</value>
+          <value>F</value>
+          <value>false</value>
+          <value>False</value>
+          <value>FALSE</value>
+          <value>t</value>
+          <value>T</value>
+          <value>true</value>
+          <value>True</value>
+          <value>TRUE</value>
+        </choice>
+      </attribute>
+
+      <!-- scheduler workflow attribute is required and must be one of the values listed -->
+      <attribute name="scheduler">
+        <choice>
+          <value>sge</value>
+          <value>lsf</value>
+          <value>lsfcray</value>
+          <value>ll</value>
+          <value>moab</value>
+          <value>moabtorque</value>
+          <value>torque</value>
+          <value>pbspro</value>
+          <value>slurm</value>
+          <value>cobalt</value>
+        </choice>
+      </attribute>
+
+      <!-- cyclethrottle workflow attribute is optional and must be a non negative integer -->
+      <optional>
+        <attribute name="cyclethrottle">
+          <data type="nonNegativeInteger"/>
+        </attribute>
+      </optional>
+
+      <!-- taskthrottle workflow attribute is optional and must be a non negative integer -->
+      <optional>
+        <attribute name="taskthrottle">
+          <data type="nonNegativeInteger"/>
+        </attribute>
+      </optional>
+
+      <!-- corethrottle workflow attribute is optional and must be a non negative integer -->
+      <optional>
+        <attribute name="corethrottle">
+          <data type="nonNegativeInteger"/>
+        </attribute>
+      </optional>
+
+      <!-- duthrottle workflow attribute is optional and must be a non negative integer -->
+      <optional>
+        <attribute name="duthrottle">
+          <data type="nonNegativeInteger"/>
+        </attribute>
+      </optional>
+
+      <!-- cyclelifespan workflow attribute is optional and must be a nonNegativeTime defined below-->
+      <optional>
+        <attribute name="cyclelifespan">
+          <ref name="nonNegativeTime"/>
+        </attribute>
+      </optional>
+
+      <!-- Order of log, cycle, task, and metatask tags does not matter -->
+      <interleave>
+
+        <!-- Must have exactly one log tag -->
+        <element name="log">
+          <optional>
+            <attribute name="verbosity">
+              <data type="nonNegativeInteger"/>
+            </attribute>
+          </optional>
+          <ref name="compoundTimeString"/>
+        </element>
+
+        <!-- Must have at least one cycledef tag -->
+        <oneOrMore>
+          <element name="cycledef">
+            <!-- group attribute is optional and must be a string -->
+            <optional>
+              <attribute name="group">
+                <data type="string"/>
+              </attribute>
+            </optional>
+            <optional>
+              <attribute name="activation_offset">
+                <ref name="time"/>
+              </attribute>
+            </optional>
+            <choice>
+              <list>
+                <ref name="cronField"/>
+                <ref name="cronField"/>
+                <ref name="cronField"/>
+                <ref name="cronField"/>
+                <ref name="cronField"/>
+                <ref name="cronField"/>
+              </list>
+              <list>
+                <ref name="YYYYMMDDHHMM"/>
+                <ref name="YYYYMMDDHHMM"/>
+                <ref name="nonNegativeTime"/>
+              </list>
+              <list>
+                <ref name="YYYYMMDDHHMM"/>
+                <ref name="nonNegativeTime"/>
+                <ref name="nonNegativeTime"/>
+              </list>
+            </choice>
+          </element>
+        </oneOrMore>
+
+        <!-- Must have at least one task or metatask tag.  Task tags must follow log and cycle tags -->
+        <oneOrMore>
+          <choice>
+            <ref name="task"/>
+            <ref name="metatask"/>
+          </choice>
+        </oneOrMore>
+
+      </interleave>
+
+    </element>  
+
+  </start>
+
+  <!-- Type definitions -->
+
+  <define name="rb_or_sh">
+    <choice>
+      <element name="rb">
+        <optional>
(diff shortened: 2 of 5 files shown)


src/uwtools/tests/test_rocoto.py
```diff
@@ -52,3 +53,14 @@ def test_write_rocoto_xml(tmp_path):
...
+
+def test_rocoto_xml_is_valid():
+    input_xml = support.fixture_path("hello_workflow.xml")
+    tree = etree.parse(input_xml)
+    root = etree.tostring(tree.getroot())
+
+    breakpoint()
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        relaxng_doc = str(resc / "schema_with_metatasks")
+    relaxng = etree.RelaxNG(file=relaxng_doc)
```
NaureenBharwaniNOAA, line 66: Getting an error message: 
`E   lxml.etree.RelaxNGParseError: unknown error
src/lxml/relaxng.pxi:100: RelaxNGParseError`

When originally passing the `relaxng_doc` I was getting the file path back as a PosixPath object and converted it to a string to pass to the RelaxNG class, where used this [documentation](https://lxml.de/api/lxml.etree.RelaxNG-class.html) to confirm using a keyword `file` to pass the path in. Any suggestions on working with this error?

I pulled `schema_with_metatasks.rng` from Chris' rocoto repo as well.

maddenp-noaa: The file is called `schema_with_metatasks.rng` -- the code above doesn't have the extension. See if that helps.

This works for me:
```
>>> from lxml import etree
>>> with open("schema_with_metatasks.rng", "r") as f:
...     schema = etree.RelaxNG(etree.parse(f))
... 
>>> 
```
That's a slightly different method that you're using -- you still need to use `resources.as_file()` to get the resources directory and then construct the filename, but maybe the issue was just the missing extension.

NaureenBharwaniNOAA: That worked, thank you!

src/uwtools/tests/fixtures/rocoto_invalid.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE workflow [
+  <!ENTITY ACCOUNT "myaccount">
+  <!ENTITY FOO "test.log">
+
+]>
+<workflow realtime="False" scheduler="slurm" >
+
```
NaureenBharwaniNOAA, line 8: Here I removed the `cycledef` tag that is required to return true from the Rocoto schema validation. Removing the `cycledef` tag will cause the validation to fail to confirm our unit tests are catching this appropriately. If helpful I can make this more obvious by changing the file name as needed to help highlight this better.
i.e. `invalid_rocoto_removed_cycledef.xml`

Line removed: `<cycledef group="howdy">202209290000 202209300000 06:00:00</cycledef>`

src/uwtools/tests/test_rocoto.py
```diff
@@ -52,3 +53,17 @@ def test_write_rocoto_xml(tmp_path):
...
+        with open(resc / "schema_with_metatasks.rng", "r", encoding="utf-8") as f:
+            schema = etree.RelaxNG(etree.parse(f))
+
+    valid_xml = support.fixture_path("hello_workflow.xml")
+    tree = etree.parse(valid_xml)
+    assert schema.validate(tree)
+
+    invalid_xml = support.fixture_path("rocoto_invalid.xml")
+    tree = etree.parse(invalid_xml)
+    assert not schema.validate(tree)
```
NaureenBharwaniNOAA, line 69: linter complained at `assert False == schema.validate(tree)`. Let me know if this is the way we want to go here.

NaureenBharwaniNOAA (commented):
Another comment, currently I have Here I removed the cycledef tag that is required to return true from the Rocoto schema validation. Removing the cycledef tag will cause the validation to fail to confirm our unit tests are catching this appropriately. If helpful I can make this more obvious by changing the file name as needed to help highlight this better.
i.e. invalid_rocoto_removed_cycledef.xml

maddenp-noaa:
> Currently I am failing the linter when importing the Module `lxml.etree`. Output message: `Module 'lxml.etree' has no 'RelaxNG' member, but source is unavailable. Consider adding this module to extension-pkg-allow-list if you want to perform analysis based on run-time introspection of living objects. (c-extension-no-member)`
> 
> I am also getting this error on the lines I use `parse` as well. Do we need to add `lxml` to the linter? From some [research](https://stackoverflow.com/questions/43280486/pylint-error-message-e1101-module-lxml-etree-has-no-strip-tags-member), it seems like we can correct these errors by adding `lxml` to our whitelist (`--extension-pkg-whitelist=lxml`)
> 
> Thoughts?

Yes, you should be able to define a whitelist in one of the pylint sections of `src/pyproject.toml`. You might have to read a bit to find out which section, the specific syntax, etc., but that's the way to go. Let me know if you get stuck.

The underlying issue is that part of `lxml` is written in C, and `pylint` cannot look into the compiled C code to see if certain calls Python is making are valid. More or less.

src/uwtools/tests/fixtures/rocoto_invalid.xml
maddenp-noaa: Seems like a good approach to me, and I like the filename. It might be good to add an XML comment saying what's going on.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: Try `assert schema.validate(tree) is False` and see if the linter is happy with that. Normally `pylint` complains about `== None`, `== True`, and `== False` and prefers `is` to `==` in these cases (maybe some others, I'm not sure). The rationale is that those values are _singleton_ objects in Python, and `is` is a stronger test than `==`; `is` means that the two things are really the exact same object in memory, which must be the case with singletons, while `==` means that two values are somehow equivalent, however equivalence is defined for those values. There's nothing really wrong with `==` that I'm aware of, but using `is` is more idiomatic.

Using `not` can be dicey in Python. I mean, definitely do not avoid using it, but understand and consider that a lot of values can be "falsey" in Python, including an empty `list`, an empty `dict`, and empty `str`, the number 0, `False`, and `None`. Probably I'm forgetting some! For example, if you know you've got some kind of collection `c` -- a `list`, a `dict`, a `set`, etc. -- saying `not c` is the idiomatic way to indicate an empty collection, much preferred to `len(c) == 0`. But when using `not`, think about what kind of values it's being applied to and whether there's any chance of them being falsey when that's not what you intend. For example, if a function argument was typed `x: Optional[list] = None`, using `not x` to try to detect when `x` wasn't passed in by the caller (i.e. has a `None` value) isn't a great idea, since a user might have passed `[]` in as a value, which is also falsey.

NaureenBharwaniNOAA:
> >

I was able to resolve the errors with the linter on my side for the `lxml` module by adding an extension package allow list under the `tool.pylint.message_control` section of `pyproject.toml` 

However, I see the linter is still failing when pushing my branch
`uwtools/tests/test_rocoto.py:8:0: E0401: Unable to import 'lxml' (import-error)`

maddenp-noaa:
> However, I see the linter is still failing when pushing my branch `uwtools/tests/test_rocoto.py:8:0: E0401: Unable to import 'lxml' (import-error)`

You'll need to add an `lxml` entry to the `requirements` -> `run` section of `recipe/meta.yaml`, as this package is now a run-time requirement of `uwtools`. If the tests pass in your development shell, you can do `conda list | grep lxml` to get the version you're currently using, and add `lxml major.minor.*` to `meta.yaml`, using the major and minor version components from your `conda list`. After you do that you can first test locally by doing:
```
$ exit # to exit your dev shell
$ conda env remove -n DEV-uwtools # to remove the conda environment underpinning your previous development shell
$ git clean -dfx # to remove any cruft from your git clone -- check 'git status' first and be sure you've committed everything
$ make devshell # to create a new development shell
```
The new dev shell should have `lxml` installed and `make test` should pass.

src/uwtools/resources/schema_with_metatasks.rng
```diff
@@ -0,0 +1,622 @@
+<grammar xmlns="http://relaxng.org/ns/structure/1.0" datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
```
christinaholtNOAA, line 1: We likely want to add a header comment here that says we copied this file from the Rocoto repository at release version 1.3.5.

src/uwtools/tests/fixtures/rocoto_invalid.xml
```diff
@@ -0,0 +1,45 @@
...
+  <!ENTITY FOO "test.log">
+
+]>
+<workflow realtime="False" scheduler="slurm" >
+  <!-- cycldef tag has intentionally been removed to fail Rocoto schema validation
+  cycledef is required and defines the set of cycles to be run for the workflow
+  -->
+  <log>/some/path/to/&FOO;</log>
+  
+  <task name="hello" cycledefs="howdy" maxtries="2" >
```
christinaholtNOAA, line 13: You might consider removing all task and metatask entries, too, just to keep content shorter.

src/uwtools/tests/test_rocoto.py
```diff
@@ -52,3 +53,17 @@ def test_write_rocoto_xml(tmp_path):
...
+        with open(resc / "schema_with_metatasks.rng", "r", encoding="utf-8") as f:
+            schema = etree.RelaxNG(etree.parse(f))
+
+    valid_xml = support.fixture_path("hello_workflow.xml")
+    tree = etree.parse(valid_xml)
+    assert schema.validate(tree)
+
+    invalid_xml = support.fixture_path("rocoto_invalid.xml")
+    tree = etree.parse(invalid_xml)
+    assert schema.validate(tree) is False
```
christinaholtNOAA, line 69: Does it make sense to simplify this line with:

```
assert not schema.validate(tree
```

christinaholtNOAA (approved):
Nice!

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: @christinaholtNOAA based on your comment elsewhere on the PR

>> Does it make sense to simplify this line with:
>> `assert not schema.validate(tree)`

recipe/meta.yaml
```diff
@@ -18,6 +18,7 @@ requirements:
     - python >=3.8,<3.12
     - python {{ release }}
     - pyyaml 6.0.*
+    - lxml 4.9.*
```
maddenp-noaa, line 21: Please sort list of required packages.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: This can be DRYed out with `parametrize`:
```
@pytest.mark.parametrize("vals", [("hello_workflow.xml", True), ("rocoto_invalid.xml", False)]
def test_rocoto_xml_is_valid(vals):
    fn, validity = vals
    # same "with" block...
    xml = support.fixture_path(fn)
    tree = etree.parse(xml)
    assert schema.validate(tree) is validity
```

maddenp-noaa (changes requested):

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: Excellent suggestion! It made the code less repetitive.

WeirAE (approved):
Great work sorting out Rocoto and all of those changes! LGTM

maddenp-noaa (approved):

PR #287, opened 2023-08-31
optional-format

maddenp-noaa:
**Description**

A post-UW-343 follow-up PR to address a handful of leftover issues.

First, though, there are some trivial changes that account for a large majority of the diffs. Once you know what you're looking at, you can just scan over these and look for the actually interesting changes, which I'll point out in inline comments.

- Many instances of magic string values, like `"yaml"` or `"!INCLUDE"` have been replaced with references to variables providing the actual string value to be used. A typo in a string variable is impossible for static-analysis tools to detect, and can be hard to debug. References to module variables (for `!INCLUDE`) and dataclass attributes (for the names of common format languages) _can_ be statically checked and mistakes detected.
- Non-public members in `uwtools.cli` have been marked private with a leading underscore.

Those two categories account for the majority of diffs. Apart from those, the upshot of the work in the PR is: Make the format-specification CLI arguments (e.g. `--input-format`) optional, as they can usually be deduced from filenames, **unless** filenames are not provided, in which case the format must be specified. Post argument-processing checks are added to detect and report issues.

**How Has This Been Tested?**

Passes `make test` locally + GitHub Actions.

### DIFF
--- a/src/uwtools/config/core.py
+++ b/src/uwtools/config/core.py
@@ -24,7 +24,7 @@ from uwtools.config.j2template import J2Template
 from uwtools.exceptions import UWConfigError
 from uwtools.logging import MSGWIDTH
 from uwtools.types import DefinitePath, OptionalPath
-from uwtools.utils.file import get_file_type, readable, writable
+from uwtools.utils.file import FORMAT, get_file_type, readable, writable
 
 MSGS = ns(
     unhashable="""
@@ -675,8 +675,8 @@ def compare_configs(
     :return: False if config files had differences, otherwise True.
     """
 
-    cfg_a = _cli_name_to_config(config_a_format)(config_a_path)
-    cfg_b = _cli_name_to_config(config_b_format)(config_b_path)
+    cfg_a = _format_to_config(config_a_format)(config_a_path)
+    cfg_b = _format_to_config(config_b_format)(config_b_path)
     logging.info("- %s", config_a_path)
     logging.info("+ %s", config_b_path)
     logging.info("-" * MSGWIDTH)
@@ -733,7 +733,7 @@ def realize_config(
     :raises: UWConfigError if errors are encountered.
     """
 
-    input_obj = _cli_name_to_config(input_format)(config_file=input_file)
+    input_obj = _format_to_config(input_format)(config_file=input_file)
     input_obj.dereference_all()
     input_obj = _realize_config_update(input_obj, values_file, values_format)
     _realize_config_check_depths(input_obj, output_format)
@@ -742,14 +742,14 @@ def realize_config(
     if dry_run:
         logging.info(input_obj)
     else:
-        _cli_name_to_config(output_format).dump_dict(path=output_file, cfg=input_obj.data)
+        _format_to_config(output_format).dump_dict(path=output_file, cfg=input_obj.data)
     return True
 
 
 # Private functions
 
 
-def _cli_name_to_config(cli_name: str) -> Type[Config]:
+def _format_to_config(cli_name: str) -> Type[Config]:
     """
     Maps a CLI format name to its corresponding Config class.
 
@@ -757,10 +757,10 @@ def _cli_name_to_config(cli_name: str) -> Type[Config]:
     :return: The appropriate Config class.
     """
     classes: Dict[str, type[Config]] = {
-        "fieldtable": FieldTableConfig,
-        "ini": INIConfig,
-        "nml": NMLConfig,
-        "yaml": YAMLConfig,
+        FORMAT.fieldtable: FieldTableConfig,
+        FORMAT.ini: INIConfig,
+        FORMAT.nml: NMLConfig,
+        FORMAT.yaml: YAMLConfig,
     }
     try:
         return classes[cli_name]
@@ -810,7 +810,7 @@ def _realize_config_update(
     if values_file:
         logging.debug("Before update, config has depth %s", input_obj.depth)
         values_format = values_format or get_file_type(values_file)
-        values_obj = _cli_name_to_config(values_format)(config_file=values_file)
+        values_obj = _format_to_config(values_format)(config_file=values_file)
         logging.debug("Values config has depth %s", values_obj.depth)
         input_obj.update_values(values_obj)
         input_obj.dereference_all()
--- a/src/uwtools/config/core.py
+++ b/src/uwtools/config/core.py
@@ -758,6 +758,22 @@ def _realize_config_update(
     return input_obj
 
 
+def _realize_config_check_depths(input_obj: Config, output_format: str) -> None:
+    """
+    Check that the depth of the input config does not exceed the output format's max.
+
+    :param input_obj: The input config.
+    :param output_format: The output format:
+    :raises: UWConfigError on excessive input-config depth.
+    """
+    if (output_format == "ini" and input_obj.depth > 2) or (
+        output_format == "nml" and input_obj.depth != 2
+    ):
+        msg = "Cannot write depth-%s input to type-'%s' output" % (input_obj.depth, output_format)
+        logging.error(msg)
+        raise UWConfigError(msg)
+
+
 def realize_config(
     input_file: OptionalPath,
     input_format: str,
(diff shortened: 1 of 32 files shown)


src/uwtools/config/core.py
```diff
@@ -325,11 +326,11 @@ def parse_include(self, ref_dict: Optional[dict] = None):
...
-            elif isinstance(value, str) and "!INCLUDE" in value:
-                filepaths = value.lstrip("!INCLUDE [").rstrip("]").split(",")
-                # Update the dictionary with the values in the included file.
-                self.update_values(self._load_paths(filepaths))
-                del ref_dict[key]
+            elif isinstance(value, str):
+                if m := re.match(r"^\s*%s\s+(.*)" % INCLUDE_TAG, value):
+                    filepaths = yaml.safe_load(m[1])
+                    self.update_values(self._load_paths(filepaths))
+                    del ref_dict[key]
```
maddenp-noaa, line 333: Use `re` and `yaml` to process include tags in config files. The behavior of the regex aligns better with how a native YAML tag (which is what is being simulated in non-YAML configs) would behave. For example, the previous version expected exactly one space between the tag and the start of the value, while an actual YAML file would permit an arbitrary number of spaces here -- as the regex does. We can also use `yaml` to parse the actual array-of-strings value.

src/uwtools/config/core.py
```diff
@@ -675,14 +676,35 @@ def compare_configs(
...
-    cfg_b = _cli_name_to_config(config_b_format)(config_b_path)
+    cfg_a = format_to_config(config_a_format)(config_a_path)
+    cfg_b = format_to_config(config_b_format)(config_b_path)
     logging.info("- %s", config_a_path)
     logging.info("+ %s", config_b_path)
     logging.info("-" * MSGWIDTH)
     return cfg_a.compare_config(cfg_b.data)
 
 
+def format_to_config(cli_name: str) -> Type[Config]:
```
maddenp-noaa, line 687: This function turns out to be useful outside of just CLI context, and so is renamed. It also now leverages the `FORMAT` object to map strings (e.g. `"nml"`) to their respective config classes (e.g. `NMLConfig`).

src/uwtools/tests/test_cli.py
```diff
@@ -209,29 +215,71 @@ def test_dispatch_template_render_yaml(caplog):
...
-def test_parse_args():
+
+def test__parse_args():
     raw_args = ["foo", "--bar", "88"]
     with patch.object(cli, "Parser") as Parser:
-        cli.parse_args(raw_args)
+        cli._parse_args(raw_args)
         Parser.assert_called_once()
         parser = Parser()
         parser.parse_args.assert_called_with(raw_args)
```
maddenp-noaa, line 248: Here and below, new tests for the now-optional (but sometimes required!) data-format CLI arguments.

src/uwtools/utils/file.py
```diff
@@ -7,13 +7,62 @@
...
+    fieldtable: str = _fieldtable
+    ini: str = _ini
+    jinja2: str = _jinja2
+    nml: str = _nml
+    sh: str = _ini
+    yaml: str = _yaml
+    yml: str = _yaml
+
+
+FORMAT = _FORMAT()
```
maddenp-noaa, line 47: Here's the dataclass object now accessed elsewhere to refer to specific strings used as config filename extensions and CLI values.

src/uwtools/config/core.py
christinaholtNOAA: Nice.

src/uwtools/tests/config/test_atparse_to_jinja2.py
```diff
@@ -21,7 +21,7 @@ def atparselines():
 
 @fixture
 def atparsefile(atparselines, tmp_path):
-    path = tmp_path / "atparse"
+    path = tmp_path / "atparse.txt"
```
christinaholtNOAA, line 24: Any particular reason for this addition? I'm only asking because I'm scared that you're going to say that something somewhere requires all config files to have a suffix.

If it's testing an unrecognized suffix, that's less frightening.

src/uwtools/tests/test_cli.py
```diff
@@ -130,77 +131,82 @@ def test_dispatch_config_realize():
...
-        dry_run=6,
+        values_needed=6,
+        dry_run=7,
     )
     with patch.object(cli.uwtools.config.templater, "render") as m:
         with patch.object(cli.sys, "argv", ["foo", "--bar", "88"]):
-            cli.dispatch_template_render(args)
+            cli._dispatch_template_render(args)
     assert m.called_once_with(args)
     assert logged(caplog, "Command: foo --bar 88")
```
christinaholtNOAA, line 211: I can't say that I understand this test. We're patching `argv` and we want to make sure those values are being processed by argparse. But then we also check that `m` (outside its scope) is called (at least?) once with `args` though?

christinaholtNOAA (approved):
This is a cleaner approach, for sure. 

I left a few comments and questions, but don't see anything wrong...I don't think.

src/uwtools/tests/config/test_atparse_to_jinja2.py
maddenp-noaa: Good question -- I thought about commenting this line and decided not to, but I'm glad you asked. Since I made `atparse` one of the canonical format strings in the `FORMAT` object, I wanted to make sure that, when someone (probably me) did a search across the codebase to make sure that standalone instances of these strings weren't reappearing, they wouldn't be thrown off by this case, where `atparse` isn't actually meaningful in the same sense that the strings in `FORMAT` are. I figured that if someone (me) saw `atparse.txt`, they (I) would realize, "Oh, that's just a filename and could be anything".

src/uwtools/tests/test_cli.py
maddenp-noaa: Good eye. This test is weird because the function it tests is weird, because that function is doing something no other function does: In addition to calling into the right place in the `uwtools` library to access the requested functionality (`uwtools.config.templater.render()`) and passing arguments that would normally have been parsed from `argv`, it's also accessing `argv` directly to log the command that was issued from the command line. This is unfortunate, because nowhere else this deep into the CLI code -- and certainly not downstream of it, as we'd previously ensured -- is `argv` being accessed: Everywhere else, in all the `_dispath_*()` functions, `argv` has already been parsed and the functions are receiving only the parsed, not raw, `args`. But we had a request to log the command-line command for re-run capability, and that's why `argv` is being accessed here.

All the tests of these `_dispatch_*()` functions are pretty basic plubming tests, just making sure that the dispatch function has access to all the arguments it needs, and that it wires them up to the library code correctly. That's what's tested here with `args`, the mock `m` standing in for `uwtools.config.templater.render()`, and the `called_once_with(args)` assertion. Totally orthogonal to that is the mocking of `argv` and the assertion that the code logs what would have appeared on the command line.

I don't like it either.

One potential improvement would be to make two separate tests to avoid the illusion that the two things being tested are related in any way, because they aren't.

Another improvement might be to factor this logging of the original command line out and do it for _all_ commands, which would make this one dispatch function less of a snowflake.

What do you think?

src/uwtools/tests/test_cli.py
```diff
@@ -12,113 +12,114 @@
 
 from uwtools import cli
 from uwtools.tests.support import logged
+from uwtools.utils.file import FORMAT
 
 # Test functions
 
 
-def test_abort(capsys):
+def test__abort(capsys):
```
NaureenBharwaniNOAA, line 20: Just want to confirm that there are two underscores here, the first for the function name definition and the second for the private member `_abort`

If this is correct, is this the standard? There is a similar example of this in PR UW-259

NaureenBharwaniNOAA (approved):
LGTM!

src/uwtools/tests/test_cli.py
maddenp-noaa: I'm going ahead with moving the command logging to the `main()` function (it's only active in `--verbose` mode) so that it applies to all commands. We can change later if needed, but for now this makes the code and tests more consistent.

maddenp-noaa:
I'm going to go ahead and merge this. Unfortunately, I found an interesting edge case in the file/format checking: For `template render` mode, _all_ arguments are optional, in all cases, because the code will take values from the environment or as key-value pairs on the command line, so `--values-format` need not be specified even if `--values-file` is _not_ specified. So, there's a latent bug there I'll need to fix. I'll see if I can do so today.

src/uwtools/tests/test_cli.py
maddenp-noaa: Yes, that's correct and, yes, I'd recommend this as a standard: It's `test_`, and then whatever the function/method name is, including another underscore if that's how the function/method is named.

Issue #286, opened 2023-08-30
UW-362 rename uwtools.config.str_to_type() & consider mandating YAML values


WeirAE:
**Description**
Strictly, str_to_type does not convert to a type, so 'reification' is a better description. YAML can also more cleanly handle acceptable inputs with `yaml.safe_load()`

**Requirements**

**Acceptance Criteria (Definition of Done)**
function has an accurate name
string values are parsed with the YAML library and acceptable inputs are synonymous with acceptable YAML values

WeirAE:
Completed by #288

PR #288, opened 2023-09-01
UW-362: rename uwtools.config.str_to_type() & consider mandating YAML values

WeirAE:
**Description**
Acceptance Criteria:
Function has an accurate name
String values are parsed with the YAML library and acceptable inputs are synonymous with acceptable YAML values

Jira issue UW-362
Fixes issue #286 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/config/core.py
+++ b/src/uwtools/config/core.py
@@ -267,7 +267,7 @@ class Config(ABC, UserDict):
 
                     # Put the full template line back together as it was, filled or not, and make a
                     # guess on its intended type.
-                    ref_dict[key] = self.str_to_type("".join(data))
+                    ref_dict[key] = self.reify_str("".join(data))
 
     def dereference_all(self) -> None:
         """
@@ -332,31 +332,14 @@ class Config(ABC, UserDict):
                     self.update_values(self._load_paths(filepaths))
                     del ref_dict[key]
 
-    def str_to_type(self, s: str) -> Union[bool, float, int, str]:
+    def reify_str(self, s: str) -> Union[bool, float, int, str]:
         """
-        Reify a string to a Python object, if possible.
-
-        Try to convert Boolean-ish values to bool objects, int-ish values to ints objects, and
-        float-ish values to float objects. Return the original string if none of these apply.
+        Reify a string to a Python object, using YAML. Jinja2 templates will be passed as-is.
 
         :param s: The string to reify.
         """
         s = s.strip("\"'")
-        if s.lower() in ["true", "yes", "yeah"]:
-            return True
-        if s.lower() in ["false", "no", "nope"]:
-            return False
-        # int
-        try:
-            return int(s)
-        except ValueError:
-            pass
-        # float
-        try:
-            return float(s)
-        except ValueError:
-            pass
-        return s
+        return s if "{{" in s or "{%" in s else yaml.safe_load(s)
 
     def update_values(self, src: Union[dict, Config], dst: Optional[Config] = None):
         """
--- a/src/uwtools/config/core.py
+++ b/src/uwtools/config/core.py
@@ -340,7 +340,7 @@ class Config(ABC, UserDict):
         """
         try:
             r = yaml.safe_load(s)
-        except (yaml.YAMLError, AttributeError):
+        except yaml.YAMLError:
             return s
         return s if type(r) in [dict, list] else r
 
--- a/src/uwtools/config/core.py
+++ b/src/uwtools/config/core.py
@@ -267,7 +267,7 @@ class Config(ABC, UserDict):
 
                     # Put the full template line back together as it was, filled or not, and make a
                     # guess on its intended type.
-                    ref_dict[key] = self.reify_str("".join(data))
+                    ref_dict[key] = self.reify_scalar_str("".join(data))
 
     def dereference_all(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


src/uwtools/config/core.py
```diff
@@ -332,31 +332,14 @@ def parse_include(self, ref_dict: Optional[dict] = None) -> None:
...
-            return int(s)
-        except ValueError:
-            pass
-        # float
-        try:
-            return float(s)
-        except ValueError:
-            pass
-        return s
+        return s if "{{" in s or "{%" in s else yaml.safe_load(s)
```
maddenp-noaa, line 342: 1. I love the simplicity of this. I think my description in Jira suggested something more complicated, like repeatedly asking `yaml` to parse the value and seeing when it changed, but now I'm not sure why. This seems right: Either `yaml` parses the `str` to a new value and returns it, or it remains a `str` and it returns it. All good, right?
2. But what about an `s` like `{ n: 88 }` or `[1, 2, 3]`? Those will get parsed into `dict` and `list` objects, and those aren't part of our return type, so should we just return `s` if we get one of these back from `safe_load()`? Or should we raise an exception here since we got something we didn't expect? It probably depends on the context from which this function is being called...
3. I also wonder if we can't simplify the Jinja2 handling, and also catch other weird values. For example, trying to `safe_load()` either `s = "{{ foo }}"` or `s = "@[foo]" both raise `yaml` exceptions (different ones, too!). Would it be a good idea to call `safe_load()` in a `try` and just return `s` if there's an error?

Considering the above, would something like this work?
```
try:
    r = yaml.safe_load(s)
except yaml.YAMLError:
    return s
return s if type(r) in [dict, list] else r
```

WeirAE: It'll also handle type-ing `int` and `float` too! I like letting the YAMLError handle it,  I definitely agree with that!
Though in regards to 2, thats sort of why I put this up as a draft - while its lovely, consistent and clean here, it will absolutely reify everything. There may be even more things we want to keep as strings, so I wanted to make sure we accounted for all of them here. Though that brings up the point that, if we're only handling a small subset, we're not really just reifying anymore and should either clarify the subset in the function name or only call it when we're handling that subset

christinaholtNOAA: I like the simplified `try/except` approach. Can we get away with removing the `strip` in the code given that we're now loading with YAML?

christinaholtNOAA (commented):
I like this simplification a lot!

And it maps directly to how all other YAML values would be read in.

src/uwtools/tests/config/test_core.py
```diff
@@ -862,14 +862,14 @@ def test_Config_characterize_values(nml_cfgobj):
...
-    assert nml_cfgobj.str_to_type("3.14") == 3.14
-    assert nml_cfgobj.str_to_type("NA") == "NA"  # no conversion
+def test_Config_reify_str(nml_cfgobj):
+    for x in ["true", "yes", "TRUE"]:
+        assert nml_cfgobj.reify_str(x) is True
+    for x in ["false", "no", "FALSE"]:
+        assert nml_cfgobj.reify_str(x) is False
+    assert nml_cfgobj.reify_str("88") == 88
+    assert nml_cfgobj.reify_str("3.14") == 3.14
+    assert nml_cfgobj.reify_str("maybe") == "maybe"  # no conversion
```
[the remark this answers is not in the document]
WeirAE: Thats fair - I put it in-line with some of the task examples, but it doesn't really need to be changed, I can put it back

src/uwtools/config/core.py
WeirAE: Turns out the list is considered an `AttributeError` not `YAMLError`, so I made it catch both

maddenp-noaa: Sorry, what's the context where an `AttributeError` is raised? Is it related to this assertion in the unit tests?
```
assert nml_cfgobj.reify_scalar_str([1, 2, 3]) == [1, 2, 3]
```
If so, that's a `list` being passed in, not a `str`, so that shouldn't be a valid use case. We wouldn't want to catch an exception caused by that input, because the caller would be misusing the function.

src/uwtools/tests/config/test_core.py
```diff
@@ -862,14 +862,16 @@ def test_Config_characterize_values(nml_cfgobj):
...
-    assert nml_cfgobj.str_to_type("NA") == "NA"  # no conversion
+def test_Config_reify_scalar_str(nml_cfgobj):
+    for x in ["true", "yes", "TRUE"]:
+        assert nml_cfgobj.reify_scalar_str(x) is True
+    for x in ["false", "no", "FALSE"]:
+        assert nml_cfgobj.reify_scalar_str(x) is False
+    assert nml_cfgobj.reify_scalar_str("88") == 88
+    assert nml_cfgobj.reify_scalar_str("3.14") == 3.14
+    assert nml_cfgobj.reify_scalar_str("NA") == "NA"  # no conversion
+    assert nml_cfgobj.reify_scalar_str("cycle: '06'") == "cycle: '06'"
```
maddenp-noaa, line 873: I know this was the example I gave, but I feel like it's kind of obvious that this whole string would remain a string while having just
```
assert nml_cfgobj.reify_scalar_str("'06'") == "'06'"
```
or even just
```
assert nml_cfgobj.reify_scalar_str("'88'") == "'88'"  # quoted int not converted
```
placed next to the other `88` test might make the intent clearer.

WeirAE: Thats fair - I was using the full entry as the example for its own context, but no need to use the tests to prove a point

src/uwtools/config/core.py
WeirAE: Oh I just processed that all wrong in my head, I've fixed it and added the fail on list as a case too

src/uwtools/tests/config/test_core.py
```diff
@@ -862,14 +862,19 @@ def test_Config_characterize_values(nml_cfgobj):
...
+def test_Config_reify_scalar_str(nml_cfgobj):
+    for x in ["true", "yes", "TRUE"]:
+        assert nml_cfgobj.reify_scalar_str(x) is True
+    for x in ["false", "no", "FALSE"]:
+        assert nml_cfgobj.reify_scalar_str(x) is False
+    assert nml_cfgobj.reify_scalar_str("88") == 88
+    assert nml_cfgobj.reify_scalar_str("'88'") == "88"  # quoted int not converted
+    assert nml_cfgobj.reify_scalar_str("3.14") == 3.14
+    assert nml_cfgobj.reify_scalar_str("NA") == "NA"  # no conversion
+    assert nml_cfgobj.reify_scalar_str("@[foo]") == "@[foo]"  # no conversion for YAML exceptions
```
maddenp-noaa, line 874: Nice one.

maddenp-noaa (approved):
Looks good!

PR #289, opened 2023-09-02
flexible-arg-post-processing

maddenp-noaa:
**Description**

I discovered some discrepancies in how arguments should be treated in various modes that made the idea of a single function to do all final checks on parsed arguments look infeasible. For example, under `uw config realize`, `--values-format` is mandatory if `--values-file` is not defined (because `uw` cannot know how it should interpret data appearing on `stdin`), or when the format cannot be deduced from a given `--values-file` argument. But under `uw template render`, `--values-format` is **not** mandatory even if `--values-file` is not defined, because the user's intent might be to use values from the environment or from key-value pairs provided on the command line; but `--values-format` **is** mandatory if `--values-file` is specified but its format cannot be deduced from the provided filename. So, in short, final validation of arguments is probably going to depend on the particular semantics of each mode.

To support that, this PR delegates the definition of those checks to each mode, so that the checks are defined in the same place as the arguments themselves. After the arguments are parsed, the checks that the selected [sub]mode specifies are carried out on the arguments.

This PR also introduces a container for canonical string values used in the CLI, to avoid repetition of hard-to-check magic string values.

**How Has This Been Tested?**

Passes `make test` locally, plus GitHub Actions.

**Checklist**

### DIFF
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -83,18 +83,18 @@ def _add_subparser_config_compare(subparsers: Subparsers) -> SubmodeChecks:
     """
     parser = _add_subparser(subparsers, STR.compare, "Compare configs")
     required = parser.add_argument_group(TITLE_REQ_ARG)
-    _add_arg_file_path(required, switch=_arg2sw(STR.file1path), helpmsg="Path to file 1")
-    _add_arg_file_path(required, switch=_arg2sw(STR.file2path), helpmsg="Path to file 2")
+    _add_arg_file_path(required, switch=_switch(STR.file1path), helpmsg="Path to file 1")
+    _add_arg_file_path(required, switch=_switch(STR.file2path), helpmsg="Path to file 2")
     optional = _basic_setup(parser)
     _add_arg_file_format(
         optional,
-        switch=_arg2sw(STR.file1fmt),
+        switch=_switch(STR.file1fmt),
         helpmsg="Format of file 1",
         choices=FORMATS,
     )
     _add_arg_file_format(
         optional,
-        switch=_arg2sw(STR.file2fmt),
+        switch=_switch(STR.file2fmt),
         helpmsg="Format of file 2",
         choices=FORMATS,
     )
@@ -368,7 +368,7 @@ def _dispatch_template_render(args: Namespace) -> bool:
 
 def _add_arg_config_file(group: Group) -> None:
     group.add_argument(
-        _arg2sw(STR.cfgfile),
+        _switch(STR.cfgfile),
         "-c",
         help="Path to config file",
         metavar="PATH",
@@ -379,7 +379,7 @@ def _add_arg_config_file(group: Group) -> None:
 
 def _add_arg_dry_run(group: Group) -> None:
     group.add_argument(
-        _arg2sw(STR.dryrun),
+        _switch(STR.dryrun),
         action="store_true",
         help="Only log info, making no changes",
     )
@@ -409,7 +409,7 @@ def _add_arg_file_path(group: Group, switch: str, helpmsg: str, required: bool =
 
 def _add_arg_input_file(group: Group, required: bool = False) -> None:
     group.add_argument(
-        _arg2sw(STR.infile),
+        _switch(STR.infile),
         "-i",
         help="Path to input file (defaults to stdin)",
         metavar="PATH",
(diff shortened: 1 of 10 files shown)


src/uwtools/cli.py
```diff
@@ -22,28 +24,32 @@
...
+        args, checks = _parse_args(sys.argv[1:])
+        for check in checks[args.mode][args.submode]:
+            check(args)
         setup_logging(quiet=args.quiet, verbose=args.verbose)
         logging.debug("Command: %s %s", Path(sys.argv[0]).name, " ".join(sys.argv[1:]))
+        modes = {
+            STR.config: _dispatch_config,
+            STR.forecast: _dispatch_forecast,
+            STR.template: _dispatch_template,
+        }
```
maddenp-noaa, line 52: This block has the first of many references to the `STR` object to obtain string values. Just in the course of preparing this PR, my editor was able to catch a number of would-be typos, e.g. `STR.tempate` was flagged immediately by my editor as an error, while there's nothing it could say about `"tempate"`.

src/uwtools/cli.py
```diff
@@ -587,45 +645,61 @@ def _parse_args(raw_args: List[str]) -> Namespace:
...
+
+
+def _switch(arg: str) -> str:
+    """
+    Convert argument name to long-form switch.
+
+    :param arg: Internal name of parsed argument.
+    :return: The long-form switch.
+    """
+    return "--%s" % arg.replace("_", "-")
```
maddenp-noaa, line 664: This function was factored out of the old `_set_formats()`.

src/uwtools/tests/test_cli.py
```diff
@@ -234,62 +351,28 @@ def test_main_raises_exception(capsys):
...
-def subparsers():
-    # Create and return a subparsers test object.
-    return Parser().add_subparsers()
-
-
 def submodes(parser: Parser) -> List[str]:
     # Return submodes (named subparsers) belonging to the given parser.
     if subparsers := parser._subparsers:
-        if action := subparsers._actions[3]:
+        if action := subparsers._actions[1]:
```
christinaholtNOAA, line 368: Hmmmm. This change seems a little magical.

christinaholtNOAA (approved):
LGTM! Seems like a more robust solution, and might even help save us from ourselves in a few ways.

src/uwtools/tests/test_cli.py
maddenp-noaa: It's awful. I haven't found a more robust way to introspect the subparser configuration. Using a private attribute is bad enough, but the fact that the location of the submodes in the `subparsers` object is apparently sensitive to the order in which arguments are added is even worse. I'll be on the lookout for something better, but this was all I was able to find so far.

christinaholtNOAA: Would a comment here offer any insight on what the number is for? Or where to find the code that would explain it?

WeirAE: Ugh, its like inspect.stack(), isn't it? I don't know any better way, myself. Might be able to drop a small comment?

maddenp-noaa: I'll add a comment so we at least know what it means, and where to look if something changes.

maddenp-noaa: I found a way to at least get rid of the magic number by filtering out what we need. It's still ugly, as `argparse` wasn't designed for this kind of introspection, but its a little shorter. I've also updated the comment to link to a StackOverflow article where someone asked about this.

src/uwtools/tests/test_cli.py
```diff
@@ -234,62 +352,26 @@ def test_main_raises_exception(capsys):
...
+# Helper functions
 
 
-# Helper functions
+def submodes(parser: Parser) -> List[str]:
+    # Return submodes (named subparsers) belonging to the given parser. For some background, see
+    # https://stackoverflow.com/questions/43688450.
+    if actions := [x for x in parser._actions if isinstance(x, _SubParsersAction)]:
+        return list(actions[0].choices.keys())
+    return []
```
christinaholtNOAA, line 371: I appreciate this change. I _think_ I'm now reading code that tells me in itself that it's looking through the parser for the first instance of a subparser action and returning that action's choices.

christinaholtNOAA (approved):
I like the modification. Thanks for all the extra work for the magic number. :)

WeirAE (approved):
Read through it, then read through it again with the changes, then once more to make sure I was actually following correctly - LGTM!

PR #290, opened 2023-09-06
UW-380 Create JSONSchema section for workflow tags

NaureenBharwaniNOAA:
**Description**

This work will capture all of the requirements for specifying the keys in the workflow-level YAML Rocoto definition. That includes:

```
attrs:

cycledefs:

entites:

log:

```

The tasks and dependencies subsections will be handled in other tickets.
* The schema developed should ensure the user has provided the required sections, that they are formatted appropriately, and entries have not been repeated (this might be a difficult bit to capture given how YAML load will treat this situation).
* If entities are referenced, they have been defined.
* Full documentation of the structure and assumptions made about the structured YAML should be available via the SRW docs by the time this ticket is scheduled. 

**Type of change**

**How Has This Been Tested?**

**Checklist**

`Co-authored-by: Emily Carpenter <Emily.l.Carpenter@noaa.gov>`


NaureenBharwaniNOAA (commented):
Currently, the addition of the entities key/value requirements is still needed to fulfill this tickets requirements

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,39 @@
...
+                "moabtorque",
+                "moab",
+                "pbspro",
+                "slurm",
+                "cobalt"
+            ]
+        }
+    }
+    "cycledefs": {
+        "type": "string",
```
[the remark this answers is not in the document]
christinaholtNOAA: I think it makes sense to try a regex solution after skimming some of the JSONSchema docs.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,39 @@
...
+        "format": "date" " " "date" " " "time"
+    }
+    "log": {
+        "type": "string"
```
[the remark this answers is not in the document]
christinaholtNOAA: I don't think it should be blank to pass the Rocoto requirement. Is there a way to indicate that it's required and non-blank?

IIR, Venita used "uri" to indicate that the contents should be somewhat pathlike. Does that make sense here, too?

christinaholtNOAA (commented):
Great progress and questions! I like what I see so far. :)

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,44 @@
...
+    }
+    "entities": {
+        "type": "string"
```
NaureenBharwaniNOAA, line 42: Entities will be a dict of key/value pairs. Here the key seems to be a string. Is this the only type we would be seeing?

For example: 
```
entities:
     foo: 1
     bar: “/some/path”
```

The value pairs type doesn't seem to need to be a string. Would this be sufficient to account for all options the user could supply? Entities from the SRW documentation are required, but looking over the schema_with_metask RelaxNG they don't seem to be included in the requirement for a valid Rocoto XML.

src/uwtools/resources/rocoto.jsonschema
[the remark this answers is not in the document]
NaureenBharwaniNOAA: From reading through the documentation this seems to be consistent that the log should include a pathlike string

src/uwtools/resources/rocoto.jsonschema
christinaholtNOAA: I don't think entities are necessarily required, per se. If they are provided under this section, they should be a dictionary of key-value pairs. I'm not sure if we want to necessarily restrict what the values could be? Perhaps only strings, integers, floats, and bools?

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,44 +1,55 @@
 {
-    "attrs": {
-        "realtime": {
+    "workflow": {
+        "attrs": {
```
maddenp-noaa, line 3: Does JSON Schema support making some of these attributes mandatory and others optional? And do we want to make the ones that are optional from Rocoto's perspective also optional in UW YAML? That second question isn't rhetorical -- it's a design decision potentially to-be-made. Looking at the Jinja2 template for the Rocoto XML doc, it looks like attributes will be created based on whatever is passed to Jinja2 to do the rendering, so I think we should define here anything that we want users to be able to use, and enforce the required/optional constraints we think appropriate.

Actually, given that, if I'm not mistaken, Jinja2 will roll out `key=value` attribute pairs based on whatever its given, probably we should not have empty keys in UW YAML just for the sake of schema conformance, as otherwise those will appear as `key=""` attributes, which Rocoto will probably reject. So, I think we'd want the Rocoto-optional attributes to be UW YAML-optional, too. I can't tell here if there's a required/optional distinction.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,44 +1,55 @@
...
-                "T",
-                "F",
-                "t",
-                "f",
-                "True",
-                "False",
-                "FALSE",
-                "TRUE"
-            ]
+            "pattern": "^\[\d{10},\d{10},(0[0-9]|1[0-9]|2[0-4]):([0-5][0-9]):([0-5][0-9])\]$"
```
maddenp-noaa, line 45: What do we envision appearing in UW YAML that would be validated by `cycledefs` here? In the Rocoto XML I see
```
<cycledef group="at_start">201907010000 201907010000 24:00:00</cycledef>
<cycledef group="forecast">201907010000 201907010000 24:00:00</cycledef>
```
So, one or more (per the RelaxNG schema) of these `<cycledef>` entities, each with an optional `group` attribute, with two `yyyymmddHHMM` items and then a `HH:MM:SS` item. How would that appear in UW YAML?

Also, does JSON Schema support defining patterns at a higher, reusable level? We've got `yyyymmddHHMM` repeated here, which already suggests that we should give it a name; and looking at the full realized XML I'm seeing a `HH:MM:SS` pattern repeated as the value of the `<wallclock>` entity under `<task>`, so we'll need that, too. I see that RelaxNG supports defining reusable patterns, and that's used in the Rocoto XML schema. Can we do that with JSON Schema, too?

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,44 +1,55 @@
...
-                "ll",
-                "torque",
-                "moabtorque",
-                "moab",
-                "pbspro",
-                "slurm",
-                "cobalt"
-            ]
+            "format", "uri"
+        }
```
maddenp-noaa, line 50: I'm confused about `log`. The Jinja2 template has
```
<log>{{ log }}</log>
```
which suggests to me that we'd supply something like a path string here (maybe involving an entity reference), but the realized XML from S3 shows
```
<log><cyclestr>/scratch1/BMC/zrtrr/uw/srw/expt_dirs/grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2/log/FV3LAM_wflow.log</cyclestr></log>
```
So I'm curious where the nested `<cyclestr>` comes from...

Oh, looking at the RelaxNG schema, it looks like the value of `<log>` can either be just text, or a `compoundTimeString`, which provides the `<cyclestr>` nested element. @christinaholtNOAA given that the canned SRW case uses this second form, but our Jinja2 template seems not to support this, what's the requirement?

christinaholtNOAA: The really challenging part of this thing is that a cyclestr can show up almost anywhere that you have a string, so it felt really tough to code into the Jinja2 template. I did have support for it in an earlier version with an elementtree, though.

maddenp-noaa: I would really not like to see an XML literal like `<cyclestr>` appear in UW YAML. If we can detect the condition(s) under which it needs to appear, having Python code insert that tag might be the least-bad interim solution while we're generating XML with Jinja2.

christinaholtNOAA: I'd be fine with that, for sure.

So if we found anything like @Y@m@d@H in the string, we'd need it, but then we still don't have a great way to treat the offset attribute.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,59 @@
...
+        "cycledefs": {
+            "type": "string",
+            "pattern": "^\[\d{10},\d{10},(0[0-9]|1[0-9]|2[0-4]):([0-5][0-9]):([0-5][0-9])\]$"
+        }
+        "log": {
+            "type": "string",
+            "format", "uri"
+        }
+        "entities": {
+            "type": "object"
```
NaureenBharwaniNOAA, line 56: Changed entities from a string type to an object type. Meaning "since the type validation keyword is object we can use the properties validation keyword to define a nested data structure." In this case for the entities we have a nested data structure of a dict. Without knowing concrete key names like `foo` or `bar` I'm not sure we can go further into the nested dict.

```
entities:
     foo: 1
     bar: “/some/path”
```

Is it sufficient to check here and confirm that entities does exist and is a dict? If we want to restrict the types that the values can be, how can we iterate over varying keys? I considered the naive approach would be to iterate over the keys and return them. Then use this information to make the schema accurate and restrict their types. Essentially filling in the missing information. After reading the problems with [this approach](https://github.com/json-schema-org/json-schema-spec/issues/1378) I'm not sure that's the way we want to go

NaureenBharwaniNOAA:
Another general question, do we want to add unit tests for rocoto.jsonschema under `src/uwtools/tests/drivers/` ?

src/uwtools/resources/rocoto.jsonschema
christinaholtNOAA: Here, the YAML spec might look like this:
```
  cycledefs:
    groupname:
      - !startstopfreq ['2022102000', ‘2023102018’, ‘06:00:00’]
    groupname2:
      - !startstopfreq ['2022102000', ‘2023102018’, ‘24:00:00’]
      - !startstopfreq ['2022102006', ‘2023102018’, ‘24:00:00’]
```

So, it would need to support an object where keys were arbitrary and values are lists of these date strings.

src/uwtools/resources/rocoto.jsonschema
christinaholtNOAA: I think it's fine for this work to leave it with just being a dictionary.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: Circling back on checking if JSON Schema supports defining patterns and reusing them:

It seems that JSONSchema supports, however it may not be _easily_. I'm finding [discussion](https://github.com/json-schema-org/json-schema-spec/issues/886) online that others want to do a similar thing here, where we can add our regex expression in a constant and reuse the variable when needed. 

There is an option to use `$ref` which after much googling will allow us to

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,72 @@
+{
+    "definitions": {
+        "dateTimePattern": {
+            "type": "string",
+            "pattern": "^(\\d{12}) (\\d{12}) (\\d{2}):(\\d{2}):(\\d{2}):(\\d{2})$"
+        }
```
NaureenBharwaniNOAA, line 6: Circling back on checking if JSON Schema supports defining patterns and reusing them:

It seems that JSONSchema does not support constants/declaring variables. However, I'm finding [discussion](https://github.com/json-schema-org/json-schema-spec/issues/886) online that others want to do a similar thing here, where we can add our regex expression in a constant and reuse the variable when needed. 

It seems to be that you we cannot _directly_ save an expressions to variable and reuse them within the schema. There is no built-in method for defining variables.

The best way to do this was to define `definitions` that holds the reusable section we want to use in multiple places in our schema. In this way we are referencing the the information to avoid duplication of code.

@maddenp-noaa @christinaholtNOAA

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,72 @@
...
+                        },
+                        "taskthrottle": {
+                            "type": "integer",
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["realtime", "scheduler"]
+                },
+                "cycledefs": {
+                    "$ref": "#/definitions/dateTimePattern"
```
NaureenBharwaniNOAA, line 58: Here is where we reference the dateTimePattern being used for the start, stop, increment method for the cycledef tag

Ex:
```
<cycledef>201101010000 201112311800 06:00:00</cycledef>
<cycledef>yyyymmddhhmm yyyymmddhhmm dd:hh:mm:ss</cycledef>
```

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,72 @@
...
+                },
+                "log": {
+                    "type": "string",
+                    "format": "uri"
+                },
+                "entities": {
+                    "type": "object"
+                }
+            },
+            "required": ["attrs", "cycledefs", "log"]
```
NaureenBharwaniNOAA, line 68: I let entities off the `required` list since they are not required to be supplied by the user

src/uwtools/resources/rocoto.jsonschema
maddenp-noaa: I'd like to see what you come up with. Anything that cuts down on repetition of schema elements that should always be the same thing would be a plus.

src/uwtools/resources/rocoto.jsonschema
maddenp-noaa: This looks like it's headed in the right direction -- maybe it's even there. I just did some light googling to try to understand, and I found https://json-schema.org/understanding-json-schema/structuring.html#defs, which kind of sounds like it's come into play here. I see a `definitions` section like what you've got in this part of the docs https://json-schema.org/understanding-json-schema/structuring.html#bundling, but that seems to be within a `$defs` block. I really don't know, but this bit of text from the docs seems pertinent to our use case:

> Sometimes we have small subschemas that are only intended for use in the current schema and it doesn’t make sense to define them as separate schemas. Although we can identify any subschema using JSON Pointers or named anchors, the `$defs` keyword gives us a standardized place to keep subschemas intended for reuse in the current schema document.

Does that come into play here?

src/uwtools/resources/rocoto.jsonschema
maddenp-noaa: Would validation fail a YAML block with `entities` if they were not at least included as _optional_ items of the schema? How does JSON Schema treat extraneous items that are not part of the schema?

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: Added a new Jira ticket to handle the `<cyclestr>` tag, [linked here](https://jira.epic.oarcloud.noaa.gov/browse/UW-386)

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: I see, there are two different options here `$ref` and `$def`. 
`$ref` is used when we want to define reusable schemas that is used to reference an external schema elsewhere in our JSONSchema docs or external files.
`$def` is used when we want to define reusable schemas locally meaning within the same JSONSchema doc. 

I agree that here we will want to go with a `$def` block over a `$ref` block.

NaureenBharwaniNOAA: I see, there are two different options here `$ref` and `$def`. 
`$ref` is used when we want to define reusable schemas that is used to reference an external schema elsewhere in our JSONSchema docs or external files.
`$def` is used when we want to define reusable schemas locally meaning within the same JSONSchema doc. 

I agree that here we will want to define a `$def` block to hold the reusable definitions.

src/uwtools/tests/config/test_validator.py
```diff
@@ -4,12 +4,15 @@
 """
 
 import json
+from importlib import resources
 from pathlib import Path
 from typing import Any, Dict
 
+from lxml import etree
```
NaureenBharwaniNOAA, line 11: Will remove this in my next pus.

src/uwtools/tests/config/test_validator.py
```diff
@@ -121,6 +124,14 @@ def test_validate_yaml_pass(config, config_file, schema, schema_file):
...
 
 
+def test_validate_workflow_tags_pass():
+    # Test a fully valid config file passes schema
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        schema_file = resc / "rocoto.jsonschema"
+    config_file = support.fixture_path("hello_workflow_tags.yaml")
+    assert validator.validate_yaml(schema_file=schema_file, config_file=config_file)
+
+
```
NaureenBharwaniNOAA, line 134: I utilized the `validate_yaml` method here that already "checks whether the given config file (yaml) conforms to the given JSON Schema spec and whether any filesystem paths it identifies do not exist." I found this to be a good place to add the unit test since it seemed to fit in nicely. Let me know if you'd like to pull it out.

One thing to note is once we have each subtask from UW-320 working properly and merged into the `rocoto.jsonschema` then we can remove these individual fixtures such as `hello_workflow_tags` and we can soimply test against `hello_workflow`. Then we can change our unit test above to reflect this as well.

src/uwtools/tests/fixtures/hello_workflow_tags.yaml
```diff
@@ -0,0 +1,11 @@
+  workflow:
```
NaureenBharwaniNOAA, line 1: Compared to `hello_workflow` which contains task, metatask and dependency tags, `hello_workflow_tags` has to include `workflow` to pass (as of right now). This is because within the `rocoto.jsonschema` we include the `workflow` tag.

src/uwtools/tests/config/test_validator.py
maddenp-noaa: > we can remove these individual fixtures ... and we can simply test against ...
Yes! As we start to get more realistic schemas and test data, we should be able to remove some placeholder tests data and concentrate on the realistic stuff. Nice.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,60 @@
...
+                        },
+                        "taskthrottle": {
+                            "type": "integer",
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["realtime", "scheduler"]
+                },
+                "cycledefs": {
+                    "$ref": "#/$defs/dateTimePattern"
```
WeirAE, line 46: Hmm, shouldn't this be 'string' and not "dateTimePattern" here and in the yaml?

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,60 @@
...
+                                "ll",
+                                "torque",
+                                "moabtorque",
+                                "moab",
+                                "pbspro",
+                                "slurm",
+                                "cobalt"
+                            ]
+                        },
+                        "cycylethrottle": {
```
WeirAE, line 34: Looks like an extra 'y' snuck in here

src/uwtools/tests/fixtures/hello_workflow_tags.yaml
```diff
@@ -0,0 +1,11 @@
...
+    attrs:
+      realtime: false
+      scheduler: slurm
```
WeirAE, line 4: Probably add 'cyclethrottle' here too just to make sure all the tags are going in and being checked?

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,69 @@
...
+                    "required": ["realtime", "scheduler"]
+                },
+                "cycledefs": {
+                    "type": "object",
+                    "properties": {
+                        "howdy": {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                                "$ref": "#/$defs/dateTimePattern"
```
WeirAE, line 52: You shouldn't need to type here since the $ref is handling that - do we need that 'howdy' section though?

src/uwtools/tests/fixtures/hello_workflow_tags.yaml
```diff
@@ -0,0 +1,12 @@
...
+      cyclethrottle: 5
+    cycledefs:
+      howdy: 
+        - 202209290000 202209300000 06:00:00
```
WeirAE, line 8: I wonder if its mad about no heading line and the dash?

NaureenBharwaniNOAA: Fixed this with Christina's solution to the `properties` key missing and changing the type from `string` to `array`

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: I was able to add in that example `hello_workflow_tags` that includes an `entities` block and it passes. It is included in the schema as `optional`. Meaning if it's not in the `required` section it is `optional`. 

I've decided to make a `hello_workflow_invalid.yaml` that tests what happens if an item that is not part of the schema and it should fail. After going through the tests for each of the `tags`: `attrs`, `cycledef`, `log`, `entities`. All fail as if an input is not as expected, EXCEPT `cycledef`.

I added a test to see if it would catch if the format of the `cycledef` did not match the regex. It does not catch that correctly. It should pass the `assert not` with true in the tests.

Not sure exactly why it is passing the regex match.

NaureenBharwaniNOAA: Tried to play around with the idea of making `groupname` and `$ref` required using `"required": ["groupname", "$ref"]`

christinaholtNOAA (approved):
LGTM.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: @christinaholtNOAA

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,69 @@
...
+                        "cyclethrottle": {
+                            "type": "integer",
+                            "minimum": 0
+                        },
+                        "taskthrottle": {
+                            "type": "integer",
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["realtime", "scheduler"]
```
NaureenBharwaniNOAA, line 43: It seems that the `required keyword here is not appropriately catching when an invalid yaml does not include `realtime` or `scheduler`

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,69 @@
...
+                },
+                "log": {
+                    "type": "string",
+                    "format": "uri"
+                },
+                "entities": {
+                    "type": "object"
+                }
+            },
+            "required": ["attrs", "cycledefs", "log"]
```
NaureenBharwaniNOAA, line 65: Here as well, the `required keyword here is not appropriately catching when an invalid yaml does not include `attrs`, `cycledef`, `log` or `entities`. However, it is catching `"required": "ref"` correctly

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: Fixed this issue

src/uwtools/tests/config/test_validator.py
```diff
@@ -126,6 +128,21 @@ def test_validate_yaml_pass(config, config_file, schema, schema_file):
...
+    config_file = support.fixture_path("hello_workflow_tags.yaml")
+    assert validator.validate_yaml(
+        schema_file=schema_file, config_file=config_file, check_path=False
+    )
+
+    config_file = support.fixture_path("hello_workflow_invalid.yaml")
+    assert not validator.validate_yaml(
+        schema_file=schema_file, config_file=config_file, check_path=False
+    )
+
```
NaureenBharwaniNOAA, line 144: Everything passes here for the tests. Before we close this ticket out, I wanted to have a discussion on if we think we should separate these two sections out into separate unit tests?

One uses a valid yaml confirming that it validates with the schema, and the other confirms that an invalid yaml fails the schema validation. I would consider separating them out since they are testing different configs. However, thinking about clean coding I'm not sure if it's useful to repeat getting the path to the schema twice. 

Perfectly fine if we feel like leaving this alone as well.

NaureenBharwaniNOAA: Actually looking at this again, I tried to perform `parameterization` on this test:

```
@pytest.mark.parametrize("fn", ["hello_workflow_tags.yaml", "hello_workflow_invalid.yaml"])
131 def test_validate_workflow_tags_pass(fn):
132     # Test a fully valid config file passes schema
133     with resources.as_file(resources.files("uwtools.resources")) as resc:
134         schema_file = resc / "rocoto.jsonschema"
135 
136     config_file = support.fixture_path(fn)
137     assert validator.validate_yaml(
138         schema_file=schema_file, config_file=config_file, check_path=False
139     )
140 
141     assert not validator.validate_yaml(
142         schema_file=schema_file, config_file=config_file, check_path=False
143     )
```

@maddenp-noaa would you consider trying to do this? I'm not sure if this is possible if I want to do an `assert not` and an `assert` with two different files. Thoughts?

maddenp-noaa: I think I'd break them up or, more likely, `parametrize` them as you're doing. Since `validator.validate_yaml()` returns a `bool`, I think we could e.g.
```
@pytest.mark.parametrize("vals", [("hello_workflow_tags.yaml", True), ("hello_workflow_invalid.yaml", False)])
def test_validate_workflow_tags(vals):
    fn, validity = vals
    with resources.as_file(resources.files("uwtools.resources")) as resc:
        schema_file = resc / "rocoto.jsonschema"
    config_file = support.fixture_path(fn)
    assert validator.validate_yaml(
        schema_file=schema_file, config_file=config_file, check_path=False
    ) is validity
```
`is validity` will be either `is True` or `is False` depending on the `vals` parameter.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,70 @@
...
+                            "type": "integer",
+                            "minimum": 0
+                        }
+                    },
+                    "required": ["realtime", "scheduler"]
+                },
+                "cycledefs": {
+                    "type": "object",
+                    "properties": {
+                        "groupname": {
```
christinaholtNOAA, line 48: I am curious about this entry...the `cycledefs` section can include any number of arbitrarily defined entries.

christinaholtNOAA (approved):
This still looks good! I just have one additional question below.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: I updated the valid `hello_workflow.yaml` to account for various different sections we could see within the `cycledef` and the tests pass accordingly. Let me know if you feel like this is sufficient. @christinaholtNOAA

maddenp-noaa (approved):

PR #292, opened 2023-09-13
UW-382 Create JSONSchema section for dependency section tags

WeirAE:
**Description**

This work will capture all of the requirements for specifying the keys in the dependency-level YAML Rocoto definition. The workflow and task/metatask subsections will be handled in other tickets

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**


WeirAE:
This is a very rough shape without recursion

christinaholtNOAA (commented):
It seems like Rocoto has handled the case of recursion in a way that may work nicely in JSONSchema files.

Notice how `dependency_tree` is defined as a reusable construct (starting on line 225 of [schema_with_metatasks.rng](https://github.com/christopherwharrop/rocoto/blob/develop/lib/workflowmgr/schema_with_metatasks.rng), then it's referenced in any of the tags that can be nested (not, and, or, nor, etc.). I believe the same should be possible in the syntax of JSONSchema if you define a similar dependency tree under `$defs`. It's only a hunch, though.

christinaholtNOAA:
That's interesting. Proceed with caution. While this can generate some nice ideas and examples, some (maybe even many) of those things are not consistent with Rocoto.

WeirAE:
Yeahhh - can't trust these robots to do anything to the letter, but its an interesting feature for sure. I think its more of a proof-of-concept and question if we want to set it that explicitly or keep it as general as the current version but with the internal `$ref` looping

WeirAE:
Hmm. It looks like one of the `properties` declarations in the schema is wrong, working on tracking it down

WeirAE:
OK, specifically its erroring on the `validator._bad_paths` check. I tried throwing it through ChatGPT to check for errors and it did not find this cause, although it did make the following suggestions:

1. The "patternProperties" keyword should be used with the "properties" keyword inside the "oneOf" objects to specify a pattern for object keys and properties. However, it's not valid to use "patternProperties" directly under "type: object" without "properties."
2. In the "some" property, you have "properties" defined. This is not valid inside an "items" definition. You can use "items" and "properties" separately.

WeirAE:
Updated with the invalid test, but its failing....on passing the validation? The example I added should fail on multiple categories     in the jsonschema, I'm not sure what the issue is

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,251 @@
...
+            },
+            "text": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        }
+      },
+      "properties": {
+        "taskdep": {
```
christinaholtNOAA, line 120: The `taskdep` entry is another one that can follow the `_arbitrary_string` suffix pattern.

Actually all of these below here can...

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,251 @@
...
+              "type": "string"
+            },
+            "right": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "left",
+            "right"
+          ]
```
christinaholtNOAA, line 161: This one should probably have additional properties for consistency.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,251 @@
...
+        "taskvalid": {
+          "type": "object",
+          "properties": {
+            "task": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        "metataskdep": {
```
christinaholtNOAA, line 197: It's hard to tell from the indentation, but it's probably good to make sure the `metataskdep` section also has an additionalProperties tag applied.

christinaholtNOAA (commented):
This is shaping up nicely! 

A couple of more general things here, and a few inline below...

The dependency tags should live only under a task heading like this:

```
workflow:
  tasks:
    task_hello:
      dependency:
```

I think we should make sure that is built into this PR with placeholders for all the content PR #293. 

I'm curious if "additional_properties = False" could be applied at a higher level as the default for all the properties to reduce some replication.

src/uwtools/resources/rocoto.jsonschema
WeirAE: Fair! Inconveniently, as far as I know, `additionalProperties` and `required` have to be specified for every `properties` entry. I'll look into any options for making it defaulted though

src/uwtools/resources/rocoto.jsonschema
WeirAE: oh wow, all of them? Ok, can do

WeirAE:
AFAIK `additionalProperties` can't be specified globally, but I'll look into it. For PR #293 , we were talking about how to handle that. I think we might have to move everything inside of `$def`/`definitions` to make referencing at arbitrary depths work. I think I can put the base part of that loop in here to make merging #293 easier

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,243 @@
...
+              "items": {
+                "$ref": "#/$defs/dateTimePattern"
+              },
+              "required": [
+                "groupname"
+              ]
+            },
+            "required": [
+              "groupname"
+            ]
```
NaureenBharwaniNOAA, line 74: Have `groupname` required twice here. I think you only need it outside the `properties` section once

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,240 @@
...
+            },
+            "^some(_.*)?$": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "minItems": 1
             },
-            "required": ["attrs", "cycledefs", "log"]
+            "^threshold(_.*)?$": {
```
christinaholtNOAA, line 230: It looks like `threshold` is an attribute of `some` in the Rocoto docs.

christinaholtNOAA (changes requested):

src/uwtools/resources/rocoto.jsonschema
WeirAE: Woah, that got shifted over somewhere along the way, let me just do a full doc recheck, thanks for catching this!

WeirAE: Ah, it also has to loop too, I'll get that in

src/uwtools/tests/fixtures/hello_workflow_invalid.yaml
```diff
@@ -9,3 +9,14 @@
       ACCOUNT: myaccount
       FOO: test.log
     log: /some/path/to/&FOO;
+
+task_get_extrn_ics:
+  dependency:
```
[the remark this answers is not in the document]
WeirAE: Updated to fit that pattern, consistent with #293

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,70 +1,251 @@
...
+          "format": "uri"
+        },
+        "entities": {
+          "type": "object"
+        },
+        "tasks": {
+          "type": "object",
+          "$ref": "#/$defs/taskPattern"
+        },
+        "dependency": {
```
christinaholtNOAA, line 238: This section will also need to go under the tasks sections and task definitions.

christinaholtNOAA (changes requested):

src/uwtools/resources/rocoto.jsonschema
WeirAE: Ah, sloppy pasting here from moving everything around, I'll move it over

christinaholtNOAA (approved):
LGTM.

src/uwtools/tests/config/test_validator.py
```diff
@@ -202,6 +213,35 @@ def test_validate_yaml_rocoto_invalid_walltime_pattern(rocoto_assets, caplog):
...
+    with patch.object(validator, "YAMLConfig") as YAMLConfig:
+        YAMLConfig().data = config
+        assert not validator.validate_yaml(**kwargs)
+        assert logged(
+            caplog,
+            "'maybe' does not match any of the regexes: "
+            "'^(and|or|not|nand|nor|xor)(_.*)?$', '^(streq|strneq)(_.*)?$', "
+            "'^cycleexistdep(_.*)?$', '^datadep(_.*)?$', '^metataskdep(_.*)?$', "
+            "'^some(_.*)?$', '^taskdep(_.*)?$', '^taskvalid(_.*)?$', '^timedep(_.*)?$'",
+        )
```
WeirAE, line 230: I really don't like this long line, but I had issues getting the first part to match I think due to the `records` in `logged()`

maddenp-noaa: I can give you a trick for doing a partial match, which `logged()` doesn't support. I'll get back to you shortly with it...

maddenp-noaa: In `test_templater.py`, there's currently this assertion using `logged()`:
```
assert logged(caplog, "roses are red, violets are blue")
```
That has to match, as you noted, a complete line of log output. If I do `import re` at the top of the test module, I could alternatively do things like
```
assert any(re.match(r"^.*red, violets.*$", record.message) for record in caplog.records)
```
to assert that `red, violets` appears on some line, anywhere on that line or, say,
```
assert any(re.match(r"^roses are red.*$", record.message) for record in caplog.records)
```
to assert that `roses are red` appears at the _start_ of some line.

If you find this useful, it might be worth adding it to `uwtools.tests.support` as a helper similar to `logged()` so that you could use it now but it would also be available for future tests without much repetition.

WeirAE: Ah, ok - hmm, could we update `logged()` to something like `return any(msg in [record.message for record in caplog.records])`? We could also move the re.match there - I don't like the way the list comprehension looks personally, though

maddenp-noaa: In re: `return any(msg in [record.message for record in caplog.records])`, I don't think so, because the expression `msg in [record.message for record in caplog.records]` evaluates to a `bool` value, and `any()` accepts an iterable as its argument:
```
>>> any(2 in [1, 2, 3])
TypeError: 'bool' object is not iterable
```
If we made `logged()` regex-based, all call sites would have to be updated to give complete, anchored regexes, e.g. `^roses are red, violets are blue$` instead of simple strings, to avoid unintentional partial matches. I feel like it's common enough that we can and want to match complete log lines that having a simple, non-regex based `logged()` is helpful and simple to use in most use cases; having something like `logged_regex()` for power-use cases might be nice to have, though -- your use case certainly begs for it.

WeirAE: Hmm, yeah, I missed that - breaking it out of the brackets to just  `return any(msg in record.message for record in caplog.records)` works by leaving the `any()` to do the evaluation, but I think I'll still break the regex off into a helper like you said though

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -114,44 +270,50 @@
       ],
       "anyOf": [
         {
-            "not": {
-                "required": ["exclusive"]
-            }
-        },
-        {
-            "not": {
-                "required": ["shared"]
```
christinaholtNOAA, line 123: I'm not following the logic on why the logic here is removed. I see below it's mainly just reformatting.

WeirAE: Hrm, could be a conflict resolution error, let me recheck everything

WeirAE: Oh weird, `jq -r` removed it... I'm overriding it for now - any idea why it would do that? @maddenp-noaa ?

maddenp-noaa: Yes: You have (or had), two `anyOf` keys at the same level. That's not possible in a JSON object (as it would not be possible in a Python `dict`), so there's something wrong with the schema -- if we've got a schema with two identical keys at the same level, then the schema is incorrectly constructed. `jq` did the right thing by removing one of them. In general, "last in wins" with these things; Python behaves the same:
```
>>> d = {"a": 1, "a": 2, "a": 3}
>>> d
{'a': 3}
```

WeirAE: Hrm, I'm struggling a bit to solve this one - @NaureenBharwaniNOAA 's implication use here from 381 is excellent otherwise. Can we stash these within `dependentSchemas`? ChatGPT only got confused when I asked it, of course.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -112,45 +268,65 @@
         "command",
         "walltime"
       ],
-      "anyOf": [
-        {
-            "not": {
-                "required": ["exclusive"]
-            }
+      "dependentSchemas": {
```
NaureenBharwaniNOAA, line 271: Nice! Glad you were able to get it working!

src/uwtools/tests/config/test_validator.py
```diff
@@ -202,6 +213,29 @@ def test_validate_yaml_rocoto_invalid_walltime_pattern(rocoto_assets, caplog):
         assert logged(caplog, "'0:01:00' does not match '^[0-9]{2}:[0-9]{2}:[0-9]{2}$'")
 
 
+def test_validate_yaml_rocoto_invalid_dependency_bool(rocoto_assets, caplog):
```
NaureenBharwaniNOAA, line 216: Last idea, do we want to place your two new test cases for `dependencies` in alphabetical order with the others I previously added? More of a question on preference. Feel free to do whatever you think looks best.

NaureenBharwaniNOAA (approved):
Everything looks great to me. Nice job!

christinaholtNOAA (approved):
LGTM.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -112,45 +268,65 @@
...
+          ]
         },
-        {
-            "not": {
-                "required": ["stdout"]
-            }
+        "stderr": {
+          "required": [
+            "stdout"
+          ]
```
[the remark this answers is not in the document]
WeirAE: With help from Paul!

src/uwtools/tests/config/test_validator.py
```diff
@@ -154,7 +165,7 @@ def test_validate_yaml_rocoto_invalid_dependency(rocoto_assets, caplog):
     with patch.object(validator, "YAMLConfig") as YAMLConfig:
         YAMLConfig().data = config
         assert not validator.validate_yaml(**kwargs)
-        assert logged(caplog, "'stderr' is a dependency of 'stdout'")
+        assert logged(caplog, "'stderr' is a required property")
```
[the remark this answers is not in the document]
WeirAE: Good point, will do!

src/uwtools/tests/config/test_validator.py
WeirAE: Good point, will do!

PR #293, opened 2023-09-13
UW-381 Create JSONSchema section for task and metatask tags

NaureenBharwaniNOAA:
**Description**

- This work will capture all of the requirements for specifying the keys and attributes in the tasks in YAML Rocoto definition.
- The workflow and dependencies subsections will be handled in other tickets.

- The schema developed should ensure the user has provided the required sections, that they are formatted appropriately, and entries have not been repeated (this might be a difficult bit to capture given how YAML load will treat this situation).
- If entities are referenced, they have been defined.
- Full documentation of the structure and assumptions made about the structured YAML should be available via the SRW docs by the time this ticket is scheduled.

**Type of change**

**How Has This Been Tested?**

**Checklist**

- [X] My changes need updates to the documentation.  I have made corresponding changes to the documentation

Co-authored-by: Paul Madden <paul.madden@noaa.gov>


maddenp-noaa:
Can we standardize on formatting our JSON code with `jq`? If so, you can use it like this:
```
jq --indent 4 . <file>
```
and capture the output. If no one feels strongly about the number of spaces, I'd recommend using `jq`'s default of 2 and omitting the `--indent 4` switch for brevity.

If we standardize on a formatter and settings, we can all contribute code that matches in style and doesn't produce diffs when edited later by someone else (because they'll format with the same tool and settings).

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -0,0 +1,102 @@
...
+        "task": {
+            "type": "object",
+            "properties": {
+                "attrs": {
+                    "type": "object",
+                    "properties": {
+                        "name": {
+                            "type": "object"
+                        },
+                        "maxtrices": {
```
maddenp-noaa, line 12: `maxtries`?

At some point -- maybe sooner rather than later -- I think we should have a representative YAML file in this repo and either have a procedure for testing it against this schema from the command line with `uw config validate`, or via a unit test. Otherwise, it's going to be hard to judge the correctness of the schema, and we may not notice typos. I think we probably need to co-evolve the schema and a representative YAML file.

maddenp-noaa (changes requested):

src/uwtools/resources/rocoto.jsonschema
christinaholtNOAA: I could add some from the SRW tests. Any idea how many we might want? Would just one be a good enough place to start?

maddenp-noaa: One would be infinitely better than zero.

NaureenBharwaniNOAA: One test would definitely be a great place to start. I also see in all the documentation that `maxtries` is actually the correct spelling. This should be updated now in the latest push.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -3,6 +3,82 @@
...
+            "properties": {
+                "attrs": {
+                    "type": "object",
+                    "properties": {
+                        "name": {
+                            "type": "object"
+                        },
+                        "maxtries": {
+                            "type": "integer",
+                            "minimum": 0
```
NaureenBharwaniNOAA, line 18: I'm seeing a contridction between the `schema_with_metatasks` and the documentation [here](https://ufs-srweather-app.readthedocs.io/en/develop/CustomizingTheWorkflow/DefineWorkflow.html#defining-a-task) and [here](http://christopherwharrop.github.io/rocoto/).

It looks like the `schema_with_metatasks` has the `maxtries` listed as:

```
      <!-- Optional task maxtries attribute -->
      <optional>
        <attribute name="maxtries">
          <data type="positiveInteger"/>          
        </attribute>
      </optional>
```
Looking at this it seems like `maxtries` should be an `integer` since it needs to be a positive `integer`, however in the documentation it is listed as a `string`. I'm trying to figure out the best way to approach this.

christinaholtNOAA (changes requested):
This is shaping up very nicely. Just a few things below.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -3,68 +3,250 @@
...
+                            "throttle": {
+                                "type": "integer",
+                                "minimum": 0
+                            },
+                            "cycledefs": {
+                                "type": "string"
+                            },
+                            "final": {
+                                "type": "string",
+                                "enum": [
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: Not sure what you mean here, could you expand a bit more? A bool for as long as attrs is an `object` then it's fine? Something I thought about looking at this is, is the yaml below valid if we do not have an `attrs` section altogether?

```
workflow:
    attrs:
      realtime: false
      scheduler: slurm
      cyclethrottle: 5
    cycledefs:
      howdy: 
        - 202209290000 202209300000 06:00:00
        - 202210200000 202310201800 02:00:00
      groupname: 
        - 202209290000 202209300000 06:00:00
      groupname2: 
        - 202209290000 202209300000 06:00:00
    entities:
      ACCOUNT: myaccount
      FOO: test.log
    log: /some/path/to/&FOO;
    tasks:
      task_hello:
        account: "&ACCOUNT;"
        command: "echo hello $person"
        walltime: 00:01:00
        memory: 512M
        deadline: 202209290000
        envar:
          name: siri
        cores: 3
        native: hello
```
Meaning should I be removing this required for `task_*` to include `attrs` at all? If `attrs` is left blank currently we get a type object none error. Seeing as everything is optional in `attrs` I'm guessing we want to remove this requirement? Only for `task_*` not for `workflow` tag

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -3,68 +3,241 @@
...
+                        "type": "string",
+                        "format": "uri"
+                    },
+                    "entities": {
+                        "type": "object"
+                    },
+                    "tasks": {
+                        "type": "object",
+                        "$ref": "#/$defs/taskPattern",
+                        "$ref": "#/$defs/metataskPattern"
```
NaureenBharwaniNOAA, line 236: Figured out I had this added in that made it a circular pattern, here for the `tasks` section I only needed `taskPattern` and then call `metataskPattern` from `taskPattern`!

src/uwtools/tests/fixtures/hello_workflow_tags.yaml
```diff
@@ -15,3 +15,32 @@
...
+        attrs:
+          cycledef: howdy
+        account: "&ACCOUNT;"
+        command: "echo hello $person"
+        walltime: 00:01:00
+        memory: 512M
+        deadline: 202209290000
+        envar:
+          name: siri
+        join: hello
```
NaureenBharwaniNOAA, line 29: Still working on making join/stderr/stdout optional and nodes/cores/native required

src/uwtools/resources/rocoto.jsonschema
WeirAE: Annoyingly, it may require maximum with minimum to count, maybe something arbitrary like:
```
"exclusiveMinimum": 0
"maximum": 9999
```
I looked at some conditional versions and I don't like how any of them came out

NaureenBharwaniNOAA: There was no upper limit requirement I found reading in the docs, but I used the solution above and it seemed to be exactly what I was looking for. I could also use `maximum: _` for an upper limit if we wanted one. Solution above seemed to be good if you're looking for something.

WeirAE: Hey, if it works it works!

src/uwtools/resources/rocoto.jsonschema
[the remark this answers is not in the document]
christinaholtNOAA: I think it works just fine if we make `attrs` optional for a `task` entry.

For the boolean part, I was suggesting that instead of a variety of true/false type strings, we just require it to be something that YAML will interpret as a bool. If we do that, we can clean up this enum section to something like this:

```
final: {
    "type": "boolean"
}
```

christinaholtNOAA (commented):
This looks great! What a convoluted way to need to specify these different tags, though!

I just left a couple minor things below.

WeirAE:
Good fixes! Just have to update the branch I think

NaureenBharwaniNOAA (commented):
Another layer:
If we remove `"additionalProperties": false` in the schema then we see that the schema does not traverse the yaml completely. This causes an issue when we have a `task` property under the `metatask` tag within the yaml because we have three required `properties` of `task` which are `command`, `walltime` and `cores`. Within `"additionalProperties": false` we see that a user could supply a yaml that does not include one of these required `properties` causing further disruption down the line.

Our two options are:
1. For now leave the `var` property off completely the schema and continue with `"additionalProperties": false` being included. This would mean a user would not be able to supply `var` at all in their yaml.
2. We can allow a user to supply `var` in their yaml and not make it `required`, then we would have to remove `"additionalProperties": false` in the schema and this would mean users could supply an invalid yaml without supplying all the `required properties` for `task` (specifically one of the three required properties could be left off).

src/uwtools/tests/config/test_validator.py
```diff
@@ -135,13 +136,123 @@ def test_validate_workflow_tags_pass(vals):
     fn, validity = vals
     with resources.as_file(resources.files("uwtools.resources")) as resc:
         schema_file = resc / "rocoto.jsonschema"
-    config_file = support.fixture_path(fn)
+    config_file = fixture_path(fn)
     assert (
         validator.validate_yaml(schema_file=schema_file, config_file=config_file, check_path=False)
         is validity
     )
```
NaureenBharwaniNOAA, line 144: I think with the addition of the `@fixture` tests that Paul fleshed out we should be able to completely remove the `test_validate_workflow_tags_pass` along with their respective`hello_workflow_tags.yaml` and `hello_workflow_invalid.yaml` files that are referenced within the test case. 

Eeventually when we merge, `hello_workflow.yaml` should remain since past work makes use of this file.

NaureenBharwaniNOAA (commented):
With some minor comments that address the test cases and any clean up comments, this PR should be good to merge shortly.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,12 +1,174 @@
...
+            "not": {
+                "required": ["shared"]
+            }
+        },
+        {
+            "not": {
+                "required": ["exclusive"]
+            }
+        }
+      ],
```
maddenp-noaa, line 128: Is this `anyOf` + `not` pattern how we specify mutually exclusive but also optional properties? As in, "You don't have to specify either of these, but if you specify one you cannot specify the other"?

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -1,12 +1,174 @@
...
+        {
+            "not": {
+                "required": ["join"]
+            }
+        }
+      ],
+      "dependencies": {
+        "stdout": ["stderr"],
+        "stderr": ["stdout"]
+      }
```
maddenp-noaa, line 144: On page https://json-schema.org/understanding-json-schema/reference/conditionals, if you search for `dependencies`, it sounds like this syntax as semi-deprecated (or, at least, belongs to an older draft) in favor of two more-specific keywords, `dependentRequired` and `dependentSchemas`. Would it be useful / possible for us to "upgrade" to that newer syntax here?

src/uwtools/tests/config/test_validator.py
```diff
@@ -135,13 +136,123 @@ def test_validate_workflow_tags_pass(vals):
...
+    schema_file, config_dict = rocoto_assets
+    with patch.object(validator, "YAMLConfig") as YAMLConfig:
+        yaml_config = YAMLConfig()
+        yaml_config.data = config_dict
+        assert (
+            validator.validate_yaml(
+                schema_file=schema_file, config_file="/not/used", check_path=False
+            )
+            is True
+        )
```
maddenp-noaa, line 253: Nice work expanding these tests!

We could DRY this out a little more by having the fixture do, say,
```
return {"schema_file": schema_file, "config_file": "/not/used", "check_path": False}, { ... same big config dict ...}
```
unpacking `rocoto_assets` as
```
kwargs, config_dict = rocoto_assets
```
in each test function, and then doing the `validate_yaml()` calls like
```
assert validator.validate_yaml(**kwargs)) is ...
```
Since we (I think) make the `validate_yaml()` call exactly the same way in each test function, we could make the keyword arguments we pass to it part of the fixture and DRY that out. That might collapse about 6 lines in each test function down to 1, as it'll be a lot shorter.

You could also consider doing `assert` and `assert not` as shorthand, though I know that's not what I did when we were pairing...

maddenp-noaa: In the fixture, if you want to be more explicit, you could also do
```
kwargs = {"schema_file": schema_file, "config_file": "/not/used", "check_path": False}
config_dict = { ... same big config dict ... }
return kwargs, config_dict
```
That also introduces some consistent names between the variables that are packed into the fixture's return value and the variables they are unpacked into in the test functions. Might be nice.

And maybe it should just be called `config` and not `config_dict` -- we usually try to avoid encoding type info in variable names. (I know, my fault from our pairing session...)

maddenp-noaa (commented):
I somehow missed this in a previous PR, but would you consider, in `uwtools.config.validator`:

1. Adding a `:param` entry for the new-ish `check_path` argument to `validate_yaml()`?
2. Changing the name of that argument to `check_paths` (plural), because it is meant to deal with potentially multiple paths -- see e.g. the fact that it calls `_bad_paths()` (plural)?

Thanks.

src/uwtools/tests/config/test_validator.py
```diff
@@ -135,13 +136,123 @@ def test_validate_workflow_tags_pass(vals):
...
+            is False
+        )
+        assert logged(caplog, "'0:01:00' does not match '^[0-9]{2}:[0-9]{2}:[0-9]{2}$'")
+
+
+def test_validate_yaml_rocoto_valid(rocoto_assets):
+    schema_file, config_dict = rocoto_assets
+    with patch.object(validator, "YAMLConfig") as YAMLConfig:
+        yaml_config = YAMLConfig()
+        yaml_config.data = config_dict
```
maddenp-noaa, line 247: One more opportunity for brevity, if you're interested: The previous two lines can be simplified to
```
YAMLConfig().data = config_dict
```
That says: "The thing that the `Mock` object `YAMLConfig` returns when it is called -- i.e. the value of the expression `YAMLConfig()` -- should have an attribute called `data`, and that attribute's value should be `config_dict`".

My interest in making these tests concise is in making it as painless as possible to add more tests if we want them -- the less code not really accomplishing anything, the better, IMO.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: Yes it is! Exactly. We don't need either `shared` or `exclusive`, but if one is supplied the other cannot be. Take a look at `stderr`, `stdout` and `join`. These 3 properties are not rewuired either, but in this case `stderr` and `stdout` have to show up as a pair if one or the other is supplied, and cannot appear with `join`. So we have a `stderr` AND `stdout` OR `join` situation. I did this with an added `dependencies` on `stdout` and `stderr`.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: I attempted to do this by using the `dependentRequired` keyword and when including that over `dependencies` it automatically passes the tests when we should see a failure if both `stdout` and `stderr` are not supplied together. I am working on this issue, but unsure why `dependentRequired` would not behavior the same as `dependencies` since from the link you supplied they seem to be a 1:1 replacement for each other. I added another test for the `dependencies` feel free to take a look at it and test it out.

src/uwtools/tests/config/test_validator.py
NaureenBharwaniNOAA: I'm trying to fully understand what you mean by:

"Since we (I think) make the validate_yaml() call exactly the same way in each test function, we could make the keyword arguments we pass to it part of the fixture and DRY that out. That might collapse about 6 lines in each test function down to 1, as it'll be a lot shorter."

I think you mean something like this:
```
kwargs = {"schema_file": schema_file, "config_file": "/not/used", "check_path": False}
config_dict = { ... same big config dict ... }
validate = 
    yaml_config = YAMLConfig()
        yaml_config.data = config
        assert not (
            validator.validate_yaml(
                **kwargs
            )
        )
return kwargs, config, validate
```

And then within the test case:
```
with patch.object(validator, "YAMLConfig") as YAMLConfig:
        validate
```

src/uwtools/tests/config/test_validator.py
```diff
@@ -128,18 +129,121 @@ def test_validate_yaml_pass(config, config_file, schema, schema_file):
...
+                }
+            }
+        }
+    }
+    return kwargs, config
+
+
+def test_validate_yaml_rocoto_invalid_dependency(rocoto_assets, caplog):
+    kwargs, config = rocoto_assets
+    config["workflow"]["tasks"]["metatask"]["task"].update({"stdout": "hello"})
```
NaureenBharwaniNOAA, line 162: Here I added a new test case on the `dependencies` for `stderr` and `stdout`. When replacing with `dependentRequired` I see we pass when we should not be. Having trouble tracking this one down and fully understanding why it's not a 1:1 replacement for each other.

src/uwtools/tests/config/test_validator.py
maddenp-noaa: What you did in https://github.com/ufs-community/workflow-tools/pull/293/files#diff-362fb561e035f3c6d293bd5b896a5fee3e568a9ca70dde56a1a6027224cb13ecR151-R157 is just what I had in mind. I think that's about as concise as we can get.

src/uwtools/resources/rocoto.jsonschema
maddenp-noaa: I wouldn't spend a lot of time on it if it doesn't fall into place easily -- we should just go with what you've got working, and we can update later if we need to. If you don't like not having it behave the way you expect, maybe spend an hour or so on it before shelving it?

NaureenBharwaniNOAA: I think the added test case on the dependencies will help catch any changes in the future on the issue. For example, if someone did change it to `dependentRequired` blindly without testing it, then the test case would catch this and it would fail. I'm happy leaving this how it is currently. The additions over the last two days have been monumental in getting this ticket to the point of where we wanted it to be at.

PR #298, opened 2023-09-18
meta.json-and-pip-deps

maddenp-noaa:
Commit `recipe/meta.json` to git to support easier installation for end users until we are able to provide pre-built packages.


christinaholtNOAA (approved):
LGTM.

PR #302, opened 2023-10-03
uw-369

maddenp-noaa:
**Description**

Provide a DRY wrapper for `subprocess` calls and update existing `subprocess` call sites.

**How Has This Been Tested?**

Local devshell `make test` + GitHub Actions checks.

**Checklist**


README.md
```diff
@@ -58,7 +58,7 @@ bash Miniforge3-Linux-aarch64.sh -bfp ~/conda
 rm Miniforge3-Linux-aarch64.sh
 source ~/conda/etc/profile.d/conda.sh
 conda activate
-conda install -y -c maddenp condev=0.4.0
+conda install -y -c maddenp condev=0.5.3
```
maddenp-noaa, line 61: Here and in the diff just below, update `condev` versions to track updates to that package. Older versions will generally work, but when I can verify that the latest release works, I'd like to suggest using that.

src/uwtools/apps/srw.py
```diff
@@ -29,15 +30,19 @@ def load_config(self, config_file: str) -> None:
...
+                    "python config_utils.py",
+                    "-c %s" % config_file,
+                    "-t $PWD/config_defaults.yaml",
+                    "-o yaml",
+                    ">%s" % f,
+                ]
+                cmd = " ".join(cmd_components)
+                result = run(cmd=cmd)
+                if not result.success:
+                    raise UWError(f"Command failed: {cmd}")
```
maddenp-noaa, line 45: Replace direct `subprocess.run()` call with a call to the new wrapper, which also checks the command's exit status and returns it to Python, so we can respond if the command failed -- in this case by raising an exception.

src/uwtools/drivers/driver.py
```diff
@@ -59,7 +59,7 @@ def resources(self, platform: dict) -> Mapping:
         """
 
     @abstractmethod
-    def run(self) -> None:
+    def run(self) -> bool:
```
maddenp-noaa, line 62: We can now have drivers' `run()` methods report success/failure.

src/uwtools/drivers/forecast.py
```diff
@@ -186,27 +188,18 @@ def run(self) -> None:
...
+            result = run(cmd=f"{batch_command} {outpath}")
         else:
-            subprocess.run(
-                f"{run_command}",
-                stderr=subprocess.STDOUT,
-                check=False,
-                shell=True,
-            )
+            result = run(cmd=run_command)
+        return result.success
```
maddenp-noaa, line 202: Presumably the caller can decide what to do if the command failed, so just return a `bool` status value rather than raising an exception.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -328,13 +322,8 @@ def test_FV3Forecast_run_dry_run(caplog, fv3_run_assets):
 def test_run_submit(fv3_run_assets):
     batch_script, config_file, config = fv3_run_assets
     with patch.object(FV3Forecast, "_validate", return_value=True):
-        with patch.object(forecast.subprocess, "run") as sprun:
+        with patch.object(forecast, "run") as run:
```
maddenp-noaa, line 325: The `"run"` being patched is now `run()` from `uwtools.utils.run`, which is `import`ed into the `uwtools.drivers.forecast` module.

src/uwtools/utils/run.py
maddenp-noaa, line 1: I borrowed this `run()` function, and its tests, from `iotaa`.

src/uwtools/drivers/forecast.py
```diff
@@ -153,9 +153,11 @@ def resources(self, platform: dict) -> Mapping:
             "walltime": "00:01:00",
         }
 
-    def run(self) -> None:
+    def run(self) -> bool:
         """
-        Runs FV3 either as a subprocess or by submitting a batch script.
+        Runs FV3 either locally or via a batch-script submission.
+
+        :return: Did the FV3 run exit with success status?
```
christinaholtNOAA, line 160: Oh, it sure does just move merrily along when the executable fails.

src/uwtools/tests/drivers/test_forecast.py
christinaholtNOAA: This is getting a bit convoluted. From the context of the `assert` below, it makes sense that the run is from `uwtools.utils.run`, but from the line `with patch.object(forecast, "run") as run:`, it's quite unclear that this isn't the `Forecast.run` method unless you read pretty hard into it.

christinaholtNOAA (approved):
I think this looks good as far as I can tell. I can confirm that my early tests with the forecast driver were not handling failures in the subprocess at all, which made it tough.

I am just slightly concerned about namespace issues with the new function `run`.  We're quickly overloading what it means to "run" something. Maybe there isn't a good way to get around this though? We do want to "run" a forecast and "run" a subprocess. In the case of the forecast `run` method though it might submit a batch job or execute a program in  subprocess.

maddenp-noaa:
@christinaholtNOAA I agree, the proliferation of "run" is confusing and not helpful. I think maybe the easiest thing would be to rename the `run()` function in the `uwtools.utils.run` module, and maybe the name of the module itself. Any thoughts on names? IIRC we cannot use `exec()` as a name because it conflicts (per `pylint`) with Python's built-in `exec()` built-in function. We could go full-on `execute()`, or... `sprun()` for "subprocess run", though something about encoding `subprocess` in the name feels bad to me. `external()` or `ext()`?

docs/environment.yml
maddenp-noaa, line 1: Now that `recipe/meta.json` lives full-time in our repo (and is not generated as-needed by `condev` and then discarded), we don't need `condev` to render the docs (it was really only used to get the version number from `recipe/meta.yaml` -- CI can now expect to find that in `recipe/meta.json`). Also, the ReadTheDocs build picked up a newer version of `sphinx` (and, I think, `sphinx-gallery`) that didn't provide the "rtd theme" (whatever that is), so I've pinned a couple packages in a way that allows the ReadTheDocs build to complete again.

christinaholtNOAA:
I am kind of leaning toward `execute` at the moment. As we discussed earlier today, it technically executes in a bash subshell, so even `bash_execute` might be pretty clear on what it's doing. I think it could be helpful to convey that it's sending the command to the system for execution.

maddenp-noaa:
I'm going to try `execute()` -- let's see what we think.

.github/scripts/build-documentation
```diff
@@ -5,7 +5,6 @@
 set -e
 source conda/etc/profile.d/conda.sh
 conda activate
-make meta # to make recipe/meta.json available to doc-build logic
```
maddenp-noaa, line 8: We no longer need this since `recipe/meta.json` is part of the repo now.

src/uwtools/drivers/forecast.py
maddenp-noaa: The exit status of NWP codes has been historically unreliable (Fortran for a long time didn't support Unix-style exit codes, but I believe it does now, and many codes now use it), but we might as well give them the benefit of the doubt and try to use them.

maddenp-noaa:
OK, I've renamed the `uwtools.utils.run` module to `uwtools.utils.processing`, and renamed the `run()` function therein to `execute()`. I think it's a lot clearer now.

PR #305, opened 2023-10-04
uw-317

maddenp-noaa:
**Description**

Add GitHub Actions logic to build and publish a conda package to the `ufs-community` channel on anaconda.org.

I thought more about how to test this, and I think a good work plan might be:

1. Merge this PR, which bumps the version number to `1.1.0`, to `develop`.
2. Merge `develop` to `main`, which should trigger the `release` CI job, which should tag the new `HEAD` commit `v1.1.0-0` (the `-0` is the conda build number -- useful, though seldom used, for publishing a new package where something about the package that is not a code change -- e.g. a change to dependency packages), and build and publish a `uwtools-1.1.0-0` package to anaconda.org.
3. After verifying the correctness of the above and fixing anything that breaks, remove the `v1.1.0-0` tag from the repo and remove the new packages from anaconda.org -- because we're not really ready to do a release yet. The `develop` branch will still have the version number at `1.1.0`, but it will be a pre-release version number until we're ready to do another merge to `main` and a _real_ release.

**Type of change**

**How Has This Been Tested?**

Will be tested upon merge to the (new) `main` branch.


recipe/meta.json
```diff
@@ -1,5 +1,5 @@
 {
-  "build": "pyh28a5fc4_0",
+  "build": "pyh233a6f4_0",
```
NaureenBharwaniNOAA, line 2: Is this a version of the build? I'm not sure exactly what this is for and after doing some basic google searching for `pyh233a6f4_0` I wasn't able to find anything online. Is this something specific to our repo? Any context would help here

.github/scripts/publish
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -eux
+
+set +ux
```
NaureenBharwaniNOAA, line 3: Could you confirm about what this is doing here? 

From this [resource](https://askubuntu.com/questions/1167965/what-does-set-x-do-in-a-bash-script), I understand that `set +x` disables the individual printing of commands/arguments when the code is being executed. So basically hiding this information from the command line. Where later on `set -x` enables it again.


This [article](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html), talks about treating unset variables/parameters as an error for the `set +u` command.

WeirAE (approved):
These are always tricky because you have to pull to test, but it looks good for this phase!

NaureenBharwaniNOAA (approved):
Some general comments to confirm my understanding. Looks good!

recipe/meta.json
christinaholtNOAA: I'm interested in this, too! How does meta.json get populated in general? Is it a manual process?

christinaholtNOAA (approved):
LGTM.

recipe/meta.json
maddenp-noaa: https://github.com/ufs-community/workflow-tools/blob/develop/recipe/README.md

PR #308, opened 2023-10-05
disconnect-jenkins

maddenp-noaa:
We still have a Jenkins CI configuration enabled, which is being triggered when I'm trying to merge `develop` -> `main`. I'd like to remove this for now to keep it from interfering with our release process. I can commit to reworking and reactivating Jenkins CI integration if we discover later that we need it.


christinaholtNOAA (approved):
This is good with me.

Issue #316, opened 2023-10-10
UW-322 validate a rocoto xml workflow


WeirAE:
**Description**
as a user, I would like to validate my rocoto xml workflow, given the rocoto native schema
Requires #315 

**Acceptance Criteria (Definition of Done)**
shall extend the cli such that users can validate a rocoto xml document
shall respond with a simple yes/no
shall support a -verbose flag to provide whatever LXML provides as validtion output

WeirAE:
completed by #317

Issue #315, opened 2023-10-10
UW-334 add a modal CLI support for the Rocoto XML tool


WeirAE:
**Description**
Add a modal CLI support for the Rocoto XML tool, required by UW-322 #316 

**Acceptance Criteria (Definition of Done)**
Shall have a CLI command to run the Rocoto XML tool
Shall provide help for the Rocoto XML tool command
Shall take UW YAML and output Rocoto XML
Shall do inline validation of the input
Shall do an inline validation of the output

WeirAE:
completed by #317

PR #317, opened 2023-10-10
UW 322 as a user, I would like to validate my rocoto xml workflow, given the rocoto native schema

WeirAE:
**Description**
Add the Rocoto XML tool to the modal CLI, including both XML generation and validation. Generation must validate both input and output. Include appropriate tests at all steps.

Due to UW-334 requiring validation, both 322 and 334 will be completed together.
Completes issues #316 and #315 

**Type of change**

**How Has This Been Tested?**

- [ ] Tests on Ubunbu OS

**Checklist**

### DIFF
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -15,6 +15,8 @@ from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
 
+from lxml import etree
+
 import uwtools.config.atparse_to_jinja2
 import uwtools.config.core
 import uwtools.config.templater
@@ -422,6 +424,14 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
         )
     else:
         success = False
+    if args.verbose: #pragma: no cover
+        errors = str(etree.RelaxNG.error_log).split("\n")
+        log_method = logging.error if len(errors) else logging.info
+        log_method(
+            "%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s"
+        )
+        for line in errors:
+            logging.error(line)
     return success
 
 
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -11,7 +11,6 @@ from argparse import _ArgumentGroup as Group
 from argparse import _SubParsersAction as Subparsers
 from dataclasses import dataclass
 from functools import partial
-from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
 
@@ -307,7 +306,7 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
     )
 
 
-# Mode Rocoto
+# Mode rocoto
 
 
 def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
@@ -386,12 +385,8 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
-    with resources.as_file(resources.files("uwtools.resources")) as resc:
-        rocoto_schema = resc / "rocoto.jsonschema"
 
-    success = uwtools.rocoto.validate_rocoto_xml(
-        input_xml=args.input_file, schema_file=rocoto_schema
-    )
+    success = uwtools.rocoto.validate_rocoto_xml(input_xml=args.input_file)
     return success
 
 
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -379,17 +379,27 @@ def _dispatch_rocoto(args: Namespace) -> bool:
 
 def _dispatch_rocoto_write(args: Namespace) -> bool:
     """
-    Dispatch logic for rocoto write submode.
+    Dispatch logic for rocoto write submode. Validate input and output.
 
     :param args: Parsed command-line args.
     """
     success = True
     if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
-        uwtools.rocoto.write_rocoto_xml(
-            input_yaml=args.input_file,
-            input_template=str(args.schema_file),
-            rendered_output=str(args.output_file),
+        valid_input = uwtools.config.validator.validate_yaml(
+            config_file=args.input_file, schema_file=args.schema_file
         )
+        if valid_input:
+            uwtools.rocoto.write_rocoto_xml(
+                input_yaml=args.input_file,
+                input_template=str(args.schema_file),
+                rendered_output=str(args.output_file),
+            )
+        else:
+            success = False
+        valid_output = uwtools.rocoto.validate_rocoto_xml(
+            input_xml=args.output_file, schema_file=args.schema_file
+        )
+        success = False if not valid_output else success
     else:
         success = False
     return success
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -11,8 +11,11 @@ from argparse import _ArgumentGroup as Group
 from argparse import _SubParsersAction as Subparsers
 from dataclasses import dataclass
 from functools import partial
+from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
+
+from lxml import etree
 
 import uwtools.config.atparse_to_jinja2
 import uwtools.config.core
@@ -303,6 +306,97 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
     ).run(
         cycle=args.cycle,
     )
+
+
+# Mode Rocoto
+
+
+def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
+    """
+    Subparser for mode: rocoto
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.rocoto, "Write and validate Rocoto XML Documents")
+    _basic_setup(parser)
+    subparsers = _add_subparsers(parser, STR.submode)
+    return {
+        STR.run: _add_subparser_rocoto_write(subparsers),
+        STR.run: _add_subparser_rocoto_validate(subparsers),
+    }
+
+
+def _add_subparser_rocoto_write(subparsers: Subparsers) -> SubmodeChecks:
+    """
+    Subparser for mode: rocoto write
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.run, "Write a Rocoto XML workflow document")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    _add_arg_schema_file(required)
+    _add_arg_output_file(required)
+    return [
+        partial(_check_file_vs_format, STR.infile, STR.infmt),
+        partial(_check_file_vs_format, STR.outfile, STR.outfmt),
+    ]
+
+
+def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
+    """
+    Subparser for mode: rocoto validate
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.validate, "Validate Rocoto XML")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_verbose(optional)
+    return [
+        partial(_check_file_vs_format, STR.infile, STR.infmt),
+    ]
+
+
+def _dispatch_rocoto(args: Namespace) -> bool:
+    """
+    Dispatch logic for rocoto mode.
+
+    :param args: Parsed command-line args.
+    """
+    return {STR.run: _dispatch_rocoto_write}[args.submode](args)
+
+
+def _dispatch_rocoto_write(args: Namespace) -> bool:
+    """
+    Dispatch logic for rocoto write submode.
+
+    :param args: Parsed command-line args.
+    """
+    rocoto_class = uwtools.rocoto.CLASSES[args.rocoto]
+    return rocoto_class.write_rocoto.xml(
+        input_yaml=args.input_file,
+        input_template=str(args.schema_file),
+        rendered_output=str(args.output_file),
+    )
+
+
+def _dispatch_rocoto_validate(args: Namespace) -> bool:
+    """
+    Dispatch logic for rocoto validate submode.
+
+    :param args: Parsed command-line args.
+    """
+    success = True
+
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        with open(args.schema_file, "r", encoding="utf-8") as f:
+            schema = etree.RelaxNG(etree.parse(f))
+    tree = etree.parse(args.input_file)
+    success = schema.validate(tree)
+
+    return "Yes" if success == True else "No"
 
 
 # Mode template
@@ -717,6 +811,7 @@ class _STR:
     quiet: str = "quiet"
     realize: str = "realize"
     render: str = "render"
+    rocoto: str = "rocoto"
     run: str = "run"
     schemafile: str = "schema_file"
     submode: str = "submode"
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -337,10 +337,7 @@ def _add_subparser_rocoto_realize(subparsers: Subparsers) -> SubmodeChecks:
     _add_arg_output_file(required)
     optional = _basic_setup(parser)
     checks = _add_args_quiet_and_verbose(optional)
-    return checks + [
-        partial(_check_file_vs_format, STR.infile, FORMAT.yaml),
-        partial(_check_file_vs_format, STR.outfile, FORMAT.rocoto),
-    ]
+    return checks
 
 
 def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
@@ -354,9 +351,7 @@ def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
     _add_arg_input_file(required)
     optional = _basic_setup(parser)
     checks = _add_args_quiet_and_verbose(optional)
-    return checks + [
-        partial(_check_file_vs_format, STR.infile, FORMAT.rocoto),
-    ]
+    return checks
 
 
 def _dispatch_rocoto(args: Namespace) -> bool:
@@ -380,7 +375,7 @@ def _dispatch_rocoto_realize(args: Namespace) -> bool:
     :param args: Parsed command-line args.
     """
     success = uwtools.rocoto.realize_rocoto_xml(
-        input_yaml=args.input_file, rendered_output=str(args.output_file)
+        input_yaml=args.input_file, rendered_output=args.output_file
     )
     return success
 
@@ -395,7 +390,7 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
         rocoto_schema = resc / "rocoto.jsonschema"
 
     success = uwtools.rocoto.validate_rocoto_xml(
-        input_xml=args.input_file, schema_file=str(rocoto_schema)
+        input_xml=args.input_file, schema_file=rocoto_schema
     )
     return success
 
@@ -824,7 +819,6 @@ class _STR:
     valsfmt: str = "values_format"
     valsneeded: str = "values_needed"
     verbose: str = "verbose"
-    write: str = "write"
 
 
 STR = _STR()
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -11,6 +11,7 @@ from argparse import _ArgumentGroup as Group
 from argparse import _SubParsersAction as Subparsers
 from dataclasses import dataclass
 from functools import partial
+from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
 
@@ -315,29 +316,28 @@ def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.rocoto, "Write and validate Rocoto XML Documents")
+    parser = _add_subparser(subparsers, STR.rocoto, "Realize and validate Rocoto XML Documents")
     _basic_setup(parser)
     subparsers = _add_subparsers(parser, STR.submode)
     return {
-        STR.write: _add_subparser_rocoto_write(subparsers),
+        STR.realize: _add_subparser_rocoto_realize(subparsers),
         STR.validate: _add_subparser_rocoto_validate(subparsers),
     }
 
 
-def _add_subparser_rocoto_write(subparsers: Subparsers) -> SubmodeChecks:
+def _add_subparser_rocoto_realize(subparsers: Subparsers) -> SubmodeChecks:
     """
-    Subparser for mode: rocoto write
+    Subparser for mode: rocoto realize
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.write, "Write a Rocoto XML workflow document")
+    parser = _add_subparser(subparsers, STR.realize, "Realize a Rocoto XML workflow document")
     required = parser.add_argument_group(TITLE_REQ_ARG)
     _add_arg_input_file(required)
-    _add_arg_schema_file(required)
     _add_arg_output_file(required)
     optional = _basic_setup(parser)
     _add_arg_input_format(optional, choices=[FORMAT.yaml])
-    _add_arg_output_format(optional, choices=[FORMAT.jinja2])
+    _add_arg_output_format(optional, choices=[FORMAT.rocoto])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
@@ -355,7 +355,7 @@ def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
     required = parser.add_argument_group(TITLE_REQ_ARG)
     _add_arg_input_file(required)
     optional = _basic_setup(parser)
-    _add_arg_input_format(optional, choices=[FORMAT.jinja2])
+    _add_arg_input_format(optional, choices=[FORMAT.rocoto])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
@@ -369,36 +369,38 @@ def _dispatch_rocoto(args: Namespace) -> bool:
     :param args: Parsed command-line args.
     """
     return {
-        STR.write: _dispatch_rocoto_write,
+        STR.realize: _dispatch_rocoto_realize,
         STR.validate: _dispatch_rocoto_validate,
     }[
         args.submode
     ](args)
 
 
-def _dispatch_rocoto_write(args: Namespace) -> bool:
+def _dispatch_rocoto_realize(args: Namespace) -> bool:
     """
-    Dispatch logic for rocoto write submode. Validate input and output.
+    Dispatch logic for rocoto realize submode. Validate input and output.
 
     :param args: Parsed command-line args.
     """
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        input_template = resc / "rocoto.jinja2"
+        output_template = resc / "rocoto.jsonschema"
+
     success = True
-    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.jinja2:
-        valid_input = uwtools.config.validator.validate_yaml(
-            config_file=args.input_file, schema_file=args.schema_file
-        )
-        if valid_input:
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
+        if uwtools.config.validator.validate_yaml(
+            config_file=args.input_file, schema_file=input_template
+        ):
             uwtools.rocoto.write_rocoto_xml(
                 input_yaml=args.input_file,
-                input_template=str(args.schema_file),
+                input_template=str(input_template),
                 rendered_output=str(args.output_file),
+            )
+            success = uwtools.rocoto.validate_rocoto_xml(
+                input_xml=args.output_file, schema_file=str(output_template)
             )
         else:
             success = False
-        valid_output = uwtools.rocoto.validate_rocoto_xml(
-            input_xml=args.output_file, schema_file=args.schema_file
-        )
-        success = False if not valid_output else success
     else:
         success = False
     return success
@@ -410,11 +412,13 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        schema_file = resc / "rocoto.jsonschema"
 
     success = True
-    if args.input_format == FORMAT.jinja2:
+    if args.input_format == FORMAT.rocoto:
         success = uwtools.rocoto.validate_rocoto_xml(
-            input_xml=args.input_file, schema_file=args.schema_file
+            input_xml=args.input_file, schema_file=str(schema_file)
         )
     else:
         success = False
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -11,11 +11,8 @@ from argparse import _ArgumentGroup as Group
 from argparse import _SubParsersAction as Subparsers
 from dataclasses import dataclass
 from functools import partial
-from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
-
-from lxml import etree
 
 import uwtools.config.atparse_to_jinja2
 import uwtools.config.core
@@ -52,6 +49,7 @@ def main() -> None:
         modes = {
             STR.config: _dispatch_config,
             STR.forecast: _dispatch_forecast,
+            STR.rocoto: _dispatch_rocoto,
             STR.template: _dispatch_template,
         }
         sys.exit(0 if modes[args.mode](args) else 1)
@@ -321,8 +319,8 @@ def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
     _basic_setup(parser)
     subparsers = _add_subparsers(parser, STR.submode)
     return {
-        STR.run: _add_subparser_rocoto_write(subparsers),
-        STR.run: _add_subparser_rocoto_validate(subparsers),
+        STR.write: _add_subparser_rocoto_write(subparsers),
+        STR.validate: _add_subparser_rocoto_validate(subparsers),
     }
 
 
@@ -332,11 +330,14 @@ def _add_subparser_rocoto_write(subparsers: Subparsers) -> SubmodeChecks:
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.run, "Write a Rocoto XML workflow document")
+    parser = _add_subparser(subparsers, STR.write, "Write a Rocoto XML workflow document")
     required = parser.add_argument_group(TITLE_REQ_ARG)
-    _add_arg_input_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_output_file(required)
+    _add_arg_input_format(optional, choices=[FORMAT.yaml])
     _add_arg_schema_file(required)
     _add_arg_output_file(required)
+    _add_arg_output_format(optional, choices=[FORMAT.rocoto])
     return [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
         partial(_check_file_vs_format, STR.outfile, STR.outfmt),
@@ -353,6 +354,7 @@ def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
     required = parser.add_argument_group(TITLE_REQ_ARG)
     _add_arg_input_file(required)
     optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.rocoto])
     _add_arg_verbose(optional)
     return [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
@@ -365,7 +367,10 @@ def _dispatch_rocoto(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
-    return {STR.run: _dispatch_rocoto_write}[args.submode](args)
+    return {
+        STR.write: _dispatch_rocoto_write,
+        STR.validate: _dispatch_rocoto_validate,
+        }[args.submode](args)
 
 
 def _dispatch_rocoto_write(args: Namespace) -> bool:
@@ -374,29 +379,29 @@ def _dispatch_rocoto_write(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
-    rocoto_class = uwtools.rocoto.CLASSES[args.rocoto]
-    return rocoto_class.write_rocoto.xml(
-        input_yaml=args.input_file,
-        input_template=str(args.schema_file),
-        rendered_output=str(args.output_file),
-    )
+    success = True
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
+        uwtools.rocoto.write_rocoto_xml(
+            input_yaml=args.input_file,
+            input_template=str(args.schema_file),
+            rendered_output=str(args.output_file),
+        )
+    else:
+        success = False
+    return success
 
 
-def _dispatch_rocoto_validate(args: Namespace) -> bool:
+def _dispatch_rocoto_validate(args: Namespace) -> str:
     """
     Dispatch logic for rocoto validate submode.
 
     :param args: Parsed command-line args.
     """
     success = True
-
-    with resources.as_file(resources.files("uwtools.resources")) as resc:
-        with open(args.schema_file, "r", encoding="utf-8") as f:
-            schema = etree.RelaxNG(etree.parse(f))
-    tree = etree.parse(args.input_file)
-    success = schema.validate(tree)
-
-    return "Yes" if success == True else "No"
+    success = uwtools.rocoto.validate_rocoto_xml(
+        input_xml=args.input_file, schema_file=args.schema_file
+    )
+    return "Yes" if success is True else "No"
 
 
 # Mode template
@@ -769,6 +774,7 @@ def _parse_args(raw_args: List[str]) -> Tuple[Namespace, Checks]:
     checks = {
         STR.config: _add_subparser_config(subparsers),
         STR.forecast: _add_subparser_forecast(subparsers),
+        STR.rocoto: _add_subparser_rocoto(subparsers),
         STR.template: _add_subparser_template(subparsers),
     }
     return parser.parse_args(raw_args), checks
@@ -822,6 +828,7 @@ class _STR:
     valsfmt: str = "values_format"
     valsneeded: str = "values_needed"
     verbose: str = "verbose"
+    write: str = "write"
 
 
 STR = _STR()
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -332,12 +332,12 @@ def _add_subparser_rocoto_write(subparsers: Subparsers) -> SubmodeChecks:
     """
     parser = _add_subparser(subparsers, STR.write, "Write a Rocoto XML workflow document")
     required = parser.add_argument_group(TITLE_REQ_ARG)
-    optional = _basic_setup(parser)
     _add_arg_input_file(required)
-    _add_arg_input_format(optional, choices=[FORMAT.yaml])
     _add_arg_schema_file(required)
     _add_arg_output_file(required)
-    _add_arg_output_format(optional, choices=[FORMAT.rocoto])
+    optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.yaml])
+    _add_arg_output_format(optional, choices=[FORMAT.jinja2])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
@@ -355,8 +355,7 @@ def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
     required = parser.add_argument_group(TITLE_REQ_ARG)
     _add_arg_input_file(required)
     optional = _basic_setup(parser)
-    _add_arg_input_format(optional, choices=[FORMAT.rocoto])
-    _add_arg_verbose(optional)
+    _add_arg_input_format(optional, choices=[FORMAT.jinja2])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
         partial(_check_file_vs_format, STR.infile, STR.infmt),
@@ -384,7 +383,7 @@ def _dispatch_rocoto_write(args: Namespace) -> bool:
     :param args: Parsed command-line args.
     """
     success = True
-    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.jinja2:
         valid_input = uwtools.config.validator.validate_yaml(
             config_file=args.input_file, schema_file=args.schema_file
         )
@@ -413,7 +412,7 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
     """
 
     success = True
-    if args.input_format == FORMAT.xml:
+    if args.input_format == FORMAT.jinja2:
         success = uwtools.rocoto.validate_rocoto_xml(
             input_xml=args.input_file, schema_file=args.schema_file
         )
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -374,7 +374,7 @@ def _dispatch_rocoto_realize(args: Namespace) -> bool:
     :param args: Parsed command-line args.
     """
     success = uwtools.rocoto.realize_rocoto_xml(
-        input_yaml=args.input_file, rendered_output=args.output_file
+        config_file=args.input_file, rendered_output=args.output_file
     )
     return success
 
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -15,8 +15,6 @@ from importlib import resources
 from pathlib import Path
 from typing import Callable, Dict, List, Tuple
 
-from lxml import etree
-
 import uwtools.config.atparse_to_jinja2
 import uwtools.config.core
 import uwtools.config.templater
@@ -338,12 +336,10 @@ def _add_subparser_rocoto_realize(subparsers: Subparsers) -> SubmodeChecks:
     _add_arg_input_file(required)
     _add_arg_output_file(required)
     optional = _basic_setup(parser)
-    _add_arg_input_format(optional, choices=[FORMAT.yaml])
-    _add_arg_output_format(optional, choices=[FORMAT.rocoto])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
-        partial(_check_file_vs_format, STR.infile, STR.infmt),
-        partial(_check_file_vs_format, STR.outfile, STR.outfmt),
+        partial(_check_file_vs_format, STR.infile, FORMAT.yaml),
+        partial(_check_file_vs_format, STR.outfile, FORMAT.rocoto),
     ]
 
 
@@ -357,10 +353,9 @@ def _add_subparser_rocoto_validate(subparsers: Subparsers) -> SubmodeChecks:
     required = parser.add_argument_group(TITLE_REQ_ARG)
     _add_arg_input_file(required)
     optional = _basic_setup(parser)
-    _add_arg_input_format(optional, choices=[FORMAT.rocoto])
     checks = _add_args_quiet_and_verbose(optional)
     return checks + [
-        partial(_check_file_vs_format, STR.infile, STR.infmt),
+        partial(_check_file_vs_format, STR.infile, FORMAT.rocoto),
     ]
 
 
@@ -384,27 +379,9 @@ def _dispatch_rocoto_realize(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
-    with resources.as_file(resources.files("uwtools.resources")) as resc:
-        input_template = resc / "rocoto.jinja2"
-        output_template = resc / "rocoto.jsonschema"
-
-    success = True
-    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
-        if uwtools.config.validator.validate_yaml(
-            config_file=args.input_file, schema_file=input_template
-        ):
-            uwtools.rocoto.write_rocoto_xml(
-                input_yaml=args.input_file,
-                input_template=str(input_template),
-                rendered_output=str(args.output_file),
-            )
-            success = uwtools.rocoto.validate_rocoto_xml(
-                input_xml=args.output_file, schema_file=str(output_template)
-            )
-        else:
-            success = False
-    else:
-        success = False
+    success = uwtools.rocoto.realize_rocoto_xml(
+        input_yaml=args.input_file, rendered_output=str(args.output_file)
+    )
     return success
 
 
@@ -415,23 +392,11 @@ def _dispatch_rocoto_validate(args: Namespace) -> bool:
     :param args: Parsed command-line args.
     """
     with resources.as_file(resources.files("uwtools.resources")) as resc:
-        schema_file = resc / "rocoto.jsonschema"
+        rocoto_schema = resc / "rocoto.jsonschema"
 
-    success = True
-    if args.input_format == FORMAT.rocoto:
-        success = uwtools.rocoto.validate_rocoto_xml(
-            input_xml=args.input_file, schema_file=str(schema_file)
-        )
-    else:
-        success = False
-    if args.verbose: #pragma: no cover
-        errors = str(etree.RelaxNG.error_log).split("\n")
-        log_method = logging.error if len(errors) else logging.info
-        log_method(
-            "%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s"
-        )
-        for line in errors:
-            logging.error(line)
+    success = uwtools.rocoto.validate_rocoto_xml(
+        input_xml=args.input_file, schema_file=str(rocoto_schema)
+    )
     return success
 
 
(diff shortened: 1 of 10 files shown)


src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+def _add_subparser_rocoto_write(subparsers: Subparsers) -> SubmodeChecks:
+    """
+    Subparser for mode: rocoto write
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.write, "Write a Rocoto XML workflow document")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    _add_arg_schema_file(required)
```
christinaholtNOAA, line 336: I don't think the user should have to provide this schema file. A rocoto YAML should always be validated against our  Rocoto JSONSchema file.

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    _add_arg_output_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.yaml])
+    _add_arg_output_format(optional, choices=[FORMAT.jinja2])
```
christinaholtNOAA, line 340: Input and output formats should always be YAML and XML, respectively.

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    """
+    Subparser for mode: rocoto validate
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.validate, "Validate Rocoto XML")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.jinja2])
```
christinaholtNOAA, line 358: The input here should always be XML.

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+                input_yaml=args.input_file,
+                input_template=str(args.schema_file),
+                rendered_output=str(args.output_file),
+            )
+        else:
+            success = False
+        valid_output = uwtools.rocoto.validate_rocoto_xml(
+            input_xml=args.output_file, schema_file=args.schema_file
+        )
+        success = False if not valid_output else success
```
christinaholtNOAA, line 401: Why not:

```
success = uwtools.rocoto.validate_rocoto_xml(
         ...
        )
```
?

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    Dispatch logic for rocoto write submode. Validate input and output.
+
+    :param args: Parsed command-line args.
+    """
+    success = True
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.jinja2:
+        valid_input = uwtools.config.validator.validate_yaml(
+            config_file=args.input_file, schema_file=args.schema_file
+        )
+        if valid_input:
```
christinaholtNOAA, line 390: This might clean up with a walrus operator:

```
if not valid_input := uwtools.config.validator.validate_yaml(
            config_file=args.input_file, schema_file=args.schema_file
        ):
    return valid_input

uwtools.rocoto.write_rocoto_xml(....)
```

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+            uwtools.rocoto.write_rocoto_xml(
+                input_yaml=args.input_file,
+                input_template=str(args.schema_file),
```
christinaholtNOAA, line 393: This template should always be the `resources/rocoto.jinja2` file.

src/uwtools/rocoto.py
```diff
@@ -42,3 +44,18 @@ def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str)
...
+
+    :param input_XML: Path to rendered XML file.
+    :param schema_file: Path to schema file.
+    """
+
+    # Validate the XML.
+    with open(schema_file, "r", encoding="utf-8") as f:
+        schema = etree.RelaxNG(etree.parse(f))
+    tree = etree.parse(input_xml)
+    return schema.validate(tree)
```
christinaholtNOAA, line 61: The AC for UW-322 asks for debug-level logging of the output of whatever lxml provides here.

src/uwtools/cli.py
```diff
@@ -305,6 +306,121 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    """
+    return {
+        STR.write: _dispatch_rocoto_write,
+        STR.validate: _dispatch_rocoto_validate,
+    }[
+        args.submode
+    ](args)
+
+
+def _dispatch_rocoto_write(args: Namespace) -> bool:
```
christinaholtNOAA, line 379: What if we considered calling this rocoto action something other than `write`? It's doing way more than that. The action may be consistent with the `realize` we use for the config mode.

christinaholtNOAA (changes requested):
I think it makes sense here that we are adding a `rocoto` mode to the CLI. The actions performed under the `rocoto` mode are going to be similar to those under the `config`, for example `validate`, `render`, or `realize`. We should keep our verbs consistent where we can.

I think that the value add of having the new submode means that we can strap down a couple of the arguments pretty tight because we really are only going to accept YAML input and get XML output. Any time the YAML is validated, it's against a very specific JSONSchema file and when the XML is validated, it's against a very specific RelaxNG schema. The user really doesn't need to be aware of any of that.

I am a little concerned, though, that the `_dispatch_rocoto_write` function is doing a lot more than its other cli counterparts. Maybe that is related to the difference in structure that we see with the rocoto module vs say the Forecast class. When the Forecast object is made, it automatically validates against its schema upon creation. 

I don't know that we want to introduce a Rocoto class, but perhaps we do want to at least move the bulk of the work from dispatch -- validate, write, validate -- to a method in the rocoto.py module to lighten the load of the cli, and to make that code more reusable beyond the cli.

src/uwtools/cli.py
WeirAE: Funny story, I had that originally and for some reason talked myself out of it in the last update - I'll fix it

src/uwtools/cli.py
WeirAE: At one point I had this as optional too, but that's fair

src/uwtools/rocoto.py
WeirAE: I had a feeling I was going to forget that one, thanks - hmm, to pass that I think this is going to need an interface, I'll think about it

src/uwtools/cli.py
WeirAE: Hmm, I think I could be convinced to go with realize, although I do agree with your top comment that its probably doing too much work for a cli function. Ah - here's an idea, I'll add a function to `rocoto.py` called `realize_rocoto_xml()` that does the inline validation and write, then call that over here as `_dispatch_rocoto_realize()`

src/uwtools/cli.py
WeirAE: My head is still in 3.7 I think, I always forget about that! I ended up removing the entire `valid_input` logic though as there was no need to use it

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.realize, "Realize a Rocoto XML workflow document")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    _add_arg_output_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.yaml])
+    _add_arg_output_format(optional, choices=[FORMAT.rocoto])
```
christinaholtNOAA, line 342: These are not actually optional. Input is always YAML, output is always XML.

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    """
+    Subparser for mode: rocoto validate
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.validate, "Validate Rocoto XML")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    optional = _basic_setup(parser)
+    _add_arg_input_format(optional, choices=[FORMAT.rocoto])
```
christinaholtNOAA, line 360: This will always need to be XML, so it isn't optional.

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+def _dispatch_rocoto_realize(args: Namespace) -> bool:
+    """
+    Dispatch logic for rocoto realize submode. Validate input and output.
+
+    :param args: Parsed command-line args.
+    """
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        input_template = resc / "rocoto.jinja2"
+        output_template = resc / "rocoto.jsonschema"
```
christinaholtNOAA, line 389: The variable `output_template` might be more descriptive as `rocoto_schema`.

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+    success = True
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
+        if uwtools.config.validator.validate_yaml(
+            config_file=args.input_file, schema_file=input_template
+        ):
```
christinaholtNOAA, line 395: Here, you should be able to define success in the if-statement with an assignment, and won't need to manage the value of `success` otherwise.

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    else:
+        success = False
+    if args.verbose: #pragma: no cover
+        errors = str(etree.RelaxNG.error_log).split("\n")
+        log_method = logging.error if len(errors) else logging.info
+        log_method(
+            "%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s"
+        )
+        for line in errors:
+            logging.error(line)
```
christinaholtNOAA, line 434: All of this should be logged directly from the `validate_rocoto_xml` method.

src/uwtools/cli.py
```diff
@@ -305,6 +309,132 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+    success = True
+    if args.input_format == FORMAT.yaml and args.output_format == FORMAT.rocoto:
+        if uwtools.config.validator.validate_yaml(
+            config_file=args.input_file, schema_file=input_template
+        ):
+            uwtools.rocoto.write_rocoto_xml(
+                input_yaml=args.input_file,
+                input_template=str(input_template),
+                rendered_output=str(args.output_file),
```
christinaholtNOAA, line 399: Why not add new functions with `Path` support?

src/uwtools/rocoto.py
```diff
@@ -26,6 +31,57 @@ def _add_jobname(tree: dict) -> None:
 
 
 # Public functions
+def realize_rocoto_xml(
+    input_yaml: str,
+    input_template: str,
+    rendered_output: str,
+    schema_file: str,
```
christinaholtNOAA, line 38: Why not hard-code the schema file in here so any user of this code gets that benefit. Then it doesn't need to be tracked in both places?

The same is applicable for the Rocoto JSONSchema.

src/uwtools/rocoto.py
```diff
@@ -26,6 +31,57 @@ def _add_jobname(tree: dict) -> None:
...
+) -> None:  # pragma: no cover
```
christinaholtNOAA, line 39: The other "realize" function return a bool. Perhaps we should keep this one consistent with those?

christinaholtNOAA (commented):
This is shaping up nicely. I left some additional suggestions below.

src/uwtools/cli.py
WeirAE: Hmm, there are "optional-but-one-choice" settings in other CLI functions, I should probably make them fully fixed as well

src/uwtools/cli.py
WeirAE: Yeah, I need this too - and the outer if is also unnecessary since there's no choice

src/uwtools/rocoto.py
WeirAE: Ah yes, thanks!

src/uwtools/rocoto.py
WeirAE: I debated this, whether to leave the function universal or not, but I think you're right in cementing it here

src/uwtools/cli.py
WeirAE: This was another debate point for me, whether to handle the errors here, offload to another method in `cli()`, or to expand `rocoto()` to accept args - expanding `rocoto()` is probably the way to go in the end

src/uwtools/cli.py
WeirAE: Hm, on re-look, this looks like it is just setting up `_check_file_vs_format` to ensure the input file is the correct format. I could

src/uwtools/cli.py
WeirAE: correction: not args, still thinking in old terms, got it!

src/uwtools/cli.py
```diff
@@ -305,6 +307,99 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    """
+    parser = _add_subparser(subparsers, STR.realize, "Realize a Rocoto XML workflow document")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    _add_arg_output_file(required)
+    optional = _basic_setup(parser)
+    checks = _add_args_quiet_and_verbose(optional)
+    return checks + [
+        partial(_check_file_vs_format, STR.infile, FORMAT.yaml),
+        partial(_check_file_vs_format, STR.outfile, FORMAT.rocoto),
```
christinaholtNOAA, line 342: I'm not sure either of these checks is necessary. This check seems to only make sure that when the user doesn't specify a file format, that it can be derived from the file name.

src/uwtools/cli.py
```diff
@@ -305,6 +307,99 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.validate, "Validate Rocoto XML")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_input_file(required)
+    optional = _basic_setup(parser)
+    checks = _add_args_quiet_and_verbose(optional)
+    return checks + [
+        partial(_check_file_vs_format, STR.infile, FORMAT.rocoto),
```
christinaholtNOAA, line 358: Same here.

src/uwtools/cli.py
```diff
@@ -727,6 +824,7 @@ class _STR:
     valsfmt: str = "values_format"
     valsneeded: str = "values_needed"
     verbose: str = "verbose"
+    write: str = "write"
```
christinaholtNOAA, line 827: This isn't needed since `validate` was adopted.

src/uwtools/rocoto.py
```diff
@@ -26,6 +48,67 @@ def _add_jobname(tree: dict) -> None:
 
 
 # Public functions
+def realize_rocoto_xml(
+    input_yaml: str,
+    rendered_output: str,
+) -> bool:  # pragma: no cover
+    """
+    Main entry point.
```
christinaholtNOAA, line 56: It seems that  many of the functions in this module are now documented as "Main entry point". Would you mind making those more descriptive now that they've multiplied?

src/uwtools/rocoto.py
```diff
@@ -26,6 +48,67 @@ def _add_jobname(tree: dict) -> None:
...
+
+    :param input_yaml: Path to YAML input file.
+    :param input_template: Path to input template file.
+    :param rendered_output: Path to write rendered XML file.
+    :param schema_file: Path to schema file.
+    """
+    values = YAMLConfig(input_yaml)
+    tasks = values["tasks"]
+    if isinstance(tasks, dict):
+        _add_jobname(tasks)
```
christinaholtNOAA, line 66: This chunk of work is repeated in the `write_rocoto_xml`. Maybe it could be pulled out into its own private function?

src/uwtools/rocoto.py
```diff
@@ -26,6 +48,67 @@ def _add_jobname(tree: dict) -> None:
...
+        schema = etree.RelaxNG(etree.parse(f))
+    tree = etree.parse(input_xml)
+    success = schema.validate(tree)
+
+    # Store validation errors in the main log.
+    errors = str(etree.RelaxNG.error_log).split("\n")
+    log_method = logging.debug if len(errors) else logging.info
+    log_method("%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s")
+    for line in errors:
+        logging.debug(line)
```
christinaholtNOAA, line 107: The logging of these errors should be tied to the `--verbose` flag.

WeirAE: So I left it as implicit-verbose with `logging.debug()` since we don't actually directly call args outside of the cli anymore, but there should be a way to pass it down, I'll think about it

src/uwtools/rocoto.py
WeirAE: ah good point, thanks!

src/uwtools/rocoto.py
```diff
@@ -2,12 +2,34 @@
...
+from lxml import etree
+
+import uwtools.config.validator
 from uwtools.config.core import YAMLConfig
 from uwtools.config.j2template import J2Template
 
 # Private functions
 
 
+def _rocoto_template() -> str: #pragma: no cover
```
NaureenBharwaniNOAA, line 17: Order alphabetically for the private functions section.
Place `def _add_jobname`  above `def _rocoto_schema` and `def _rocoto_template`.

src/uwtools/tests/test_cli.py
```diff
@@ -289,6 +289,57 @@ def test__dispatch_forecast_run():
...
+    [
+        (STR.realize, "_dispatch_rocoto_realize"),
+        (STR.validate, "_dispatch_rocoto_validate"),
+    ],
+)
+def test__dispatch_rocoto(params):
+    submode, funcname = params
+    args = ns()
+    vars(args).update({STR.submode: submode})
+    with patch.object(cli, funcname) as m:
```
NaureenBharwaniNOAA, line 303: What does `m` stand for here? Placeholder? Not sure if I know the background

NaureenBharwaniNOAA (commented):
Just a couple stylistic comments/questions for ordering of functions and naming conventions.

src/uwtools/cli.py
```diff
@@ -305,6 +307,99 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    Dispatch logic for rocoto validate submode.
+
+    :param args: Parsed command-line args.
+    """
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        rocoto_schema = resc / "rocoto.jsonschema"
+
+    success = uwtools.rocoto.validate_rocoto_xml(
+        input_xml=args.input_file, schema_file=str(rocoto_schema)
+    )
```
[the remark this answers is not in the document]
WeirAE: haha we just flipped this one earlier - I'm torn, personally

src/uwtools/rocoto.py
```diff
@@ -26,6 +48,67 @@ def _add_jobname(tree: dict) -> None:
 
 
 # Public functions
+def realize_rocoto_xml(
+    input_yaml: str,
```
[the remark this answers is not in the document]
WeirAE: I'm trying to have it both ways - leaving the `write()` function to accept an external config and any input, but making `realize()` the interface that only uses our Rocoto schemas. Honestly it might be a mistake to try for it and just fix everything to Rocoto-only

src/uwtools/tests/test_cli.py
WeirAE: yeah these `m` calls are patching in placeholder modules - but our book would probably tell us to stop trying to save bytes and just call them `module` haha. I'll see if I can track all of these down and change them

src/uwtools/rocoto.py
WeirAE: Thanks! I had put these in as `@properties` to put first, but of course that doesn't work outside of a class so back to alphabetical

src/uwtools/rocoto.py
```diff
@@ -26,6 +31,57 @@ def _add_jobname(tree: dict) -> None:
...
+    input_template: str,
+    rendered_output: str,
```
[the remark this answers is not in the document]
WeirAE: Annoyingly `J2Template()` takes `str` but I can put the `str()` call there and have more options elsewhere without having to be restricted

src/uwtools/rocoto.py
```diff
@@ -25,20 +32,113 @@ def _add_jobname(tree: dict) -> None:
...
+def validate_rocoto_xml(input_xml: OptionalPath = None, schema_file: OptionalPath = None) -> bool:
+    """
+    Given a rendered XML file, validate it against the Rocoto schema.
+
+    :param input_XML: Path to rendered XML file.
+    :param schema_file: Path to schema file.
+    """
+
+    # Validate the XML.
+    with open(str(schema_file), "r", encoding="utf-8") as f:
```
WeirAE, line 112: I don't like using `str()` for `with open()` here but I had some issues with `with writable()` that I'm working on

src/uwtools/tests/test_cli.py
WeirAE: Looks like they were just here, so quick fix!

src/uwtools/rocoto.py
WeirAE: Fixed, just had a weird combination of options before that weren't working

src/uwtools/cli.py
WeirAE: Thank you! Moved everything to OptionalPath and so far only needed to hard-code string once

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+    """
+    Given a rendered XML file, validate it against the Rocoto schema.
+
+    :param input_XML: Path to rendered XML file.
+    :param schema_file: Path to schema file.
+    """
+
+    # Validate the XML.
+    with readable(schema_file) as f:
+        schema = etree.RelaxNG(etree.parse(f))
```
maddenp-noaa, line 114: I'm not sure that `readable()` is appropriate here: It is if we want to give users an option to pipe in a schema via `stdin`, but I think from other comments/discussion we were thinking that *our* schema, the one in `resources/` is the one to use, and that it might not make sense to give users the ability to override that -- which makes sense to me. If that's what we collectively think, than should we have both of these functions that the CLI is dispatching to simply obtain the correct file via `resources/` and not present the user with a choice they shouldn't be making? Apologies if I missed something in discussion on this point, though...

src/uwtools/rocoto.py
maddenp-noaa: If `--verbose` is specified on the command line, then the logger will accept `logging,.debug()` messages and emit them to `stderr`; if not, it will not. So, I don't think there's a need to pass `args.verbose` down: I think it's only needed to configure the logger, which then handles would-be conditional behavior (to log or not to log) based on the level set.

But do we really want to force the user to say `--verbose` to see the validation errors? If they don't say `--verbose` and see `88 Rocoto validation errors found`, is that really useful without the extra info? If the behavior of our YAML validation against JSON Schema to show the errors without the `--verbose` flag? I'm not sure, but I thought so... If so, should we be consistent here?

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+    return success
+
+
+def write_rocoto_xml(
+    input_yaml: OptionalPath = None,
+    input_template: OptionalPath = None,
+    rendered_output: OptionalPath = None,
+) -> None:
+    """
+    Main entry point. Render the given YAML file to XML using the given template.
```
maddenp-noaa, line 134: What do we mean by `Main entry point.` here? Seems maybe unnecessary.

maddenp-noaa (changes requested):
I still want to review the unit tests, but thought I would share comments on the main code first to potentially save time.

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+            return True
+        logging.error("Rocoto validation errors identified in %s", rendered_output)
+        return False
+    logging.error("YAML validation errors identified in %s", input_yaml)
+    return False
+
+
+def _rocoto_template() -> str:
+    """
+    The path to the file containing the template to validate the config file against.
```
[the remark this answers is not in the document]
WeirAE: I'm using this like a `@property` which may not be the way to go, so suggestions are appreciated!

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+            rendered_output=rendered_output,
+        )
+        # Validate the XML.
+        if validate_rocoto_xml(input_xml=rendered_output, schema_file=rocoto_template):
+            # If no issues were detected, report success.
+            return True
+        logging.error("Rocoto validation errors identified in %s", rendered_output)
+        return False
+    logging.error("YAML validation errors identified in %s", input_yaml)
+    return False
```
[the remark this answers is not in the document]
WeirAE: Mmm, good point, thats a better way to handle it

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+    rendered_output: OptionalPath = None,
+) -> bool:  # pragma: no cover
+    """
+    Realize the given YAML file to XML, using fixed Rocoto RelaxNG schema templates. Validate both
+    the YAML and the XML. External functions should call this function.
+
+    :param input_yaml: Path to YAML input file.
+    :param input_template: Path to input template file.
+    :param rendered_output: Path to write rendered XML file.
+    :param schema_file: Path to schema file.
```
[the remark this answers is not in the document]
WeirAE: Oops, left them in, thanks!

src/uwtools/cli.py
```diff
@@ -305,6 +307,94 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+    Dispatch logic for rocoto validate submode.
+
+    :param args: Parsed command-line args.
+    """
+    with resources.as_file(resources.files("uwtools.resources")) as resc:
+        rocoto_schema = resc / "rocoto.jsonschema"
+
+    success = uwtools.rocoto.validate_rocoto_xml(
+        input_xml=args.input_file, schema_file=rocoto_schema
+    )
```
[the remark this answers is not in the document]
WeirAE: I was allowing it to stay open so a user could use the tool to validate any schema, but maybe we should just tuck the whole validation away, why would we need a separate validation when we're ever only double-checking our own output?

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
     tasks = values["tasks"]
     if isinstance(tasks, dict):
         _add_jobname(tasks)
 
+
+# Public functions
+def realize_rocoto_xml(
+    input_yaml: OptionalPath = None,
+    rendered_output: OptionalPath = None,
+) -> bool:  # pragma: no cover
```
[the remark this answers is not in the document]
WeirAE: ah, I knew I had left one behind! Sorry, thanks!

src/uwtools/cli.py
```diff
@@ -305,6 +307,94 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
     )
 
 
+# Mode Rocoto
```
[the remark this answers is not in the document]
WeirAE: Pick away! Nit-less code is better code :)

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: Ah ok, yes!

src/uwtools/rocoto.py
maddenp-noaa: Yes, I was confused about what we were validating here -- this is the XML output, not the YAML input. But I think the crucial question remains: Do we want to present users with an option to override the official RelaxNG schema that we took from the Rocoto repo? I can't immediately see why, so I'm on team YAGNI on this one.

christinaholtNOAA (changes requested):

src/uwtools/rocoto.py
maddenp-noaa: @christinaholtNOAA In that 2nd paragraph, are you saying that's what we _have_ or that's what we _want_?

My $.02 is that it is what we _have_ but *not* what we _want_. I think we'd want to always report the errors to the user. By way of comparison, in `uwtools.config.validator.validate_yaml()`, when we're validating the UW YAML with our `.jsonschema` schema, we call `logging.error()` when validation errors are encountered -- not `logging.debug()`.

WeirAE: That's fair - do we even need a --verbose flag then?

src/uwtools/rocoto.py
```diff
@@ -25,20 +33,113 @@ def _add_jobname(tree: dict) -> None:
...
+    """
+    Realize the given YAML file to XML, using fixed Rocoto RelaxNG schema templates. Validate both
+    the YAML and the XML. External functions should call this function.
```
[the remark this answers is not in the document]
WeirAE: Honestly, I was probably thinking 'declare this for users' but thats a weak design that should make functions public or private. I'm going to drop that, and the talk about "fixed templates", the schema is the schema, I don't need to overstate it!

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: Still working on this one

src/uwtools/cli.py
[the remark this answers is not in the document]
WeirAE: Yeah, after all of this discussion, I've removed it as an option at all, which also removes it from the CLI

src/uwtools/rocoto.py
WeirAE: Yep, removing all the schema options!

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: Handling it with `tempfile.NamedTemporaryFile()` and `shutil.move()`, much nicer now

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: Yeah, I think you're right, I've just changed all of the `input_yaml` calls to `config_file`

src/uwtools/cli.py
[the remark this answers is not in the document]
WeirAE: Now that I've taken it away as an option at all, I think I'd like to leave it as `_rocoto_schema()` for the property only

src/uwtools/rocoto.py
maddenp-noaa: I'd vote for keeping `--verbose` 1. for consistency, as it's currently present in all modes (I think), and 2. because at any point someone might want to add some `logging.debug()` messages, in which case it would be convenient for the plumbing to be in place.

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,110 @@ def _add_jobname(tree: dict) -> None:
...
+    with resources.as_file(resources.files("uwtools.resources")) as path:
+        return (path / "rocoto.jsonschema").as_posix()
+
+
+def _rocoto_template() -> str:
+    """
+    The path to the file containing the Rocoto workflow document template to render.
+    """
+    with resources.as_file(resources.files("uwtools.resources")) as path:
+        # return (path / "rocoto.jinja2").as_posix()
```
maddenp-noaa, line 65: Commented-out code can be removed.

src/uwtools/tests/test_rocoto.py
```diff
@@ -44,26 +45,65 @@ def test_add_jobname():
...
-    support.compare_files(expected, output)
+    with patch.object(rocoto, "validate_rocoto_xml", value=True):
+        with patch.object(rocoto.uwtools.config.validator, "_bad_paths", return_value=None):
+            with resources.as_file(resources.files("uwtools.tests.fixtures")) as path:
+                input_yaml = (path / fn).as_posix()
+                result = rocoto.realize_rocoto_xml(input_yaml=input_yaml, rendered_output=output)
+    assert result is validity
+
+
+def test_realize_rocoto_invalid_xml():
```
WeirAE, line 84: It looks like the last outstanding issue is why this test is not successfully covering `rocoto.py` [lines 98-99](https://github.com/WeirAE/workflow-tools/blob/aa912672d1b16dabc4ca3e8d0fc439a1c31912db/src/uwtools/rocoto.py#L98)

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,110 @@ def _add_jobname(tree: dict) -> None:
...
 
 
-# Public functions
-def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str) -> None:
+def _add_tasks(
+    input_yaml: OptionalPath = None,
+) -> None:
     """
-    Main entry point.
+    Define "task" elements in the given config tree and request a "jobname" for each.
```
maddenp-noaa, line 42: I'm not sure about this docstring:
1. I wouldn't think to call a path to a YAML file a "config tree".
2. I don't see where "task" elements are being defined -- they're just there or not there, right?
Maybe something like
> Load YAML config and add job names to each defined workflow task.

?

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,110 @@ def _add_jobname(tree: dict) -> None:
...
+            write_rocoto_xml(
+                config_file=config_file,
+                rendered_output=temp_file.name,
+            )
+            # Validate the XML.
+            if validate_rocoto_xml(input_xml=temp_file.name):
+                # If no issues were detected, save temp file and report success.
+                shutil.move(temp_file.name, str(rendered_output))
+                return True
+        logging.error("Rocoto validation errors identified in %s", temp_file.name)
```
maddenp-noaa, line 98: Nice.

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,110 @@ def _add_jobname(tree: dict) -> None:
...
+        return False
+    logging.error("YAML validation errors identified in %s", config_file)
+    return False
+
+
+def validate_rocoto_xml(input_xml: OptionalPath = None) -> bool:
+    """
+    Given a rendered XML file, validate it against the Rocoto schema.
+
+    :param input_XML: Path to rendered XML file.
```
maddenp-noaa, line 108: `input_XML` -> `input_xml`

src/uwtools/rocoto.py
maddenp-noaa: I'm getting confused about this one -- can we discuss briefly post-standup today?

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,110 @@ def _add_jobname(tree: dict) -> None:
...
+    errors = str(etree.RelaxNG.error_log).split("\n")
+    log_method = logging.error if len(errors) else logging.info
+    log_method("%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s")
+    for line in errors:
+        logging.error(line)
+
+    return success
+
+
+def write_rocoto_xml(
```
maddenp-noaa, line 128: Do we call this function from _outside_ this module, apart from in tests? I don't think so... If not, it should be marked private.

maddenp-noaa (changes requested):
A few more requests / detail for discussion before reviewing the tests...

src/uwtools/rocoto.py
WeirAE: Ah, I missed one, thanks

src/uwtools/rocoto.py
WeirAE: Yeah, when I moved this out, I did not go back and review the docstring here. Something like that is better, I'll fix it

src/uwtools/rocoto.py
WeirAE: Kind of an extension of our earlier discussions about docstrings that said "main entry" or "functions access this", etc. I should bite the bullet and put everything not called `realize()` private

maddenp-noaa: I think we have two public functions, which need to be callable from the CLI: `realize_rocoto_xml()` and `validate_rocoto_xml()`. True?

src/uwtools/rocoto.py
maddenp-noaa: I'll summarize my misgivings (misunderstandings?) here as a prelude to our upcoming meeting:

1. In other modes, we permit provision of the "primary" input on `stdin` as an alternative to providing an on-disk file. For example, with `uw template render`, the Jinja2 template to be rendered can be specified via the _optional_ `--input-file` argument, or it can be piped in on `stdin`. The latter means someone could do something like `curl -s https://foo.com/path/to/template.jinja2 | uw template render` and render a template pulled from the internet. Why? Who knows, but it's the "Unix way" and can be handy in lots of circumstances. It could also be `cat`ed or produced however, and piped in. My (possibly wrong) understanding of `uw rocoto validate` is that we have a would-be Rocoto XML document that we want to validate, so to me this would be the "primary" input of this mode, so it ought to -- for consistency with other modes -- be an optional argument, and taken from `stdin` if not specified, which is what `readable()` provides: Either a readable handle to an opened file, if a filename was provided, or a readable handle to `stdin` if not. So I kind of expected to see `--input-file` optional, and its value handled by `readable()` for consistency with other modes.
2. I think `_rocoto_template()` is probably misnamed: The file it returns a path to is not a template, it's a schema. Of course, the name `_rocoto_schema()` is already taken, so might we have something like `_rocoto_schema()` -> `_rocoto_schema_yaml()` and `_rocoto_template()` -> `_rocoto_schema_xml()`? Is that what the files at the two returned paths are, namely the schema for a Rocoto workflow document in YAML form, and a Rocoto workflow document in XML form?
3. Finally, I wouldn't expect to see `readable()` used to get a handle to the Rocoto XML schema file. I mean, it _works_, but the use of that function -- if you know what that function does -- gives the appearance that that file might be provided on `stdin`, which is not the case with the schema file. Per point 1 above, I'd hope to see the XML document optionally available on `stdin`, so would expect to see `readable()` used to get a readable handle to it; but not to the schema file. The schema file is simply an on-disk file, so I'd just expect to see a `with open(...) ...` block on that one. Again, `readable()` works because we'd never _not_ provide a real filename, but I think `readable()` sends the wrong message. I _think_ that if you look for the existing `with readable()` statements in the codebase, you'll find that they're all tied to inputs that might be unspecified, and where `stdin` will be used instead; while there are other `with open()` calls for cases where the input will always be a file. See e.g. `uwtools.config.validator.validate_yaml()`, where `with open()` is used to read in the schema file. I think adhering to this convention will remind us which files may be streams, and which must always be files.

src/uwtools/tests/test_rocoto.py
WeirAE: Fixed it - had to fix my `with patch.object()` usage

src/uwtools/rocoto.py
WeirAE: Looking through, I may have to adjust functions to `with readable()` all the way up to `config.YAMLConfig()` but we'll definitely talk about it as a group

WeirAE: Feedback from the offline discussion:
1) We will be moving forward with converting to streams, but that will move to another PR to cover a broader scope
2&3) Agreed and added, as above
also, reiterating the replacement of `str()` with `Path()` which is now a requirement of this PR and is in progress
Thanks to @maddenp-noaa, @christinaholtNOAA and @NaureenBharwaniNOAA for their time today!

WeirAE: Hmm, `j2template()` and `config()` are fixed, I think the only remnant `str` paths may be in `forecast()` now and some redundant stragglers in tests

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,111 @@ def _add_jobname(tree: dict) -> None:
...
+
+
+# Public functions
+def realize_rocoto_xml(
+    config_file: OptionalPath,
+    rendered_output: DefinitePath,
+) -> bool:
+    """
+    Realize the given YAML file to XML, using the Rocoto RelaxNG schema. Validate both the YAML and
+    the XML.
```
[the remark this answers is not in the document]
WeirAE: This is probably a first draft error left from me mixing my schemas and templates again. I like the points you've been making to stick to generic facts, I know the book would be quite miffed at me over-specifying my docs again

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,111 @@ def _add_jobname(tree: dict) -> None:
...
+    with resources.as_file(resources.files("uwtools.resources")) as path:
+        return path / "schema_with_metatasks.rng"
+
+
+def _write_rocoto_xml(
+    config_file: OptionalPath,
+    rendered_output: DefinitePath,
+) -> None:
+    """
+    Render the given YAML file to XML using the given template.
```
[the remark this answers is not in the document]
WeirAE: Oh, you're right! That's left over from when it was taking a template - just more proof that the book was right about comment/docstring problems

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,111 @@ def _add_jobname(tree: dict) -> None:
...
+    return False
+
+
+def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
+    """
+    Given a rendered XML file, validate it against the Rocoto schema.
+
+    :param input_xml: Path to rendered XML file.
+    """
+    rocoto_template = _rocoto_schema_xml()
```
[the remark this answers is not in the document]
WeirAE: I'll move them inline - honestly, its separated here because I was having an issue where it was not returning the value but trying to call the function and erroring, but that may have been a separate implementation error of mine

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,111 @@ def _add_jobname(tree: dict) -> None:
...
+    rendered_output: DefinitePath,
+) -> None:
+    """
+    Render the given YAML file to XML using the given template.
+
+    :param config_file: Path to YAML input file.
+    :param rendered_output: Path to write rendered XML file.
+    """
+
+    rocoto_template = _rocoto_schema_xml()
```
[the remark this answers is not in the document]
WeirAE: Errors on `rocoto.jinja2` are:
`lxml.etree.XMLSyntaxError: Start tag expected, '<' not found`
`jinja2.exceptions.UndefinedError: 'entities' is undefined`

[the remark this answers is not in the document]
maddenp-noaa: What's the context in which you're seeing those errors? I mean, what are you doing/running when you see those errors?

The `lxml` error looks something like what I'd expect to see if `lxml` was given the RelaxNG schema and asked to validate `rocoto.jinja`, as that file is not well-formed XML and has Jinja2 syntax where `lxml` would expect XML syntax.

The `jinja2` error looks something like what I'd expect to see if an attempt was made to _render_ `rocoto.jinja2` without supplying the necessary data.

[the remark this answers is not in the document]
WeirAE: The first is from `test_rocoto_xml_is_valid` and the second is from `test__write_rocoto_xml` , `test_realize_rocoto_invalid_xml` , and `test_realize_rocoto_xml`. 
Strictly, that write test should be the same as it is currently if it is given rocoto.jinja2, let me look more>
For the validations, its validating `hello_workflow.xml` against the schema. `hello_workflow.xml` is consistent with `hello_workflow.yaml`, so there's something funny in the validation then in reading the schema?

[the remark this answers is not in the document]
WeirAE: Ah. `schema_with_metatasks.rng` is the schema while `rocoto.jinja2` is the template, so I should use all 3. Not sure why 'write' is not working now with that error though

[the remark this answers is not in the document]
maddenp-noaa: If you find yourself blocked, let me know (and push all your changed to your fork) and I'll see if I can replicate the failures and suggest anything.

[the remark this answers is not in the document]
WeirAE: Thanks - I've pushed the update with schema and template and errors. It looks the same so it should work, but I may have been looking at it too long to see it now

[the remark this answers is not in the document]
maddenp-noaa: `rocoto.jinja2` contains the reference `entities.items()`, meaning that the "values" dict passed in to the `J2Template` constructor needs to have a top-level `entities` key. But, IIRC you added a top-level `workflow` key to the test-fixture file `hello_workflow.yaml`, so you might try changing the expression `YAMLConfig(config_file).data` to `YAMLConfig(config_file).data["workflow"]` in `_write_rocoto_xml()`, so that the same `dict` as before is being passed in.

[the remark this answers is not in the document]
maddenp-noaa: I think there's also an error in the updated tests, because when I run `make unittest`, the file `src/uwtools/tests/fixtures/rocoto_invalid.xml` is _changed by the tests_. Test-fixture files should not be modified by tests.

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,117 @@ def _add_jobname(tree: dict) -> None:
             _add_jobname(subtree)
 
 
-# Public functions
-def write_rocoto_xml(input_yaml: str, input_template: str, rendered_output: str) -> None:
+def _add_tasks(
```
christinaholtNOAA, line 38: This name is misleading. It doesn't add tasks to the YAML. Since this calls `_add_jobname` for all tasks, perhaps it could be `_add_jobnames` or `_add_jobname_to_tasks`?

Also, the function isn't returning the updated values, so I'm not sure these jobnames are going to be present when you write the xml.

christinaholtNOAA (commented):
I haven't finished the review between meetings, but wanted to go ahead and leave the item below.

src/uwtools/rocoto.py
WeirAE: Ah, thank you!

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: whew, got that fixed, not too bad! Now I'm having issues with patching through the XML file, I've updated with my changes

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,118 @@ def _add_jobname(tree: dict) -> None:
...
+# Public functions
+def realize_rocoto_xml(
+    config_file: OptionalPath,
+    rendered_output: DefinitePath,
+) -> bool:
+    """
+    Realize the Rocoto workflow defined in the given YAML as XML. Validate both the YAML input and
+    XML output.
+
+    :param input_yaml: Path to YAML input file.
```
christinaholtNOAA, line 104: The argument in the function signature is `config_file`.

src/uwtools/tests/test_cli.py
```diff
@@ -281,22 +281,73 @@ def test__dispatch_forecast_run():
...
+    with patch.object(cli.uwtools.rocoto, "realize_rocoto_xml", return_value=False):
+        assert cli._dispatch_rocoto_realize(args) is False
+
+
+def test__dispatch_rocoto_validate_xml():
+    args = ns()
+    vars(args).update({STR.infile: 1, STR.infmt: FORMAT.rocoto, STR.verbose: False})
+    with patch.object(cli.uwtools.rocoto, "validate_rocoto_xml") as module:
+        cli._dispatch_rocoto_validate(args)
+    assert module.called_once_with(args)
```
christinaholtNOAA, line 333: This seems to be called like this:

```
uwtools.rocoto.validate_rocoto_xml(input_xml=args.input_file)
```

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: Yikes, had to learn a lot about mocking and context managers! However, it works now!

src/uwtools/utils/file.py
```diff
@@ -39,7 +40,9 @@ class _FORMAT:
     ini: str = _ini
     jinja2: str = _jinja2
     nml: str = _nml
+    rocoto: str = _xml
     sh: str = _ini
+    xml: str = _xml
```
christinaholtNOAA, line 45: I may have missed it, but I don't think `FORMAT.xml` was ever used and `FORMAT.rocoto` was used arbitrarily, so may not be needed here, either.

src/uwtools/tests/test_cli.py
WeirAE: Caught by the cli letting me override too much again, thanks

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,118 @@ def _add_jobname(tree: dict) -> None:
...
+    rendered_output: DefinitePath,
+) -> None:
+    """
+    Render the Rocoto workflow defined in the given YAML to XML.
+
+    :param config_file: Path to YAML input file.
+    :param rendered_output: Path to write rendered XML file.
+    """
+
+    values = _add_jobname_to_tasks(config_file)
```
maddenp-noaa, line 88: I'm not sure on this one, but... The only call to `_write_rocoto_xml()` I see is via `_realize_rocoto_xml()`, which itself calls `_add_jobname_to_tasks()`. So, presumably the workflow YAML has already been updated with job names by the time we get here. Do we need this call? It probably does no harm, but it seems potentially redundant, and so possibly confusing.

src/uwtools/cli.py
```diff
@@ -305,6 +306,89 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
...
+
+def _add_subparser_rocoto_realize(subparsers: Subparsers) -> SubmodeChecks:
+    """
+    Subparser for mode: rocoto realize
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.realize, "Realize a Rocoto XML workflow document")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_output_file(required)
```
maddenp-noaa, line 335: The flip side of read-input-file vs read-from-stdin option would be a write-output-file vs write-to-stdout option. Ideally we'd want to be able to e.g.
```
cat some.yaml | uw rocoto realize >some.xml
```
making use of the stdin/stdout streams. That's probably beyond the scope of this PR, but would you mind adding a backlog task for it?

src/uwtools/rocoto.py
maddenp-noaa: This one looks good to me now -- please feel free to resolve the conversation if you have no misgivings.

src/uwtools/tests/test_cli.py
```diff
@@ -192,17 +192,17 @@ def test__dispatch_config(params):
     submode, funcname = params
     args = ns()
     vars(args).update({STR.submode: submode})
-    with patch.object(cli, funcname) as m:
+    with patch.object(cli, funcname) as module:
```
maddenp-noaa, line 195: `module` is probably not a good name here, as the object being patched is not a module, but rather a function. I think I meant the `m` to mean `mock`, but obviously that was also a bad name since it led to this confusion!

`func` might be a good name here, as that's an idiomatic Python name for a variable that refers to a function object.

The same comment applies similarly elsewhere in this file, though `func` will not be appropriate in all cases.

src/uwtools/tests/test_cli.py
```diff
@@ -235,12 +235,12 @@ def test__dispatch_config_translate_arparse_to_jinja2():
...
     )
-    with patch.object(cli.uwtools.config.atparse_to_jinja2, "convert") as m:
+    with patch.object(cli.uwtools.config.atparse_to_jinja2, "convert") as module:
         cli._dispatch_config_translate(args)
-    assert m.called_once_with(args)
+    assert module.called_once_with(args)
 
 
-def test_dispath_config_translate_unsupported():
+def test__dispatch_config_translate_unsupported():
```
maddenp-noaa, line 243: Nice catch, thanks.

src/uwtools/tests/test_cli.py
```diff
@@ -192,17 +192,17 @@ def test__dispatch_config(params):
...
         cli._dispatch_config(args)
-    assert m.called_once_with(args)
+    assert module.called_once_with(args)
 
 
 def test__dispatch_config_compare():
     args = ns()
     vars(args).update({STR.file1path: 1, STR.file1fmt: 2, STR.file2path: 3, STR.file2fmt: 4})
-    with patch.object(cli.uwtools.config.core, "compare_configs") as m:
+    with patch.object(cli.uwtools.config.core, "compare_configs") as module:
```
maddenp-noaa, line 203: Typically, where the name of the object being mocked is known, we've just used that name as the name of the mock, so I'd suggest `... as compare_configs` in this case.

src/uwtools/tests/test_cli.py
```diff
@@ -219,9 +219,9 @@ def test__dispatch_config_realize():
             STR.dryrun: 8,
         }
     )
-    with patch.object(cli.uwtools.config.core, "realize_config") as m:
+    with patch.object(cli.uwtools.config.core, "realize_config") as module:
```
maddenp-noaa, line 222: Suggestion: `... as realize_config`

src/uwtools/tests/test_cli.py
```diff
@@ -235,12 +235,12 @@ def test__dispatch_config_translate_arparse_to_jinja2():
             STR.dryrun: 5,
         }
     )
-    with patch.object(cli.uwtools.config.atparse_to_jinja2, "convert") as m:
+    with patch.object(cli.uwtools.config.atparse_to_jinja2, "convert") as module:
```
maddenp-noaa, line 238: `... as convert`

src/uwtools/tests/test_cli.py
```diff
@@ -251,12 +251,12 @@ def test_dispath_config_translate_unsupported():
 def test__dispatch_config_validate_yaml():
     args = ns()
     vars(args).update({STR.infile: 1, STR.infmt: FORMAT.yaml, STR.schemafile: 3})
-    with patch.object(cli.uwtools.config.validator, "validate_yaml") as m:
+    with patch.object(cli.uwtools.config.validator, "validate_yaml") as module:
```
maddenp-noaa, line 254: `... as validate_yaml`

src/uwtools/tests/test_cli.py
```diff
@@ -281,22 +281,73 @@ def test__dispatch_forecast_run():
         forecast_model="foo",
     )
     vars(args).update({STR.cfgfile: 1, "forecast_model": "foo"})
-    with patch.object(cli.uwtools.drivers.forecast, "FooForecast", create=True) as m:
+    with patch.object(cli.uwtools.drivers.forecast, "FooForecast", create=True) as module:
```
maddenp-noaa, line 284: You get the idea.

src/uwtools/tests/test_cli.py
```diff
@@ -281,22 +281,73 @@ def test__dispatch_forecast_run():
...
+    args = ns()
+    vars(args).update({STR.submode: submode})
+    with patch.object(cli, funcname) as module:
+        cli._dispatch_rocoto(args)
+    assert module.called_once_with(args)
+
+
+def test__dispatch_rocoto_realize():
+    args = ns()
+    vars(args).update({STR.infile: 1, STR.outfile: 2})
```
maddenp-noaa, line 310: Just a comment, NOT a request for a change in this PR: We should probably someday expand the tests to also _not_ provide optional args and demonstrate that `argparse` doesn't complain and that, e.g. `args.input_file` is `None`. Please don't bother about this now, but I see now that the CLI tests I wrote previously could have been stronger.

src/uwtools/rocoto.py
```diff
@@ -25,20 +35,118 @@ def _add_jobname(tree: dict) -> None:
...
+        return False
+    logging.error("YAML validation errors identified in %s", config_file)
+    return False
+
+
+def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
+    """
+    Given a rendered XML file, validate it against the Rocoto schema.
+
+    :param input_xml: Path to rendered XML file.
```
maddenp-noaa, line 134: Would be nice to have a `:return:` docstring clause describing the meaning of the return value.

src/uwtools/tests/test_rocoto.py
```diff
@@ -44,26 +46,69 @@ def test_add_jobname():
...
+def test__write_rocoto_xml(tmp_path):
+    config_file = support.fixture_path("hello_workflow.yaml")
+    output = tmp_path / "rendered.xml"
+    config = YAMLConfig(config_file)
+
+    with patch.object(rocoto, "_add_jobname_to_tasks", return_value=config):
+        rocoto._write_rocoto_xml(config_file=config_file, rendered_output=output)
+
+    expected = support.fixture_path("hello_workflow.xml")
+    support.compare_files(expected, output)
```
maddenp-noaa, line 114: Do we maybe need to make this an `assert` statement to actually make it fail if the files do not compare? Can you force it to fail now by interfering with one of the files?

maddenp-noaa (changes requested):
Almost there, I think! Most of these are pretty trivial, and I have finally reviewed all the way through the unit tests.

src/uwtools/tests/test_rocoto.py
WeirAE: Assert is a stronger test regardless, but it sure fails when I add that! It looks like our example log path in `hello_workflow.yaml` was not getting written to the XML, but also its not drilling down to the third level - we can get the first task and metatask names, but the nested task is not pulling the name - I'm working on this one

src/uwtools/rocoto.py
WeirAE: Oh, true, we should only need it at the `_write()`, this one should be able to go

maddenp-noaa: It's possible that, when unit-testing this function directly, you might need to call `_add_jobname_to_tasks()` in the test if that's necessary for it to pass. Maybe it's not, though.

src/uwtools/tests/test_rocoto.py
WeirAE: Ha, just had to stop patching the jobname, thanks!

maddenp-noaa (approved):

WeirAE:
Tasks UW-412 and UW-413 added to the backlog to expand tests and update output streams, respectively

christinaholtNOAA (approved):

PR #320, opened 2023-10-20
switchable-logging

maddenp-noaa:
**Description**

As a prelude to API work, add a capability to replace the default `uwtools` logger (i.e. the Python root logger) with an arbitrary logger. This will allow API users to configure `uwtools` to use any logger they like, to obtain custom formatting, logging to a file, to suppress logging entirely, etc.

There are a small number of changes in this PR that are not duplicates. Most of the diffs consist of replacing `import logging` statements with `from uwtools.logging import log`, and `loging.info()` (also .`debug()`, `.error()`, etc.) with `log.info()` (etc.) calls. See inline PR comments.

**Type of change**

**How Has This Been Tested?**

`make test` in a dev shell + GitHub Actions.
**Checklist**

### DIFF
--- a/src/uwtools/apps/srw.py
+++ b/src/uwtools/apps/srw.py
@@ -2,11 +2,11 @@
 This file contains the specific drivers for a particular app, using the facade pattern base class.
 """
 
-import logging
 import shutil
 
 from uwtools.drivers.facade import Facade
 from uwtools.exceptions import UWError
+from uwtools.logging import log
 from uwtools.utils.file import FORMAT, get_file_type
 from uwtools.utils.processing import execute
 
@@ -47,7 +47,7 @@ class SRW210(Facade):
             shutil.copy2(config_file, "config.yaml")
         else:
             msg = f"Bad file type -- {file_type}. Cannot load configuration!"
-            logging.critical(msg)
+            log.critical(msg)
             raise ValueError(msg)
 
     def validate_config(self, config_file: str) -> None:
--- a/src/uwtools/logging.py
+++ b/src/uwtools/logging.py
@@ -23,7 +23,7 @@ class _Logger:
     """
 
     def __init__(self) -> None:
-        self.logger = logging.getLogger()  # default to Python base logger.
+        self.logger = logging.getLogger()  # default to Python root logger.
 
     def __getattr__(self, attr: str) -> Any:
         """
--- a/src/uwtools/logging.py
+++ b/src/uwtools/logging.py
@@ -5,6 +5,7 @@ Logging support.
 import logging
 import os
 import sys
+from typing import Any
 
 # The logging prefix
 #
@@ -15,7 +16,26 @@ import sys
 
 MSGWIDTH = 69
 
-log = logging.getLogger()  # default to Python base logger
+
+class _Logger:
+    """
+    Support for swappable loggers.
+    """
+
+    def __init__(self) -> None:
+        self.logger = logging.getLogger()  # default to Python base logger.
+
+    def __getattr__(self, attr: str) -> Any:
+        """
+        Delegate attribute access to the currently-used logger.
+
+        :param attr: The attribute to access.
+        :returns: The requested attribute.
+        """
+        return getattr(self.logger, attr)
+
+
+log = _Logger()
 
 
 def setup_logging(quiet: bool = False, verbose: bool = False) -> None:
@@ -46,5 +66,4 @@ def use_logger(logger: logging.Logger) -> None:
 
     :param logger: The logger to log to.
     """
-    global log  # pylint: disable=global-statement
-    log = logger
+    log.logger = logger
--- a/src/uwtools/logging.py
+++ b/src/uwtools/logging.py
@@ -15,6 +15,8 @@ import sys
 
 MSGWIDTH = 69
 
+log = logging.getLogger()  # default to Python base logger
+
 
 def setup_logging(quiet: bool = False, verbose: bool = False) -> None:
     """
@@ -36,3 +38,13 @@ def setup_logging(quiet: bool = False, verbose: bool = False) -> None:
         **({"filename": os.devnull} if quiet else {}),
     }
     logging.basicConfig(**kwargs)
+
+
+def use_logger(logger: logging.Logger) -> None:
+    """
+    Log hereafter via the given logger.
+
+    :param logger: The logger to log to.
+    """
+    global log  # pylint: disable=global-statement
+    log = logger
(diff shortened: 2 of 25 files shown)


src/uwtools/apps/srw.py
maddenp-noaa, line 1: The two diffs in this module are representative of most that follow: Either `import` updates, or updates to logging calls.

src/uwtools/logging.py
```diff
@@ -16,6 +17,27 @@
 MSGWIDTH = 69
 
 
+class _Logger:
+    """
+    Support for swappable loggers.
+    """
+
+    def __init__(self) -> None:
+        self.logger = logging.getLogger()  # default to Python root logger.
```
maddenp-noaa, line 26: When a `uwtools.logging._Logger` object is instantiated, it will initially be connected to the Python root logger, which is what we want in CLI mode.

src/uwtools/logging.py
```diff
@@ -16,6 +17,27 @@
...
+
+    def __getattr__(self, attr: str) -> Any:
+        """
+        Delegate attribute access to the currently-used logger.
+
+        :param attr: The attribute to access.
+        :returns: The requested attribute.
+        """
+        return getattr(self.logger, attr)
```
maddenp-noaa, line 35: This method permits instances of `uwtools.logging._Logger` to respond to arbitrary attribute lookups (e.g. a lookup for an `info` attribute in an expression like `log.info("hello")`) and pass them on to the currently-connected logger.

src/uwtools/logging.py
```diff
@@ -16,6 +17,27 @@
...
+
+
+log = _Logger()
```
maddenp-noaa, line 38: This is the singleton `_Logger` object to be used throughout the lifetime of a `uwtools` CLI invocation or API session.

src/uwtools/logging.py
```diff
@@ -36,3 +58,12 @@ def setup_logging(quiet: bool = False, verbose: bool = False) -> None:
...
     logging.basicConfig(**kwargs)
+
+
+def use_logger(logger: logging.Logger) -> None:
+    """
+    Log hereafter via the given logger.
+
+    :param logger: The logger to log to.
+    """
+    log.logger = logger
```
maddenp-noaa, line 69: This is the public method to be used by API users to pass in a logger customized for their needs.

src/uwtools/tests/test_logging.py
```diff
@@ -50,3 +50,13 @@ def test_setup_logging_verbose():
...
         )
+
+
+def test_use_logger():
+    with patch.object(uwtools.logging, "log", uwtools.logging._Logger()):
+        # Initially, uwtools logging uses the Python root logger:
+        assert uwtools.logging.log.logger == logging.getLogger()
+        # But the logger can be swapped to use a logger of choice:
+        test_logger = logging.getLogger("test-logger")
+        uwtools.logging.use_logger(test_logger)
```
christinaholtNOAA, line 61: So the user who plans to use their own logger and our API would just call this once they set up their own logger and then don't have to worry about it again?

christinaholtNOAA (approved):
LGTM. One clarification question below.

maddenp-noaa:
I meant to show this earlier, but here's an example of how this would work. Since we don't really have API code yet, I first tweaked the code behind `uw template render` so that default logging is set up (this would be done by the CLI, but API users are on their own):
```
% git diff -U0 src/uwtools/config/templater.py
diff --git a/src/uwtools/config/templater.py b/src/uwtools/config/templater.py
index be1ebef..ad5d417 100644
--- a/src/uwtools/config/templater.py
+++ b/src/uwtools/config/templater.py
@@ -33,0 +34,2 @@ def render(
+    from uwtools.logging import setup_logging
+    setup_logging(verbose=True)
```
With that in place, I can simulate an API call from the command line, too, and see the `uwtools` default logging:
```
% echo "hello {{ recipient }}" | recipient=world python -c "from uwtools.config.templater import render; render(None, None, None)"
[2023-10-19T21:29:51]    DEBUG Internal arguments:
[2023-10-19T21:29:51]    DEBUG ---------------------------------------------------------------------
[2023-10-19T21:29:51]    DEBUG       input_file: None
[2023-10-19T21:29:51]    DEBUG      output_file: None
[2023-10-19T21:29:51]    DEBUG      values_file: None
[2023-10-19T21:29:51]    DEBUG    values_format: None
[2023-10-19T21:29:51]    DEBUG        overrides: None
[2023-10-19T21:29:51]    DEBUG    values_needed: False
[2023-10-19T21:29:51]    DEBUG          dry_run: False
[2023-10-19T21:29:51]    DEBUG    setup_logging: <function setup_logging at 0x7f41c85f3560>
[2023-10-19T21:29:51]    DEBUG ---------------------------------------------------------------------
[2023-10-19T21:29:51]    DEBUG Initial values taken from environment
hello world
```
(Ignore the `setup_logging` entry in the `Internal arguments` block -- this is a side effect of the patching shown in the diff and would not be there in the final code.)

Now, instead, I can simulate a user setting up a custom logger, then calls `use_logger()` to have `uwtools` use it. This one replaces the timestamp with the word FOO. An API user would, of course, do this in their own code, but I just hacked it into `uwtools`:
```
% git diff -U0 src/uwtools/config/templater.py
diff --git a/src/uwtools/config/templater.py b/src/uwtools/config/templater.py
index be1ebef..cce2170 100644
--- a/src/uwtools/config/templater.py
+++ b/src/uwtools/config/templater.py
@@ -33,0 +34,10 @@ def render(
+    import logging
+    from uwtools.logging import log, use_logger
+    logger = logging.getLogger("foo")
+    for handler in logger.handlers:
+        logger.removeHandler(handler)
+    logger.setLevel(logging.DEBUG)
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("FOO - %(message)s"))
+    logger.addHandler(handler)
+    use_logger(logger)
```
API simulation:
```
% echo "hello {{ recipient }}" | recipient=world python -c "from uwtools.config.templater import render; render(None, None, None)"
FOO - Internal arguments:
FOO - ---------------------------------------------------------------------
FOO -       input_file: None
FOO -      output_file: None
FOO -      values_file: None
FOO -    values_format: None
FOO -        overrides: None
FOO -    values_needed: False
FOO -          dry_run: False
FOO -          logging: <module 'logging' from '/home/maddenp/sw/conda/envs/DEV-uwtools/lib/python3.11/logging/__init__.py'>
FOO -              log: <uwtools.logging._Logger object at 0x7f3cbd893310>
FOO -       use_logger: <function use_logger at 0x7f3cbdaf3740>
FOO -           logger: <Logger foo (DEBUG)>
FOO -          handler: <StreamHandler <stderr> (NOTSET)>
FOO - ---------------------------------------------------------------------
FOO - Initial values taken from environment
hello world
```
(Again, ignore the items in the `Internal arguments` block following `dry_run` -- those are side effects of this demo.)

PR #322, opened 2023-10-23
Add an MPAS driver 

christinaholtNOAA:
**Description**

Adds a second forecast driver for MPAS. To do that, some refactoring was needed:

* The forecast module now contains a Forecast base class to hold the common methods to both forecast subclasses.
* The forecast driver tests were split into classes for Forecast, FV3Forecast, and MPASForecast.
* Some additional tests were needed after the last forecast overhaul missed them.
* Start using Path paths in a few places where it wasn't explicit in the drivers.

**Type of change**

**How Has This Been Tested?**

Standard test suite.

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -9,6 +9,7 @@ import sys
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from datetime import datetime
+from pathlib import Path
 from typing import Any, Dict, Optional, Type, Union
 
 from uwtools.config import validator
@@ -115,7 +116,9 @@ class Driver(ABC):
 
     @staticmethod
     def stage_files(
-        run_directory: str, files_to_stage: Dict[str, Union[list, str]], link_files: bool = False
+        run_directory: DefinitePath,
+        files_to_stage: Dict[str, Union[list, str]],
+        link_files: bool = False,
     ) -> None:
         """
         Creates destination files in run directory and copies or links contents from the source path
@@ -127,9 +130,10 @@ class Driver(ABC):
             (values).
         :param link_files: Whether to link or copy the files.
         """
+        run_directory = Path(run_directory)
         link_or_copy = os.symlink if link_files else shutil.copyfile
         for dst_rel_path, src_path_or_paths in files_to_stage.items():
-            dst_path = os.path.join(run_directory, dst_rel_path)
+            dst_path = run_directory / dst_rel_path
             if isinstance(src_path_or_paths, list):
                 Driver.stage_files(
                     dst_path,
@@ -137,7 +141,7 @@ class Driver(ABC):
                     link_files,
                 )
             else:
-                link_or_copy(src_path_or_paths, dst_path)  # type: ignore
+                link_or_copy(src_path_or_paths, str(dst_path))  # type: ignore
                 msg = f"File {src_path_or_paths} staged as {dst_path}"
                 logging.info(msg)
 
@@ -155,6 +159,7 @@ class Driver(ABC):
 
         # Exit program with error if caller chooses to quit.
 
+        run_dir = Path(run_dir)
         if exist_act == "quit" and run_dir.is_dir():
             logging.critical("User chose quit option when creating directory")
             sys.exit(1)
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -112,7 +112,7 @@ class Driver(ABC):
 
     @property
     @abstractmethod
-    def schema_file(self) -> str:
+    def schema_file(self) -> Path:
         """
         The path to the file containing the schema to validate the config file against.
         """
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -167,13 +167,13 @@ class Driver(ABC):
 
         run_dir = Path(run_directory)
         if exist_act == "quit" and run_dir.is_dir():
-            logging.critical("User chose quit option when creating directory")
+            log.critical("User chose quit option when creating directory")
             sys.exit(1)
 
         # Delete or rename directory if it exists.
         handle_existing(run_dir, exist_act)
 
-        logging.info("Creating directory: %s", run_dir)
+        log.info("Creating directory: %s", run_dir)
         os.makedirs(run_dir)
 
     @staticmethod
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -52,15 +52,15 @@ class Driver(ABC):
         :return: The batch script object with all run commands needed for executing the program.
         """
 
-    @staticmethod
-    def create_directory_structure(run_directory: DefinitePath, exist_act="delete"):
+    @classmethod
+    def create_directory_structure(cls, run_directory: DefinitePath, exist_act="delete"):
         """
         Creates the run directory for the forecast.
 
         :param run_directory: Path of desired run directory.
         :param exist_act: caller-specified action to delete, rename, or quit.
         """
-        Driver._create_run_directory(run_directory, exist_act)
+        cls._create_run_directory(run_directory, exist_act)
 
     @abstractmethod
     def output(self) -> None:
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -49,6 +49,12 @@ class Driver(ABC):
         :return: The batch script object with all run commands needed for executing the program.
         """
 
+    def create_directory_structure(self, run_directory: DefinitePath, exist_act="delete"):
+        """
+        Creates the run directory for the forecast
+        """
+        self._create_run_directory(run_directory, exist_act)
+
     @abstractmethod
     def output(self) -> None:
         """
@@ -133,6 +139,28 @@ class Driver(ABC):
                 logging.info(msg)
 
     # Private methods
+
+    @staticmethod
+    def _create_run_directory(run_dir: DefinitePath, exist_act="delete") -> None:
+
+        """
+        Makes the run directory for the driver. If it already exists, handles it based on the
+        caller instruction.
+        """
+        if exist_act not in ["delete", "rename", "quit"]:
+            raise ValueError(f"Bad argument: {exist_act}")
+
+        # Exit program with error if caller chooses to quit.
+
+        if exist_act == "quit" and os.path.isdir(run_dir):
+            logging.critical("User chose quit option when creating directory")
+            sys.exit(1)
+
+        # Delete or rename directory if it exists.
+        handle_existing(run_dir, exist_act)
+
+        logging.info("Creating directory: %s", run_dir)
+        os.makedirs(run_dir)
 
     @staticmethod
     def _create_user_updated_config(
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -56,6 +56,9 @@ class Driver(ABC):
     def create_directory_structure(run_directory: DefinitePath, exist_act="delete"):
         """
         Creates the run directory for the forecast.
+
+        :param run_directory: Path of desired run directory.
+        :param exist_act: caller-specified action to delete, rename, or quit.
         """
         Driver._create_run_directory(run_directory, exist_act)
 
@@ -80,7 +83,7 @@ class Driver(ABC):
     @abstractmethod
     def run(self, cycle: datetime) -> bool:
         """
-        Run the NWP tool.
+        Run the NWP tool for a given date and time.
 
         :param cycle: The time stamp of the cycle to run.
         :return: Did the driver exit with success status?
@@ -148,18 +151,21 @@ class Driver(ABC):
     # Private methods
 
     @staticmethod
-    def _create_run_directory(run_dir: DefinitePath, exist_act="delete") -> None:
+    def _create_run_directory(run_directory: DefinitePath, exist_act="delete") -> None:
         """
         Makes the run directory for the driver.
 
-        If it already exists, handles it based on the caller instruction.
+        If it already exists, handles it based on the caller-specified action.
+
+        :param run_directory: Path of desired run directory.
+        :param exist_act: caller-specified action to delete, rename, or quit.
         """
         if exist_act not in ["delete", "rename", "quit"]:
             raise ValueError(f"Bad argument: {exist_act}")
 
         # Exit program with error if caller chooses to quit.
 
-        run_dir = Path(run_dir)
+        run_dir = Path(run_directory)
         if exist_act == "quit" and run_dir.is_dir():
             logging.critical("User chose quit option when creating directory")
             sys.exit(1)
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -5,6 +5,7 @@ Provides an abstract class representing drivers for various NWP tools.
 import logging
 import os
 import shutil
+import sys
 from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from datetime import datetime
@@ -13,7 +14,8 @@ from typing import Any, Dict, Optional, Type, Union
 from uwtools.config import validator
 from uwtools.config.core import Config, YAMLConfig
 from uwtools.scheduler import BatchScript, JobScheduler
-from uwtools.types import OptionalPath
+from uwtools.types import DefinitePath, OptionalPath
+from uwtools.utils.file import handle_existing
 
 
 class Driver(ABC):
@@ -49,11 +51,12 @@ class Driver(ABC):
         :return: The batch script object with all run commands needed for executing the program.
         """
 
-    def create_directory_structure(self, run_directory: DefinitePath, exist_act="delete"):
+    @staticmethod
+    def create_directory_structure(run_directory: DefinitePath, exist_act="delete"):
         """
-        Creates the run directory for the forecast
+        Creates the run directory for the forecast.
         """
-        self._create_run_directory(run_directory, exist_act)
+        Driver._create_run_directory(run_directory, exist_act)
 
     @abstractmethod
     def output(self) -> None:
@@ -142,17 +145,17 @@ class Driver(ABC):
 
     @staticmethod
     def _create_run_directory(run_dir: DefinitePath, exist_act="delete") -> None:
-
         """
-        Makes the run directory for the driver. If it already exists, handles it based on the
-        caller instruction.
+        Makes the run directory for the driver.
+
+        If it already exists, handles it based on the caller instruction.
         """
         if exist_act not in ["delete", "rename", "quit"]:
             raise ValueError(f"Bad argument: {exist_act}")
 
         # Exit program with error if caller chooses to quit.
 
-        if exist_act == "quit" and os.path.isdir(run_dir):
+        if exist_act == "quit" and run_dir.is_dir():
             logging.critical("User chose quit option when creating directory")
             sys.exit(1)
 
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -6,6 +6,7 @@ Drivers for forecast models.
 import logging
 import os
 import sys
+from abc import ABC, abstractmethod
 from collections.abc import Mapping
 from datetime import datetime
 from importlib import resources
@@ -13,15 +14,15 @@ from pathlib import Path
 from typing import Dict, Optional
 
 from uwtools.config.core import FieldTableConfig, NMLConfig, YAMLConfig
+from uwtools.config.j2template import J2Template
 from uwtools.drivers.driver import Driver
 from uwtools.scheduler import BatchScript
 from uwtools.types import DefinitePath, OptionalPath
-from uwtools.utils.file import change_dir, handle_existing
+from uwtools.utils.file import change_dir, readable, writable
 from uwtools.utils.processing import execute
 
 
-class Forecast(Driver):
-
+class Forecast(Driver, ABC):
     """
     The base class for any forecast driver.
     """
@@ -80,13 +81,14 @@ class Forecast(Driver):
         """
         Runs FV3 either locally or via a batch-script submission.
 
+        :param cycle: the date and start time for the forecast
         :return: Did the FV3 run exit with success status?
         """
         # Prepare directories.
-        run_directory = self._config["run_dir"]
+        run_directory = Path(self._config["run_dir"])
         self.create_directory_structure(run_directory, "delete")
 
-        self._prepare_config_files(Path(run_directory))
+        self._prepare_config_files(cycle, Path(run_directory))
 
         self._config["cycle-dependent"].update(self._define_boundary_files())
 
@@ -114,7 +116,7 @@ class Forecast(Driver):
             print(full_cmd, file=sys.stdout)
             return True
 
-        with change_dir(run_dir):
+        with change_dir(run_directory):
             result = execute(cmd=full_cmd)
         return result.success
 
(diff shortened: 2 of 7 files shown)


src/uwtools/drivers/forecast.py
```diff
@@ -184,17 +116,10 @@ def run(self, cycle: datetime) -> bool:
             print(full_cmd, file=sys.stdout)
             return True
 
-        result = execute(cmd=full_cmd)
+        with change_dir(run_directory):
+            result = execute(cmd=full_cmd)
```
christinaholtNOAA, line 120: The executables should typically be run from the directory where all the data is staged.

src/uwtools/drivers/forecast.py
```diff
@@ -256,4 +299,93 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+
+        template_file = self._config["streams"]["template"]
+        values = self._config["streams"]["vars"]
+
+        with readable(template_file) as f:
+            template_str = f.read()
+
+        template = J2Template(values=values, template_str=template_str)
+        with writable(output_path) as f:
+            print(template.render(), file=f)
```
christinaholtNOAA, line 355: The streams files are XML, and I'm currently treating them as templates to make it fairly straightforward. We may consider switching that approach at some point.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -82,6 +100,153 @@ def schema():
 """.strip()
 
 
+def test_create_directory_structure(tmp_path):
```
christinaholtNOAA, line 103: As the methods became more generic and moved up in the structure to the base class, the tests needed to be testing those base class methods directly.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -1,26 +1,29 @@
-# pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
+# pylint: disable=missing-class-docstring, missing-function-docstring,protected-access,redefined-outer-name
```
christinaholtNOAA, line 1: Because this file was reorganized into classes, it will be easier to see important diffs with the "hide whitespace" setting enabled (under the gear dropdown at the top of the PR "Files changed" tab)

src/uwtools/utils/file.py
```diff
@@ -47,6 +47,22 @@ class _FORMAT:
 FORMAT = _FORMAT()
 
 
+@contextmanager
+def change_dir(path: Path) -> Iterator[None]:
```
christinaholtNOAA, line 51: A more pythonic way to capture `pushd/popd` behavior.

src/uwtools/resources/MPASForecast.jsonschema
maddenp-noaa, line 1: Perhaps too idealistic/optimistic for now, but: I wish we didn't have two schemas for YAML-based forecast descriptions. Is there a way to 1. describe the general properties of a forecast, 2. specify a specific forecast model, and then 3. have logic that interprets the general description in the context of that model to produce the required run-time assets? That's essentially what compilers do: Take a high-level description of a sequence of operations (aka programming-language syntax), combine it with knowledge of the abilities and requirements of a specific CPU, and emit a set of CPU-specific instructions that carry out the high-level program. That's what I would hope UW YAML could become. Having a different language for each model may be a pragmatic place to start, but are there fundamental/technical reasons why we couldn't generalize, abstract and, well, _unify_?

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -82,6 +100,153 @@ def schema():
...
+    with raises(ValueError):
+        Driver.create_directory_structure(run_directory="/some/path", exist_act="foo")
+
+
+def test_run_cmd_expected(mpi_config, tmp_path):
+    config_file = tmp_path / "config.yaml"
+    YAMLConfig.dump_dict(config_file, mpi_config)
+    with patch.object(Driver, "_validate", return_value=True):
+        driver = ConcreteDriver(config_file=config_file)
+    driver._config = driver._experiment_config.get("component", {})
```
maddenp-noaa, line 120: Is there doubt here as to whether `.get()` is going to find `component`? I'm surprised by the use of `.get()` instead of `[]`, and the supply of a default value. Same question about the next test function...

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -82,6 +100,153 @@ def schema():
...
+    with raises(KeyError):
+        driver.run_cmd()
+
+
+def test_scheduler():
+    config_file = fixture_path("fruit_config.yaml")
+    with patch.object(Driver, "_validate", return_value=True):
+        concretedriver = ConcreteDriver(config_file=config_file)
+        # pylint: disable=pointless-statement
+        with patch.object(driver.JobScheduler, "get_scheduler") as _get_scheduler:
```
maddenp-noaa, line 140: Why the apparent private name for the `_get_scheduler` mock (that is, vs just `get_scheduler`)? If we _did_ need a different name for a local variable, e.g. to avoid conflicting with a Python keyword, I think the idiom would be an underscore suffix rather than a prefix.

maddenp-cu (commented):
I'm still reviewing the unit tests and thought I'd submit this batch of comments in case you want to look at them before I have a chance to finish the review.

src/uwtools/drivers/forecast.py
christinaholtNOAA: Yep! I can make that change. I think `execute()` came around in between me doing this originally for MPAS and now and I didn't make that connection.

src/uwtools/drivers/forecast.py
```diff
@@ -150,13 +81,14 @@ def run(self, cycle: datetime) -> bool:
...
+        :param cycle: the date and start time for the forecast
         :return: Did the FV3 run exit with success status?
         """
         # Prepare directories.
-        run_directory = self._config["run_dir"]
+        run_directory = Path(self._config["run_dir"])
         self.create_directory_structure(run_directory, "delete")
 
-        self._prepare_config_files(Path(run_directory))
+        self._prepare_config_files(cycle, Path(run_directory))
```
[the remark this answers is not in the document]
maddenp-noaa: I see, I think: "config files" in your comment referred to the _component_ config files. We might need some better nomenclature conventions here to differentiate. UW does a LOT with "config files", generally referring to its inputs, rather than its outputs. It might be helpful to have a different term we can use for the component driver configs. I don't have an idea at the moment, but a good name could be helpful.

src/uwtools/drivers/driver.py
```diff
@@ -49,6 +52,13 @@ def batch_script(self) -> BatchScript:
         :return: The batch script object with all run commands needed for executing the program.
         """
 
+    @staticmethod
+    def create_directory_structure(run_directory: DefinitePath, exist_act="delete"):
+        """
+        Creates the run directory for the forecast.
+        """
+        Driver._create_run_directory(run_directory, exist_act)
```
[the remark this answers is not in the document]
maddenp-noaa: I see it now: The default implementation only creates the run directory; `FV3Forecast` overrides it to do more. Thanks.

src/uwtools/drivers/forecast.py
```diff
@@ -209,6 +134,124 @@ def _boundary_hours(self, lbcs_config: Dict) -> tuple[int, int, int]:
...
+        """
+        Creates the run directory and adds subdirectories INPUT and RESTART. Verifies creation of
+        all directories.
+
+        :param run_directory: Path of desired run directory.
+        :param exist_act: Could be any of 'delete', 'rename', 'quit'. Sets how the program responds
+            to a preexisting run directory. The default is to delete the old run directory.
+        """
+        run_directory = Path(run_directory)
+        Driver._create_run_directory(run_directory, exist_act)  # pylint: disable=protected-access
```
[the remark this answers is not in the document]
maddenp-noaa: You're right, I see it now. Out of curiosity, I tried it as a `@classmethod` instead, and that seems maybe like an improvement:
```
% git diff -U0
diff --git a/src/uwtools/drivers/driver.py b/src/uwtools/drivers/driver.py
index 73c304a..3cc603c 100644
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -55,2 +55,2 @@ class Driver(ABC):
-    @staticmethod
-    def create_directory_structure(run_directory: DefinitePath, exist_act="delete"):
+    @classmethod
+    def create_directory_structure(cls, run_directory: DefinitePath, exist_act="delete"):
@@ -63 +63 @@ class Driver(ABC):
-        Driver._create_run_directory(run_directory, exist_act)
+        cls._create_run_directory(run_directory, exist_act)
diff --git a/src/uwtools/drivers/forecast.py b/src/uwtools/drivers/forecast.py
index de56b3d..71dc9f4 100644
--- a/src/uwtools/drivers/forecast.py
+++ b/src/uwtools/drivers/forecast.py
@@ -185,2 +185,4 @@ class FV3Forecast(Forecast):
-    @staticmethod
-    def create_directory_structure(run_directory: DefinitePath, exist_act: str = "delete") -> None:
+    @classmethod
+    def create_directory_structure(
+        cls, run_directory: DefinitePath, exist_act: str = "delete"
+    ) -> None:
@@ -196 +198 @@ class FV3Forecast(Forecast):
-        Driver._create_run_directory(run_directory, exist_act)  # pylint: disable=protected-access
+        cls._create_run_directory(run_directory, exist_act)
```
Gets rid of the `pylint` suppressor and passes all tests.

src/uwtools/resources/MPASForecast.jsonschema
maddenp-noaa: I believe that, yes, there's support for referencing other schemas -- see e.g. https://json-schema.org/understanding-json-schema/structuring

It occurred to me this morning that it might still be necessary to have separate config blocks for power-user control of features that one model might offer that another does not. For example, the way MPAS configures its output streams, and one one can do with them, probably doesn't map 1:1 onto what e.g. FV3 does. So, I can imagine that a cross-model config language / schema might provide a way to specify the frequency at which history files should be written to disk -- a setting that probably applies to any forecast model -- it might be useful to have an ancillary config block and associated schema for, say, MPAS that exposes some edge-case / power-user features applicable only to MPAS. But most users could probably ignore this and use the more generic model-agnostic "forecast" language. (If a model didn't support configurable history frequency, UW could also opt to just emit a warning message saying "Model Foo only supports hourly history output; setting ignored". Each model's driver could decide how to respond to / implement / ignore configuration information.)

src/uwtools/tests/drivers/test_driver.py
maddenp-noaa: That's the reason for a lot of weird things I do. That and ignorance.

src/uwtools/drivers/forecast.py
```diff
@@ -184,31 +116,155 @@ def run(self, cycle: datetime) -> bool:
...
+        start_time = cycle.strftime("%Y-%m-%d_%H:%M:%S")
+        date_values = {
+            "start_year": cycle.strftime("%Y"),
+            "start_month": cycle.strftime("%m"),
+            "start_day": cycle.strftime("%d"),
+            "start_hour": cycle.strftime("%H"),
+            "start_minute": cycle.strftime("%M"),
+            "start_second": cycle.strftime("%S"),
+        }
+        log.info(f"Updating namelist date values to start at: {start_time}")
```
maddenp-noaa, line 238: Is `model_configure` the file that is kind-of-but-not-really a namelist file? If so, should we say something other than "namelist" here? Or, even if I'm wrong about that, might it be helpful to log that this is about the `model_configure` file, or even e.g.
```
log.info(f"Updating date values to start at {start_time} in {output_path}")
```
?

src/uwtools/drivers/forecast.py
```diff
@@ -184,31 +116,155 @@ def run(self, cycle: datetime) -> bool:
...
+        config_obj = YAMLConfig(output_path)
+        config_obj.update_values(date_values)
+        config_obj.dump(output_path)
+
+    def create_namelist(self, output_path: OptionalPath) -> None:
+        """
+        Uses an object with user supplied values and an optional namelist base file to create an
+        output namelist file. Will "dereference" the base file.
```
maddenp-noaa, line 246: What does "optional namelist base file" refer to here?

Also, I don't see anything about dereferencing being done here. If that's something that's implicit in the use of `NMLConfig`, then IMO we shouldn't mention it here: That class's behavior should be documented in that class.

src/uwtools/drivers/forecast.py
```diff
@@ -256,4 +316,106 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+        """
+        Initialize the Forecast Driver.
+        """
+
+        super().__init__(config_file=config_file, dry_run=dry_run, batch_script=batch_script)
+
+    def create_namelist(self, cycle: datetime, output_path: OptionalPath) -> None:
+        """
+        Uses an object with user supplied values and an optional namelist base file to create an
+        output namelist file. Will "dereference" the base file.
```
maddenp-noaa, line 339: Similar questions/comments in this method about
1. "optional namelist base file" and "dereference" in this docstring
2. Whether we can do what's needed in a single `NMLConfig` call by setting up `config_values` first.

src/uwtools/drivers/forecast.py
```diff
@@ -256,4 +316,106 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+        """
+        No boundary files are currently needed for MPAS global support.
+
+        :return: A dict of boundary file names mapped to source input file paths
+        """
+        return {}
+
+    def _mpi_env_variables(self, delimiter=" ") -> str:
+        """
+        Sets the environment variables needed for running the mpi command.
```
maddenp-noaa, line 399: `mpi` -> `MPI`

src/uwtools/drivers/forecast.py
```diff
@@ -256,4 +316,106 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+
+        :param delimiter: The delimiter to be used between items in the configuration dictionary.
+        :return: A bash-formatted string of MPI environment variables.
+        """
+        return delimiter.join([f"{k}={v}" for k, v in self._config.get("mpi_settings", {}).items()])
```
maddenp-noaa, line 404: Just trying to get my bearings: Is `mpi_settings` something that should appear in `MPASForecast.jsonschema`? I don't see the `additionalProperties` key in that schema, so I suppose additional, unrecognized content might be tolerated, but that's probably not what we'd want. I was wondering, again, if this item was really optional (i.e. why we're using `.get()` with a `{}` default instead of `[]`), and this time thought that -- maybe unlike, say, a block defining the content of `model_configure` for FV3 -- this might truly be optional, so thought I'd see if the schema shed any light on this. But it did not.

maddenp-cu (commented):
Another, smaller, batch of comments/questions... Still not finished reviewing the unit tests, sorry, getting there.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -83,13 +101,160 @@ def schema():
...
+        driver.run_cmd()
+
+
+def test_scheduler():
+    config_file = fixture_path("fruit_config.yaml")
+    with patch.object(Driver, "_validate", return_value=True):
+        concretedriver = ConcreteDriver(config_file=config_file)
+        # pylint: disable=pointless-statement
+        with patch.object(driver.JobScheduler, "get_scheduler") as get_scheduler:
+            concretedriver.scheduler
```
maddenp-noaa, line 142: If I'm reading right, `Driver.scheduler` is a `@property` method that returns the value returned by calling `JobScheduler.get_scheduler()`. We've mocked the latter, so we can control what it returns. It might be useful to demonstrate that `.scheduler` is essentially a wrapper around `.get_scheduler()`, which could maybe be done like
```
def test_scheduler():
    with patch.object(Driver, "_validate", return_value=True):
        concretedriver = ConcreteDriver(config_file=fixture_path("fruit_config.yaml"))
        with patch.object(concretedriver, "resources", return_value="bar"):
            with patch.object(driver.JobScheduler, "get_scheduler") as get_scheduler:
                get_scheduler.return_value = "foo"
                assert concretedriver.scheduler == "foo"
                get_scheduler.assert_called_once_with("bar")
```
That demonstrates that the value of `.scheduler` is based on the value returned by `.get_scheduler()`, and that `.get_scheduler()` gets called with whatever `.resources()` returns, which we can also control. No more `pointless-statement`.

maddenp-noaa: Or maybe we don't need to mock `.resources()` -- looks like we know it's going to return `{}` in our tests. So
```
ef test_scheduler():
    with patch.object(Driver, "_validate", return_value=True):
        concretedriver = ConcreteDriver(config_file=fixture_path("fruit_config.yaml"))
        with patch.object(driver.JobScheduler, "get_scheduler") as get_scheduler:
            get_scheduler.return_value = "foo"
            assert concretedriver.scheduler == "foo"
            get_scheduler.assert_called_once_with({})
```

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -83,13 +101,160 @@ def schema():
...
+            for src_file in files_to_stage[dst_fn]:
+                Path(src_file).touch()
+        else:
+            fixed_src_path = src_directory / Path(src_path).name
+            files_to_stage[dst_fn] = str(fixed_src_path)
+            fixed_src_path.touch()
+    # Test that none of the destination files exist yet:
+    for dst_fn in files_to_stage.keys():
+        assert not (run_directory / dst_fn).is_file()
+    # Ask a forecast object to stage the files to the run directory:
```
maddenp-noaa, line 179: Is `forecast object` accurate in this comment? It looks to me like we're just calling `Driver` static methods.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -83,13 +101,160 @@ def schema():
...
+            assert link_or_file(run_directory / dst_rel_path)
+
+
+@pytest.mark.parametrize("exist_act", ["delete", "rename"])
+def test__create_run_directory_exists(exist_act, tmp_path):
+    run_dir = tmp_path / "run"
+    Driver._create_run_directory(run_dir, exist_act)
+    Driver._create_run_directory(run_dir, exist_act)
+    run_dirs = len(glob.glob((tmp_path / "run*").as_posix()))
+    assert run_dirs == 1 if exist_act == "delete" else run_dirs == 2
```
maddenp-noaa, line 198: Nice test.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -34,157 +37,162 @@ def test_batch_script():
...
+
+        # Test rename behavior when run directory exists.
+        FV3Forecast.create_directory_structure(rundir, "rename")
+        copy_directory = next(tmp_path.glob("%s_*" % rundir.name))
+        assert (copy_directory / "RESTART").is_dir()
+
+        # Test quit behavior when run directory exists.
+        with raises(SystemExit) as pytest_wrapped_e:
+            FV3Forecast.create_directory_structure(rundir, "quit")
+        assert pytest_wrapped_e.type == SystemExit
```
maddenp-noaa, line 77: Not a rhetorical question: Is this assertion moot given the `with raises(SystemExit)` above? Did we already establish that, or is this assertion asserting something different?

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -34,157 +37,162 @@ def test_batch_script():
...
+        rundir = tmp_path / "rundir"
+
+        # Test delete behavior when run directory does not exist.
+        FV3Forecast.create_directory_structure(rundir, "delete")
+        assert (rundir / "RESTART").is_dir()
+
+        # Create a file in the run directory.
+        test_file = rundir / "test.txt"
+        test_file.touch()
+        assert test_file.is_file()
```
maddenp-noaa, line 61: Would it make sense to factor these blocks out into individual tests?  The seem kind of independent and, if they do depend on the previous blocks working, maybe they shoudln't? There could be a class-level fixture like
```
@fixture
def rundir(self, tmp_path):
    return tmp_path / "rundir"
```
I count maybe 5 potential individual test in this method. I don't feel strongly about it, just thought it was worth checking.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -1,27 +1,30 @@
...
-from uwtools.drivers.forecast import FV3Forecast
+from uwtools.drivers.forecast import FV3Forecast, MPASForecast
 from uwtools.logging import log
 from uwtools.tests.support import compare_files, fixture_path
+from uwtools.utils.file import readable, writable
 
 
-def test_batch_script():
-    expected = """
+class TestForecast:
```
maddenp-noaa, line 25: I just got around to expanding the diff on this file and saw that you beat me to trying out `pytest` classes! Nice.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -34,157 +37,162 @@ def test_batch_script():
...
+
+    def test_create_namelist_call_private(self, tmp_path):
+        basefile = str(tmp_path / "base.yaml")
+        infile = fixture_path("forecast.yaml")
+        outfile = str(tmp_path / "out.yaml")
+        for path in infile, basefile:
+            Path(path).touch()
+        with patch.object(Driver, "_create_user_updated_config") as _create_user_updated_config:
+            with patch.object(FV3Forecast, "_validate", return_value=True):
+                # pylint: disable=unused-variable
```
maddenp-noaa, line 171: I think we could remove `as YAMLConfig` on the next line and then this `pylint` suppressor shouldn't be necessary.

Also, whenever possible, these suppressors that apply to one line should be an end-of-line comment, as otherwise they will apply to... I _think_ all the code in the block or scope, depending on where the suppressor is.

Or so I would have written a few minutes ago except that TIL, per [the docs](https://pylint.pycqa.org/en/latest/user_guide/messages/message_control.html#block-disables), that `pylint: disable-next=` disables violations only on _the following line_. But note the `disable-next` instead of `disable`.

The scoping rules on that linked-to page look useful. I need to review them myself.

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
christinaholtNOAA: I've made a change here to `_prepare_input_parm`, but we can certainly keep iterating. The [docs for FV3](https://ufs-weather-model.readthedocs.io/en/ufs-v1.0.0/InputsOutputs.html#inputs-and-outputs) call them "input files", but that seemed too broad as input data files are handled somewhere else in our code.

src/uwtools/drivers/forecast.py
```diff
@@ -184,31 +116,155 @@ def run(self, cycle: datetime) -> bool:
...
+            "start_month": cycle.strftime("%m"),
+            "start_day": cycle.strftime("%d"),
+            "start_hour": cycle.strftime("%H"),
+            "start_minute": cycle.strftime("%M"),
+            "start_second": cycle.strftime("%S"),
+        }
+        log.info(f"Updating namelist date values to start at: {start_time}")
+        config_obj = YAMLConfig(output_path)
+        config_obj.update_values(date_values)
+        config_obj.dump(output_path)
```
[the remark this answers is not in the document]
christinaholtNOAA: I think I misunderstood. I am updating this for both namelist and model_config.

src/uwtools/drivers/forecast.py
christinaholtNOAA: I think probably I should add it there, but really it would just be a generic object to hold arbitrary environment variables.

src/uwtools/tests/drivers/test_forecast.py
christinaholtNOAA: I believe it is moot.

src/uwtools/tests/drivers/test_forecast.py
christinaholtNOAA: Good call. I think it makes it clearer. 

I left the permissions check tied to the first creation test, but separated out the others.

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
maddenp-noaa: Could we consider not propagating the industry-NON-standard term `parm` here? `param` would be better IMO.

src/uwtools/drivers/forecast.py
```diff
@@ -184,31 +116,155 @@ def run(self, cycle: datetime) -> bool:
...
+
+    def create_field_table(self, output_path: OptionalPath) -> None:
+        """
+        Uses the forecast config object to create a Field Table.
+
+        :param output_path: Optional location of output field table.
+        """
+        self._create_user_updated_config(
+            config_class=FieldTableConfig,
+            config_values=self._config.get("field_table", {}),
```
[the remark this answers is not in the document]
maddenp-noaa: Just to be clear, and if I'm not mistaken, if a `field_table` key does not exist in `self._config`, this call will simply write an empty file to disk. Is that intended?

christinaholtNOAA:
I'm going to tuck this one away until I can come back to it.

Issue #323, opened 2023-10-23
UW-413 Ensure CLI functions accept file and stdin streams for both input and output


WeirAE:
**Description**
Streams for stdin were replaced last PR using `OptionalPath` and `readable()` while replacing `str() `or `PosixPath()` - however writes were not updated. 

**Requirements**
Calls to DefinitePath and Str will need to be updated to OptionalPath and tests should also be expanded as necessary to ensure this works correctly. Intended behavior is something like `cat some.yaml | uw rocoto realize >some.xml`

**Acceptance Criteria (Definition of Done)**
`DefinitePath` and `Str` typing of input and output paths replaced with `OptionalPath`
All existing tests pass and new tests added as necessary

WeirAE:
Completed by #324

PR #324, opened 2023-10-23
UW-413: Ensure CLI functions accept file and stdin streams for both input and output

WeirAE:
**Description**
Calls to `DefinitePath` and `Str` will need to be updated to `OptionalPath` and tests should also be expanded as necessary to ensure this works correctly.

Created from PR #316 , Fixes issue #323 

**Type of change**

**How Has This Been Tested?**

- [x] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/config/j2template.py
+++ b/src/uwtools/config/j2template.py
@@ -8,7 +8,7 @@ from typing import List, Optional, Set
 from jinja2 import BaseLoader, Environment, FileSystemLoader, Template, meta
 
 from uwtools.logging import log
-from uwtools.types import DefinitePath, OptionalPath
+from uwtools.types import OptionalPath
 from uwtools.utils.file import readable, writable
 
 
--- a/src/uwtools/config/j2template.py
+++ b/src/uwtools/config/j2template.py
@@ -8,8 +8,8 @@ from typing import List, Optional, Set
 
 from jinja2 import BaseLoader, Environment, FileSystemLoader, Template, meta
 
-from uwtools.types import DefinitePath, OptionalPath
-from uwtools.utils.file import readable
+from uwtools.types import OptionalPath
+from uwtools.utils.file import readable, writable
 
 
 class J2Template:
@@ -43,7 +43,7 @@ class J2Template:
 
     # Public methods
 
-    def dump(self, output_path: DefinitePath) -> None:
+    def dump(self, output_path: OptionalPath) -> None:
         """
         Write rendered template to the path provided.
 
@@ -51,7 +51,7 @@ class J2Template:
         """
         msg = f"Writing rendered template to output file: {output_path}"
         logging.debug(msg)
-        with open(output_path, "w+", encoding="utf-8") as f:
+        with writable(output_path) as f:
             print(self.render(), file=f)
 
     def render(self) -> str:
--- a/src/uwtools/config/j2template.py
+++ b/src/uwtools/config/j2template.py
@@ -8,8 +8,8 @@ from typing import List, Optional, Set
 
 from jinja2 import BaseLoader, Environment, FileSystemLoader, Template, meta
 
-from uwtools.types import OptionalPath
-from uwtools.utils.file import readable, writable
+from uwtools.types import DefinitePath, OptionalPath
+from uwtools.utils.file import readable
 
 
 class J2Template:
@@ -43,7 +43,7 @@ class J2Template:
 
     # Public methods
 
-    def dump(self, output_path: OptionalPath) -> None:
+    def dump(self, output_path: DefinitePath) -> None:
         """
         Write rendered template to the path provided.
 
@@ -51,7 +51,7 @@ class J2Template:
         """
         msg = f"Writing rendered template to output file: {output_path}"
         logging.debug(msg)
-        with writable(output_path) as f:
+        with open(output_path, "w+", encoding="utf-8") as f:
             print(self.render(), file=f)
 
     def render(self) -> str:
--- a/src/uwtools/config/j2template.py
+++ b/src/uwtools/config/j2template.py
@@ -8,7 +8,7 @@ from typing import List, Optional, Set
 from jinja2 import BaseLoader, Environment, FileSystemLoader, Template, meta
 
 from uwtools.logging import log
-from uwtools.types import OptionalPath
+from uwtools.types import DefinitePath, OptionalPath
 from uwtools.utils.file import readable, writable
 
 
@@ -79,7 +79,7 @@ class J2Template:
 
     # Private methods
 
-    def _load_file(self, template_path: OptionalPath) -> Template:
+    def _load_file(self, template_path: DefinitePath) -> Template:
         """
         Load the Jinja2 template from the file provided.
 
--- a/src/uwtools/rocoto.py
+++ b/src/uwtools/rocoto.py
@@ -2,15 +2,15 @@
 Support for creating Rocoto XML workflow documents.
 """
 
-import shutil
 import tempfile
 from importlib import resources
+from pathlib import Path
 
 from lxml import etree
 
-import uwtools.config.validator
 from uwtools.config.core import YAMLConfig
 from uwtools.config.j2template import J2Template
+from uwtools.config.validator import validate_yaml
 from uwtools.logging import log
 from uwtools.types import DefinitePath, OptionalPath
 from uwtools.utils.file import readable
@@ -95,7 +95,7 @@ def _write_rocoto_xml(
 # Public functions
 def realize_rocoto_xml(
     config_file: OptionalPath,
-    rendered_output: OptionalPath,
+    rendered_output: OptionalPath = None,
 ) -> bool:
     """
     Realize the Rocoto workflow defined in the given YAML as XML. Validate both the YAML input and
@@ -106,25 +106,23 @@ def realize_rocoto_xml(
     :return: Did the input and output files conform to theirr schemas?
     """
 
-    # Validate the YAML.
-    if uwtools.config.validator.validate_yaml(
-        config_file=config_file, schema_file=_rocoto_schema_yaml()
-    ):
-        # Render the template to a temporary file.
-        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
-            _write_rocoto_xml(
-                config_file=config_file,
-                rendered_output=temp_file.name,
-            )
-            # Validate the XML.
-            if validate_rocoto_xml(input_xml=temp_file.name):
-                # If no issues were detected, save temp file and report success.
-                shutil.move(temp_file.name, str(rendered_output))
-                return True
+    if not validate_yaml(config_file=config_file, schema_file=_rocoto_schema_yaml()):
+        log.error("YAML validation errors identified in %s", config_file)
+        return False
+
+    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+        _write_rocoto_xml(config_file=config_file, rendered_output=temp_file.name)
+
+    if not validate_rocoto_xml(input_xml=temp_file.name):
         log.error("Rocoto validation errors identified in %s", temp_file.name)
         return False
-    log.error("YAML validation errors identified in %s", config_file)
-    return False
+
+    if rendered_output is None:
+        with open(temp_file.name, "r", encoding="utf-8") as f:
+            print(f.read())
+    else:
+        Path(temp_file.name).rename(rendered_output)
+    return True
 
 
 def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
--- a/src/uwtools/rocoto.py
+++ b/src/uwtools/rocoto.py
@@ -76,33 +76,33 @@ def _rocoto_template_xml() -> DefinitePath:
 
 def _write_rocoto_xml(
     config_file: OptionalPath,
-    rendered_output: OptionalPath,
+    output_file: OptionalPath,
 ) -> None:
     """
     Render the Rocoto workflow defined in the given YAML to XML.
 
     :param config_file: Path to YAML input file.
-    :param rendered_output: Path to write rendered XML file.
+    :param output_file: Path to write rendered XML file.
     """
 
     values = _add_jobname_to_tasks(config_file)
 
     # Render the template.
     template = J2Template(values=values.data, template_path=_rocoto_template_xml())
-    template.dump(output_path=rendered_output)
+    template.dump(output_path=output_file)
 
 
 # Public functions
 def realize_rocoto_xml(
     config_file: OptionalPath,
-    rendered_output: OptionalPath = None,
+    output_file: OptionalPath = None,
 ) -> bool:
     """
     Realize the Rocoto workflow defined in the given YAML as XML. Validate both the YAML input and
     XML output.
 
     :param config_file: Path to YAML input file.
-    :param rendered_output: Path to write rendered XML file.
+    :param output_file: Path to write rendered XML file.
     :return: Did the input and output files conform to theirr schemas?
     """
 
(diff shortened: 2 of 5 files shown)


src/uwtools/rocoto.py
```diff
@@ -119,7 +119,7 @@ def realize_rocoto_xml(
             # Validate the XML.
             if validate_rocoto_xml(input_xml=temp_file.name):
                 # If no issues were detected, save temp file and report success.
-                shutil.move(temp_file.name, rendered_output)
+                shutil.move(temp_file.name, str(rendered_output))
```
WeirAE, line 122: I was hoping to replace `shutil.move` with `Path.rename` but since the temp file is a string path, this seemed to work better in this situation

WeirAE: reverting this change, so not an issue for this PR

WeirAE:
After some offline discussion, `j2template` and `atparse_to_jinja2` will stay unchanged so as to manage the scope of this PR

src/uwtools/rocoto.py
christinaholtNOAA: This is only going to work this way if the output is a file, no? Wouldn't it need some special treatment if it were stdout?

src/uwtools/config/j2template.py
```diff
@@ -43,15 +43,15 @@ def __init__(
...
+    def dump(self, output_path: OptionalPath) -> None:
         """
         Write rendered template to the path provided.
 
         :param output_path: Path to file to write.
         """
         msg = f"Writing rendered template to output file: {output_path}"
         logging.debug(msg)
-        with open(output_path, "w+", encoding="utf-8") as f:
+        with writable(output_path) as f:
```
christinaholtNOAA: So, this is necessary to implement the rocoto piece you're trying for in this PR. I had assumed in our conversation before seeing the PR that you were handling the behavior of all the `uw template` functions, too, which I think would be out of scope for this work. 

I don't see a connection to the atparse bit, so I think that can be left off, unless I'm missing a connection. Def let me know if that's the case!

christinaholtNOAA (changes requested):

src/uwtools/config/j2template.py
WeirAE: That's how I ended up expanding initially, but this can be bypassed with converting to str earlier - I should have that in the most recent update

src/uwtools/rocoto.py
WeirAE: So I thought so, I looked at a few options on file handling, but this seemed to work correctly instantiating the output file. However, if there's a better way I'll happily make it better here!

src/uwtools/config/j2template.py
christinaholtNOAA: It's not the technicality of passing a string or a Path I'm concerned with here. If the user wants to output to stdout instead of a file (the OptionalPath change), there is no way to achieve that without this writeable piece.

src/uwtools/rocoto.py
maddenp-noaa: Not quite related to the above discussion, but why use `.name` on `temp_file`, which gives a relative path instead of an absolute one?
```
>>> from pathlib import Path
>>> p = Path("/tmp/foo")
>>> p.is_file()
True
>>> p.name
'foo'
```
Any reason this wouldn't work with just `temp_file` if `shutil.move` accepts a `Path`, or `str(temp_file)` if not?

Or, maybe
```
Path(temp_file).rename(rendered_output)
```
?

maddenp-noaa: I'm also curious about the question @christinaholtNOAA raised: Will this work correctly if input is `stdin` and output is `stdout`? The creation of an on-disk temp file would probably best be conditional on the user wanting an on-disk final file; an in-memory / streaming use case should probably not create files at all. It may be functionally correct (and even acceptable) if it does create a temporary file and then remove it, but not as clean as it could be.

Could we see a CLI demo (output pasted here as a comment) of e.g. `cat config.yaml | uw rocoto realize >rocoto.xml` (I'm inventing file names, and maybe other CLI arguments/switches are needed, not sure) and show that `rocoto.xml` is created with the expected content? I'll see if I can try it myself, too.

maddenp-noaa: I see what you mean: `temp_file` is not a `Path` object, so `.name` doesn't mean what I thought it did. On the kind of object it **is**, (Python says: `<class 'tempfile._TemporaryFileWrapper'>`), `.name` is indeed the absolute path. Carry on!

src/uwtools/config/j2template.py
WeirAE: Happily restored!

src/uwtools/rocoto.py
WeirAE: You're right, of course - it seemed to be ok, but it is creating very wrong after being more rigorous. If I pipe in the input, I get `'NoneType' object has no attribute 'items'`. If I redirect out, it creates a log file called 'None' and gives `ERROR <attribute 'error_log' of 'lxml.etree._Validator' objects>`

WeirAE: If I change the input/output definitions in `cli.py` with `type=FileType('w'), default=sys.stdout,` it at least doesn't None me, but I get `[Errno 2] No such file or directory: "<_io.TextIOWrapper name='<stdin>' mode='r' encoding='utf-8'>"`

maddenp-noaa: I don't think we'd want to delegate that level of control to `cli.py`. IMO it should just take an optional string, so that `args.output_file` will either the the user-specified path or, if they don't specify one, `None`.

Would a light refactor of `realize_rocoto_xml()` help? I think the code might be clearer anyway without two nested `if` blocks, and what about, just behaviorally speaking:

1. Validate the YAML and return `False` if it fails. (Do we need more logging than what `validate_yaml()` provides? Honest question, I don't know.)
2. Render the template to a temp file.
3. Validate the template and return `False` if it fails.
4. Now that we've got valid XML on disk in the temp file, we can (a) Rename it if argument `rendered_output` is not `None`, or (b) Read it into a string and `print()` it otherwise. We don't have to use `writable()` if it's _less_ convenient.

What do you think?

maddenp-noaa: Not really tested, but maybe something like
```
    if not validate_yaml(config_file=config_file, schema_file=_rocoto_schema_yaml()):
        log.error("YAML validation errors identified in %s", config_file)
        return False

    with tempfile.NamedTemporaryFile(delete=False) as temp_file:
        _write_rocoto_xml(config_file=config_file, rendered_output=temp_file.name)

    if not validate_rocoto_xml(input_xml=temp_file.name):
        log.error("Rocoto validation errors identified in %s", temp_file.name)
        return False

    if rendered_output is None:
        with open(temp_file.name, "r", encoding="utf-8") as f:
            print(f.read())
    else:
        Path(temp_file.name).rename(rendered_output)
    return True
```
That depends on import tweaks:
```
-import shutil
+from pathlib import Path
-import uwtools.config.validator
+from uwtools.config.validator import validate_yaml
```
which then wants unit-test patch:
```
% git diff -U0 src/uwtools/tests/test_rocoto.py
diff --git a/src/uwtools/tests/test_rocoto.py b/src/uwtools/tests/test_rocoto.py
index a0031ff..08f1601 100644
--- a/src/uwtools/tests/test_rocoto.py
+++ b/src/uwtools/tests/test_rocoto.py
@@ -13,0 +14 @@ from uwtools import rocoto
+from uwtools.config import validator
@@ -78 +79 @@ def test_realize_rocoto_xml(vals, tmp_path):
-        with patch.object(rocoto.uwtools.config.validator, "_bad_paths", return_value=None):
+        with patch.object(validator, "_bad_paths", return_value=None):
@@ -89 +90 @@ def test_realize_rocoto_invalid_xml():
-        with patch.object(rocoto.uwtools.config.validator, "_bad_paths", return_value=None):
+        with patch.object(validator, "_bad_paths", return_value=None):
```

WeirAE: Hrm, it does work with the current tests in this form, but on the command line it still gives me the same errors, unfortunately. I'm not sure we can get away with not handling broader scope arguments in the cli if we want to pipe in and out of it since its expecting `type=str` there

maddenp-noaa: > it does work with the current tests in this form, but on the command line it still gives me the same errors
That could mean that the unit tests are not robust enough. Can we devise a unit test that exercises the bug, to avoid regression once we fix it?

As for a fix, could you give me a simple recipe to reproduce the issue via the checked-in code in your fork? I could take a look.

WeirAE: I'm pondering how to test piping in the unit tests, I've got a few things to try.

For the errors, in the current code you will need to edit `hello_workflow.yaml` with an explicit path in the `log` tag. For me, line 11 is   `log: /home/vboxuser/Documents/work/workflow-tools/`
I have 3 tests from your suggestion:
1) `cat ~/Documents/work/workflow-tools/src/uwtools/tests/fixtures/hello_workflow.yaml| uw rocoto realize >some.xml`
stdin/stdout - errors with `'NoneType' object has no attribute 'items'`; this seems to be the cli not reading the pipe and setting the object to None.
2) `uw rocoto realize -i ~/Documents/work/workflow-tools/src/uwtools/tests/fixtures/hello_workflow.yaml >~/Documents/work/workflow-tools/src/uwtools/tests/fixtures/some.xml`
stdout only - errors with `ERROR <attribute 'error_log' of 'lxml.etree._Validator' objects>` this seems to be the XML validator failing I think with the inability to read the xml file. I think the redirect is failing so the temp file is creating and passing a null object.
3) `cat ~/Documents/work/workflow-tools/src/uwtools/tests/fixtures/hello_workflow.yaml| uw rocoto realize -o ~/Documents/work/workflow-tools/src/uwtools/tests/fixtures/some.xml`
stdin only - same error as 1

src/uwtools/rocoto.py
```diff
@@ -106,25 +106,24 @@ def realize_rocoto_xml(
...
         return False
-    log.error("YAML validation errors identified in %s", config_file)
-    return False
+
+    if rendered_output is None:
+        with open(temp_file.name, "r", encoding="utf-8") as f:
+            print(f.read())
+        Path(temp_file.name).unlink()
+    else:
+        Path(temp_file.name).rename(rendered_output)
```
christinaholtNOAA, line 125: Why not use "writeable" here to stay consistent with other "dump" methods (untested):

```
with open(temp_file.name, "r", encoding="utf-8") as f:
    content = f.read()
    with writable(rendered_output):
        print(content, file=f)
```

I think that may also help with the context manager and will clean up once this is done.

src/uwtools/rocoto.py
```diff
@@ -134,20 +133,14 @@ def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
...
-    log_method("%s Rocoto validation error%s found", len(errors), "" if len(errors) == 1 else "s")
-    for line in errors:
-        log.error(line)
-
-    return success
+    valid = schema.validate(tree)
+    nerr = len(schema.error_log)
+    log_method = log.info if valid else log.error
+    log_method("%s Rocoto validation error%s found", nerr, "" if nerr == 1 else "s")
+    for err in list(schema.error_log):
```
christinaholtNOAA, line 144: Is `list` needed here? `schema.error_log` is already an iterable, no?

src/uwtools/rocoto.py
```diff
@@ -106,25 +106,24 @@ def realize_rocoto_xml(
...
-            # Validate the XML.
-            if validate_rocoto_xml(input_xml=temp_file.name):
-                # If no issues were detected, save temp file and report success.
-                shutil.move(temp_file.name, rendered_output)
-                return True
+    if not validate_yaml(config_file=config_file, schema_file=_rocoto_schema_yaml()):
+        log.error("YAML validation errors identified in %s", config_file)
+        return False
+
+    temp_file = tempfile.NamedTemporaryFile(delete=False)  # pylint: disable=consider-using-with
```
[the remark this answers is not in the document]
maddenp-noaa: I contributed some of these lines so can comment.

The context manager automatically deletes the temporary file when its block ends. We could potentially move the statements below inside the block (I assume that the auto-delete won't complain if the file has already disappeared having been moved/renamed, but we'd have to check). But maybe we should use `mkstemp()` instead, which doesn't provide a context manager, so that we wouldn't need this ugly pylint suppressor. I'll work with @WeirAE to see.

src/uwtools/rocoto.py
WeirAE: Resolved with #325

src/uwtools/rocoto.py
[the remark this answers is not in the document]
maddenp-noaa: I don't think the context manager will work: It _will_ eventually exit and remove the temp file, and we want the temp file to remain on disk if it fails XML validation so that the user can inspect it.

src/uwtools/config/j2template.py
```diff
@@ -43,15 +43,15 @@ def __init__(
 
     # Public methods
 
-    def dump(self, output_path: DefinitePath) -> None:
+    def dump(self, output_path: OptionalPath) -> None:
```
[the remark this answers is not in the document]
WeirAE: Oh fair point, I agree!

src/uwtools/rocoto.py
```diff
@@ -76,7 +76,7 @@ def _rocoto_template_xml() -> DefinitePath:
 
 def _write_rocoto_xml(
     config_file: OptionalPath,
-    rendered_output: DefinitePath,
+    rendered_output: OptionalPath,
```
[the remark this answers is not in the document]
WeirAE: Hmm, while I get 'rendered' as a descriptor since we're creating it from a template, you have a fair point that its the path to the file that we're actually using. I think this would be a clearer change

src/uwtools/rocoto.py
WeirAE: Hmm - from the [discussion](https://github.com/ufs-community/workflow-tools/pull/324/#discussion_r1370785648) above with @maddenp-noaa that led to this, it made sense at the time to leave the writeable and context manager out for simplicity, although its a fair point to maintain consistency with other functions. It should work fine either way

src/uwtools/rocoto.py
WeirAE: Nope, turns out its non-iterable too, annoyingly

src/uwtools/rocoto.py
maddenp-noaa: Expanding on what I mentioned in the comment I made a few minutes ago, I think we cannot use the context manager because, if XML validation fails, the file needs to be left on disk for the user to inspect, and the context manager will always remove it. We could perhaps rename it to prevent the context manager from deleting it (if that works), but the whole point of using these methods / classes from Python's `tempfile` package is to ensure that a new temporary file does not conflict with an existing file, so it's not safe to arbitrarily rename files. We should only use filenames that the user supplies (if something gets clobbered, that's on them), or techniques that guarantee that we won't clobber something (like the `tempfile` tools do).

I think we could do, say,
```
with open(temp_file, "r", encoding="utf-8") as f_in:
    with writable(rendered_output) as f_out:
        print(f_in.read(), file=f_out)
Path(temp_file).unlink()
```
but we'd be opening, reading, and writing to create a file identical to `temp_file`, which we could just rename. I do like the simplicity and clarity of this approach, and maybe doing some tiny unnecessary file IO is worth it for that. Thoughts?

WeirAE: My preference is to keep it simpler, given the choice

maddenp-noaa: Yes, I think I prefer the simpler version, too. Trying to avoid the extra file IO is premature optimization. So, I think the code block I suggested above is worth a try.

christinaholtNOAA: Sure thing. It's nice to have the discussion of pros/cons of alternatives.

src/uwtools/rocoto.py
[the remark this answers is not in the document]
WeirAE: `mkstemp()` version has been added in the last push, with fixed tests - I like this best of the 3

src/uwtools/rocoto.py
WeirAE: Reopened this conversation as we're having issues in `validate`

WeirAE: solved by reverting to `etree.fromstring()`

maddenp-noaa (approved):
Looks good! I tried various invocations of `uw rocoto realize` and `uw rocoto validate`, mixing usage of `-i` / `-o` and `stdin` / `stdout` streaming and everything worked as expected. We can now do e.g.
```
% uw rocoto realize -i src/uwtools/tests/fixtures/hello_workflow.yaml | head
[2023-10-26T18:20:03]     INFO 0 schema-validation errors found
[2023-10-26T18:20:03]     INFO 0 Rocoto validation errors found
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE workflow [
  <!ENTITY ACCOUNT "myaccount">
  <!ENTITY FOO "test.log">

]>
<workflow realtime="False" scheduler="slurm" >
  <cycledef group="howdy">202209290000 202209300000 06:00:00</cycledef>

  <log>/some/path/to/&FOO;</log>
```
as well as
```
% cat src/uwtools/tests/fixtures/hello_workflow.yaml | uw rocoto realize | head
[2023-10-26T18:20:33]     INFO 0 schema-validation errors found
[2023-10-26T18:20:33]     INFO 0 Rocoto validation errors found
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE workflow [
  <!ENTITY ACCOUNT "myaccount">
  <!ENTITY FOO "test.log">

]>
<workflow realtime="False" scheduler="slurm" >
  <cycledef group="howdy">202209290000 202209300000 06:00:00</cycledef>

  <log>/some/path/to/&FOO;</log>
```
and can pipeline invocations together like
```
% cat src/uwtools/tests/fixtures/hello_workflow.yaml | uw rocoto realize -q | uw rocoto validate
[2023-10-26T18:21:03]     INFO 0 Rocoto validation errors found
```

christinaholtNOAA (approved):
Nice work!

PR #325, opened 2023-10-25
rereadable-stdin

maddenp-noaa:
Two small things in this PR to assist with [UW-413](https://jira.epic.oarcloud.noaa.gov/browse/UW-413):

1. Address [UW-414](https://jira.epic.oarcloud.noaa.gov/browse/UW-414) by removing `_bad_paths()` and updating associated code, and
2. Supporting multiple reads of the same data from `stdin`.

In re item 2., although it's probably not optimal design, UW currently reads the same config file multiple times. Best practice would be to minimize file IO and re-use already-read in-memory data, but that wouldn't be easy to address quickly. (But we should keep it in mind as a potential design goal going forward IMO).

However, stdin is a stream that, once read, is exhausted and cannot be read again. In `realize_rocoto_xml()` in module `uwtools.rocoto`, we currently read the config (file) once via `validate_yaml()`, then again via the `YAMLConfig(input_yaml)` call in `_add_jobname_to_tasks()`, which is called by `_write_rocoto_xml()`. That's fine, if not optimal, with on-disk files, but doesn't work when reading from a stream. Ideally, we'd read the stream into a string once and access the string as needed, but that might require some significant re-plumbing. For now, provide a proxy in front of stdin that permits the data read from that stream to be re-read as often as needed.

### DIFF
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -37,7 +37,6 @@ disable = [
   "too-many-instance-attributes",
   "too-many-lines",
   "too-many-locals",
-  "too-many-statements",
   "unnecessary-lambda-assignment",
   "use-dict-literal",
 ]
--- a/src/uwtools/tests/utils/test_file.py
+++ b/src/uwtools/tests/utils/test_file.py
@@ -1,10 +1,11 @@
-# pylint: disable=missing-function-docstring,redefined-outer-name
+# pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
 """
 Tests for uwtools.utils.file module.
 """
 
 import sys
 from datetime import datetime as dt
+from io import StringIO
 from unittest.mock import patch
 
 import pytest
@@ -22,6 +23,35 @@ def assets(tmp_path):
     renamed = rundir.parent / ("rundir_%s" % now.strftime("%Y%m%d_%H%M%S"))
     assert not renamed.is_dir()
     return now, renamed, rundir
+
+
+def test_StdinProxy():
+    msg = "proxying stdin"
+    with patch.object(sys, "stdin", new=StringIO(msg)):
+        assert sys.stdin.read() == msg
+        # Reading from stdin a second time yields no input, as the stream has been exhausted:
+        assert sys.stdin.read() == ""
+    with patch.object(sys, "stdin", new=StringIO(msg)):
+        sp = file.StdinProxy()
+        assert sp.read() == msg
+        # But the stdin proxy can be read multiple times:
+        assert sp.read() == msg
+
+
+def test__stdinproxy():
+    file._stdinproxy.cache_clear()
+    msg0 = "hello world"
+    msg1 = "bonjour monde"
+    # Unsurprisingly, the first read from stdin finds the expected message:
+    with patch.object(sys, "stdin", new=StringIO(msg0)):
+        assert file._stdinproxy().read() == msg0
+    # But after re-patching stdin with a new message, a second read returns the old message:
+    with patch.object(sys, "stdin", new=StringIO(msg1)):
+        assert file._stdinproxy().read() == msg0  # <-- the OLD message
+    # However, if the cache is cleared, the second message is then read:
+    file._stdinproxy.cache_clear()
+    with patch.object(sys, "stdin", new=StringIO(msg1)):
+        assert file._stdinproxy().read() == msg1  # <-- the NEW message
 
 
 def test_get_file_type():
--- a/src/uwtools/tests/utils/test_file.py
+++ b/src/uwtools/tests/utils/test_file.py
@@ -103,7 +103,7 @@ def test_readable_file(tmp_path):
 
 def test_readable_nofile():
     with file.readable() as f:
-        assert f is sys.stdin
+        assert hasattr(f, "read")
 
 
 def test_writable_file(tmp_path):
--- a/src/uwtools/utils/file.py
+++ b/src/uwtools/utils/file.py
@@ -8,8 +8,10 @@ import sys
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime as dt
+from functools import cache
+from io import StringIO
 from pathlib import Path
-from typing import IO, Generator
+from typing import IO, Any, Generator, Union
 
 from uwtools.logging import log
 from uwtools.types import DefinitePath, OptionalPath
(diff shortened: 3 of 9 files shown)


src/uwtools/tests/utils/test_file.py
maddenp-noaa, line 1: New and updated tests related to `StdinProxy`.

src/uwtools/tests/utils/test_file.py
```diff
@@ -103,7 +133,7 @@ def test_readable_file(tmp_path):
 
 def test_readable_nofile():
     with file.readable() as f:
-        assert f is sys.stdin
+        assert hasattr(f, "read")
```
maddenp-noaa, line 136: The object yielded from `readable()` will no longer **be** `stdin`, but will be `read()`-able.

src/uwtools/utils/file.py
```diff
@@ -48,6 +50,33 @@ class _FORMAT:
...
+    Reads stdin once but permits multiple reads of its data.
+    """
+
+    def __init__(self) -> None:
+        self._stdin = sys.stdin.read()
+        self._reset()
+
+    def __getattr__(self, attr: str) -> Any:
+        self._reset()
+        return getattr(self._stringio, attr)
```
maddenp-noaa, line 64: `__getattr__()` delegates attribute accesses (e.g. `.read()`) to the `StringIO` object referred to by `self._stringio`. `self._reset()` is called first, though, to ensure that the `StringIO` object is reset to its pristine state -- in particular, that it is ready to `.read()` again.

src/uwtools/utils/file.py
```diff
@@ -48,6 +50,33 @@ class _FORMAT:
...
+
+    def __iter__(self):
+        self._reset()
+        for line in self._stringio.read().split("\n"):
+            yield line
```
maddenp-noaa, line 69: This supports e.g. iteration over the object yielded from `readable()`, e.g.
```
with readable(path) as f:
    for line in f:
        print(f)
```

src/uwtools/utils/file.py
WeirAE: Much more effective than what I was hammering away at, we'll need this again for sure

WeirAE (approved):
Thanks for going in and fixing all of this - we can use this for upcoming work and refactor some other functions to use this. LGTM!

src/pyproject.toml
```diff
@@ -37,7 +37,6 @@ disable = [
   "too-many-instance-attributes",
   "too-many-lines",
   "too-many-locals",
-  "too-many-statements",
```
maddenp-noaa, line 40: This is no longer actively suppressing any `pylint` complaints.

src/uwtools/tests/utils/test_file.py
```diff
@@ -24,6 +25,35 @@ def assets(tmp_path):
...
 
+def test_StdinProxy():
+    msg = "proxying stdin"
+    with patch.object(sys, "stdin", new=StringIO(msg)):
+        assert sys.stdin.read() == msg
+        # Reading from stdin a second time yields no input, as the stream has been exhausted:
+        assert sys.stdin.read() == ""
+    with patch.object(sys, "stdin", new=StringIO(msg)):
+        sp = file.StdinProxy()
+        assert sp.read() == msg
```
NaureenBharwaniNOAA, line 36: Just want to confirm my understanding on these lines here. Using ```sys.stdin.read()``` causes the standard input to only be able to be read once, so when we read it again there is nothing to call here. Meaning it clears the cache after one use?

Once we use the ```StdinProxy()``` class this makes it so we can read the standard input multiple times hence why we have to clear the cache when we run each this each time i.e. in ```test_core.py```

src/uwtools/tests/utils/test_file.py
```diff
@@ -24,6 +25,35 @@ def assets(tmp_path):
...
+    # Unsurprisingly, the first read from stdin finds the expected message:
+    with patch.object(sys, "stdin", new=StringIO(msg0)):
+        assert file._stdinproxy().read() == msg0
+    # But after re-patching stdin with a new message, a second read returns the old message:
+    with patch.object(sys, "stdin", new=StringIO(msg1)):
+        assert file._stdinproxy().read() == msg0  # <-- the OLD message
+    # However, if the cache is cleared, the second message is then read:
+    file._stdinproxy.cache_clear()
+    with patch.object(sys, "stdin", new=StringIO(msg1)):
+        assert file._stdinproxy().read() == msg1  # <-- the NEW message
```
NaureenBharwaniNOAA, line 54: Super interesting use, comments were definitely useful here to understand what was happening in these lines of code.

NaureenBharwaniNOAA (approved):
I like this version better than using ```check_paths``` or ```bad_paths```, which would make it harder to expand on later as our code base gets larger. 

Looks good to me!

src/uwtools/tests/utils/test_file.py
```diff
@@ -24,6 +25,35 @@ def assets(tmp_path):
...
+        # Reading from stdin a second time yields no input, as the stream has been exhausted:
+        assert sys.stdin.read() == ""
+    with patch.object(sys, "stdin", new=StringIO(msg)):
+        sp = file.StdinProxy()
+        assert sp.read() == msg
+        # But the stdin proxy can be read multiple times:
+        assert sp.read() == msg
+
+
+def test__stdinproxy():
```
christinaholtNOAA, line 41: I really enjoy the commenting style in this test.

christinaholtNOAA (approved):
Nice addition. LGTM.

src/uwtools/tests/utils/test_file.py
maddenp-noaa: To be pedantic, using `sys.stdin.read()` doesn't _cause_ the standard input to only be able to be read once; the standard input is a stream which, be definition, can only be read once, regardless of what you do with it. You can think of it as a queue, with data being pushed on one end by some other process (e.g. by the `cat` command in something like `cat rocoto.xml | uw rocoto validate`, but it could also be a human typing on the keyboard), and then being popped off the other side (by `uw` reading it). But, yes, when we read it again, there's nothing there -- we've already read everything with the first `.read()`. No caching is involved in any of that.

The `StdinProxy` object reads (and exhausts) that stream one time, as usual, but stores it as a string in an instance variable. When the holder of a `StdinProxy` object calls `.read()` on it, `StdinProxy` streams that string to the caller in a way (mostly) indistinguishable from a real, live read from `stdin`. But users aren't expected to interact with `StdinProxy` objects directly; `readable()` handles that by calling the `_stdinproxy()` function, and it's this function that caches a single `StdinProxy` object, and its cache that needs to be reset so that a new `StdinProxy` object, which reads a new (mocked) `stdin` stream created by a unit test, can be created. Clearing the cache shouldn't be necessary during normal `uwtools` use, as we wouldn't normally read `stdin` more than once. But the unit tests need to.

Issue #318, opened 2023-10-10
Forecast run method needs refactor.


christinaholtNOAA:
I find this method confusing, possibly due to it trying to do too many things. There are IMO too many conditionals on `self._dry_run`, and "dry run" is hardly dry -- it creates files and directories on disk, it just doesn't submit a batch job. I'm confused by `self.batch_script()` vs `self._batch_script (reminds me of my `run` problem earlier today).

I may just have DAGs on the brain, but it's kind of making me want that technology here: If a user just wanted to see the run command, they should be able to invoke a task that provides just that, rather than pretending to do a run but doing it in dry-run mode, and getting the command printed on `stdout` but also getting a populated run directory that they don't want. We're not ready for that, though.

Why does the value returned by `run_cmd()` need to be augmented? If the `pre_run` information is necessary for running the executable, shouldn't it be part of what `run_cmd()` returns? Is what `run_cmd()` returns actually useful on its own? Honest question, I can't think of an answer right now.

I feel like this could probably be refactored just in OO style to be cleaner and more single-responsibility focused, but I don't think we should do it in this PR if what we have here works. I just wanted to share that I'm feeling some tech debt here. Please don't change anything based on these comments.

_Originally posted by @maddenp-noaa in https://github.com/ufs-community/workflow-tools/pull/291#discussion_r1346747329_

PR #327, opened 2023-10-30
UW-408: Refactor of forecast's run method

elcarpenterNOAA:
The refactor of the run method includes:
- adding more logging messages when using `dry_run`
- replacing many os.path calls with pathlib calls instead
- delegating the work within the original `run()` into either `_run_via_batch_submission` or `_run_via_local_execution`
- testing for `_run_via_batch_submission` and `_run_via_local_execution`
- updated test for `run` method

Issue #318


src/uwtools/drivers/forecast.py
```diff
@@ -77,14 +82,20 @@ def create_directory_structure(run_directory: DefinitePath, exist_act: str = "de
...
 
         for subdir in ("INPUT", "RESTART"):
             path = os.path.join(run_directory, subdir)
-            log.info("Creating directory: %s", path)
-            os.makedirs(path)
+            if dry_run:
+                log.info("Would create directory: %s", path)
+            else:
+                log.info("Creating directory: %s", path)
+                os.makedirs(path)
```
maddenp-noaa, line 98: nitpick: In `stage_files()`, earlier, we log an after-the-fact message about an action we're taking, while here we log a before-the-fact message. It would probably be a nice design to be consistent about this. I think before-the-fact messaging is helpful because, if the operation fails (e.g. throws and exception that halts the program), we've just logged something about the thing that failed; while an after-the-fact message won't appear at all and the user might be left trying to understand what the program was doing when it failed.

src/uwtools/drivers/forecast.py
```diff
@@ -231,14 +231,47 @@ def _define_boundary_files(self) -> Dict[str, str]:
...
+        """
+        batch_script = self.batch_script()
+        assert self._batch_script is not None
+        outpath = Path(run_directory) / self._batch_script
+
+        if self._dry_run:
+            # Apply switch to allow user to view the run command of config.
+            # This will not run the job.
+            log.info("Batch Script:")
+            batch_script.dump(None)
```
maddenp-noaa, line 246: Here, too, we should log the lines of the batch script instead of printing them to `stdout`. I also notice that we call the batch script `bs` elsewhere, and `batch_script` here, so those could be harmonized. Better yet: Can we factor this "log the lines of a batch script in dry-run mode" logic out? Then everything would be in sync.

src/uwtools/drivers/forecast.py
```diff
@@ -231,14 +231,47 @@ def _define_boundary_files(self) -> Dict[str, str]:
 
         return boundary_files
 
+    def _prepare_and_run_batch_script(self, run_directory: str) -> bool:
```
maddenp-noaa, line 234: I learned from _Clean Code_ that the presence of the word "and" in a function/method name might indicate that the function/method could usefully be split into two. These do seem like potentially two distinct operations that don't necessarily belong together: It seems like it might be useful to create a batch script without submitting it, or to submit a batch script that we already have for whatever reason. I'm ok with not changing this in this PR, but it does seem like a split might make sense.

src/uwtools/drivers/forecast.py
```diff
@@ -231,14 +231,47 @@ def _define_boundary_files(self) -> Dict[str, str]:
...
+
+    def _prepare_and_execute_command(self) -> bool:
+        """
+        Collects the necessary MPI environment variables in order to construct full run command.
+        """
+        pre_run = self._mpi_env_variables(" ")
+        full_cmd = f"{pre_run} {self.run_cmd()}"
+        if self._dry_run:
+            log.info("Would run: ")
+            print(full_cmd, file=sys.stdout)
```
maddenp-noaa, line 272: This should also be logged instead of `print()`ed in dry-run mode. And no need for the space after the colon in the logged string in the preceding command: There will be a newline there anyway.

src/uwtools/drivers/forecast.py
```diff
@@ -231,14 +231,47 @@ def _define_boundary_files(self) -> Dict[str, str]:
...
-        self.create_namelist(run_directory / "input.nml")
+        if self._dry_run:
+            for call in ("field_table", "model_configure", "input.nml"):
+                log.info(f"Would prepare: {run_directory}/{call}")
+        else:
+            self.create_field_table(run_directory / "field_table")
+            self.create_model_configure(run_directory / "model_configure")
+            self.create_namelist(run_directory / "input.nml")
+
+    def _prepare_and_execute_command(self) -> bool:
```
maddenp-noaa, line 264: This PR adds some new methods, so we should write targeted unit tests for those. Even if we have 100% coverage due to the fact that these methods get called by other methods that we _do_ have targeted test for, each method should really have at least one test that calls it directly. See e.g. the tests recently added for the `_RocotoXML` class: There's a test, for each method in the class, that explicitly, directly invokes that method, independent of anything else.

maddenp-noaa (changes requested):

maddenp-noaa:
I made some comments / requested some changes -- I'd be happy to pair on any of this! I also wasn't sure (because I am lazy and didn't check) whether the questions in the PR description are still pertinent. I'll circle back to that, maybe in our team meet this morning.

src/uwtools/drivers/forecast.py
```diff
@@ -50,10 +50,15 @@ def batch_script(self) -> BatchScript:
         bs = self.scheduler.batch_script
         bs.append(pre_run)
         bs.append(self.run_cmd())
+        if self._dry_run:
+            log.info("Batch Script: ")
+            bs.dump(None)
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: Thanks for this comment! It makes it much easier to understand the differences between ```stdout```, ```stdin```, and ```stderr```

src/uwtools/drivers/forecast.py
```diff
@@ -152,40 +180,12 @@ def run(self, cycle: datetime) -> bool:
...
-        pre_run = self._mpi_env_variables(" ")
-        full_cmd = f"{pre_run} {self.run_cmd()}"
-        if self._dry_run:
-            log.info("Would run: ")
-            print(full_cmd, file=sys.stdout)
-            return True
-
-        result = execute(cmd=full_cmd)
-        return result.success
+        return self._prepare_and_execute_command()
```
christinaholtNOAA, line 188: This method cleaned up pretty nice!

src/uwtools/drivers/forecast.py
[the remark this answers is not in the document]
elcarpenterNOAA: I've tried updating the respective dry run test using `assert logged(caplog, run_expected)` and I'm struggling to see why it's failing at the moment. I've pushed the most recent changes with the failing test.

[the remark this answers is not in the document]
maddenp-noaa: I'll fetch your branch and take a look -- more soon.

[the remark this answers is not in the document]
maddenp-noaa: I think what's happening is that `logged()` determines whether its `msg` argument appears in _some_ line of logged output (in some log "record" in logging parlance). But `run_expected` is a big string with _all_ the lines of logged output, joined with newline characters.

This works for me:
```
% git diff -U0
diff --git a/src/uwtools/tests/drivers/test_forecast.py b/src/uwtools/tests/drivers/test_forecast.py
index 125b8e0..1b9b7bb 100644
--- a/src/uwtools/tests/drivers/test_forecast.py
+++ b/src/uwtools/tests/drivers/test_forecast.py
@@ -347 +347 @@ def test_FV3Forecast_run_dry_run(caplog, fv3_mpi_assets, fv3_run_assets, with_ba
-        run_expected = "\n".join(batch_components)
+        run_expected = batch_components
@@ -350 +350 @@ def test_FV3Forecast_run_dry_run(caplog, fv3_mpi_assets, fv3_run_assets, with_ba
-        run_expected = " ".join(fv3_mpi_assets)
+        run_expected = [" ".join(fv3_mpi_assets)]
@@ -356,2 +356,2 @@ def test_FV3Forecast_run_dry_run(caplog, fv3_mpi_assets, fv3_run_assets, with_ba
-    assert logged(caplog, run_expected)
-    # assert run_expected in capsys.readouterr().out
+    for line in run_expected:
+        assert logged(caplog, line)
```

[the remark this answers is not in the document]
maddenp-noaa: Maybe `expected_lines` would be a good name instead of `run_expected`, to emphasize the fact that it should contain plural lines, suggesting a `list` object.

elcarpenterNOAA:
@maddenp-noaa Currently having issues with the local execution test with the current failure message below. What's the best work around for not having srun locally? 

> -------------------------------------------- Captured log call --------------------------------------------
INFO     root:processing.py:30 Executing: KMP_AFFINITY=scatter OMP_NUM_THREADS=1 OMP_STACKSIZE=512m MPI_TYPE_DEPTH=20 ESMF_RUNTIME_COMPLIANCECHECK=OFF:depth=4 srun --export=NONE test_exec.py
ERROR    root:processing.py:45   Failed with status: 127
ERROR    root:processing.py:49   Output:
ERROR    root:processing.py:51     /bin/sh: srun: command not found
ERROR    root:processing.py:51
=================================== short test summary info ===================================
FAILED uwtools/tests/drivers/test_forecast.py::test__run_via_local_execution - AssertionError: Expected 'execute' to be called once. Called 0 times.

src/uwtools/drivers/forecast.py
elcarpenterNOAA: I've reordered the log messages in stage_files to make them before-the-fact, but if I've missed something, please let me know.

maddenp-noaa:
> What's the best work around for not having srun locally?

That looks like it's actually trying to execute a command in a subprocess. I think the choices are either to mock the function that provides the command to run and have it run something harmless like `/bin/true`, or mock `execute()` so that the real one doesn't get called, and make assertions stating that the mocked `execute()` was called with the expected arguments. The mocked `execute()` can be forced to return either `True` or `False` if we need to test both cases (but maybe we don't), or maybe just `True` to simulate a successful command execution. Let me know if you want to pair -- mocks can be a little tricky if you haven't worked much with them.

src/uwtools/drivers/driver.py
```diff
@@ -81,7 +81,7 @@ def run_cmd(self) -> str:
         """
```
elcarpenterNOAA, line 81: Does anyone have a better name suggestion for `run_cmd`? I updated the return statement to not be as misleading, but not sure if it's best to just leave it be for now.

maddenp-noaa: I think we should leave it for now. I'm having a hard time thinking of something better. It's kind of true that this is the command to run the program itself -- I think the other items that get prepended to the command are just environment-variable settings and not really part of the command per se. I mean, for example, that there's no real difference between
```
FOO=88 BAR=99 mpirun -n 8 program
```
and
```
export FOO=88
export BAR=99
mpirun -n 8 program
```
The environment variables can be set on the same line as the program invocation, or separately, so they're kind of not really part of the command, if that makes sense. So maybe `run_cmd()` isn't so inaccurate after all. That said, I **do** think your docstring update is an improvement.

While we're here, though, could you change the body of this function as follows and run the tests to see if it's ok?
```
components = [
    self._platform_config.get("mpicmd"),
    *self._config["runtime_info"].get("mpi_args"),
    self._config["exec_name"],
]
return " ".join(filter(None, components))
```
Things I think are bad with the current body:

1. A local variable called `run_cmd` has the same name as the method, and isn't actually the run command at all.
2. A local variable called `run_time_args` is set to the value of something called `mpi_args`.
3. The local variables are unnecessarily set in an order different from how they appear in the final command.
4. If `mpicmd` or `mpi_args` don't exist in the config, the final command will contain anomalous whitespace due to empty-string values.

This change eliminates all those issues. (I still don't like that `mpicmd` isn't `mpi_cmd` to match `mpi_args` in the config, but that's something for another day.)

Could you see if that works? If it's any trouble, just ignore it and I'll fix it in a separate PR.

elcarpenterNOAA: It did not work because there are integers in the command, not just strings.

maddenp-noaa: I see. Well, let's punt on that. It's a fairly easy fix, but I'll implement it elsewhere. Sorry for the diversion.

WeirAE: Agreed - Honestly I struggled with this in the original version and didn't find anything better

src/uwtools/drivers/forecast.py
```diff
@@ -253,5 +246,51 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+        """
+        if self._dry_run:
+            for call in ("field_table", "model_configure", "input.nml"):
+                log.info(f"Would prepare: {run_directory}/{call}")
+        else:
+            self.create_field_table(run_directory / "field_table")
+            self.create_model_configure(run_directory / "model_configure")
+            self.create_namelist(run_directory / "input.nml")
+
+    def _run_via_batch_submission(self) -> Tuple[bool, List[str]]:
```
WeirAE, line 261: This is certainly a better way, nicely done

src/uwtools/drivers/forecast.py
```diff
@@ -151,40 +178,15 @@ def run(self, cycle: datetime) -> bool:
...
-
-            batch_script.dump(outpath)
-            return self.scheduler.submit_job(outpath)
-
-        pre_run = self._mpi_env_variables(" ")
-        full_cmd = f"{pre_run} {self.run_cmd()}"
+        if self._batch_script:
+            method = self._run_via_batch_submission()
+        else:
+            method = self._run_via_local_execution()
```
maddenp-noaa, line 184: `self._run_via_local_execution()` is an actual _call_ of the method, rather than just a reference to the method itself. So the value of that expression is not a "method", but whatever it is that that function returns. It's ok to do it this way, but then the LHS should probably not be `method` but `status, output`. Or it could be, say, `result` and then `status, output = method` would change to `status, output = result`.

But how about the more concise
```
method = self._run_via_batch_submission if self._batch_script else self._run_via_local_execution
status, output = method()
```
This sets the variable `method` to point to one of the two private methods, and then the expression `method()` performs the call and assigns the results of the call to the `status` and `output` variables.

Or we could have the single line
```
status, output = self._run_via_batch_submission() if self._batch_script else self._run_via_local_execution()
```
`black` will reformat that, since it's too long for one line, but that one statement should do.

src/uwtools/drivers/forecast.py
```diff
@@ -253,5 +246,51 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+
+        assert self._batch_script is not None
+        outpath = run_directory / self._batch_script
+
+        batch_script.dump(outpath)
+        return self.scheduler.submit_job(outpath), batch_lines
+
+    def _run_via_local_execution(self) -> Tuple[bool, List[str]]:
+        """
+        Collects the necessary MPI environment variables in order to construct full run command.
```
maddenp-noaa, line 282: It does more than this, right? It actually executes the command, unless we're in dry-run mode.

src/uwtools/drivers/forecast.py
```diff
@@ -253,5 +246,51 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+        else:
+            self.create_field_table(run_directory / "field_table")
+            self.create_model_configure(run_directory / "model_configure")
+            self.create_namelist(run_directory / "input.nml")
+
+    def _run_via_batch_submission(self) -> Tuple[bool, List[str]]:
+        """
+        Prepares and runs batch script.
+
+        :return: A tuple that contains a boolean of success status of the FV3 run and a list of
```
maddenp-noaa, line 265: Isn't the success status not that of the FV3 run but, rather, that of the batch-job submission?

I think "submits a batch script" would be a more familiar phrasing than "runs".

elcarpenterNOAA: The batch-job submission returns `result.success` of `execute()` though, so I agree the first sentence can be improved, but do you think it's all that inaccurate for the return still?

maddenp-noaa: I'd probably write
```
:return: A tuple containing the success status of submitting the job to the batch system, and a list of strings that make up the batch script.
```
Unless I'm confused about something, the `execute()` that gets called here runs e.g. `sbatch <script>` in a subshell, a command that would submit the batch script to the batch system. We'd be getting back the success status of that operation -- which could fail if e.g. the batch script contained bad information. But as for FV3 itself, it might be literally hours before it runs: The batch job may be queued for a long time. When it runs, _that_ script will run FV3, and it would have to record the success status of the actual FV3 run. That's not available to `uw` at this place and time in the code.

src/uwtools/drivers/forecast.py
```diff
@@ -253,5 +246,52 @@ def _mpi_env_variables(self, delimiter: str = " ") -> str:
...
+
+        batch_script.dump(outpath)
+        return self.scheduler.submit_job(outpath), batch_lines
+
+    def _run_via_local_execution(self) -> Tuple[bool, List[str]]:
+        """
+        Collects the necessary MPI environment variables in order to construct full run command,
+        then executes said command.
+
+        :return: A tuple containing a boolean of the success status of the FV3 run and a list of
```
maddenp-noaa, line 285: **This** one is accurate about the success status, because there's no batch system involved: It's running FV3 directly and immediately, and so would know about it's status.

maddenp-noaa (approved):
Nice work. Thanks for letting me pair on some of it.

elcarpenterNOAA:
@maddenp-noaa Thank you! I learned a lot this week from pairing with you.

WeirAE (approved):
Much better than I had left it, nicely done working with all of the feedback and making it work. LGTM

PR #328, opened 2023-10-31
post-uw-376-items

maddenp-noaa:
Three small updates:

1. I realized that it's not necessary to instantiate a `@dataclass` object, at least for our use case: The members of the class itself can be used directly, which leads to a minor simplification.
2. The `README.md` needed a couple of minor changes for accuracy.
3. Following up on a great question from @christinaholtNOAA in my previous PR, I found that it was indeed useful to organize the tests of class `uwtools.rocoto._RocotoXML` into a pytest class.

### DIFF
--- a/src/uwtools/tests/test_rocoto.py
+++ b/src/uwtools/tests/test_rocoto.py
@@ -66,26 +66,26 @@ class Test_RocotoXML:
     def root(self):
         return rocoto.Element("root")
 
-    def test__RocotoXML__doctype_entities(self, instance):
+    def test__doctype_entities(self, instance):
         assert '<!ENTITY ACCOUNT "myaccount">' in instance._doctype
         assert '<!ENTITY FOO "test.log">' in instance._doctype
 
-    def test__RocotoXML__doctype_entities_none(self, instance):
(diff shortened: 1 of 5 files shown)


src/uwtools/tests/test_rocoto.py
```diff
@@ -63,158 +52,160 @@ def test_validate_rocoto_xml(vals):
...
-    del instance._config["workflow"]["entities"]
-    assert instance._doctype is None
+    @fixture
+    def instance(self, assets):
+        cfgfile, _ = assets
+        return rocoto._RocotoXML(config_file=cfgfile)
 
+    @fixture
+    def root(self):
+        return rocoto.Element("root")
```
maddenp-noaa, line 67: ^^^ Here are the two relocated fixtures.

Below this line in this module, and despite the huge diff, nothing has actually changed except that 1. The method names are simplified, removing the `_RocotoXML` component, as that is now obvious due to their membership in this class; and 2. Each function signature now includes the mandatory `self` argument.

WeirAE (approved):
Super clean test changes, LGTM

Issue #329, opened 2023-10-31
UW-412: expand tests to also not provide optional args


WeirAE:
**Description**
We should probably someday expand the tests to also _not_ provide optional args.

**Requirements**
Tests demonstrate that `argparse` doesn't complain and that, eg. `args.input_file` is `None`

**Acceptance Criteria (Definition of Done)**
New tests pass with optional args set to None
existing tests pass

WeirAE:
Completed by #330

PR #330, opened 2023-10-31
UW-412: expand tests to also not provide optional args

WeirAE:
**Description**
Tests demonstrate that argparse doesn't complain and that, eg. args.input_file is None

Fixes issue #329 

**Type of change**

**How Has This Been Tested?**
- [ ] Tests on Ubuntu OS

**Checklist**

### DIFF
--- a/src/uwtools/tests/test_cli.py
+++ b/src/uwtools/tests/test_cli.py
@@ -388,9 +388,9 @@ def test__dispatch_template(params):
 
 def test__dispatch_template_render_no_optional():
     args, _ = cli._parse_args("uw template render".split()[1:])
-    with patch.object(cli.uwtools.config.templater, STR.render) as templater:
+    with patch.object(cli.uwtools.config.templater, "render") as render:
         cli._dispatch_template_render(args)
-    assert templater.called_once_with(args)
+    assert render.called_once_with(args)
 
 
 def test__dispatch_template_render_yaml():
@@ -406,9 +406,9 @@ def test__dispatch_template_render_yaml():
             STR.dryrun: 7,
         }
     )
-    with patch.object(cli.uwtools.config.templater, STR.render) as templater:
+    with patch.object(cli.uwtools.config.templater, "render") as render:
         cli._dispatch_template_render(args)
-    assert templater.called_once_with(args)
+    assert render.called_once_with(args)
 
 
 @pytest.mark.parametrize("quiet", [True])
--- a/src/uwtools/tests/test_cli.py
+++ b/src/uwtools/tests/test_cli.py
@@ -226,19 +226,7 @@ def test__dispatch_config_realize():
 
 
 def test__dispatch_config_realize_no_optional():
-    args = ns()
-    vars(args).update(
-        {
-            STR.infile: None,
-            STR.infmt: None,
-            STR.outfile: None,
-            STR.outfmt: None,
-            STR.valsfile: 5,
-            STR.valsfmt: None,
-            STR.valsneeded: None,
-            STR.dryrun: None,
-        }
-    )
+    args, _ = cli._parse_args("uw config realize --values-file /foo.vals".split()[1:])
     with patch.object(cli.uwtools.config.core, "realize_config") as realize_config:
         cli._dispatch_config_realize(args)
     assert realize_config.called_once_with(args)
@@ -261,16 +249,7 @@ def test__dispatch_config_translate_atparse_to_jinja2():
 
 
 def test__dispatch_config_translate_no_optional():
-    args = ns()
-    vars(args).update(
-        {
-            STR.infile: None,
-            STR.infmt: None,
-            STR.outfile: None,
-            STR.outfmt: None,
-            STR.dryrun: None,
-        }
-    )
+    args, _ = cli._parse_args("uw config translate".split()[1:])
     with patch.object(cli.uwtools.config.atparse_to_jinja2, "convert") as convert:
         cli._dispatch_config_translate(args)
     assert convert.called_once_with(args)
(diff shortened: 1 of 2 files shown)


WeirAE:
Capturing feedback from pair programming:
1) use of None was not the intended behavior, nor was the specific namespace uses
- converted to correct argparse expected input and reverted changes
2) calling STR.render and naming templater is inappropriate for the intended use
- calling "render" and naming the same consistently with other tests

src/uwtools/tests/test_cli.py
```diff
@@ -237,23 +253,37 @@ def test__dispatch_config_translate_arparse_to_jinja2():
...
+
 def test__dispatch_config_translate_unsupported():
     args = {STR.infile: 1, STR.infmt: "jpg", STR.outfile: 3, STR.outfmt: "png", STR.dryrun: 5}
     assert cli._dispatch_config_translate(args) is False
 
 
-def test__dispatch_config_validate_yaml():
-    args = {STR.infile: 1, STR.infmt: FORMAT.yaml, STR.schemafile: 3}
+def test__dispatch_config_validate_no_optional():
+    args, _ = cli._parse_args("uw config validate --schema-file /foo.schema".split()[1:])
```
WeirAE, line 269: The issue here appears to be [this conditional](https://github.com/ufs-community/workflow-tools/blob/ac5879ad2c863d57f7966532d1a0056e9a7facdb/src/uwtools/cli.py#L227C5-L227C5) on the file format. I'm not sure why `args[STR.infmt]` is not working but `args.input_format` worked before. 
For that matter, do we even want it to be confirming the format, or just accepting an override? If we do, then it should be required if input_file is provided, or perform the type check for the user instead of taking the option

maddenp-noaa: The line of code you've linked to is in `_dispatch_config_translate()`, not `_dispatch_config_validate()` under test here. Do you mean [this](https://github.com/ufs-community/workflow-tools/blob/ac5879ad2c863d57f7966532d1a0056e9a7facdb/src/uwtools/cli.py#L245)?

If so, it seems to me like we have to check that the format is YAML before we call `uwtools.config.validator.validate_yaml()`. That's currently the only data format we have a schema for (except Rocoto XML, but it's got its own CLI [sub]mode -- we could also support it here, but that's a question for another day).

Although the input-format CLI argument is technically optional from the perspective of `argparse`, there's a post-parse check that is supposed to ensure that, if an input file wasn't specified (i.e. we are reading `stdin`), the input format **is**, as it will otherwise be impossible to deduce the format from a non-existent filename extension.

Do you have an idea for changing this? I'm not sure I understood if you were proposing something.

WeirAE: AH sorry, you're right - I think this was my misunderstanding of everything `_check_file_vs_format` was doing and how we were breaking the format variable before. Its already doing my proposal of making it check the format if not provided, but it has to take a real path not a mocked `None`

src/uwtools/tests/test_cli.py
```diff
@@ -329,6 +388,29 @@ def test__dispatch_template(params):
     func.assert_called_once_with(args)
 
 
+def test__dispatch_template_render_no_optional():
+    args = {
+        STR.infile: None,
+        STR.outfile: None,
+        STR.valsfile: None,
+        STR.valsfmt: None,
+        STR.keyvalpairs: "",
```
christinaholtNOAA, line 397: Was keyvalpairs supposed to be an empty list when not provided?

src/uwtools/tests/test_cli.py
```diff
@@ -225,7 +225,16 @@ def test__dispatch_config_realize():
 
 
 def test__dispatch_config_realize_no_optional():
-    args, _ = cli._parse_args("uw config realize --values-file /foo.vals".split()[1:])
```
christinaholtNOAA: Out of curiosity, why did you decide to revert back to this method? We had some pretty good arguments going for supportability and readability earlier today for using the `_parse_args` approach.

christinaholtNOAA (approved):

src/uwtools/tests/test_cli.py
WeirAE: The parsing that creates STR.keyvalpairs would create an empty list if not given anything, so when going back to this args{} method requires defining it as empty since we don't allow it to accept None from talking earlier

src/uwtools/tests/test_cli.py
WeirAE: From an offline conversation, "In a test function, if we have
args, _ = cli._parse_args("uw config translate".split()[1:])
We're throwing away the checks by assigning them to _. We keep the args. uwtools.cli.main() keeps both and then runs the check functions, including _check_file_vs_format() in this case. But, if our test either calls _parse_args() and discards the second return value, or if we just construct an argument dict ourselves, the "format is optional if filename is specified, required otherwise" check never runs."

maddenp-noaa: I think maybe we're all synced up due to Slack conversation, but the upshot, from my perspective is: If we use `_parse_args`, we really should use `STR` values instead of re-introducing literal strings like `config`, `realize`, `values-file`, etc., and that all gets so messy that I think just constructing the `args` `dict` manually is IMO cleaner. That also avoids the possibility that `_parse_args()` has a bug that makes should-be-unrelated tests fail.  I tried `_parse_args()` out on my branch and didn't like where it was going, so I dropped that and told @WeirAE that I thought he should, too.

src/uwtools/tests/test_cli.py
maddenp-noaa: `[]` is also empty and would be more natural to see here since that's what `argparse` would provide. We do often provide bogus and not even implausible values in tests when they'll only be received by a mock, but in this case I think that `_dict_from_key_eq_val_strings()` will actually receive and operate on this `""` value and, though it will _work_ since `""` is both iterable and empty, I think it might surprise future readers, while `[]` should be more expected.

maddenp-noaa (approved):
I'd like to see that `""` -> `[]` argument change, but other than that this looks good to me.

src/uwtools/tests/test_cli.py
WeirAE: Interestingly I had `[]` originally but it gives `error: Need type annotation for "args"`. The typing for `config_items` in `_dict_from_key_eq_val_strings` is `List[str]`, even though the docstring reads "Strings in the form key=value"

maddenp-noaa: `args: dict = {` should do it.

WeirAE: Interesting, so then the linter is mad because its not understanding the implicit typing for `[]` ? Thanks, I'll have to file that away for future reference

PR #331, opened 2023-10-31
cli-test-fixup

maddenp-noaa:
Fixes per group discussion today. See inline comments.

### DIFF
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -42,9 +42,9 @@ def main() -> None:
     setup_logging(quiet=True)
     try:
         args, checks = _parse_args(sys.argv[1:])
-        for check in checks[args.mode][args.submode]:
+        for check in checks[vars(args)[STR.mode]][vars(args)[STR.submode]]:
             check(args)
-        setup_logging(quiet=args.quiet, verbose=args.verbose)
+        setup_logging(quiet=vars(args)[STR.quiet], verbose=vars(args)[STR.verbose])
         log.debug("Command: %s %s", Path(sys.argv[0]).name, " ".join(sys.argv[1:]))
         modes = {
             STR.config: _dispatch_config,
@@ -52,7 +52,7 @@ def main() -> None:
             STR.rocoto: _dispatch_rocoto,
             STR.template: _dispatch_template,
         }
-        sys.exit(0 if modes[args.mode](args) else 1)
+        sys.exit(0 if modes[vars(args)[STR.mode]](args) else 1)
     except Exception as e:  # pylint: disable=broad-exception-caught
         _abort(str(e))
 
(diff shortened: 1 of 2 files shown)


src/uwtools/cli.py
```diff
@@ -42,17 +42,17 @@ def main() -> None:
     setup_logging(quiet=True)
     try:
         args, checks = _parse_args(sys.argv[1:])
-        for check in checks[args.mode][args.submode]:
+        for check in checks[vars(args)[STR.mode]][vars(args)[STR.submode]]:
```
maddenp-noaa, line 45: Replace references to `args` members where the expression uses the equivalent of an uncheckable string literal with a reference to the `STR` object, which can be verified by the linter. Many repetitions of this pattern follow.

src/uwtools/cli.py
```diff
@@ -479,7 +481,7 @@ def _add_arg_config_file(group: Group) -> None:
 
 def _add_arg_cycle(group: Group) -> None:
     group.add_argument(
-        "--cycle",
+        _switch(STR.cycle),
```
maddenp-noaa, line 484: A string literal slipped in here -- relocated to `STR`.

src/uwtools/cli.py
```diff
@@ -298,12 +300,12 @@ def _dispatch_forecast_run(args: Namespace) -> bool:
 
     :param args: Parsed command-line args.
     """
-    forecast_class = uwtools.drivers.forecast.CLASSES[args.forecast_model]
+    forecast_class = uwtools.drivers.forecast.CLASSES[vars(args)[STR.model]]
```
maddenp-noaa, line 303: I think `forecast_model` was replaced with `model` somewhere along the line but this didn't get updated. Fix, and fixed corresponding unit test.

src/uwtools/tests/test_cli.py
```diff
@@ -195,15 +195,20 @@ def test__dispatch_config(params):
...
     vars(args).update({STR.file1path: 1, STR.file1fmt: 2, STR.file2path: 3, STR.file2fmt: 4})
     with patch.object(cli.uwtools.config.core, "compare_configs") as compare_configs:
         cli._dispatch_config_compare(args)
-    assert compare_configs.called_once_with(args)
+    compare_configs.assert_called_once_with(
+        config_a_path=args.file_1_path,
+        config_a_format=args.file_1_format,
+        config_b_path=args.file_2_path,
+        config_b_format=args.file_2_format,
+    )
```
maddenp-noaa, line 211: Fixed several instances of improper assertions:
1. Replaced `assert` statements with invocations of mock objects' `asssert_called_once_with()` methods.
2. With the previous change in place, tests (correctly) started failing, so updated expected call arguments with the correct values.

This pattern repeats throughout the remainder of this module.

WeirAE (approved):
Whew, there were even more than expected, I'm glad we took the time to look through everything - LGTM

src/uwtools/cli.py
christinaholtNOAA: 😬 My bad!

christinaholtNOAA (approved):
Thanks for the quick turnaround on the tedious work. Looks good.

christinaholtNOAA (approved):
Even better!

PR #332, opened 2023-11-01
UW-365

maddenp-noaa:
**Description**

- Remove the GitHub Actions workflow for creating local documentation that was apparently never published. We will focus on our ReadTheDocs based documentation and related tooling.
- Add a ReadTheDocs build-status badge to `README.md`, and pull the _Documentation_ section near the top for more obvious access.
- Remove obsolete links from the bottom of `README.md`.
- Rename the `Make Package` Actions workflow to `Test`, as that's what we actually use it for: To proactively test our proposed merged to the `develop` branch. The fact that the tests are accomplished by making a package is an implementation detail. At one point I think I thought that we would extend this workflow to support releases, but we implemented a separate `Release` workflow specifically for that purpose. Given all that, I think the name `Test` makes the workflow's purpose clearer.
- Remove `src/uwtools/files` and related tests. We can mine the git history for any of this code -- none of which we currently use, or which is accessible from any code exposed to end users -- if we need it later.
- Clean up various `__init__.py` files and update `import` statements to import the packages at their actual paths. There are cases where `import`ing packages, modules, classes, functions, etc., into `__init__.py` to create a more convenient namespace for users makes sense, but `uwtools` probably doesn't need that overhead currently.

**Type of change**

- [x] Clean up

**How Has This Been Tested?**

`make test` locally + Actions.

**Checklist**


README.md
```diff
@@ -1,10 +1,15 @@
-[![Documentation](https://github.com/ufs-community/workflow-tools/actions/workflows/docs.yaml/badge.svg)](https://github.com/ufs-community/workflow-tools/actions/workflows/docs.yaml)
+[![docs](https://readthedocs.org/projects/unified-workflow/badge/?version=latest)](https://readthedocs.org/projects/unified-workflow/builds/)
```
maddenp-noaa, line 1: This is the ReadTheDocs build-status badge.

NaureenBharwaniNOAA (approved):
Simple changes and a lot of needed cleanup here. Aligned very well with the task discussed in Slack. Once it fully passes, it looks good to me!

.github/workflows/test.yaml
```diff
@@ -10,8 +10,8 @@ on:
     branches:
       - develop
 jobs:
-  make-package:
-    name: Make Package
```
NaureenBharwaniNOAA, line 14: After discussion in Slack, I'm trying to fully understand GitHub workflows/Actions setup.
Specifically the `workflow/make-package.yaml` connection to `workflow/scripts/make-package`. If we are re-naming this to `Test`, do we also need to rename the script to `Test`?

WeirAE (approved):
LGTM

.github/workflows/test.yaml
maddenp-noaa: Yes, the `run` entries refer to scripts in the `scripts/` directory. Maybe they can be other things (maybe arbitrary shell commands?), I'm not sure. The scripts can be shared between different workflows, so provide a good way to keep the overall Actions codebase DRY.

.github/scripts/install-conda
```diff
@@ -7,5 +7,5 @@ bash $(basename $url) -bfp conda
   set -e +ux
   source conda/etc/profile.d/conda.sh
   conda activate
-  conda install -q -y -c maddenp --repodata-fn repodata.json anaconda-client condev
+  conda install -q -y -c maddenp --repodata-fn repodata.json anaconda-client condev=0.6.0
```
maddenp-noaa, line 10: This `condev` version will facilitate our CI-based unformatted-code checks.

christinaholtNOAA (approved):
LGTM!

PR #340, opened 2023-11-08
UW-370

maddenp-noaa:
Please be sure to expand files in the _Files changed_ view to see all comments...

Here are the highlights of this PR:

- `uwtools.config.core` was broken up, into: 1. `uwtools.config.tools`, providing, essentially, the `compare_configs()` and `realize_config()` public functions; 2. A hierarchy of modules under `uwtools.config.formats`, providing `YAMLConfig`, `NMLConfig`, et al, each in its own module; and 3. `uwtools.config.support` housing shared functions.
- All ~~(?)~~ Jinja2-related functionality is now located in `uwtools.config.jinja2`. From our group discussion today, I understand now why an attempt was made to keep template-rendering logic independent of the specific Jinja2 implementation, but I'm not sure this was serving us well given that we're fairly heavily invested in the specifics of Jinja2, and because it's hard to think of a realistic alternative. Jinja2 references had already spread beyond `j2template.py`, though I wouldn't call them "widespread", so there _was_ a decently clean interface between UW and Jinja2. In some sense I feel bad about removing that, but I don't know if it was really useful. I'm interested in hearing thoughts on this.
- There's nominally some Jinja2 in `atparse_to_jinja2`, but it's just string manipulation that _produces_ Jinja2 syntax. ~~We might consider whether this should be folded into `uwtools.config.jinja2`, and whether Jinja2 offers any API support for creating syntax (I have no idea).~~ [I checked, and Jinja2 doesn't seem to offer anything for creation of templates -- just for rendering them.]
- ~~There's also some Jinja2 code in `uwtools.rocoto`. We should probably look at seeing if `dereference()` could be called upon to do what's needed there, which would eliminate 1 of 2 `imports` of `jinja2` in the UW codebase (discounting possibly imports in tests).~~ [I updated `uwtools.rocoto` to use `uwtools.config.jinja2.dereference()`.]
- Having _everything_ Jinja2-related in a single module might help if, against all odds, we did eventually find that we need to use some other templating engine.
- Maybe `uwtools.config.jinja2` should just be `uwtools.jinja2`, alongside `uwtools.rocoto`. It's not really a "config" in the sense we mean with `YAMLConfig`, `NMLConfig`, etc. Thoughts?
- The `dereference()` function also now lives in `uwtools.config.jinja2`, as it's intimately tied to Jinja2.
- I think breaking the various `Config` classes out into their own modules is kind of nice, but am open to suggestions. Also open to suggestions on any and all names.
- ~~In addition to maybe checking out my branch an _looking_ at the code, a good guide to how little has actually changed, despite all the diffs, might be to look at the tests, which have not changed much. I waited on breaking up the unit tests into smaller modules so that this comparison could be made, but will do so if we agree that we should proceed with this PR.~~

EDIT

- Another structure option would be to re-combine the contents of `src/uwtools/config/formats` (`base.py` etc.) into a single `src/uwtools/config/formats.py`. That would be fewer files/directories, but one bigger `formats.py` file -- though that would still be smaller than the old `core.py` due to the refactoring to `tools.py` and `support.py`. In one sense, it might be kind of awkward to have a whole e.g. `nml.py` file just to contain the `NMLConfig` class -- when that class could live in a `formats.py` module with its siblings. Then again, if we ever had more Fortran namelist-specific code, `nml.py` would be a sensible place for it to live. Same for the other formats. I don't feel strongly either way.

EDIT

I refactored the unit tests, so this should be good-to-go. This branch has 19 more unit tests than `develop`, mostly due to granular testing of the refactored `dereference()` function.


src/uwtools/tests/config/test_core.py
```diff
@@ -171,53 +175,13 @@ def test_dereference():
...
-type_prob: '{{ list_a / \"a\" }}'  # TypeError
-""",
-            file=f,
-        )
-    cfgobj = core.YAMLConfig(config_file=path)
-    cfgobj.dereference()
-    log.info("HELLO")
-    raised = [record.message for record in caplog.records if "raised" in record.message]
-    assert "ZeroDivisionError" in raised[0]
-    assert "TypeError" in raised[1]
```
maddenp-noaa, line 211: I'll replace this test with a new one that catches divide-by-zero behavior.

src/uwtools/tests/config/test_core.py
```diff
@@ -867,37 +824,30 @@ def test_Config_characterize_values(nml_cfgobj):
...
+#     for x in ["false", "no", "FALSE"]:
+#         assert nml_cfgobj.reify_scalar_str(x) is False
+#     assert nml_cfgobj.reify_scalar_str("88") == 88
+#     assert nml_cfgobj.reify_scalar_str("'88'") == "88"  # quoted int not converted
+#     assert nml_cfgobj.reify_scalar_str("3.14") == 3.14
+#     assert nml_cfgobj.reify_scalar_str("NA") == "NA"  # no conversion
+#     assert nml_cfgobj.reify_scalar_str("@[foo]") == "@[foo]"  # no conversion for YAML exceptions
+#     with raises(AttributeError) as e:
+#         nml_cfgobj.reify_scalar_str([1, 2, 3])
+#     assert "'list' object has no attribute 'read'" in str(e.value)  # Exception on unintended list
```
maddenp-noaa, line 839: The method-under-test became a function in `uwtools.config.jinja2` -- I'll reactivate this test when breaking this module up to mirror the actual library modules.

src/uwtools/tests/fixtures/simple.ini
```diff
@@ -4,4 +4,3 @@ fruit = banana
 vegetable = tomato
 how_many = 12
 dressing = balsamic
-
```
maddenp-noaa, line 7: The dump logic in `INIConfig` no longer emits an extra trailing newline, so the fixture files were updated.

src/uwtools/config/j2template.py
```diff
@@ -1,127 +0,0 @@
-"""
```
maddenp-noaa, line 1: This all moved into `uwtools.config.jinja2`.

src/uwtools/tests/fixtures/gfs.yaml
```diff
@@ -11,7 +11,7 @@ fcst:
   y_grid_points: 100
   length: 12
   freq: 3
-  output_hours: '{% for h in range(0, length, freq) %}{{ "%d " % h }}{% endfor %}'
+  output_hours: '{% for h in range(0, fcst.length, fcst.freq) %}{{ "%d " % h }}{% endfor %}'
```
maddenp-noaa, line 14: This one feels like trouble...

It seems to have been the case that, during dereferencing, the "context" object used to obtain values to render Jinja2 syntax was constructed in such a way that lookups could be done either (a) By a path-from-the-top reference to a value, like the new version here, `fcst.length`; or (b) By a values-surrounding-me reference, like the old version here, `length`. The (new version) values object used by this test looks like
```
model: gfs
target: ufs-weather-model
fcst:
  horizontal_resolution: c768
  vertical_resolution: 64
  x_grid_points: 100
  y_grid_points: 100
  length: 12
  freq: 3
  output_hours: '{% for h in range(0, fcst.length, fcst.freq) %}{{ "%d " % h }}{% endfor %}'
executable: '{{ UFSEXEC }}'
undefined_env: '{{ NOPE }}'
filetype: gfs
datapath: '{{ [experiment_dir, current_cycle] | path_join }}'
filename_core: 'fv_core.res.nc'
testupdate: '{{ updatethis }}'
updatethis: 'testpassed'
grid_stats:
  total_ens_points: '{{ (grid_stats.num_ens_members * grid_stats.total_points ) // 1 }}'
  num_ens_members: 30
  points_per_level: '{{ grid_stats.total_ens_points // grid_stats.num_ens_members // fcst.vertical_resolution }}'
  ref_fcst: '{{ fcst.vertical_resolution }}'
  total_points: '{{ fcst.x_grid_points * fcst.y_grid_points * fcst.vertical_resolution }}'
```
I think this seemed to work due to the old code
```
values = {**os.environ, **ref_dict, **full_dict}
```
which would define a new values context to use for Jinja2 rendering. In the case being discussed here, `full_dict` would be the object shown above, and `ref_dict` would be the subtree
```
horizontal_resolution: c768
vertical_resolution: 64
x_grid_points: 100
y_grid_points: 100
length: 12
freq: 3
output_hours: '{% for h in range(0, fcst.length, fcst.freq) %}{{ "%d " % h }}{% endfor %}'
```
Given that, `values` would end up being: The shell environment, updated by `ref_dict`, updated by `full_dict`. Each update clobbers existing keys where they overlap. In the case of `os.environ`, that's fine. But `full_dict` clobbering keys placed at the top level by `ref_dict` is not. Take a simple example:
```
foo:
    bar: 88
    baz: "{{ bar }}"
bar: 99
```
It's ambiguous what `{{ bar }}` refers to. It will end up being `99`, because `99` is the value in what would be `full_dict`, which will clobber the `88` placed at the top level by what would be `ref_dict`.

While this local reference behavior seems convenient, it also seems to me error-prone and hard to reason about. Isn't it perhaps better to require explicit, unambiguous references in Jinja2 expressions?

src/uwtools/config/formats/ini.py
```diff
@@ -0,0 +1,94 @@
...
+        if cfgdepth == 2:
+            parser.read_dict(cfg)
+            parser.write(s, space_around_delimiters=opts.space if opts else True)
+        else:
+            space = " " if not opts or opts.space else ""
+            for key, value in cfg.items():
+                print(f"{key}{space}={space}{value}", file=s)
+        with writable(path) as f:
+            print(s.getvalue().strip(), file=f)
+        s.close()
```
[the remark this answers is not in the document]
maddenp-noaa: Also, I think much of this would be moot and could be simplified if/when we create a `ShConfig` class to draw a distinction between ini and sh. I'll be creating a ticket for that.

WeirAE:
I for one like the idea of moving jinja next to rocoto in the top level, if not moving them off to a broader 'templates', and trying to roll in atparse. I'm still working on a version that preserves the j2template behavior under my PR, but I'm not personally attached to it

src/uwtools/config/formats/base.py
```diff
@@ -0,0 +1,221 @@
...
+    def dereference_all(self) -> None:
+        """
+        Render as much Jinja2 syntax as possible.
+        """
+        while True:
+            new = dereference(val=self.data, context={**os.environ, **self.data})
+            if new == self.data:
+                break
+            assert isinstance(new, dict)
+            self.data = new
```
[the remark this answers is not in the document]
WeirAE: Agreed, This works better simplified too, merging the two in original is a little unwieldly

src/uwtools/config/jinja2.py
```diff
@@ -0,0 +1,283 @@
...
+    if isinstance(val, str):
+        template = _register_filters(Environment(undefined=DebugUndefined)).from_string(f"'{val}'")
+        rendered = None
+        try:
+            rendered = template.render(**context)
+        except:  # pylint: disable=bare-except
+            pass
+        if rendered:
+            return _reify_scalar_str(yaml.safe_load(rendered))
+    return val
```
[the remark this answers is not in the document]
WeirAE: The more I mess with the original, the more I'm preferring this simplified implementation

src/uwtools/config/support.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    lookup = {
+        FORMAT.fieldtable: "FieldTableConfig",
+        FORMAT.ini: "INIConfig",
+        FORMAT.nml: "NMLConfig",
+        FORMAT.yaml: "YAMLConfig",
+    }
+    if not fmt in lookup:
+        raise log_and_error("Format '%s' should be one of: %s" % (fmt, ", ".join(lookup)))
+    return getattr(import_module(f"uwtools.config.formats.{fmt}"), lookup[fmt])
```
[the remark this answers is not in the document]
WeirAE: Ahhh, I think I had been doing it wrong originally trying to make a deferred import, this is better!

src/uwtools/tests/fixtures/gfs.yaml
maddenp-noaa: This is OBE. I have restored the global/local functionality in the updated version, and added an explicit unit test for it. I still am unsure whether this behavior is a good idea.

src/uwtools/tests/config/test_core.py
maddenp-noaa: Test reactivated.

src/uwtools/tests/config/test_core.py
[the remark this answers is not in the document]
maddenp-noaa: Tests are now refactored.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -0,0 +1,149 @@
...
+        "'a': '{{ b.c + 11 }}', 'b': {'c': 66}, 'd': '{{ X }}'",
+        "'a': 77, 'b': {'c': 66}, 'd': '{{ X }}'",
+    ]:
+        assert regex_logged(caplog, excerpt)
+    assert config == {"foo": 88, "a": 77, "b": {"c": 66}, "d": "{{ X }}"}
+
+
+# NB: Need direct test for parse_include().
+
+# NB: Need direct test for update_values().
```
maddenp-noaa, line 131: These two methods need direct tests. I tried, but failed, to write them. I think I need a refresher on how the `!INCLUDE` directive is meant to behave. I'm going to check the user guide to see if it's described, but everything I tried didn't work as I expected it to, so I'm confused. It's ok for now: The code is transitively covered by other tests, but it should have direct tests.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -0,0 +1,149 @@
...
+    outfile = tmp_path / f"test_{fmt1.lower()}to{fmt2.lower()}_dump.{fmt2}"
+    reference = fixture_path(f"simple.{fmt2}")
+    cfgin = tools.format_to_config(fmt1)(fixture_path(f"simple.{fmt1}"))
+    tools.format_to_config(fmt2).dump_dict(path=outfile, cfg=cfgin.data)
+    with open(reference, "r", encoding="utf-8") as f1:
+        reflines = [line.strip().replace("'", "") for line in f1]
+    with open(outfile, "r", encoding="utf-8") as f2:
+        outlines = [line.strip().replace("'", "") for line in f2]
+    for line1, line2 in zip(reflines, outlines):
+        assert line1 == line2
```
maddenp-noaa, line 149: This is a funny (but valuable) test that doesn't seem to belong anywhere: It tests the interaction of various formats and so doesn't really belong to the test modules for any of those individual formats. I've put it here in `test_base.py`, which seems reasonable to me.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+def test_dereference_list():
+    val = ["{{ n }}", "{{ s }}"]
+    assert jinja2.dereference(val=val, context={"n": 88, "s": "foo"}) == [88, "foo"]
+
+
+def test_dereference_local_values():
+    # Rendering can use values from the local contents of the enclosing dict, but are shadowed by
+    # values from the top-level context object.
```
maddenp-noaa, line 79: Here's the test for the global/local values-lookup behavior.

src/uwtools/tests/fixtures/gfs.yaml
```diff
@@ -1,32 +0,0 @@
-# Model-related definitions
```
maddenp-noaa, line 1: This file was only serving the tests of the `dereference()` function, which now has many granular tests, each with their own test data.

src/uwtools/config/formats/base.py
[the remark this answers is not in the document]
WeirAE: I'm grudgingly on team `formats` in lieu of `config.config` or a better choice

src/uwtools/config/formats/base.py
```diff
@@ -0,0 +1,222 @@
...
+        :param parent: Parent key.
+        """
+        complete: List[str] = []
+        empty: List[str] = []
+        template: List[str] = []
+        for key, val in values.items():
+            if isinstance(val, dict):
+                complete.append(f"    {parent}{key}")
+                c, e, t = self.characterize_values(val, f"{parent}{key}.")
+                complete, empty, template = complete + c, empty + e, template + t
```
WeirAE, line 89: this right here is the perfect use of single letters with context!

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -0,0 +1,177 @@
...
+                "dressing": "poppyseed",
+            }
+        }
+    )
+    config.parse_include()
+
+    assert config["fruit"] == "papaya"
+    assert config["how_many"] == 17
+    assert config["config"]["meat"] == "beef"
+    assert len(config["config"]) == 2
```
maddenp-noaa, line 151: Thanks to @WeirAE for this test and the one below!

WeirAE (approved):
I've gone through this version a few times - thank you for jumping on this when it got far beyond the initial estimate. LGTM, of course

src/uwtools/tests/config/formats/__init__.py
NaureenBharwaniNOAA, line 1: Could you remind me why we need these `__init__` files are needed?

I remember quite a bit ago we talked about why these `__init__.py` files are created and why they are left blank, but haven't been able to find the information to reference the reasoning.

maddenp-noaa: An `__init__.py` file turns a directory into a Python _package_. For example, because `src/uwtools/config/formats/__init__.py` exists, a Python package/namespace `uwtools.config.formats` exists. The `.py` files within a package directory are _modules_. Without the `__init__.py` file, Python does not recognize a directory as a package, and certain import and package/module-discovery mechanisms (e.g. the way `pytest` automatically locates all these modules) will not work.

It's possible for `__init__.py` to _not_ be blank, too. For example, if `src/uwtools/config/formats/__init__.py` contained a line
```
```
then another module could do
```
from uwtools.config.formats import YAMLConfig
```
instead of
```
```
So, the `__init__.py` files are sometimes used to construct synthetic namespaces by importing code into them and creating a more convenient place to import from. For example, a hypothetical `api/__init__.py` might `import` a bunch of different functions, classes, constants, etc. from other packages, modules, etc., into a single module so that a user could `import uwtools.api` and have access to just the bits of `uwtools` that we want to expose as an API.

And, finally, another use case for `__init__.py` is to just contain the code in question. See e.g. https://github.com/maddenp/iotaa/blob/main/src/iotaa/__init__.py, which makes it possible for a user to `import iotaa` and get access to all the code in that file.

NaureenBharwaniNOAA (approved):
Took a bit of time, but the PR looks good to me! It definitely seems like this ticket became a lot more work than initially thought. Nice job guys!

christinaholtNOAA (approved):
LGTM. I'm sorry for the lengthy delay in my review.

PR #341, opened 2023-11-09
UW-415

NaureenBharwaniNOAA:
**Current WIP:**
- I am adding the `datadep` tag first under the `dependency` element. I started by creating a test called `test__data_dependency` that replicates a lot of the work in `test__task_dependency`. I'm working through issues surrounding that the `taskdep` calls `_set_attrs` and does most of the work to sort out the nested `dict` under the `task_dep` tag. This is possible because `taskdep` will always have an `attribs` keyword.`datadep` does not have the same `attribs` keyword, but does have a couple specific keywords like `age`and `minsize`, however these both seem to be optional which makes the `datadep` tag also optional altogether. 

- I would like to add a new method called `add_dependency`, this would require some work to then factor out `task_dependency` into it's own entity. Under the `add_dependency` method we could then call the respective `dependency` tags broken up into new methods. This would add quite a bit of new smaller methods specifically one for each tag under the `dependency` element. There is room for overlap here, but that will not be known until getting into the details. Thoughts?

**UPDATE:**
I discovered a nice breakup for the list of tags seen for the `dependency` element. 

Main types:
1. Operands - `taskdep`, `datadep` and `timedep`
- Each of the `operands` will have their own individual needs. The `taskdep` operand was already covered in UW-376.
2. Operators - `and`, `or`, `not`, `xor`, `nand` and `nor`
- Each of the operators can see any `operator` or `operand`. Once an `operand` is found this should be the end of the nested loop as long as there are no other `operators` at the same level.
3. Misc - `strneq`, `streq`, `True` and `False`
- Each of these are required to be added to the ENTITY header, which was already taken care of in UW-376. We will still need to account for this within the YAML I believe.

Knowing this information, it would be beneficial for 3 main methods to be called `_add_operand`, `_add_operator` (this will be clean since we can simply have an `if ANY OPERATOR found` loop again clause, `else ANY OPERAND found` go to the `_add_operand` method), and a misc method.

**UPDATE:**
We have now removed the option for handling the `True` and `False` `<dependency>` elements.

**Description**

As discussed in [#326](https://github.com/ufs-community/workflow-tools/pull/326#discussion_r1374706969)

              "More work is needed here to handle the possibilities of the `<dependency>` element. This currently only reproduces what our existing tests require, which is an acceptance criterion of the task. I'll see what I have time for in this sprint, but we might need a follow-on task to flesh this out."

 

**Requirements**
Implement new test cases that cover all general categories of how the `<dependency>` element is built such as:

- Some have "text" attributes like `<left>"foo"</left>`
- Some have element attributes like `<taskdep task="foo"/>`
- Some have both text attributes and element attributes like `<datadep age=05:00>/some/path/to/data</datadep>`
- Some have subelements like `<or>`, `<streq>`, etc., but no element attributes or text attributes.

 

GitHub Issue Linked Here: https://github.com/ufs-community/workflow-tools/issues/339

**Type of change**

**How Has This Been Tested?**

`make test` and Actions

**Checklist**


src/uwtools/rocoto.py
```diff
@@ -187,9 +187,18 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+            elif tag == STR.datadep:
+                self._add_data_dependency(e, config[STR.datatdep], block)
             else:
                 raise UWConfigError("Unhandled dependency type %s" % tag)
 
+    def _add_data_dependency(self, e: Element, config: dict, block: dict) -> None:
+        """
+        ???
+        """
+        breakpoint()
```
NaureenBharwaniNOAA, line 199: Just for my own use. Will be removed later.

src/uwtools/rocoto.py
```diff
@@ -187,9 +187,18 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
             tag, _ = self._tag_name(key)
             if tag == STR.taskdep:
                 self._set_attrs(SubElement(e, STR.taskdep), block)
+            elif tag == STR.datadep:
+                self._add_data_dependency(e, config[STR.datatdep], block)
             else:
                 raise UWConfigError("Unhandled dependency type %s" % tag)
 
+    def _add_data_dependency(self, e: Element, config: dict, block: dict) -> None:
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: That makes perfect sense to me. I actually went through `test_rocoto` and found `_add_task_dependency` was quite deeply embedded in all the tests already and that out of simplicity it might just be easier to keep it the way it is currently and just factor out `taskdep` where appropriate.

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,25 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        ???
+        """
         for key, block in config.items():
             tag, _ = self._tag_name(key)
             if tag == STR.taskdep:
                 self._set_attrs(SubElement(e, STR.taskdep), block)
-            else:
-                raise UWConfigError("Unhandled dependency type %s" % tag)
+            elif tag == STR.datadep:
+                pass
```
NaureenBharwaniNOAA, line 204: Still trying to figure out this one. I want to set the attrs on `datadep` however the `taskdep` currently calls `_set_attrs` and within that method we see `config[STR.attrbs].items()`. In `datatdep` we will not see an `attrbs` element making it hard to navigate currently.

maddenp-noaa: Maybe you mentioned that there _are_ attributes to set, but that they're just not in a sub-dict called `attrs`. Is that right? If so, you might be able to still use `_set_attrs()` by constructing a dict literal, populated from the attributes in `block`, and passing it in via the `config` parameter. So, the call here might look like
```
self._set_attrs(SubElement(e, STR.datadep), config={"attrs": {???: block[???], ???: block[???], ...}})
```
where the `???`s are keys in the `datadep` block. Is that realistic?

maddenp-noaa: > adding an `attrs` key to the dict
Maybe be careful about actually modifying the original config dict, as that could have unintended consequences elsewhere. You could either use `deepcopy()` from Python's `copy` module to get a copy that would be safe to mutate (and then discard later), or construct a small dict on the fly containing just what you need and passing that down.

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,26 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+    def _add_task_dependency_operand(self, e: Element, config: dict) -> None:
+        """
+        ???
+        """
         for key, block in config.items():
             tag, _ = self._tag_name(key)
             if tag == STR.taskdep:
                 self._set_attrs(SubElement(e, STR.taskdep), block)
+            elif tag == STR.datadep:
+                self._set_attrs(SubElement(e, STR.datadep), config={"attrs": block})
```
NaureenBharwaniNOAA, line 202: @maddenp-noaa Here I added `attrs` to the beginning of the rest of the `dict` contained in `block`. I didn't update the config variable itself, but what is being passed to `_set_attrs()`. I don't think this would have an effect of the `config` being modified.

I can do as you suggested and add `deepcopy()` before calling the `_set_attrs()` method like below:
```
deepcopy(config)
self._set_attrs(SubElement(e, STR.datadep), config={"attrs": block})
```

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,26 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
             else:
-                raise UWConfigError("Unhandled dependency type %s" % tag)
+                pass
+                # self._set_attrs(SubElement(e, STR.timedep), config={"attrs": block})
```
NaureenBharwaniNOAA, line 205: Something else I am currently working on is the `timedep` tag. Here is an example of what a `timedep` tag would look like 
`<timedep><cyclestr offset="&DEADLINE;">@Y@m@d@H@M@S</cyclestr></timedep>`

We will not be handling `cyclestr` until a later ticket planned for sprint 10.5. I think for now the idea would be to just include 
`<timedep> offset="&DEADLINE;"</timedep>`
as an example of what should be covered by this ticket.

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,20 @@ def test__add_task_dependency_fail(self, instance, root):
...
+        assert datadep.tag == "datadep"
+        assert datadep.get("age") == "120"
+
+    def test__add_time_dependency(self, instance, root):
+        config = {"timedep": {"offset": "&DEADLINE;"}}
+        instance._add_task_dependency_operand(e=root, config=config)
+        timedep = root[0]
+        assert timedep.tag == "timedep"
+        assert timedep.get("offset") == "&DEADLINE;"
+
```
NaureenBharwaniNOAA, line 145: I split `taskdep`, `datadep` and `timedep` up into their own test cases. 

I can merge these under a `test__add_operand_dependency` if that is more ideal. 

Currently `test__add_task_dependency`, `test__add_data_dependency` and `test__add_time_dependency` each call `_add_task_dependency_operand` individually and leads to quite a bit of duplication.

src/uwtools/rocoto.py
maddenp-noaa: What you have here (I mean the code, not the comment) looks good: `block` is part of `config`, but since `_set_attrs()` doesn't modify anything, this is all read-only access and safe. It doesn't look to me like the `deepcopy()` is needed.

If this code now does what you need it do, it raises the question for me: Should we make the `datadep` YAML block have an `attrs` subtree, the same way `taskdep` apparently does? Do we have an unnecessary asymmetry here between how `taskdep` and `datadep` are represented in UW YAML? This is under our control (I think), so we should be able to make it as consistent as we like.

Just to motivate what I wrote in the first paragraph, note that a modifed subtree would be reflected in its parent object, as the whole tree is just a bunch of `dict` objects wired up together:
```
>>> config = {"a": "apple", "f": {"f1": "rose", "f2": "tulip"}, "b": "banana"}
>>> def f(config):
...   for key, value in config["flowers"].items():
...     config["flowers"][key] = value.upper()
... 
>>> f({"flowers": config["f"]})
>>> config
{'a': 'apple', 'f': {'f1': 'ROSE', 'f2': 'TULIP'}, 'b': 'banana'}
```
So, even though a new `dict` was constructed and passed to `f()`, part of that `dict`'s structure was shared with the original `config` and so, when `f()` modified that, `config` saw the modifications. Apologies if that's obvious.

NaureenBharwaniNOAA:
Another thing to note will be working with the operators -- `and`, `or`, `not`, `xor`, `nand` and `nor`. 

`and`, `or` and `not` are reserved words in python so I haven't added them to the  `dataclass` of string literals yet since I'm still working out how to deal with these cases.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: I like the granular tests, one per dependency type. If the three tests were more than a few lines each and shared a lot in common, I'd maybe lean toward DRY and try to use `@pytest.mark.parametrize` to fold them into a single test function, but otherwise I think having them separate makes reading and maintenance easier.

maddenp-noaa: Well, the tests are right in front of my face so I should have just read them. Doing so now, I do think that I would try to summarize them via `@pytest.mark.parametrize`. There seem to be just a couple differences that could be passed in to a generalized test-function body as parameters.

maddenp-noaa:
@NaureenBharwaniNOAA A common solution for the reserved-word problem is to append an underscore to the variable, e.g. `or` becomes `or_`.

src/uwtools/rocoto.py
maddenp-noaa: Seems reasonable for now, as long as it passes the RELAX NG schema check: Is it ok with `<timedep>` having arbitrary text content instead of a `<cyclestr>` child?

src/uwtools/rocoto.py
NaureenBharwaniNOAA: > Should we make the `datadep` YAML block have an `attrs` subtree, the same way `taskdep` apparently does? Do we have an unnecessary asymmetry here between how `taskdep` and `datadep` are represented in UW YAML?

I'm not sure if this will work as intended. Within the YAML that a user supplies to us the `datadep` tag does not have an `attrs` keyword. Only `task_*` or `workflow` tags will contain an `attrs` keyword. Similar to `datadep`, the `timedep` will also not contain an `attrs` keyword and so on with the remainder of the operators as well. I believe the asymmetry between `taskdep` and `datadep` is a necessary one due to how the Rocoto workflow is structured. 

From the [SRW documentation](https://ufs-srweather-app.readthedocs.io/en/develop/CustomizingTheWorkflow/DefineWorkflow.html#defining-dependencies) we have "`attrs`: Any of the attributes to the task tag in Rocoto. This is meant to be a subdictionary defining any of the Rocoto-supported attributes, where the key is the name of the attribute, and the value is what Rocoto expects. Attributes might include any combination of the following: cycledefs, maxtries, throttle, or final."

I believe the overlap comes into play when the sources reference the "attributes" for a tag. Although they are the attributes the keyword `attrs` will not appear in the YAML unlike how it does for the `workflow` and `taskdep` tags.

maddenp-noaa: > the keyword `attrs` will not appear in the YAML
But we get to define what appears in the YAML, right? No one is dictating that to us. We could for example decide that any `<element>` in Rocoto XML that takes attributes, e.g. `<element foo=88 bar="baz">`, will have an `attrs` subtree in YAML that contains those attributes. I'm not necessarily saying that's what we should do, but what would stand in the way of us doing it?

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,25 @@ def test__add_task_dependency_fail(self, instance, root):
...
+        assert dependency.tag == validity_tag
+        assert dependency.get(validity_key) == validity_val
+
+    def test__add_time_dependency(self, instance, root):
+        config = {"timedep": {"offset": "&DEADLINE;"}}
+        instance._add_task_dependency_operand(e=root, config=config)
+        timedep = root[0]
+        assert timedep.tag == "timedep"
+        assert timedep.get("offset") == "&DEADLINE;"
+
```
NaureenBharwaniNOAA, line 150: Here's my first attempt of combining these two test cases by using parametrization. Quite a bit going on and still working out the logistics of it. Thought I'd share to confirm the direction we're going in.

maddenp-noaa: Stacking the decorators like that results in the Cartesian product of all those values. Since there are 2 items in each decorator, we'd get 2 x 2 x 2 = 8 different tests, one in which (for example), `config` will be `{"datadep": {"age": "120"}}`, `validity_tag` will be `timedep`, and `validity_items` will be `("age", "120")`, which I think would not be a combination.

What about something like
```
    @pytest.mark.parametrize(
        "config", [{"datadep": {"age": "120"}}, {"timedep": {"offset": "&DEADLINE;"}}]
    )
    def test__add_task_dependency_operand(self, config, instance, root):
        instance._add_task_dependency_operand(e=root, config=config)
        element = root[0]
        tag = list(config.keys())[0]
        assert element.tag == tag
        for attr, value in config[tag].items():
            assert element.get(attr) == value
```
? That implies just two tests.

maddenp-noaa: The `for` loop in my suggestion may not be even needed -- but it'd be useful if there were multiple attributes to test and would work even if two tags had different numbers of attributes.

maddenp-noaa: Here's one step cleaner, I think:
```
    @pytest.mark.parametrize(
        "config", [{"datadep": {"age": "120"}}, {"timedep": {"offset": "&DEADLINE;"}}]
    )
    def test__add_task_dependency_operand(self, config, instance, root):
        instance._add_task_dependency_operand(e=root, config=config)
        element = root[0]
        for tag, attrs in config.items():
            assert element.tag == tag
            for attr, value in attrs.items():
                assert element.get(attr) == value
```

NaureenBharwaniNOAA: A couple questions/design choices here:
1. The ordering of the input parameters here should be lexicographical. 
i.e. ```def test__add_task_dependency_operand(self, config, instance, root)```

2. The ordering of the test cases here is not lexicographical and this is because it follows the ordering of how the tags appear in the YAML/workflow. I left this consistent by simply adding the new test case below this test case: ```test__add_task_dependency```

3. I also left ```taskdep`` within the ```test__add_task_dependency``` test case and not factoring it out into the ```def test__add_task_dependency_operand```. I felt like within this case it looked fine staying within the current one. And looks cleaner altogether.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: It is currently happy with the arbitrary text instead of a ```<cyclestr>``` child and passes the RELAXNG schema check.

We will need to add this note to add additional criteria to the current Rocoto XML ticket that covers ```<cyclestr>``` or create a separate ticket since one involves modifying the JSONSchema and this requires updating the ```<dependency>``` element here. I feel like there is also an order to first modifying the JSONSchema and then completing this task. This should be a very minor change once the schema is updated.

NaureenBharwaniNOAA:
> @NaureenBharwaniNOAA A common solution for the reserved-word problem is to append an underscore to the variable, e.g. `or` becomes `or_`.

Interesting, I found a couple different ways to do it with a ```\``` appended to the front of it or using ```escape()```. Neither of those seemed to work for me, let me give this one a try.

src/uwtools/rocoto.py
maddenp-noaa: I agree with you about updating the schema first, then implementing the code generation. Considering our sprint review yesterday, separate small tasks is probably preferable to one larger one.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: 1. Since it's arbitrary which order the task arguments appear (except for `self` for test methods rather than standalone functions), I try to default to lexicographical ordering.
2. Fair enough -- there are often good reasons to depart from convention. [PEP8 agrees.](https://peps.python.org/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds).

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,51 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+    def _add_task_dependency_boolean(self, e: Element, config: dict) -> None:
+        """
+        ???
+        """
+        for key, block in config.items():
+            tag, _ = self._tag_name(key)
+            if tag == STR.True_:
+                self._set_attrs(SubElement(e, STR.True_), config={"attrs": block})
+            else:
+                self._set_attrs(SubElement(e, STR.False_), config={"attrs": block})
```
NaureenBharwaniNOAA, line 230: After taking a look at the Rocoto XML documentation the boolean operators ```True``` and ```False``` will need to be taken care of in the ENTITY section I believe.

Here is what the boolean operators look like when added to the workflow:
```
<?xml version="1.0"?>
<!DOCTYPE workflow
[
 ...
  <!ENTITY RUN_GSI "</true>">
]>

...

<dependency>
  &RUN_GSI;
</dependency>
```

There is no actual str literal ```True``` or ```False``` being added to the ```<dependency>``` element, only the ```value``` of boolean operator. 

Maybe adding a condition in the ```def _doctype()``` method would be the appropriate place for this:
```
if entities := self._config[STR.workflow].get(STR.entities):
    tags = (f'  <!ENTITY {key} "{val}">' for key, val in entities.items())
```

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,74 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+            elif tag == STR.nand:
+                self._set_attrs(SubElement(e, STR.nand), config={"attrs": block})
+            elif tag == STR.nor:
+                self._set_attrs(SubElement(e, STR.nor), config={"attrs": block})
+            elif tag == STR.not_:
+                self._set_attrs(SubElement(e, STR.not_), config={"attrs": block})
+            elif tag == STR.or_:
+                self._set_attrs(SubElement(e, STR.or_), config={"attrs": block})
+            elif tag == STR.xor:
+                self._set_attrs(SubElement(e, STR.xor), config={"attrs": block})
```
NaureenBharwaniNOAA, line 229: Probably a better and more clean way to do this vs. a huge if/elif/else loop. I will have to add in some logic to drill down into the next block of the config to see if an operand, operator, or other tag is present. I was thinking about doing this search by adding a dict of each and simply calling the dict. An example sketch below:

 ```
dict_operands = [datadep, taskdep, timedep]
dict_operators = [and, nand, nor, not, or, xor]
dict_other = [streq, strneq, True, False]

if tag in dict.operands:
    _add_task_dependency_operands()
```

File: src/uwtools/rocoto.py (hunk 8 above)
Comment (maddenp-noaa) [reply src/uwtools/rocoto.py:253]: I think setting `</true>` or `</false>` via an entity is just a shortcut for users who are manually creating their Rocoto workflow XML documents so that, rather than going down into the XML document to find and toggle the desired literal e.g. `</true>` or `</false>` tag, they can have a named entity (variable) that they can toggle at the top of the document. This would also support a DRY document: If they had the same sort of dependency in multiple tasks, they could control all of then by toggling a single entity.

But I think this is orthogonal to what you're currently doing, as users should -- as long as we continue trying to translate _all_ Rocoto functionality from UW YAML to Rocoto XML -- be able to specify a boolean dependency explicitly if they want to do that, and those would need to be rolled out by our code as actual `</true>` and `</false>` elements in the XML tree.

I think the excerpt from the Rocoto docs you show above would be supportable by making it possible to specify an entity reference as the `text` of a `<dependency>` element. That way, a user could specify e.g. `&RUN_GSI;` as the text, and then define an entity mapping `RUN_GSI` to `</true>` -- and we already have support for declaring arbitrary entities. I think "making it possible" means ensuring that the JSON Schema for UW YAML supports having something like a `text` setting for a `dependency` subtree in the Rocoto section of UW YAML. We might need to ensure that the user-supplied value regex-matches an entity reference, as I doubt the RELAX NG schema permits truly arbitrary text between `<dependency>` and `</dependency` -- it's probably expecting other `<>` elements, so whatever the user supplies as a value had better expand to that.

This is actually kind of tricky because hand-written XML could contain entity references _anywhere_, and if they expand to valid XML, they'd pass RELAX NG schema validation. But I doubt we have support in the Rocoto section of UW YAML for arbitrary entity references appearing pretty much anywhere. So maybe we don't want to support that. It'd be a good question for discussion.

Comment (maddenp-noaa) [reply src/uwtools/rocoto.py:253]: Maybe more explicit, I think it **is** valid to have e.g. `<dependency></true></dependency>` in Rocoto XML. The user could write that literally, or the `</true>` could be interpolated in via entity expansion like `<dependency>&RUN_GSI;</dependency>`.

It seems like Rocoto's goal here is to provide an easy way to turn pieces of functionality on and off, so maybe we need to figure out how the same goal could be achieved in UW YAML. For example, we could decide that every task or metatask block in the Rocoto section of UW YAML can have an optional `enabled` flag with a boolean value, which would default to `true` (`True` in Python) if not defined. Then, when a user generates a Rocoto XML document, tasks explicitly marked `enabled: false` would simply not appear in the XML output. A downside to that is that the user would have to use UW to regenerate the XML document if they wanted to enable that task; they could not (easily) just edit the XML, because the task wouldn't be there at all. There are more complicated things that could be done, too, involving automatically injecting control entities for all tasks, and updating all tasks with `<dependency>` elements that reference the associated entities and default to `</true>`. Those entities would have to be auto-generated by UW in a way that could never conflict with user-defined entities, and the `<dependency>` elements would have to be `<and>`ed with any other dependencies. It could be quite tricky.

File: src/uwtools/rocoto.py (hunk 9 above)
Comment (maddenp-noaa) [reply src/uwtools/rocoto.py:229]: Something like
```
if tag in (STR.and_, STR.nand, < the others >, STR.xor):
    self._set_attrs(SubElement(e, tag), config={"attrs": block})
```
? Or `if tag in dict_operators:` if you have those already defined! Same idea as what you're considering for dispatching to `_add_task_dependency_operands()`.

Comment (maddenp-noaa) [reply src/uwtools/rocoto.py:229]: What is `dict_` in `dict_operands` et al? I guess I would expect something like
```
operands = (STR.datadep, STR.taskdep, STR.timedep)
operators = (STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor)
other = (STR.streq, STR.strneq, STR.true, STR.false)
```
But I might be missing something...

Incidentally, using tuples `(...)` instead of lists `[...]` is a nice idiom when you want to express (and actually enforce) that these collections _canot be mutated_ by code. The two data structures behave the same in most read-only contexts, but you can never accidentally mutate a tuple.

Comment (NaureenBharwaniNOAA) [reply src/uwtools/rocoto.py:229]: Originally my thought was that `dict_` would be the list of items contained by that category, where category here is operator, operands, etc. But working through this logic again, a dict would not be the best way to get this point across and would lead to confusion overall. 

I also realize that this might not be necessary:
```
if tag in (STR.and_, STR.nand, < the others >, STR.xor):
    self._set_attrs(SubElement(e, tag), config={"attrs": block})
```
I currently have a messier version of this implemented and am getting back an error of
```
E               assert "{'offset': '&DEADLINE;'}" == {'offset': '&DEADLINE;'}
``` 
This is due to in `_set_attrs` we take the remainder of the config and convert it to a `str` --> `e.set(attr, str(val))`.
Maybe the better idea is if we are find one of these operators we will again call the `_add_task_dependency_operator` case which will be the recursive part. This will continue until we find a tag that ends the loop like `datadep` or `taskdep`.

Overall I think changing the idea to something like
```
operands = (STR.datadep, STR.taskdep, STR.timedep)
operators = (STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor)
other = (STR.streq, STR.strneq, STR.true, STR.false)

if tag in operands:
   # this will end the recursion
   _add_task_dependency_operands()

elif tag in operators:
    # recursion will continue
   _add_task_dependency_operands()

else: 
   # this will end the recursion
   ...
```

File: src/uwtools/rocoto.py [hunk 10]
```diff
@@ -183,12 +183,73 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        other = (STR.streq, STR.strneq) 
+
+        for key, block in config.items():
+            tag, _ = self._tag_name(key)
+            if tag in operators:
+                self._add_task_dependency_operator(e, config=config)
+            elif tag in operands:
+                self._add_task_dependency_operand(e, config=config)
+            elif tag in other:
+                self._add_task_dependency_strequality(e, config=config)
```
NaureenBharwaniNOAA, line 227: I implemented a newer version that we discussed for not having a giant if/elif/else loop. I am running into a problem with tracking down the depth of recursion. It's still a WIP but a good start for now.

src/uwtools/rocoto.py
maddenp-noaa: Looks like you're on the right path with handling (and, crucially, ending) the recursion. Should the `elif` block call an `_operators()` method rather than `_operands()`, given that the check if on `tag in operators`? Maybe just a typo in your pseudocode...

src/uwtools/rocoto.py
maddenp-noaa: I'll have to see the final code, but two potential options for dealing with something like that are: 1. Add conditional logic on the receiver side to deal with inputs with different structure, or 2. Massage the input before passing it so that it conforms to what the receiver is expecting. The latter is the favored approach of the "confident programming" camp: Minimize conditional behavior by taking steps to ensure that invariants are maintained. The suggestion I made earlier, to construct a `{"attrs": ...}` object and passing it to a receiver that expects an `attrs` key follows this line of thought. Your call on what's best based on what you're actually dealing with (sometimes the conditional logic just _moves_ and doesn't disappear, and if it both moves and _multiplies_ it may not be a good solution), but it's something to consider. In general, it's always worth considering whether conditional logic could be replaced by introduction of and adherence to convention -- which is why I asked whether we should just require an `attrs` key in UW YAML any time an XML element has attributes, as that convention would make treatment of all elements consistent.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: Yes it should be `operators`, thank you for that catch

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I'll work on adding an `attrs` key to the `datatdep`, and `timedep` tags. I think those should be the only two we need since the operators (and, or, not, etc.) will contain one of these tags or others nested inside.

src/uwtools/rocoto.py
```diff
@@ -183,12 +183,73 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        """
+        ???
+        """
+        breakpoint()
+        for key, block in config.items():
+            tag, _ = self._tag_name(key)
+            if tag == STR.true:
+                self._set_attrs(SubElement(e, STR.true), config={"attrs": block})
+            else:
+                self._set_attrs(SubElement(e, STR.false), config={"attrs": block})
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: After discussion in UW Group I'll be removing the `true` / `false` tags.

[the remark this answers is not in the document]
maddenp-noaa: It would be good to remove them from the JSON Schema, too, so that they don't slip through to the Python code.

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -161,7 +161,18 @@
...
-          "$ref": "#/$defs/TimePattern"
+          "additionalProperties": false,
+          "properties": {
+            "attrs": {
+              "additionalProperties": false,
+              "properties": {
+                "offset": {
+                  "type": "string"
+                }
+              },
```
NaureenBharwaniNOAA, line 172: I updated the schema to expand on the `timedep` tag. Here we need to take into account for the `offset` key within `timedep`. I also added `attrs` to `timedep` which is what we talked about UW Group this week.

I did not update `datadep` as it already had the `attrs` item within the schema and had all the appropriate tags needed.

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,43 @@ def test__add_task_dependency_fail(self, instance, root):
...
+    @pytest.mark.parametrize(
+        "config", [{"datadep": {"attrs": {"age": "120"}}}, {"timedep": {"attrs": {"offset": "&DEADLINE;"}}}]
+    )
+    def test__add_task_dependency_operand(self, config, instance, root):
+        instance._add_task_dependency_operand(e=root, config=config)
+        element = root[0]
+        for tag, attrs in config.items():
+            assert tag == element.tag
+            for attr, val in attrs.items():
+                assert element.get(attr) == val
```
NaureenBharwaniNOAA, line 141: I am having a bit of a problem tracking this error, where I get an error of
```
AssertionError: assert None == {'age': '120'}
AssertionError: assert None == {'offset': '&DEADLINE;'}
```
When using the debugger I haven't been able to trace down what is happening since I am having a hard time accessing the `element` variable. I included a little snippet from the debugger itself below:

```
> /Users/naureenbharwani/Desktop/UFS/workflow-tools/src/uwtools/tests/test_rocoto.py(139)test__add_task_dependency_operand()
(Pdb) p config
{'datadep': {'attrs': {'age': '120'}}}
(Pdb) n
-> for tag, attrs in config.items():
(Pdb) n
-> assert tag == element.tag
(Pdb) p tag
'datadep'
(Pdb) p element.tag
'datadep'
(Pdb) n
-> for attr, val in attrs.items():
(Pdb) n
-> assert element.get(attr) == val
(Pdb) p attr
'attrs'
(Pdb) p val
{'age': '120'}
(Pdb) p element.get(attr)
**None**
(Pdb) p attr
'attrs'
(Pdb) 

```

src/uwtools/rocoto.py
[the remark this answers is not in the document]
NaureenBharwaniNOAA: It looks like they were never included in the JSONSchema to begin with. The only `false` entries I have coming up are appropriate for not allowing `additionalProperties`. 

In the Relax NG schema I do see them included, I'll remove them from there now.

src/uwtools/resources/schema_with_metatasks.rng
```diff
@@ -257,12 +257,6 @@
         <empty />
       </element>
       <ref name="rb_or_sh" />
-      <element name="true">
-        <empty />
-      </element>
-      <element name="false">
-        <empty />
-      </element>
```
NaureenBharwaniNOAA, line 265: Removed `true`/`false` tags from the schema. They already did not exist in the JSONSchema.

src/uwtools/resources/schema_with_metatasks.rng
```diff
@@ -336,6 +330,9 @@
         <ref name="compoundTimeString" />
       </element>
       <element name="timedep">
+        <attribute name="offset">
+          <data type="string" />
+        </attribute>
```
NaureenBharwaniNOAA, line 335: Added missing `offset` attr to `timedep`

src/uwtools/resources/schema_with_metatasks.rng
maddenp-noaa: Apart from formatting, I don't think we'd want to change the RELAX NG schema: It should still be able to validate a Rocoto document, even if it's not one that we produce. I don't think it should do any harm to leave the true/false entities in the RELAX NG schema; we just don't want them in the JSON Schema.

src/uwtools/resources/schema_with_metatasks.rng
maddenp-noaa: Here, too, I don't think we should be modifying the RELAX NG schema. The original schema (identical to the one in the Rocoto git repo except that the formatting is tidied up) should be able to validate any Rocoto XML documents we produce, as well as ones we don't produce. This schema is the definition of what constitutes a valid Rocoto XML document, so I can't see why we'd change it. Please let me know if I'm missing something.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: "...`timedep` is just an alias (ref) for compoundTimeString. Do we have a representation for that in JSON Schema yet?"

Yes we have `timedep` referencing the `TimePattern` definition in the JSONSchema where it matches a specific regex expression for `^(\\d{2}:){3}\\d{2}$`. I don't believe this covers any arbitrary text string, and in the JSONSchema we can only reference one definition in this case`"$ref": "#/$defs/TimePattern"`.

src/uwtools/resources/schema_with_metatasks.rng
NaureenBharwaniNOAA: Sounds good, undoing these changes very quick

src/uwtools/resources/rocoto.jsonschema
maddenp-noaa: I don't think we have to deal with it in this PR, but here's what Rocoto actually allows for `compoundTimeString`: https://github.com/christopherwharrop/rocoto/blob/develop/lib/workflowmgr/schema_with_metatasks.rng#L203-L210 -- i.e. any combination of one or more blocks of text and `<cyclestr>` elements. The `<cyclestr>` element itself has an `offset` attribute, which has a value that may be what our `TimePattern` is trying to emulate. Rocoto's spec for the `time` pattern also looks somewhat more complicated/flexible (https://github.com/christopherwharrop/rocoto/blob/develop/lib/workflowmgr/schema_with_metatasks.rng#L181-L188) than what we have for `TimePattern`.

It's not clear to me what the pieces of text inside `compoundTimeString` are supposed to be. As far as I can tell from the RELAX NG schema, they could be anything at all, though that seems weird. I mean, would Rocoto really accept, say,
```
<dependency>
  <timedep>
    Why
    <cyclestr offset="12:05">Hello</cyclestr>
    There
  </timedep>
</dependency>
```
It looks like its schema would accept that, but maybe Rocoto would crash. Who knows? Seems weird. The Rocoto docs say
> The value between the start and end of the <timedep> tag is a time in yyyymmddhhmmss format.
but I don't think that's enforced by its schema, if I'm reading the RELAX NG code correctly. And the example code where that quote appears is
```
<timedep><cyclestr offset="&DEADLINE;">@Y@m@d@H@M@S</cyclestr></timedep>
```
so I'm not sure `<timedep>` is really meant to be valid without `<cyclestr>`, whatever the schema says.

I think for this PR a `<timedep>` that accepts whatever text content should be fine. But it doesn't have an `offset` attribute as far as I can tell -- that belongs to the `<cyclestr>` tag, and we'll have to deal with that soon enough, as there's a separate Jira ticket for that. But our JSON Schema `TimePattern` is definitely not the same thing as the RELAX NG schema's `compoundTimeString`.

NaureenBharwaniNOAA: Yes I agree, the `TimePattern` is distinctly different than the `compoundTimeString`. Currently we have ticket UW-386, which I think would need to address much of the needs you just outlined for us.

Then I think what we are talking is if we want `timedep` to accept any text content than we would need to change the schema to use a reference definition of `anyStringPattern`. For example I would think we would update the lines highlight on this comment to be:

```
          "properties": {
            "attrs": {
                  "$ref": "#/$defs/anyStringPattern"
              },
```

where `anyStringPattern` would be:
```
    "anyStringPattern": {
      "pattern": ".+", --> This would match any string content as long as it has 1 character in it
      "type": "string"
    },
```

src/uwtools/tests/test_rocoto.py
maddenp-noaa:
```
(Pdb) ll
132  	    @pytest.mark.parametrize(
133  	        "config", [{"datadep": {"attrs": {"age": "120"}}}, {"timedep": {"attrs": {"offset": "&DEADLINE;"}}}]
134  	    )
135  	    def test__add_task_dependency_operand(self, config, instance, root):
136  	        breakpoint()
137  	        instance._add_task_dependency_operand(e=root, config=config)
138  	        element = root[0]
139  	        for tag, attrs in config.items():
140  ->	            assert tag == element.tag
141  	            for attr, val in attrs.items():
142  	                assert element.get(attr) == val
(Pdb) p tag, attrs
('datadep', {'attrs': {'age': '120'}})
```
I think `attrs` isn't what's needed. Try
```
% git diff -U0
diff --git a/src/uwtools/tests/test_rocoto.py b/src/uwtools/tests/test_rocoto.py
index a73272f..f48e606 100644
--- a/src/uwtools/tests/test_rocoto.py
+++ b/src/uwtools/tests/test_rocoto.py
@@ -140 +140 @@ class Test_RocotoXML:
-            for attr, val in attrs.items():
+            for attr, val in attrs["attrs"].items():
```

src/uwtools/resources/rocoto.jsonschema
```diff
@@ -161,7 +161,15 @@
...
         },
         "^timedep(_.*)?$": {
-          "$ref": "#/$defs/TimePattern"
+          "additionalProperties": false,
+          "properties": {
+            "attrs": {
+              "additionalProperties": false,
+              "properties": {
+                "$ref": "#/$defs/TimePattern"
+              }
```
NaureenBharwaniNOAA, line 170: Removed the `offset` attr on `timedep` to currently leave it open to a `TimePattern`, but once we have more discussion we can update this to a definition reference that only references any string pattern instead. I also added the `attrs` block in here above the definition reference. Passes the schema fine currently but I think we will need to do more tests if to confirm this is accurately doing what we want. Looking at all our tests for our jsonschema in `test_validator` our configs only go up until the `taskdep` `<dependency>` element.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: Looking at the json schema there are only three pre-defined definitions that are `TimePattern`, `datePattern`, `dateTimePatten`. 

Another approach here would be leaving it as:
```
        "^timedep(_.*)?$": {
          "type": "string"
        }
```
This would mean that whatever `attrs` are within `timedep` will just be treated as a single string. Seemed to be the most simple way to handle this case.

NaureenBharwaniNOAA: Also if we decide to go this route, the `TimePattern` definition is only being used by the `timedep` tag. It might be worth leaving changing `timedep` to the above example, and consider remving `TimePattern` for now, which could in the future come back pending how `<cyclestr>` is handled.

maddenp-noaa: I like this idea, because I think `TimePattern` may already not be fully compatible with the RELAX NG schema's definition. If `TimePattern` isn't used elsewhere, I like the simplification of removing it now and then re-implementing it when we add `<cyclestr>` support, which will need it. We can then revisit the RNG schema to guide implementation. At that point, I think we should also consider just making our JSON Schema `timedep` a reference to whatever we call our implementation of `compoundTimeString`, which is how the RNG schema defines it.

NaureenBharwaniNOAA: I like this idea too. It gives more flexibility on how to handle `<cyclestr>` later to who will be working the ticket covering this discussion.

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,44 @@ def test__add_task_dependency_fail(self, instance, root):
...
+    @pytest.mark.parametrize(
+        "config", [{"and": {"datadep": {"attrs": {"age": "120"}}, "timedep": {"attrs": {"offset": "&DEADLINE;"}}}}]
+    )
+    def test__add_task_dependency_operator(self, config, instance, root):
+        instance._add_task_dependency_operator(e=root, config=config)
+        element = root[0]
+        for tag, attrs in config.items():
+            assert tag == next(iter(config))
+            #for attr, val in attrs.items():
+            #    assert element.get(attr) == val
```
NaureenBharwaniNOAA, line 153: Working on this final test case. Since the beginning key in the dict will not be `datadep`, `timedep`, or any other tags that will end the recursion, it seems that when it grabs `element = root[0]` it runs into a problem checking the rest of the `tags`/`attrs` later on. 

This gets even worse when nesting more operators ie..  `and` and `or` before we find one of these tags that stop the recursion. I think most of my issue is coming from not being able to properly manipulate the `element` variable in the debugger so I haven't been able to track down exactly what is happening.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,60 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+                self._add_task_dependency_operator(e, config={key: block})
+            elif tag in [STR.streq, STR.strneq]:
+                self._add_task_dependency_strequality(e, config={key: block})
+            else:
+                raise UWConfigError("Unhandled dependency type %s" % tag)
+
+    def _add_task_dependency_operand(self, e: Element, config: dict) -> None:
+        """
+        ???
+        """
```
christinaholtNOAA, line 199: If you pass in `{tag: block}` here instead of `{key: block}`, it might simplify what you need to reprocess in the helper method (suggestion is untested):
```
def _add_task_dependency_operand(self, e: Element, config: dict) -> None:
    for tag, block in config.items():
        self._set_attrs(SubElement(e, tag), block)
```

Edit: If you passed `key` and `block` as two separate arguments, it would simplify even further to:

```
def _add_task_dependency_operand(self, e: Element, key, block):
    self._set_attrs(SubElement(e, tag), block)
```

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,60 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+                self._set_attrs(SubElement(e, STR.timedep), block)
+
+    def _add_task_dependency_operator(self, e: Element, config: dict) -> None:
+        """
+        ???
+        """
+        operands = (STR.datadep, STR.taskdep, STR.timedep)
+        operators = (STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor)
+        strequality = (STR.streq, STR.strneq)
+
```
christinaholtNOAA, line 216: I'd imagine you'd want to add a subelement to `e` from `config` contents here. For example, if this method gets called from `_add_task_dependency`, `e` will be the top level `dependency` tag. If config contains an operator, we'd need to add that operator as a subelement to `dependency`.

Then, once the subelement is added, we need to go through the `block` to see how the next elements should be handled. So in the next line (217), you'd loop over `block` instead of `config`.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,60 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
         :param config: Configuration data for this element.
         """
         e = SubElement(e, STR.dependency)
+        for key, block in config.items():
+            tag, _ = self._tag_name(key)
+            if tag in [STR.taskdep, STR.datadep, STR.timedep]:
+                self._add_task_dependency_operand(e, config={key: block})
```
christinaholtNOAA, line 188: Given some of my comments on how the data you're passing to these methods might be used, I'm curious if it makes more sense to skip the `config` part, and pass `key` and `block` as two separate arguments.

christinaholtNOAA (commented):
Just a few suggestions/thoughts below following up on discussion in SRW Group today.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,60 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        ???
+        """
         for key, block in config.items():
             tag, _ = self._tag_name(key)
             if tag == STR.taskdep:
                 self._set_attrs(SubElement(e, STR.taskdep), block)
+            elif tag == STR.datadep:
+                self._set_attrs(SubElement(e, STR.datadep), block)
+            else:
+                self._set_attrs(SubElement(e, STR.timedep), block)
```
maddenp-noaa, line 207: Is this whole `if` block equivalent to
```
self._set_attrs(SubElement(e, tag, block))
```
? We needed the `if` previously because we weren't ready to deal with all the possibilities, but aren't we now?

NaureenBharwaniNOAA: I incorporated much of these changes/comments and simplified even further. I think it can still be breakdown more. I'm pushing what I currently have at the moment. Simplifying the methods down further leads me to edit the tests quite a bit and I wanted some opinions on which way to take those. I'll comment on that shortly as well.

Happy to take even more feedback since it'll be a continuous pairing down the methods/tests at this point. After adding much of your suggestions I was able to get back to 100% coverage as well. Thank you! @christinaholtNOAA @maddenp-noaa

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,47 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
         :param config: Configuration data for this element.
         """
         e = SubElement(e, STR.dependency)
-        for key, block in config.items():
-            tag, _ = self._tag_name(key)
```
NaureenBharwaniNOAA, line 186: I took this line out on all following methods, but reading a bit more into this, I think it would be useful to add this line back in on my methods.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,47 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
-            if tag == STR.taskdep:
-                self._set_attrs(SubElement(e, STR.taskdep), block)
+        for tag, block in config.items():
+            if tag in [STR.taskdep, STR.datadep, STR.timedep]:
+                self._add_task_dependency_operand(e, tag=tag, block=block)
+            elif tag in [STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor]:
+                self._add_task_dependency_operator(e, config={tag: block})
```
NaureenBharwaniNOAA, line 189: I left this as `config` because when iterating over the `config` in the operand method, we will only want to be passing the net `block` to the recurse the method again. We do not want to be passing every `tag` again or this leads to a never ending loop. However, we need to pass the tag to add to the `dependency` tree as Christina pointed out.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,47 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        """
+        ???
+        """
+        operands = (STR.datadep, STR.taskdep, STR.timedep)
+        operators = (STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor)
+        strequality = (STR.streq, STR.strneq)
+
+        for tag, block in config.items():
+            if tag in operators:
+                e = SubElement(e, tag)
```
NaureenBharwaniNOAA, line 211: Here is that call to add a subelement to `e`.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,47 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+                self._add_task_dependency_operator(e, config=block)
```
NaureenBharwaniNOAA, line 212: This is the recursive call I mentioned, where we want only to supply the `block` to the `operator` method and not the `tag` again.

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,70 @@ def test__add_task_dependency_fail(self, instance, root):
...
             instance._add_task_dependency(e=root, config=config)
 
+    def test__add_task_dependency_and(self, instance, root):
+        config = {"and": {"or": {"datadep": {"attrs": {"age": "120"}}}}}
+        instance._add_task_dependency(e=root, config=config)
+        dependency = root[0]
+        assert dependency.tag == "dependency"
+        and_ = dependency[0]
+        assert and_.tag == "and"
+        # assert and_.get("age") == "120"
```
NaureenBharwaniNOAA, line 139: Still working out this test case. With the updates I made to the methods, I believe I have to do a similar thing to this test as I did to `test__add_task_dependency_operand`, where I need to loop over `for attr, val in block["attrs"].items():`. I currently get `None` back as the error message for `and_.get("age")`

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,70 @@ def test__add_task_dependency_fail(self, instance, root):
...
+
+    @pytest.mark.parametrize(
+        "config",
+        [
+            ("streq", {"attrs": {"left": "&RUN_GSI;", "right": "YES"}}),
+            ("strneq", {"attrs": {"left": "&RUN_GSI;", "right": "YES"}}),
+        ],
+    )
+    def test__add_task_dependency_strequality(self, config, instance, root):
+        tag, block = config
```
NaureenBharwaniNOAA, line 189: This is the second option I mentioned above.

src/uwtools/rocoto.py
christinaholtNOAA: I see what you mean. Since this is recursive, the dicts may be larger in subsequent calls than they are on this first call, so make sure it can handle an entire full dict.

christinaholtNOAA (commented):
This is looking great. I'd suggest moving to a full PR.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,56 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
         :param config: Configuration data for this element.
         """
         e = SubElement(e, STR.dependency)
-        for key, block in config.items():
-            tag, _ = self._tag_name(key)
-            if tag == STR.taskdep:
-                self._set_attrs(SubElement(e, STR.taskdep), block)
+        for tag, block in config.items():
+            tag, _ = self._tag_name(tag)
+            if tag in [STR.taskdep, STR.datadep, STR.timedep]:
```
maddenp-noaa, line 187: The same collection exists -- here as a `list`, there as a `tuple` -- in `__add_task_dependency_operator()`. It would be good to DRY that out as either a private class constant or a private `@property` method. Same for the other categories just below.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,56 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+        """
+        :param e: The parent element to add the new element to.
+        :param tag: Configuration new element to add.
+        :param block: Configuration data for the tag.
+        """
+        self._set_attrs(SubElement(e, tag), block)
+
+    def _add_task_dependency_operator(self, e: Element, config: dict) -> None:
+        """
+        Add an operator  element to the <dependency>.
```
maddenp-noaa, line 206: nitpick: double space between `operator` and `element`

src/uwtools/rocoto.py
maddenp-noaa: @christinaholtNOAA How would the dicts get larger? Aren't we processing smaller and smaller subtrees? Or do you just mean that the number of keys at the top level might be larger?

src/uwtools/rocoto.py
maddenp-noaa: Is an assignment ok here? It looks to me like `SubElement()` returns the newly-created element so, in a `for` loop like this, wouldn't these calls create a chain of elements, rather than attaching multiple child elements to a single parent? I mean, in pseudocode, something like
```
e is E1
e = SubElement(e, tag2)
e is now E2, and we have E1 -> E2
e = SubElement(e, tag3)
e is now E3, and we have E1 -> E2 -> E3
etc.
```
Without the assignment, the identify of `e` wouldn't change and the loop would just add (potentially) multiple children to `e`.

Should this statement perhaps relocate to `_add_task_dependency_operator()`? Other methods' (`_add_metatask()`, `_add_task()`, `_add_task_dependency()`, etc.) bodies start with this statement, to create the child element and to redefine `e` _for the duration of the method call_ for subsequent processing.

src/uwtools/rocoto.py
```diff
@@ -182,13 +182,56 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+            tag, _ = self._tag_name(tag)
+            if tag in operators:
+                e = SubElement(e, tag)
+                self._add_task_dependency_operator(e, config=block)
+            elif tag in operands:
+                self._add_task_dependency_operand(e, tag=tag, block=block)
+            elif tag in strequality:
+                self._add_task_dependency_strequality(e, tag=tag, block=block)
+            else:
+                raise UWConfigError("Unhandled dependency type %s" % tag)
```
maddenp-noaa, line 225: I think it's probably a good idea to keep this check here, but just to be sure: We think we're handling all dependency types as of this PR, right? Our schema checks should ensure that we don't get invalid inputs, but a final check never hurts.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I think here I included within the for loop because of how the looping works. We could start the method with a call to add each of these elements to redine `e` and that would require passing the tag along to the method as we do at the beginning of the method call for others as you pointed out. Then ultimately we would discard the rest of the usage for `tag` since for recursion we only care about the `block`.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: I added this check here and test case for raising an error as well. We should now be handling all dependency types. Happy to remove if we decide that our schema should catch this instead. I added a test case for this to confirm 100% coverage.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: Should be fixed with the latest push.

src/uwtools/rocoto.py
```diff
@@ -174,21 +174,82 @@ def _add_task(self, e: Element, config: dict, taskname: str) -> None:
...
+    def _dependency_constants(self):
+        """
+        Returns a tuple for each of the resepctive dependency types.
+        """
+        operands = (STR.datadep, STR.taskdep, STR.timedep)
+        operators = (STR.and_, STR.nand, STR.nor, STR.not_, STR.or_, STR.xor)
+        strequality = (STR.streq, STR.strneq)
+
+        return operands, operators, strequality
+
```
NaureenBharwaniNOAA, line 187: I added this `@property` method to address DRY out comment. I passed the return value from the call in `_add_task_dependency` to `_add_task_dependency_operator` to reuse this information.

src/uwtools/rocoto.py
maddenp-noaa: OK, I understand now why you have the `SubElement()` call here, in the body of the loop, but I think the assignment is a bug, which maybe our test data would only exercise if it contained operators nested deeply enough. I think we could maybe just
```
self._add_task_dependency_operator(SubElement(e, tag), config=block)
```
instead of having the `e = SubElement(e, tag)` statement. That would have the effect of 1. Adding the child element, and 2. Passing that child element down in the recursive call, but would not change the identity of `e` within the `for` loop, which is what I think we wouldn't want to do. What do you think?

Also, do we need to pass `dependency_constants` as an argument? I think maybe not -- all methods should be able to access `self._dependency_constants` the same way `_add_task_dependency()` does.

src/uwtools/rocoto.py
maddenp-noaa: I think it's good to keep it. Maybe someday we can arrange things so we're confident that all config data passes through schema validation but, for example, and at the moment, users can monkey with the content of `Config` objects by accessing its `data` member directly, which would bypass schema validation.

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: I haven't been able to figure this one out. Since `attrs` is nested deeper into the config I attempted to use `next(iter(config))` but it began to look quite messy.

src/uwtools/resources/rocoto.jsonschema
NaureenBharwaniNOAA: I'm thinking that we might need to add this as well to the Rocoto XML `<cyclestr>` handling ticket. However, I think at this point we will need to split UW-386 and decide what should be done in each appropriately. Something I will bring up in backlog grooming tomorrow.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: I'll fetch your branch and take a look.

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,70 @@ def test__add_task_dependency_fail(self, instance, root):
...
+    def test__add_task_dependency_streq(self, instance, root):
+        config = {"streq": {"attrs": {"left": "&RUN_GSI;", "right": "YES"}}}
+        instance._add_task_dependency(e=root, config=config)
+        dependency = root[0]
+        assert dependency.tag == "dependency"
+        streq = dependency[0]
+        assert streq.tag == "streq"
+        assert streq.get("left") == "&RUN_GSI;"
+
+    # pylint: disable=line-too-long
```
maddenp-noaa, line 149: I removed this line and the other on line 163 and all tests pass, so I think these are not needed.

Also, there's special `pylint` syntax (`pytlint: disable-next=`) for suppressing a complaint about just _the following_ line, and that's what we'd probably want to put on a line that we just couldn't wrap. A directive like this will suppress the warning for everything in the scope, which in this case would, I think, mean the entire `Test_RocotoXML` class.

NaureenBharwaniNOAA: I removed those as well. I originally used them to work on my test cases and now that they'll _almost_ finalized we can remove them. I'll push those changes now and make this a full PR.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: Updated this in the newest push.

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: All comments should be addressed now and this is the final outstanding issue.

src/uwtools/rocoto.py
```diff
@@ -174,21 +174,75 @@ def _add_task(self, e: Element, config: dict, taskname: str) -> None:
         if STR.dependency in config:
             self._add_task_dependency(e, config[STR.dependency])
 
+    @property
+    def _dependency_constants(self):
+        """
+        Returns a tuple for each of the resepctive dependency types.
```
NaureenBharwaniNOAA, line 180: Fixed this and will be in the next push.

NaureenBharwaniNOAA:
General comment on the PR as a whole, after adding new methods and test cases and taking a look at the code as a whole, what do we think about ordering the methods for both `uwtools.rocoto` and `uwtools.tests.test_rocoto` lexicographically?

It made sense to originally order them based on where they appear in the workflow, and now I think this makes it harder to navigate for finding the correct method/test. Especially for someone who isn't familiar with the workflow ordering, this could be irrelevant to them.

src/uwtools/tests/test_rocoto.py
maddenp-noaa: `age` is not an attribute of `and`, it's an attribute of `datadep`, which is two levels deeper:
```
(Pdb) ll
132  	    def test__add_task_dependency_and(self, instance, root):
133  	        config = {"and": {"or_get_obs": {"datadep": {"attrs": {"age": "120"}}}}}
134  	        breakpoint()
135  	        instance._add_task_dependency(e=root, config=config)
136  	        dependency = root[0]
137  	        assert dependency.tag == "dependency"
138  	        and_ = dependency[0]
139  	        assert and_.tag == "and"
140  ->	        assert and_.get("age") == "120"
(Pdb) p and_.getchildren()[0].getchildren()[0].get("age") == "120"
True
```

src/uwtools/tests/test_rocoto.py
```diff
@@ -129,6 +129,67 @@ def test__add_task_dependency_fail(self, instance, root):
...
+            {"and": {"or": {"datadep": {"attrs": {"age": "120"}}}}},
+            {"and": {"strneq": {"attrs": {"left": "&RUN_GSI;", "right": "YES"}}}},
+        ],
+    )
+    def test__add_task_dependency_operator(self, config, instance, root):
+        instance._add_task_dependency_operator(e=root, config=config)
+        for tag, _ in config.items():
+            assert tag == next(iter(config))
+
+    def test__add_task_dependency__operand_fail(self, instance, root):
```
NaureenBharwaniNOAA, line 173: Extra `_` has been removed and will be in the next push.

maddenp-noaa:
My rule of thumb is: When in doubt: sort. It sounds like you're in doubt :)

src/uwtools/tests/test_rocoto.py
NaureenBharwaniNOAA: Interesting use of the `getchildren()` attr on the `and_` tag. I was doing something similar with:

``` next(iter(and_))[0]```

I'll try this instead for my test case. It'll also be easier to breakdown in the long run vs. nested `next(iter(...))` which were looking messy.

Comment (maddenp-noaa) [reply src/uwtools/tests/test_rocoto.py]: If nothing else, `.getchildren()` might make the intent a little clearer.

File: src/uwtools/tests/test_rocoto.py [hunk 35]
```diff
@@ -124,11 +105,84 @@ def test__add_task_dependency(self, instance, root):
...
         assert taskdep.get("task") == "foo"
 
+    def test__add_task_dependency_and(self, instance, root):
+        config = {"and": {"or_get_obs": {"datadep": {"attrs": {"age": "120"}}}}}
+        instance._add_task_dependency(e=root, config=config)
+        dependency = root[0]
+        assert dependency.tag == "dependency"
+        and_ = dependency[0]
+        assert and_.tag == "and"
+        assert and_.getchildren()[0].getchildren()[0].get("age") == "120"
```
NaureenBharwaniNOAA, line 115: Added `getchildren()` solution in here and looks clearer as suggested.

src/uwtools/rocoto.py
```diff
@@ -244,6 +264,39 @@ def _add_workflow_tasks(self, e: Element, config: dict) -> None:
...
+
+        :param config_file: Path to the YAML config (defaults to stdin).
+        """
+        if not validate_yaml(
+            config_file=config_file, schema_file=resource_pathobj("rocoto.jsonschema")
+        ):
+            raise UWConfigError("YAML validation errors identified in %s" % config_file)
+
+    @property
+    def _dependency_constants(self):
```
maddenp-noaa, line 279: I wouldn't mind seeing a `-> Tuple[Tuple, Tuple, Tuple]` annotation for the return type on the method signature. It looks a bit silly, but it's accurate.

maddenp-noaa (approved):
I made one final request for a type annotation, but everything otherwise looks great to me. Nice work.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: That brings up an excellent point. We are importing `Tuple` like so ```from typing import Optional, Tuple```. The return type on the signature I can update to ```-> Tuple[Tuple, Tuple, Tuple]``` accordingly. 

Previously when I was passing `dependency_constants` in as an argument I used `tuple` as the input type. I wanted to confirm that here the difference between `Tuple` vs. `tuple` is lower case `tuple` is being used a type in python. Whereas `Tuple` is imported from the typing module. I was trying to find more context to why we would prefer one over the other.

I'll update that return type now.

src/uwtools/rocoto.py
```diff
@@ -181,14 +158,57 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+            if tag in operators:
+                self._add_task_dependency_operator(SubElement(e, tag), config=block)
+            elif tag in operands:
+                self._add_task_dependency_operand(e, tag=tag, block=block)
+            elif tag in strequality:
+                self._add_task_dependency_strequality(e, tag=tag, block=block)
             else:
                 raise UWConfigError("Unhandled dependency type %s" % tag)
 
+    def _add_task_dependency_strequality(self, e: Element, tag: str, block: dict) -> None:
```
elcarpenterNOAA, line 204: Can we sort all of the arguments alphabetically as well?

elcarpenterNOAA (approved):
I'd also like to see type checking added to `_dependency_constants()` as well, but otherwise LGTM.

src/uwtools/rocoto.py
maddenp-noaa: You can use the built-in type names (`dict`, `list`, `tuple`, etc.) as types as long as you don't want to be more specific than that -- i.e. you're ok with "a dict of anything at all", "a list of anything at all", etc.

The ones imported from `typing` let you add square brackets and more precise information. So, for example, we could just say `-> tuple` or `-> Tuple`, which would be equivalent, but would mean that the tuples could contain anything; or `Tuple[tuple, tuple, tuple]` (lower-case, built-in versions this time) would be equivalent to what I suggested.

Or we could go even more specific: `Tuple[Tuple[str, str, str], Tuple[str, str, str, str, str, str], Tuple[str, str]]`. That would say everything there is to say, down to how many `str` elements each tuple should have, but seemed a bit much to me.

I think there's also a middle ground like `Tuple[Tuple[str, ...], Tuple[str, ...], Tuple[str, ...]]` which at least says that all the inner tuples must contain only strings, but we're not staying how many. For convenience, we might define a custom type, like `StrTuple = Tuple[str, ...]`, and then we could say `Tuple[StrTuple, StrTuple, StrTuple]`.

I'm not sure what's best.

src/uwtools/rocoto.py
maddenp-noaa: I have to (partially) oppose that suggestion, with rationale: It's a common API design strategy to present arguments in a consistent way so that users have a good idea what to expect. For example, if you've got a collection of functions/methods that all operate on the same kind of thing, but maybe with various extra arguments specific to each's functionality, a consistent design would keep the common argument(s) first in the argument list, so that that would be an invariant across the whole collection. That's (mostly? completely?) the case here, where we've got a bunch of methods that all operate on `Element` objects. The `Element` argument should come first, because it's the invariant across all the methods. If there were multiple arguments that are used across all the calls, sorting those might make sense, though it might also make sense to order them according to some kind of abstract hierarchy -- e.g. if you had functions that all did things related to flowers and their colors, you might have `foo(flower, color, ..., ...)` with `flower` and `color` in the "wrong" order because "flower" is somehow a larger or more crucial concept the "color" for whatever domain we're working in.

I want to write something more, but am going to submit this first to make sure, hopefully, that we don't merge this PR with the `Element` argument `e` demoted...

maddenp-noaa: Another reason for ordering arguments in a specific way is to use what in functional programming is called "currying" or "partial application". Python supports this with the `partial()` function from the `functools` library. Currying allows you to turn a function with `n` arguments into a function with `n-1` arguments, where the first argument is fixed by the currying operation. For example, in Python, you might have
```
>>> def flower(name, color):
...     print(f"This {name} is {color}.")
... 
>>> flower("rose", "red")
This rose is red.
```
But if you wanted to do a lot of things with roses, you might
```
>>> from functools import partial
>>> rose = partial(flower, "rose")
>>> rose("white")
This rose is white.
>>> rose("yellow")
This rose is yellow.
```
This pattern is very common in functional programming, and quite useful in Python sometimes. Success, though, requires that arguments be ordered in such a way that you can "fix" the leading argument with `partial()` in a useful way, such that you usually want the argument(s) representing the "main concept(s)" to come first. You can, of course, curry multiple times to derive more and more specific functions.

So, fully ordering argument lists lexicographically may not make sense is some cases, and I think our case here is one of those. Past the initial `Element` argument I'm not sure I care, but if e.g. every method take `e`, and then _a lot_ of elements take a `config`, and just _some_ methods take further arguments, `config` might similarly be such a common concept that users will find it ergonomic to see `f(e, config, ... other args ...)`

src/uwtools/rocoto.py
```diff
@@ -181,14 +158,57 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
             else:
                 raise UWConfigError("Unhandled dependency type %s" % tag)
 
+    def _add_task_dependency_strequality(self, e: Element, block: dict, tag: str) -> None:
+        """
+        :param e: The parent element to add the new element to.
+        :param block: Configuration data for the tag.
+        :param tag: Configuration new element to add.
+        """
+        self._set_attrs(SubElement(e, tag), block)
```
christinaholtNOAA, line 210: It's interesting that this and the operand method both do exactly the same thing. Do we want to consider combining those?

christinaholtNOAA (approved):
This looks good to me. Just one question below. Do with that what you like.

src/uwtools/rocoto.py
NaureenBharwaniNOAA: Pushing now that updates these changes.

src/uwtools/rocoto.py
```diff
@@ -181,14 +158,49 @@ def _add_task_dependency(self, e: Element, config: dict) -> None:
...
+            elif tag in operators:
+                self._add_task_dependency_operand_operator(e, config={tag: block})
+            elif tag in strequality:
+                self._add_task_dependency_strequality(e, block=block, tag=tag)
+            else:
+                raise UWConfigError("Unhandled dependency type %s" % tag)
+
+    def _add_task_dependency_operand_operator(self, e: Element, config: dict) -> None:
+        """
+        Add an operator element to the <dependency>.
```
maddenp-noaa, line 177: Docstring is now out-of-date.

PR #342, opened 2023-11-09
ci-improvements

maddenp-noaa:
In this PR:

- We recently extended `condev-shell` to allow running single commands in a development shell instead of getting an interactive shell. We initially used that to run the formatting checks. Here, we extend our use to support running our code-quality tests on each supported Python release, without requiring separate package builds for each of those reelases. While there seems to be a lot of variability in GitHub Actions run times for the same processes, such that it's hard to know what to expect on average without a lot of similar runs, anecdotally this update appears to reduce the the `Test` CI workflow's duration by something between 25% and 50%.
- We had previously been using the `conda_build_config.yaml` "variants" file to build one conda package for each supported Python release -- not for the benefit of package users, but to ensure that our tests (which are executed as part of the package-build process) ran against each Python release. Now that the tests use `condev-shell` to exercise the supported Python releases, we can build a single conda package that supports a _range_ of Python releases.


christinaholtNOAA (approved):
LGTM.

recipe/meta.json
maddenp-noaa, line 1: This file is updated by running `make meta`, and is based on `meta.yaml` -- see below.

.github/scripts/test.sh
```diff
@@ -0,0 +1,23 @@
+# Actions invokes script with: bash -e <script>
+
```
maddenp-noaa, line 2: Rather than running `make package` to run the code-quality tests, run `make test` in a development shell, using a technique similar to what we used in `format-check.sh`.

Here, the development shell gets re-used for each Python release to test, but with the correct Python interpreter version installed for each test, and the `uw` entry-point script updated to point to the new Python interpreter. 

(Unfortunately, `setuptools` hardcodes the full path to the Python interpreter in the entry-point scripts, e.g. `#!/home/maddenp/sw/conda/envs/DEV-uwtools/bin/python3.11`, so that has to be updated via the `pip` command seen below when the Python version changes.)

[Here is a sample CI run of this script.](https://github.com/ufs-community/workflow-tools/actions/runs/6815371576/job/18534320286#step:4:1)

.github/scripts/install-conda.sh
```diff
@@ -1,11 +1,12 @@
 # Actions invokes script with: bash -e <script>
```
maddenp-noaa, line 2: Just a bit more explicit: Use the conda install location specified in `common.sh`; do not rely on `cd`, but rather use explicit, absolute paths; and expand conda flags for better readability.

.github/scripts/format-check.sh
```diff
@@ -12,7 +12,7 @@ unformatted() {
   return 0
 }
 
-source /tmp/conda/etc/profile.d/conda.sh
+source $(dirname ${BASH_SOURCE[0]})/common.sh
+source $CONDADIR/etc/profile.d/conda.sh
```
christinaholtNOAA, line 16: Does it make sense to add this full path to conda.sh to the common file, too? It might reduce some duplication since we have this new file that helps with that.

christinaholtNOAA (approved):
Just one nitpick. LGTM still.

.github/scripts/format-check.sh
maddenp-noaa: Good idea, will do.

elcarpenterNOAA (approved):
LGTM!

NaureenBharwaniNOAA (approved):
Nice job. Looks good.

WeirAE (approved):
Bonus would be adding the full path to conda.sh as Christina recommended, but LGTM

PR #343, opened 2023-11-14
Update PR template

maddenp-noaa:
Simplify PR template and add emphasis to documentation. Add an update type for refactoring efforts. See inline comments.


.github/pull_request_template.md
```diff
@@ -1,61 +1,36 @@
...
-- [ ] Bug fix (non-breaking change which fixes an issue)
-- [ ] New feature (non-breaking change which adds functionality)
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- [ ] This change requires a documentation update
+Select one or more:
 
-**How Has This Been Tested?**
+- [ ] Bug fix (corrects a known issue)
+- [ ] Code maintenance (refactoring, etc.)
+- [ ] New feature (adds functionality)
```
maddenp-noaa, line 23: Does this accurately capture the kinds of updates we make? In particular, the `Code maintenance` option seems to me to describe many PRs: It improves the code base without either adding functionality or fixing issues.

.github/pull_request_template.md
```diff
@@ -1,61 +1,36 @@
...
-- [ ] Any dependent changes have been merged and published
+**Documentation**
 
-<!-- 
- If there is a co-author attribution:
+Affirm:
 
-Update the final commit message (when squashing and merging) to include  `Co-authored-by: name <name@example.com>` (e.g. `Co-authored-by: Jane Doe <jane_doe@example.com>`).  Each co-author must have their own line, and the email address used must be the email address connected with their GitHub account.
--->
+- [ ] I have reviewed the documentation and have made any updates necessitated by this change.
```
maddenp-noaa, line 36: I hope that, over the course of the next couple sprints, the docs will more accurately reflect the current state of the project, such that this check will be pertinent.

.github/pull_request_template.md
```diff
@@ -1,61 +1,36 @@
...
-- [ ] I have performed a self-review of my own code
-- [ ] I have commented my code, particularly in hard-to-understand areas
-- [ ] My changes need updates to the documentation.  I have made corresponding changes to the documentation
-- [ ] My changes generate no new warnings
-- [ ] New and existing tests pass with my changes
-- [ ] Any dependent changes have been merged and published
+**Documentation**
 
-<!-- 
- If there is a co-author attribution:
```
maddenp-noaa, line 58: Apart from being polite, what does/did this accomplish? I couldn't find any observable effect. It would, of course, be nice to acknowledge help from colleagues, but I'm not sure that merits an explicit PR step. Does anyone know differently?

.github/pull_request_template.md
```diff
@@ -1,61 +1,36 @@
...
+- [ ] This is a non-breaking change (existing functionality continues to work as expected)
+- [ ] This is a breaking change (changes existing functionality)
 
-- [ ] My code follows the style guidelines of this project
-- [ ] I have performed a self-review of my own code
-- [ ] I have commented my code, particularly in hard-to-understand areas
-- [ ] My changes need updates to the documentation.  I have made corresponding changes to the documentation
-- [ ] My changes generate no new warnings
-- [ ] New and existing tests pass with my changes
-- [ ] Any dependent changes have been merged and published
```
maddenp-noaa, line 55: - Some of this is enforced by code-quality checks at this point.
- Style guidelines -- at least those that can't be enforced by by automated tools -- are more, IMO, a subject for peer review, as style is so subjective and case-specific.
- Documentation requirements has its own sections now.
- I've never understood the "dependent changes" item.

.github/pull_request_template.md
NaureenBharwaniNOAA: I feel like this was mostly relevant when there was a constant collaboration on tickets. i.e. Emily and I working together on tickets during our time on the team as interns, same can be said for Janet and Emily working as interns on the team and for future intern collaboration. Also was useful for the first 2 months when Emily and I were new full-time developers with the team.

For current purposes, I agree it's nice to be polite and if discussion is needed on a closed ticket where the main author has the left the team but the co-author has not.

.github/pull_request_template.md
NaureenBharwaniNOAA: I assumed dependent changes meant task A is merged before task B because it is needed to pass the current test cases, etc. This would make sense if test coverage wasn't 100%. I agree with your ideas of removing these since they are covered in other checks now.

.github/pull_request_template.md
maddenp-noaa: So do you think the goal here was just to record in the PR who else worked on it? I got the feeling that maybe there was some kind of automation that was processing this, given the explicit instructions of exact format, one co-author per line, email address has to be the GitHub-registered email address, etc. Why all that specificity otherwise?

NaureenBharwaniNOAA: Yes, I would think that the goal in the past was to record in the PR who else had worked on it. I believe the intended goal is/was to push your final commit to GitHub with the co-author as an option included on the command line. I think this was an idea implemented during the spring semester right before we were brought on to the team.

.github/pull_request_template.md
WeirAE: Hm, the other question is whether this is redundant with the git labels - which we definitely should be more strict about if we cut this out

.github/pull_request_template.md
```diff
@@ -1,61 +1,36 @@
...
 
-<!-- Please describe the tests that you ran to verify your changes. -->
-<!-- Provide instructions so we can reproduce. -->
-<!-- Please also list any relevant details for your test configuration -->
+**Impact**
 
-<!-- Use the following as a guide to list your tests and delete options that are not relevant. Expand as necessary. -->
-<!--
-- [ ] pytests in GitHub actions.
-- [ ] Tests on XYZ OS/HPC/CSP
```
WeirAE, line 44: This was somewhat left over from early evaluations of platforms, which we may need to re-include if we have to do any manual platform tests outside of Jenkins

.github/pull_request_template.md
WeirAE: Yep - like a lot of the other comments, they were at times more for user onboarding and defining the standards than providing information to reviewers. I think we're at a good place to transition to more external-facing options

WeirAE (approved):
Yep, all good changes for where we're at now, thanks for streamlining this!

.github/pull_request_template.md
maddenp-noaa: That's interesting. If we wanted to go that way, we could replace this list with an `Affirm` checkbox stating that you've set the label to one of, say, three values. That seems like a good idea if we can get team buy-in.

.github/pull_request_template.md
maddenp-noaa: This is a good point, though I can't immediately see why there would be anything platform-specific in UW that would require testing outside of Actions. Even the kinds of integration tests I think we'd eventually want to to run at least on `develop` -> `main` merges to create releases should probably be testable in a generic Linux container like Actions provides. Integration tests for downstream apps are where I'd expect to see platform concerns. Please point out anything I might be missing...

.github/pull_request_template.md
maddenp-noaa: I thought of a couple more potential categories (because I was wondering how _this_ PR might be tagged):

- Documentation updates
- Tooling (CI, packaging, revision control, code-quality tools, etc.)

.github/pull_request_template.md
WeirAE: True! That would certainly be for integration tests. I think some will have to be run on specific tier 1 machines rather than just a generic container. For what its worth, SRW PRs ask which specific machines were tested on (which was simplified here), while the main UFS specifically asks "I have run the full RT suite on either Hera/Cheyenne AND have attached the log to this PR below this line"

.github/pull_request_template.md
elcarpenterNOAA: Janet and I would alternate who did the driving ticket by ticket, so when we had a PR, only one of our names would be on it, despite pairing for every part of it. This was our way of giving credit to the other person. I agree it's unnecessary now. If we have two interns come onto the team in the future, I would think the better approach is to simply show them how to work with remote forks.

.github/pull_request_template.md
maddenp-noaa: It occurred to me: Replacing this checklist with a request that submitters tag the PR would presumably depend on their rights to add labels. @WeirAE I reopened https://github.com/maddenp-noaa/uwtools/pull/1 -- can you see if you can add labels to it?

WeirAE: Oh, true - Nope, can't add anything there, that would be a problem!

maddenp-noaa: It was a good idea, but probably not a substitute for this checklist. I'll add my other two items here and we can keep the checklist.

.github/pull_request_template.md
christinaholtNOAA: I do want to make sure that when we do have people pair, that they have demonstrable evidence that they were involved in contributions during their time on the project.

christinaholtNOAA (approved):
Looks great!

.github/pull_request_template.md
maddenp-noaa: That makes sense. The PR has an _Assignees_ attribute, where you can add people who worked on the PR. Would that be a workable mechanism?

maddenp-noaa: GitHub describes that field as
> Assignees clarify who is working on specific issues and pull requests.
Sounds about right to me.

maddenp-noaa: I'm going to add it. We can remove it if we don't like it.

PR #346, opened 2023-11-16
UW-371: refactor of render() 

elcarpenterNOAA:
**Description**

`uwtools.config.jinja2.render()` needed a refactor, so I created four helper functions to share its responsibilities. I also created 8 new unit tests: 4 are for the helper functions themselves and the other 4 are to make sure that `render()` is calling those functions. 

I'll add review comments for questions I have.

**Type**

Select one or more:

- [ ] Documentation

**Impact**

Select one:

**Checklist**

Affirm:

### DIFF
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -201,11 +201,11 @@ def render(
     if missing:
         return _log_missing_values(missing)
 
-    # In dry-run mode, log the rendered template. Otherwisre, write rendered template to file.
+    # In dry-run mode, log the rendered template. Otherwise, write rendered template to file.
     return (
         _dry_run_template(template.render())
         if dry_run
-        else _write_template_to_file(output_file, template.render())
+        else _write_template(output_file, template.render())
     )
 
 
@@ -231,7 +231,7 @@ def _log_missing_values(missing: List[str]) -> bool:
 
     :param missing: A list containing the undeclared variables that do not have a corresponding
         match in values.
-    :return: Unable to sucessfully render template.
+    :return: Unable to successfully render template.
     """
 
     log.error("Required value(s) not provided:")
@@ -331,13 +331,13 @@ def _values_needed(undeclared_variables: Set[str]) -> bool:
     return True
 
 
-def _write_template_to_file(output_file: OptionalPath, rendered_template: str) -> bool:
+def _write_template(output_file: OptionalPath, rendered_template: str) -> bool:
     """
-    Write the rendered template to file.
+    Write the rendered template.
 
     :param output_file: Path to the file to write the rendered Jinja2 template to.
     :param rendered_template: A string containing a rendered Jinja2 template.
-    :return: The successful writing of the rendered template to the outfile.
+    :return: The successful writing of the rendered template.
     """
     with writable(output_file) as f:
         print(rendered_template, file=f)
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -175,6 +175,9 @@ def render(
     :param values_needed: Just issue a report about variables needed to render the template?
     :param dry_run: Run in dry-run mode?
     """
+
+    # Render template.
+
     _report(locals())
     values = _set_up_values_obj(
         values_file=values_file, values_format=values_format, overrides=overrides
@@ -188,38 +191,52 @@ def render(
     # then return.
 
     if values_needed:
-        log.info("Value(s) needed to render this template are:")
-        for var in sorted(undeclared_variables):
-            log.info(var)
-        return True
+        return _values_needed(undeclared_variables)
 
     # Check for missing values required to render the template. If found, report them and raise an
     # exception.
 
     missing = [var for var in undeclared_variables if var not in values.keys()]
     if missing:
-        msg = "Required value(s) not provided:"
-        log.error(msg)
-        for key in missing:
-            log.error(key)
-        return False
+        return _log_missing_values(missing)
 
-    # In dry-run mode, display the rendered template and then return.
-
-    if dry_run:
-        rendered_template = template.render()
-        for line in rendered_template.split("\n"):
-            log.info(line)
-        return True
-
-    # Write rendered template to file.
-
-    with writable(output_file) as f:
-        print(template.render(), file=f)
-    return True
+    # In dry-run mode, log the rendered template. Otherwisre, write rendered template to file.
+    return (
+        _dry_run_template(template.render())
+        if dry_run
+        else _write_template_to_file(output_file, template.render())
+    )
 
 
 # Private functions
+
+
+def _dry_run_template(rendered_template: str) -> bool:
+    """
+    Log the rendered template and then return as successful.
+
+    :param rendered_template: A string containing a rendered Jinja2 template.
+    :return: The successful logging of the template.
+    """
+
+    for line in rendered_template.split("\n"):
+        log.info(line)
+    return True
+
+
+def _log_missing_values(missing: List[str]) -> bool:
+    """
+    Log values missing from template and raise an exception.
+
+    :param missing: A list containing the undeclared variables that do not have a corresponding
+        match in values.
+    :return: Unable to sucessfully render template.
+    """
+
+    log.error("Required value(s) not provided:")
+    for key in missing:
+        log.error(key)
+    return False
 
 
 def _register_filters(env: Environment) -> Environment:
@@ -295,3 +312,32 @@ def _set_up_values_obj(
         values.update(overrides)
         log.debug("Updated values with overrides: %s", " ".join(overrides))
     return values
+
+
+def _values_needed(undeclared_variables: Set[str]) -> bool:
+    """
+    Log variables needed to render the template.
+
+    :param undeclared_variables: A set containing the variables needed to render the template.
+    :return: Successfully logged values needed.
+    """
+
+    # If a report of variables required to render the template was requested, make that report and
+    # then return.
+    log.info("Value(s) needed to render this template are:")
+    for var in sorted(undeclared_variables):
+        log.info(var)
+    return True
+
+
+def _write_template_to_file(output_file: OptionalPath, rendered_template: str) -> bool:
+    """
+    Write the rendered template to file.
+
+    :param output_file: Path to the file to write the rendered Jinja2 template to.
+    :param rendered_template: A string containing a rendered Jinja2 template.
+    :return: The successful writing of the rendered template to the outfile.
+    """
+    with writable(output_file) as f:
+        print(rendered_template, file=f)
+    return True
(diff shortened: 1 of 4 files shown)


src/uwtools/config/jinja2.py
```diff
@@ -166,15 +166,19 @@ def render(
...
     :param keq_eq_val_pairs: "key=value" strings to supplement values-file values.
     :param values_needed: Just issue a report about variables needed to render the template?
     :param dry_run: Run in dry-run mode?
+    :return: Jinja2 template was successfully rendered.
     """
+
+    # Render template.
+
     _report(locals())
     values = _set_up_values_obj(
```
elcarpenterNOAA, line 183: These steps are needed in all instances. I did not create a separate helper function for this section because of the number of arguments and returns that would need to be passed, but I'm not sure that was the right move.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -187,15 +187,55 @@ def test_render(values_file, template, tmp_path):
...
 
 
-def test_render_dry_run(caplog, values_file, template):
+def test_render_calls__dry_run(template, tmp_path, values_file):
+    outfile = str(tmp_path / "out.txt")
+    with patch.object(jinja2, "_dry_run_template") as dr:
+        render_helper(
+            input_file=template, values_file=values_file, output_file=outfile, dry_run=True
+        )
+        dr.assert_called_once_with("roses are red, violets are blue")
```
elcarpenterNOAA, line 196: I could not successfully use `assert_called_once_with(cmd=ANY)`, which is why these are all broken up into separate unit tests. I'd love suggestions for shortening or, better yet, parametrizing them.

maddenp-noaa: The `assert_called_once_with()` method (and its ilk -- there's a whole family of related methods) on `Mock` objects are picky about their arguments; in particular, they're literal about whether positional or keyword arguments were used. So, if we've got a function `foo()` and it's mocked as `mock_foo`, `mock_foo.assert_called_once_with(88)` would pass if the code-under-test called `foo(88)`, but not if it called `foo(n=88)`. In the latter case, the assertion would need to be `mock_foo.assert_called_once_with(n=88)`.

For `_dry_run_template()`, the actual call looks like
```
_dry_run_template(template.render())
```
so `dr.assert_called_once_with(ANY)` should pass, since a positional argument is used. If the `_dry_run_template()` call _did_ use a keyword argument, it would be `_dry_run_template(rendered_template=template.render())`, and so the corresponding assertion could be `dr.assert_called_once_with(rendered_template=ANY)`. The `cmd=` keyword argument in your comment probably came from another piece of test code you were looking at, testing a function/method whose argument really was called `cmd`.

Hopefully that helps on that item.

There is some value, though, in being more specific than `ANY` in some cases. `ANY` is great when the argument's actual value truly is something that the test shouldn't know or care about. But even in a case where e.g. it would be good to assert that a specific _type_ of argument is used in the call (like, it matters that it's a `dict` or a `list`), `ANY` might not be strong enough. That seems maybe like the case here.

Maybe something similar to this untested pseudocode could work?
```
@pytest.mark.parametrize("func_arg,func_name,render_extra_kwargs", [
    ("roses are red, violets are blue", "_log_missing_values", {"dry_run": True}), 
    ({"roses_color", "violets_color"}, "_values_needed", {"values_needed": True}),
])
def test_render_makes_expected_calls(func_arg, func_name, render_extra_kwargs, template, tmp_path, values_file):
    outfile = str(tmp_path / "out.txt")
    render_kwargs = {"input_file": template, "values_file": values_file, "output_file": outfile, **render_extra_args}
    with patch.object(jinja2, func_name) as mock:
        render_helper(**render_kwargs)
    mock.assert_called_once_with(func_arg)
```
Something like that might cover 3 of the 4 tests (I only did 2 there). One of them has that special `del` making it asymmetrical with the others, so it might have to stay separate. Or, you could have the `parametrize`d argument also provide a 4th parameter, which would be the name of a key to delete, which would be `None` for 3 of the tests and `"roses_color"` for the other. And the test code could have an `if` condition that only does that config-file mutation for the test that provides a key to delete. Then maybe all 4 tests become symmetrical.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -217,6 +257,20 @@ def test_render_values_needed(caplog, values_file, template):
         assert logged(caplog, var)
 
 
+def test__dry_run_template(caplog):
+    jinja2._dry_run_template("roses are red, violets are blue")
+    assert logged(caplog, "roses are red, violets are blue")
```
maddenp-noaa, line 262: It would be stronger if the argument were two lines, to exercise the fact that the function-under-test splits on newlines and loops on those. So e.g.
```
jinja2._dry_run_template("roses are red\nviolets are blue")
assert logged(caplog, "roses are red")
assert logged(caplog, "violets are blue")
```
since those two lines would be logged separately.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -187,15 +187,55 @@ def test_render(values_file, template, tmp_path):
...
+def test_render_dry_run(caplog, template, values_file):
     log.setLevel(logging.INFO)
     render_helper(
         input_file=template, values_file=values_file, output_file="/dev/null", dry_run=True
     )
     assert logged(caplog, "roses are red, violets are blue")
 
 
-def test_render_values_missing(caplog, values_file, template):
+def test_render_values_missing(caplog, template, values_file):
```
maddenp-noaa, line 238: Thanks for the tidy-up!

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -244,6 +298,22 @@ def test__set_up_config_obj_file(values_file):
...
+    assert logged(caplog, "Value(s) needed to render this template are:")
+    assert logged(caplog, "roses_color")
+    assert logged(caplog, "lavender_smell")
+
+
+def test__write_template_to_file(tmp_path):
+    outfile = str(tmp_path / "out.txt")
+    jinja2._write_template_to_file(outfile, "roses are red, violets are blue")
+    with open(outfile, "r", encoding="utf-8") as f:
+        assert f.read().strip() == "roses are red, violets are blue"
```
maddenp-noaa, line 314: In line with my earlier comment about the fact that we might be writing to `stdout`, this is a perfect test of the case where we really are writing to a file. It would also be great to have a second, similar test where `outfile` is `None`, in which case the output will be written to `stdout` and you can see it and assert on it by receiving the `capsys` (instead of `caplog`) fixture. If you search the unit tests for `capsys` you'll find examples to crib from.

maddenp-noaa (changes requested):
The refactor looks great. I only commented on two typos and then a potential naming improvement.

I made a couple easy suggestions about the unit tests (which also look great -- nice use of very test-specific test data), and responded to your question, which I don't think necessarily has to lead to a change, but you can decide.

src/uwtools/config/jinja2.py
```diff
@@ -188,38 +192,52 @@ def render(
...
     missing = [var for var in undeclared_variables if var not in values.keys()]
     if missing:
-        msg = "Required value(s) not provided:"
-        log.error(msg)
-        for key in missing:
-            log.error(key)
-        return False
+        return _log_missing_values(missing)
+
+    # In dry-run mode, log the rendered template. Otherwisre, write rendered template to file.
```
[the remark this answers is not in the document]
elcarpenterNOAA: Thank you for the explanation! I had questioned it earlier when I added the type checking.

src/uwtools/tests/config/test_jinja2.py
elcarpenterNOAA: How could we include a test for `_write_template` here? It would require either `outfile` or `tmp_path` as an argument, and both would be deemed as undefined variables.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -244,6 +299,29 @@ def test__set_up_config_obj_file(values_file):
...
+
+
+def test__write_template_stdout(capsys):
+    jinja2._write_template(None, "roses are red, violets are blue")
+    actual = capsys.readouterr().out
+    expected = "roses are red, violets are blue"
+    assert actual.strip() == expected
+
```
elcarpenterNOAA, line 323: I added the second test for `_write_template()` and fixed some spelling mistakes found elsewhere.

src/uwtools/tests/config/test_jinja2.py
maddenp-noaa: Ah, yes. I had to try it to understand, but I see it now. This _could_ be a job for `ANY`, and the following works for me, but I'm not sure if the complexity is worth it. Maybe a few smaller tests with a little repetition are better. You be the judge.
```
@pytest.mark.parametrize(
    "del_key,func_args,func_name,render_extra_kwargs",
    [
        (None, ["roses are red, violets are blue"], "_dry_run_template", {"dry_run": True}),
        ("roses_color", [["roses_color"]], "_log_missing_values", {"dry_run": True}),
        (None, [{"roses_color", "violets_color"}], "_values_needed", {"values_needed": True}),
        (None, [ANY, "roses are red, violets are blue"], "_write_template", {}),
    ],
)
def test_render_makes_expected_calls(
    del_key, func_args, func_name, render_extra_kwargs, template, tmp_path, values_file
):
    if del_key:
        with open(values_file, "r", encoding="utf-8") as f:
            cfgobj = yaml.safe_load(f.read())
        del cfgobj["roses_color"]
        with open(values_file, "w", encoding="utf-8") as f:
            f.write(yaml.dump(cfgobj))
    output_file = str(tmp_path / "out.txt")
    render_kwargs = {
        "input_file": template,
        "output_file": output_file,
        "values_file": values_file,
        **render_extra_kwargs,
    }
    with patch.object(jinja2, func_name) as mock:
        render_helper(**render_kwargs)
    mock.assert_called_once_with(*func_args)
```
I think I might go with smaller, separate tests. This feels like it makes the test too tricky and hard to understand.

elcarpenterNOAA: I'm sorry about that - I need to work on my personal verbose setting. 

Although I like the idea of just having one test to get 4 of the 5 tests needed (the exclusion being `write_template()`'s `stdout` test), I agree that this seems a wee bit too complicated. If I wasn't the developer on this ticket, it would take me a few minutes to figure out everything that's happening here.

maddenp-noaa: > I need to work on my personal verbose setting.
No critique implied there -- after I played with the code I understood exactly what you meant, which you accurately expressed. It's difficult to write prose talking about code in an unambiguous way; I try and am sure I fail often.

maddenp-noaa (approved):
Nice work.

NaureenBharwaniNOAA (approved):
Nice addition of clean tests and great catch of the spelling errors. LGTM.

Issue #347, opened 2023-11-17
UW-418: Split .ini & .sh handling


WeirAE:
**Description**
Currently uw treats ini and sh configs as essentially interchangeable, but they have important differences. for example, in ini files, spaces may appear around the = sign (and our code has an option to add or suppress those spaces) while sh files (being in reality just bash/sh code) must not. also ini files are two-level, comprising sections containing key-calue pares while sh files are one-level, containing only key-value pairs with no structure.

**Requirements**
duplicate INIConfig as SHConfig
remove bash info from INIConfig
modify dump_dict methods for appropriate depths
add FORMAT.sh to uwtools.utils.file.FORMAT
update core

**Acceptance Criteria (Definition of Done)**
1. sh-related logic has been factored out to a new shconfig class
2. there are comprehensive, direct tests for shconfig and its methods
3. iniconfig has been simplified and its (possibly also simplified) tests pass
4. the CLI has been updated to provide separate ini and sh format options

WeirAE:
Completed by #348

PR #348, opened 2023-11-17
UW-418: Split .ini & .sh handling

WeirAE:
**Description**

INI and SH differences should be handled uniquely. This separates bash handling out of INIConfig to SHConfig and updates affected classes and tests accordingly

Completes issue #347 

**Type**

Select one or more:

- [x] Documentation

**Impact**

Select one:

**Checklist**

Affirm:

### DIFF
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -1,4 +1,4 @@
-# pylint: disable=duplicate-code, unnecessary-comprehension
+# pylint: disable=duplicate-code
 import configparser
 from io import StringIO
 
@@ -38,7 +38,7 @@ class INIConfig(Config):
         cfg = configparser.ConfigParser()
         with readable(config_file) as f:
             cfg.read_string(f.read())
-        return {s: {k: v for k, v in cfg[s].items()} for s in cfg.sections()}
+        return {s: dict(cfg[s].items()) for s in cfg.sections()}
 
     # Public methods
 
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -3,7 +3,7 @@ import configparser
 from io import StringIO
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import config_sections, depth
+from uwtools.config.support import config_sections
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -81,7 +81,7 @@ class INIConfig(Config):
         parser = configparser.ConfigParser()
         s = StringIO()
         cfgdepth = depth(cfg)
-        assert cfgdepth == 2  # 2 => .ini
+        assert cfgdepth in (1, 2)  # 2 => .ini
         parser.read_dict(cfg)
         parser.write(s, space_around_delimiters=opts.space if opts else True)
         with writable(path) as f:
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -1,10 +1,11 @@
+# pylint: disable=duplicate-code
 import configparser
 from io import StringIO
 from types import SimpleNamespace as ns
 from typing import Optional
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import depth
+from uwtools.config.support import config_sections, depth
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
@@ -13,10 +14,11 @@ class INIConfig(Config):
     Concrete class to handle INI config files.
     """
 
+    _MAXDEPTH = 2
+
     def __init__(
         self,
         config_file: str,
-        space_around_delimiters: bool = True,
     ):
         """
         Construct an INIConfig object.
@@ -24,10 +26,8 @@ class INIConfig(Config):
         Spaces may be included for INI format, but should be excluded for bash.
 
         :param config_file: Path to the config file to load.
-        :param space_around_delimiters: Include spaces around delimiters?
         """
         super().__init__(config_file)
-        self.space_around_delimiters = space_around_delimiters
         self.parse_include()
 
     # Private methods
@@ -40,20 +40,12 @@ class INIConfig(Config):
 
         :param config_file: Path to config file to load.
         """
-        # The protected _sections method is the most straightforward way to get at the dict
-        # representation of the parse config.
-
         cfg = configparser.ConfigParser()
-        cfg.optionxform = str  # type: ignore
-        sections = cfg._sections  # type: ignore # pylint: disable=protected-access
+        sections = config_sections(cfg)
         with readable(config_file) as f:
             raw = f.read()
-        try:
             cfg.read_string(raw)
-            return dict(sections)
-        except configparser.MissingSectionHeaderError:
-            cfg.read_string("[top]\n" + raw)
-            return dict(sections.get("top"))
+            return sections
 
     # Public methods
 
@@ -63,10 +55,10 @@ class INIConfig(Config):
 
         :param path: Path to dump config to.
         """
-        INIConfig.dump_dict(path, self.data, ns(space=self.space_around_delimiters))
+        INIConfig.dump_dict(path, self.data, space=True)
 
     @staticmethod
-    def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
+    def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None, **kwargs) -> None:
         """
         Dumps a provided config dictionary in INI format.
 
@@ -78,12 +70,12 @@ class INIConfig(Config):
         # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
+        assert depth(cfg) == INIConfig._MAXDEPTH
+
         parser = configparser.ConfigParser()
         s = StringIO()
-        cfgdepth = depth(cfg)
-        assert cfgdepth in (1, 2)  # 2 => .ini
         parser.read_dict(cfg)
-        parser.write(s, space_around_delimiters=opts.space if opts else True)
+        parser.write(s, space_around_delimiters=kwargs.get("space", True))
         with writable(path) as f:
             print(s.getvalue().strip(), file=f)
         s.close()
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -1,9 +1,8 @@
-# pylint: disable=duplicate-code
+# pylint: disable=duplicate-code, unnecessary-comprehension
 import configparser
 from io import StringIO
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import config_sections
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
@@ -37,11 +36,9 @@ class INIConfig(Config):
         :param config_file: Path to config file to load.
         """
         cfg = configparser.ConfigParser()
-        sections = config_sections(cfg)
         with readable(config_file) as f:
-            raw = f.read()
-            cfg.read_string(raw)
-        return sections
+            cfg.read_string(f.read())
+        return {s: {k: v for k, v in cfg[s].items()} for s in cfg.sections()}
 
     # Public methods
 
@@ -69,7 +66,7 @@ class INIConfig(Config):
         parser = configparser.ConfigParser()
         s = StringIO()
         parser.read_dict(cfg)
-        parser.write(s, space_around_delimiters=True)
+        parser.write(s)
         with writable(path) as f:
             print(s.getvalue().strip(), file=f)
         s.close()
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -12,7 +12,7 @@ class INIConfig(Config):
     Concrete class to handle INI config files.
     """
 
-    _DEPTH = 2
+    DEPTH = 2
 
     def __init__(
         self,
@@ -65,7 +65,6 @@ class INIConfig(Config):
         # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
-        assert depth(cfg) == INIConfig._DEPTH
 
         parser = configparser.ConfigParser()
         s = StringIO()
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -1,8 +1,6 @@
 # pylint: disable=duplicate-code
 import configparser
 from io import StringIO
-from types import SimpleNamespace as ns
-from typing import Optional
 
 from uwtools.config.formats.base import Config
 from uwtools.config.support import config_sections, depth
@@ -14,7 +12,7 @@ class INIConfig(Config):
     Concrete class to handle INI config files.
     """
 
-    _MAXDEPTH = 2
+    _DEPTH = 2
 
     def __init__(
         self,
@@ -22,8 +20,6 @@ class INIConfig(Config):
     ):
         """
         Construct an INIConfig object.
-
-        Spaces may be included for INI format, but should be excluded for bash.
 
         :param config_file: Path to the config file to load.
         """
@@ -45,7 +41,7 @@ class INIConfig(Config):
         with readable(config_file) as f:
             raw = f.read()
             cfg.read_string(raw)
-            return sections
+        return sections
 
     # Public methods
 
@@ -55,27 +51,26 @@ class INIConfig(Config):
 
         :param path: Path to dump config to.
         """
-        INIConfig.dump_dict(path, self.data, space=True)
+        self.dump_dict(path, self.data)
 
     @staticmethod
-    def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None, **kwargs) -> None:
+    def dump_dict(path: OptionalPath, cfg: dict) -> None:
         """
         Dumps a provided config dictionary in INI format.
 
         :param path: Path to dump config to.
         :param cfg: The in-memory config object to dump.
-        :param space_around_delimiters: Place spaces around delimiters?
         """
         # Configparser adds a newline after each section, presumably to create nice-looking output
         # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
-        assert depth(cfg) == INIConfig._MAXDEPTH
+        assert depth(cfg) == INIConfig._DEPTH
 
         parser = configparser.ConfigParser()
         s = StringIO()
         parser.read_dict(cfg)
-        parser.write(s, space_around_delimiters=kwargs.get("space", True))
+        parser.write(s, space_around_delimiters=True)
         with writable(path) as f:
             print(s.getvalue().strip(), file=f)
         s.close()
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -8,33 +8,30 @@ from uwtools.config.support import depth
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
-class INIConfig(Config):
+class SHConfig(Config):
     """
-    Concrete class to handle INI config files.
+    Concrete class to handle bash config files.
     """
 
     def __init__(
         self,
         config_file: str,
-        space_around_delimiters: bool = True,
     ):
         """
-        Construct an INIConfig object.
+        Construct a SHConfig object.
 
-        Spaces may be included for INI format, but should be excluded for bash.
+        Spaces should be excluded for bash.
 
         :param config_file: Path to the config file to load.
-        :param space_around_delimiters: Include spaces around delimiters?
         """
         super().__init__(config_file)
-        self.space_around_delimiters = space_around_delimiters
         self.parse_include()
 
     # Private methods
 
     def _load(self, config_file: OptionalPath) -> dict:
         """
-        Reads and parses an INI file.
+        Reads and parses a bash file.
 
         See docs for Config._load().
 
@@ -59,31 +56,25 @@ class INIConfig(Config):
 
     def dump(self, path: OptionalPath) -> None:
         """
-        Dumps the config in INI format.
+        Dumps the config in bash format.
 
         :param path: Path to dump config to.
         """
-        INIConfig.dump_dict(path, self.data, ns(space=self.space_around_delimiters))
+        SHConfig.dump_dict(path, self.data, ns(space=False))
 
     @staticmethod
     def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
         """
-        Dumps a provided config dictionary in INI format.
+        Dumps a provided config dictionary in bash format.
 
         :param path: Path to dump config to.
         :param cfg: The in-memory config object to dump.
-        :param space_around_delimiters: Place spaces around delimiters?
         """
-        # Configparser adds a newline after each section, presumably to create nice-looking output
-        # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
-        # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
-        # memory, then strip the trailing newline.
-        parser = configparser.ConfigParser()
         s = StringIO()
         cfgdepth = depth(cfg)
-        assert cfgdepth == 2  # 2 => .ini
-        parser.read_dict(cfg)
-        parser.write(s, space_around_delimiters=opts.space if opts else True)
+        assert cfgdepth == 1  # 1 => .sh
+        for key, value in cfg.items():
+            print(f"{key}={value}", file=s)
         with writable(path) as f:
             print(s.getvalue().strip(), file=f)
         s.close()
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -12,7 +12,7 @@ class SHConfig(Config):
     Concrete class to handle bash config files.
     """
 
-    _DEPTH = 1
+    DEPTH = 1
 
     def __init__(
         self,
@@ -61,7 +61,6 @@ class SHConfig(Config):
         :param path: Path to dump config to.
         :param cfg: The in-memory config object to dump.
         """
-        assert depth(cfg) == SHConfig._DEPTH
 
         s = StringIO()
         for key, value in cfg.items():
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -3,7 +3,7 @@ import configparser
 from io import StringIO
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import config_sections, depth
+from uwtools.config.support import config_sections
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -1,9 +1,8 @@
-# pylint: disable=duplicate-code
+# pylint: disable=duplicate-code, unnecessary-comprehension
 import configparser
 from io import StringIO
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import config_sections
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
@@ -37,11 +36,10 @@ class SHConfig(Config):
         :param config_file: Path to config file to load.
         """
         cfg = configparser.ConfigParser()
-        sections = config_sections(cfg)
+        section = "top"
         with readable(config_file) as f:
-            raw = f.read()
-            cfg.read_string("[top]\n" + raw)
-            return dict(sections.get("top"))
+            cfg.read_string(f"[{section}]\n" + f.read())
+        return {k: v for k, v in cfg[section].items()}
 
     # Public methods
 
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -1,10 +1,11 @@
+# pylint: disable=duplicate-code
 import configparser
 from io import StringIO
 from types import SimpleNamespace as ns
 from typing import Optional
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import depth
+from uwtools.config.support import config_sections, depth
 from uwtools.utils.file import OptionalPath, readable, writable
 
 
@@ -13,14 +14,14 @@ class SHConfig(Config):
     Concrete class to handle bash config files.
     """
 
+    _MAXDEPTH = 1
+
     def __init__(
         self,
         config_file: str,
     ):
         """
         Construct a SHConfig object.
-
-        Spaces should be excluded for bash.
 
         :param config_file: Path to the config file to load.
         """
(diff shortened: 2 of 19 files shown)


src/uwtools/config/tools.py
```diff
@@ -112,8 +112,10 @@ def _realize_config_check_depths(input_obj: Config, output_format: str) -> None:
     :param output_format: The output format:
     :raises: UWConfigError on excessive input-config depth.
     """
-    if (output_format == FORMAT.ini and input_obj.depth > 2) or (
-        output_format == FORMAT.nml and input_obj.depth != 2
+    if (
+        (output_format == FORMAT.ini and input_obj.depth > 2)
+        or (output_format == FORMAT.nml and input_obj.depth != 2)
+        or (output_format == FORMAT.sh and input_obj.depth != 1)
```
WeirAE, line 118: Technically this is a different ticket to account for this, but it fit in here

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -159,8 +159,8 @@ def test_update_values(config):
     assert config == {"foo": 88, "a": "11", "b": "12", "c": "13"}
 
 
-@pytest.mark.parametrize("fmt1", [FORMAT.ini, FORMAT.nml, FORMAT.yaml])
-@pytest.mark.parametrize("fmt2", [FORMAT.ini, FORMAT.nml, FORMAT.yaml])
+@pytest.mark.parametrize("fmt1", [FORMAT.ini, FORMAT.nml, FORMAT.sh, FORMAT.yaml])
```
WeirAE, line 162: These tests are currently failing due to the single-level of bash files. This was missed before and will need to be handled now. Still working on how I want to do that, however.

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,80 @@
...
+        # representation of the parse config.
+
+        cfg = configparser.ConfigParser()
+        cfg.optionxform = str  # type: ignore
+        sections = cfg._sections  # type: ignore # pylint: disable=protected-access
+        with readable(config_file) as f:
+            raw = f.read()
+        try:
+            cfg.read_string(raw)
+            return dict(sections)
```
WeirAE, line 50: I'm getting a lack of coverage on this line and I have not been able to figure out why. However, it does look like we have not tested any of the `_load()` functions in any of the formats directly, so that may need to be added

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @staticmethod
+    def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
+        """
+        Dumps a provided config dictionary in bash format.
+
+        :param path: Path to dump config to.
+        :param cfg: The in-memory config object to dump.
+        """
+        s = StringIO()
```
WeirAE, line 73: I kept the `StringIO()` read behavior here even though I could use the direct file printing. I like preserving the symmetry with INI, personally.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -75,14 +75,14 @@ def test_characterize_values(config):
...
+@pytest.mark.parametrize("fmt", [FORMAT.ini, FORMAT.nml, FORMAT.sh, FORMAT.yaml])
 def test_compare_config(caplog, fmt, salad_base):
     """
     Compare two config objects.
     """
     log.setLevel(logging.INFO)
     cfgobj = tools.format_to_config(fmt)(fixture_path(f"simple.{fmt}"))
-    if fmt == FORMAT.ini:
+    if fmt == FORMAT.ini or FORMAT.sh:
         salad_base["salad"]["how_many"] = "12"  # str "12" (not int 12) for ini
```
maddenp-noaa, line 86: Maybe update trailing comment `for ini` -> `for ini or sh`.

maddenp-noaa (changes requested):
First round of comments, though all the library code, but not yet to the tests.

I haven't checked, but: Is there anything in the CLI that need an update?

src/uwtools/config/formats/ini.py
```diff
@@ -81,13 +81,9 @@ def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
...
         s = StringIO()
         cfgdepth = depth(cfg)
-        assert cfgdepth in (1, 2)  # 1 => .sh, 2 => .ini
-        if cfgdepth == 2:
-            parser.read_dict(cfg)
-            parser.write(s, space_around_delimiters=opts.space if opts else True)
-        else:
-            for key, value in cfg.items():
-                print(f"{key}={value}", file=s)
+        assert cfgdepth in (1, 2)  # 2 => .ini
```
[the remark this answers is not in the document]
WeirAE: So I had done that, but switched back since INI can technically take 1 or 2 levels. However, given that I'm going to have to make the output depth controlled, I definitely like the idea of overriding to let it define a `MAXDEPTH` I can interrogate later!

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+
+    def __init__(
+        self,
+        config_file: str,
+    ):
+        """
+        Construct a SHConfig object.
+
+        Spaces should be excluded for bash.
```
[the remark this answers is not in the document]
WeirAE: True, probably overstated now

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,80 @@
...
+        See docs for Config._load().
+
+        :param config_file: Path to config file to load.
+        """
+        # The protected _sections method is the most straightforward way to get at the dict
+        # representation of the parse config.
+
+        cfg = configparser.ConfigParser()
+        cfg.optionxform = str  # type: ignore
+        sections = cfg._sections  # type: ignore # pylint: disable=protected-access
```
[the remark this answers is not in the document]
WeirAE: Ah, good idea!

src/uwtools/config/formats/sh.py
WeirAE: I had gone back and forth on that, but I think you're right on this one

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
maddenp-noaa: I wasn't aware that INI can take 1 level. Can you show an example?

src/uwtools/tests/config/formats/test_base.py
WeirAE: I think the above suggestion of defining `MAXDEPTH` will let me code the writes better! Work in progress

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
WeirAE: Ah thanks,  I meant to remove that one

src/uwtools/tests/config/formats/test_base.py
maddenp-noaa: It could be a module- or class-level constant, or a `@property` method. I think the latter two are probably more appropriate so that, if you've got a reference to a `Config` instance, you can get its maximum depth without referring back the module where the instance is defined. Between class constant and `@property` method, its probably a matter of taste.

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
WeirAE: Strictly, afaik, the INI format doesn't require a header per se, although `ConfigParser` does. I think you're right to say that there's no point in supporting a special case though, or more in enforcing what we want to see. I'll clean that one up

src/uwtools/config/formats/ini.py
```diff
@@ -81,13 +81,9 @@ def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
...
+        parser.read_dict(cfg)
+        parser.write(s, space_around_delimiters=opts.space if opts else True)
```
christinaholtNOAA, line 86: What happens here if `opts` doesn't include a `space` attribute?

Do we want to stick with the optional namespace object for these optional args, or should we go with the Python-standard kwargs? This would (IMO) simplify to:

```
parser.write(s, space_around_delimiters=kwargs.get('space', True))
```

src/uwtools/tests/config/formats/test_sh.py
```diff
@@ -0,0 +1,42 @@
+# pylint: disable=duplicate-code
+"""
+Tests for uwtools.config.formats.ini module.
```
maddenp-noaa, line 3: `ini` -> `sh`

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
christinaholtNOAA: I don't even think we need the `space_around_delimiters` argument at this point. It's another piece that's meant to differentiate between ini and bash and could now be hard coded in the methods where it's used in the `dump` method. You've done that in the new bash dump method, but not in the original ini.

src/uwtools/config/tools.py
christinaholtNOAA: Shouldn't we make sure that INI format is also exactly 2?

src/uwtools/tests/config/formats/test_sh.py
```diff
@@ -0,0 +1,42 @@
...
+    outfile = tmp_path / "outfile.sh"
+    cfgobj = SHConfig(infile)
+    expected: Dict[str, Any] = {
+        **salad_base["salad"],
+        "how_many": "12",
+    }
+    assert cfgobj == expected
+    cfgobj.dump(outfile)
+    assert filecmp.cmp(infile, outfile)
+    cfgobj.update({"dressing": ["ranch", "italian"]})
```
christinaholtNOAA, line 40: Lines 40-41 don't seem to be related at all to our handling of bash files. This is testing updating config objects. IIR, we decided not to handle lists in bash, so if we wanted to keep this test, we might consider ensuring that it raises an error. 

If we want to capture that as a future work ticket, that's cool, but I'd suggest removing this misleading test here. Unless, of course, I'm missing something (entirely possible).

The same applies to the ini counterpart.

src/uwtools/tests/fixtures/fruit_config.ini
```diff
@@ -0,0 +1,4 @@
+fruit = papaya
+vegetable = eggplant
+how_many = 17
+dressing = ranch
```
christinaholtNOAA, line 4: I do think we want to go the standard path of having sections in the ini files.

src/uwtools/config/formats/ini.py
WeirAE: I definitely prefer kwargs, I'll get that in

src/uwtools/tests/config/formats/test_sh.py
WeirAE: No, you're right, its not doing anything here - but actually, thats a good point, it would be good to keep it but ensure it raises an error

src/uwtools/config/tools.py
WeirAE: Yep, updating that as part of fixing ini to 2 everywhere, thanks!

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
WeirAE: I think INI doesn't require spaces, although we could enforce it to here

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Public methods
+
+    def dump(self, path: OptionalPath) -> None:
+        """
+        Dumps the config in bash format.
+
+        :param path: Path to dump config to.
+        """
+        SHConfig.dump_dict(path, self.data, ns(space=False))
```
maddenp-noaa, line 63: Riffing on a comment made elsewhere, we don't need to pass in any value for `space` because it's not accessed in the called moethod. We currently _do_ need to pass in at least an empty namespace (`ns()`), but that also disappears if we switch to a `**kwargs` argument, which I think is a good idea.

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
maddenp-noaa: I'm all for making an informed decision and eliminating vectors for bikeshedding: Spaces look good and are encouraged in almost every language where they're optional. (Hello Bourne and Bash shells, you outliers.)

src/uwtools/tests/config/formats/test_sh.py
maddenp-noaa: I wrote something like this in Slack, but here it is again: I think the issue around how -- or whether -- we should try to convert between formats with incompatible type representations (see also: the `2_00:00:00` MPAS XML -> YAML issue) is going to keep coming up.

One painful (but potentially valuable) way around it might be to support not file _formats_ but specific well-known _files_ used by various NWP components, such that we could define a type mapping and do the right thing where possible (e.g. convert `2_00:00:00` to `str` instead of `int` because we'd _define_ the fact that the place in an MPAS XML streams document where this value appears must be interpreted as a string). That would move `uwtools` away from providing general-purpose config-file translation but, as we're finding out, that's a tricky problem, and maybe not one that serves UFS, which doesn't deal with general file formats but with specific component-configuration files.

Another way around it might be to not do this at all but, rather, flow information downstream from UW YAML to component-configuration files. That seems to me like the main thing we want to accomplish. What's the use case for, say, supplementing YAML data with content from a Fortran namelist? Does that really arise in the course of running a UFS experiment?

No need to belabor the question now, and I'm not suggesting a change here, but I think we'll have to think more about this.

src/uwtools/config/formats/ini.py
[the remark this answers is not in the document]
WeirAE: Going through a few times, we do kind of bluntly check depth compatibility, but a more rigorous treatment would be good for UW-420. I'm also not sure how I feel about currently having namelists fully incompatible - I would like to be able to convert them personally, but I don't see an actual need for it, so its probably ok to keep here for now.

src/uwtools/tests/config/formats/test_sh.py
WeirAE: I wonder if we should make an enabler task for the facade to thoroughly evaluate our functions compared to specific file cases, with this and similar issues in mind. We've kind of started that when we updated the Rocoto handling, I'm sure we're missing more things

src/uwtools/config/support.py
```diff
@@ -8,6 +9,16 @@
 INCLUDE_TAG = "!INCLUDE"
 
 
+def config_sections(config: configparser.ConfigParser):
+    """
+    Access the _sections method of a config object.
```
maddenp-noaa, line 14: `:param` entry for `config`, `:return` entry, and `->` return type on function signature, please.

src/uwtools/config/tools.py
```diff
@@ -112,8 +112,8 @@ def _realize_config_check_depths(input_obj: Config, output_format: str) -> None:
     :param output_format: The output format:
     :raises: UWConfigError on excessive input-config depth.
     """
-    if (output_format == FORMAT.ini and input_obj.depth > 2) or (
-        output_format == FORMAT.nml and input_obj.depth != 2
+    if (output_format == (FORMAT.ini or FORMAT.nml) and input_obj.depth != 2) or (
+        output_format == FORMAT.sh and input_obj.depth != 1
```
maddenp-noaa, line 116: Seeing this, I'm wondering if the `_MAXDEPTH` attribute of the format classes should be `MAXDEPTH` instead, and those values used here.

I'm also wondering if the `MAX` part is what we really want, as it's also `MIN`, right? Just thinking out loud but, except for YAML, which AFAIK has no maximum depth, the other formats all have a specific depth that they must be, neither more or less. Should we just call it `DEPTH`?

src/uwtools/tests/config/formats/test_ini.py
```diff
@@ -1,44 +1,25 @@
...
-    cfgobj.dump(outfile)
-    assert filecmp.cmp(infile, outfile)
-    cfgobj.update({"dressing": ["ranch", "italian"]})
-    expected["dressing"] = ["ranch", "italian"]
-    assert cfgobj == expected
-
-
 def test_parse_include():
     """
     Test that non-YAML handles include tags properly for INI with no sections.
```
maddenp-noaa, line 16: I know that it's not new, but what does "non-YAML" add to this docstring (and in the sh tests below)? And `no sections` is inaccurate, right? Could this be simplified? (`no sections` is accurate for the sh tests, I think.)

src/uwtools/tests/config/formats/test_sh.py
```diff
@@ -0,0 +1,42 @@
...
+
+from uwtools.config.formats.sh import SHConfig
+from uwtools.tests.support import fixture_path
+
+# Tests
+
+
+def test_parse_include():
+    """
+    Test that non-YAML handles include tags properly for bash with no sections.
```
maddenp-noaa, line 17: I mentioned it in the library code, too, but I think we should use `sh` instead of `bash` as the latter is too specific. We're really talking about the entire Bourne-shell family, which includes `sh`, `bash`, `ksh`, `dash`, maybe `zsh`. `sh` is a good generic identifier for all those, since it refers to the original Bourne shell.

src/uwtools/tests/config/test_support.py
```diff
@@ -12,13 +13,21 @@
...
 from uwtools.logging import log
 from uwtools.tests.support import logged
 from uwtools.utils.file import FORMAT
 
 
+def test_config_sections():
+    cfg = configparser.ConfigParser()
+    sections = support.config_sections(cfg)
+    cfg.read_string("[top]\n")
+    assert not dict(sections.get("top"))
```
maddenp-noaa, line 28: Could we do something here that demonstrates a little better what the code-under-test is providing? This seems like kind of a degenerate case. We could e.g.
```
(Pdb) !cfg.read_string("[flower]\nroses=red\n[fruit]\na=apple\n")
(Pdb) p sections
{'flower': {'roses': 'red'}, 'fruit': {'a': 'apple'}}
```
and make some assertions about that.

src/uwtools/utils/file.py
```diff
@@ -30,19 +30,20 @@ class FORMAT:
     _ini: str = "ini"
     _jinja2: str = "jinja2"
     _nml: str = "nml"
+    _sh: str = "sh"
```
maddenp-noaa, line 33: `uwtools.cli` contains the line
```
FORMATS = [FORMAT.ini, FORMAT.nml, FORMAT.yaml]
```
Does this need to be updated to include `FORMAT.sh`?

maddenp-noaa (changes requested):

src/uwtools/utils/file.py
WeirAE: AH, I knew I would miss one, thanks!

src/uwtools/tests/config/test_support.py
WeirAE: Ah, thats true - I left this as a blunt version of the format test, but it'll certainly be more meaningful with some detail

src/uwtools/tests/config/formats/test_sh.py
WeirAE: Mmm, very good point

src/uwtools/tests/config/test_support.py
maddenp-noaa: 100% coverage is necessary but not sufficient :)

src/uwtools/config/tools.py
WeirAE: Even more annoyingly, this goes directly to FORMAT without accessing the concrete class, which is why I ended going back to leaving it specific and not just `input_obj._MAXDEPTH`. I actually added a limit to YAML to cheekily stop absurd subtask-ing, but that could be dropped and just leave it as _DEPTH

I thought about how to call it here, but I think it would have to be modified to accept the concrete class and do the evaluation, or move DEPTH to a property in FORMAT

src/uwtools/config/formats/yaml.py
```diff
@@ -39,6 +39,8 @@ class YAMLConfig(Config):
     Concrete class to handle YAML config files.
     """
 
+    _MAXDEPTH = 8
```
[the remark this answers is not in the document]
WeirAE: I took the deepest reference we had, added one and decided that was already reaching illegible depths and should be stopped there. Honestly, personal opinion, could make it `None` more fairly

src/uwtools/config/support.py
WeirAE: I was pondering this one, the _section method is kind of weird in terms of type

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @staticmethod
+    def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
+        """
+        Dumps a provided config dictionary in bash format.
+
+        :param path: Path to dump config to.
+        :param cfg: The in-memory config object to dump.
+        """
+        assert depth(cfg) == SHConfig._MAXDEPTH
```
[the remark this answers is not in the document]
WeirAE: Making this method call 'self' was causing me all kinds of issues in the overloading

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # Public methods
+
+    def dump(self, path: OptionalPath) -> None:
+        """
+        Dumps the config in bash format.
+
+        :param path: Path to dump config to.
+        """
+        SHConfig.dump_dict(path, self.data, ns(space=False))
```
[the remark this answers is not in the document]
WeirAE: ah, true!

src/uwtools/config/formats/ini.py
```diff
@@ -78,16 +70,12 @@ def dump_dict(path: OptionalPath, cfg: dict, opts: Optional[ns] = None) -> None:
...
-        cfgdepth = depth(cfg)
-        assert cfgdepth in (1, 2)  # 1 => .sh, 2 => .ini
-        if cfgdepth == 2:
-            parser.read_dict(cfg)
-            parser.write(s, space_around_delimiters=opts.space if opts else True)
-        else:
-            for key, value in cfg.items():
-                print(f"{key}={value}", file=s)
+        parser.read_dict(cfg)
+        parser.write(s, space_around_delimiters=kwargs.get("space", True))
```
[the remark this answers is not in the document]
WeirAE: I split the difference of using kwargs and limiting, but just making it fixed is fine

src/uwtools/config/tools.py
maddenp-noaa: Not sure if we're on the same page, but I was was thinking of something like
```
if (output_format == FORMAT.ini and input_obj.depth != INIConfig.DEPTH) or
    (output_format == FORMAT.nml and input_obj.depth != NMLConfig.DEPTH) or
    (output_format == FORMAT.sh and input_obj.depth != SHConfig.DEPTH): ...
```
But we've also got `uwtools.config.support.format_to_config()` at our disposal, so I think we could have something like
```
cfgclass = format_to_config(output_format)
if cfgclass.DEPTH and input_obj.depth != cfgclass.DEPTH: ...
```
The first clause in the `if` accounts for the fact that `YAMLConfig` should have `DEPTH` of `None`, and the second clause catches depth mismatch. Would that work?

src/uwtools/config/support.py
WeirAE: Hrm, its a 'dictionary of key types', which is giving the error `expected "SupportsKeysAndGetItem[object, str]"` when I try to type it, which isn't terribly helpful. [This](https://stackoverflow.com/questions/68907243/a-dictionary-with-a-given-key-type-cannot-be-assigned-as-a-dictionary-of-objects) StackOverflow entry suggests it may be a covariance issue?

src/uwtools/config/tools.py
WeirAE: Ah, yes - I had at first tried to go with `if(input_obj.depth !<= input_obj._MAXDEPTH)` for a clean single line, but that was the wrong object type. I think using `format_to_config` is the right way, though I think we can leave it to just `if input_obj.depth != cfgclass.DEPTH:`

src/uwtools/config/support.py
maddenp-noaa: Maybe just ignore this for now, then. I may do a small PR to replace this with something that doesn't try to use private methods to get what we want.

src/uwtools/config/tools.py
WeirAE: Nope, you're right, got to define None depths

src/uwtools/config/formats/sh.py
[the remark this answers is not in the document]
WeirAE: Looking at this again, do we actually want to be checking depths in a dump method?  Now that I think about it, this was really just part of the INI/SH parameters and doesn't accomplish anything useful here - I'm removing it for involving the depth check in a more meaningful way elsewhere

src/uwtools/config/support.py
WeirAE: I had an alternate version for a minute that went straight to pulling the `dict` using the open `sections()` method but it was causing more issues and went back to this one. I'll see if I can come up with a solution too

maddenp-noaa: Since INI is a fixed number of levels, we don't even need recursion to produce a real `dict` from a `ConfigParser` object:
```
>>> from configparser import ConfigParser
>>> c = ConfigParser()
>>> c.read_string("[fruit]\na=apple\nb=banana\n")
>>> {s: {k: v for k, v in c[s].items()} for s in c.sections()}
{'fruit': {'a': 'apple', 'b': 'banana'}}
```
Using that should negate the need for those `pylint` and `type` suppressors, should make the type annotation trivial (`-> Dict[str, Dict[str, str]]` I think), and should simplify things on the calling side, too: Create the `ConfigParser`, read into it, then convert it to a `dict`.

src/uwtools/config/formats/ini.py
```diff
@@ -40,20 +36,12 @@ def _load(self, config_file: OptionalPath) -> dict:
...
-        # representation of the parse config.
-
         cfg = configparser.ConfigParser()
-        cfg.optionxform = str  # type: ignore
-        sections = cfg._sections  # type: ignore # pylint: disable=protected-access
+        sections = config_sections(cfg)
         with readable(config_file) as f:
             raw = f.read()
-        try:
             cfg.read_string(raw)
```
maddenp-noaa, line 43: Please outdent one level.

src/uwtools/config/formats/ini.py
```diff
@@ -40,20 +36,12 @@ def _load(self, config_file: OptionalPath) -> dict:
...
-            return dict(sections)
-        except configparser.MissingSectionHeaderError:
-            cfg.read_string("[top]\n" + raw)
-            return dict(sections.get("top"))
+        return sections
```
maddenp-noaa, line 44: If the `ConfigParser` -> `dict` one-liner I posted elsewhere works, I think a logical flow here might look like
```
cfg = configparser.ConfigParser()
with readable(config_file) as f:
    cfg.read_string(f.read())
return {s: {k: v for k, v in cfg[s].items()} for s in cfg.sections()}
```
Then we don't need `config_sections()` at all.

The implementation in `SHConfig` might be something like
```
cfg = configparser.ConfigParser()
section = "top"
with readable(config_file) as f:
    cfg.read_string(f"[{section}]\n" + f.read())
return {k: v for k, v in cfg[section].items()}
```
I think...

src/uwtools/config/support.py
maddenp-noaa: As I commented elsewhere, this entire function could easily disappear by simplifying the call sites and inlining simple `ConfigParser` -> `dict` expressions there.

src/uwtools/tests/config/formats/test_sh.py
```diff
@@ -0,0 +1,42 @@
...
+    cfgobj = SHConfig(fixture_path("include_files.sh"))
+    assert cfgobj["fruit"] == "papaya"
+    assert cfgobj["how_many"] == "17"
+    assert cfgobj["meat"] == "beef"
+    assert len(cfgobj) == 5
+
+
+def test_bash(salad_base, tmp_path):
+    """
+    Test that bash config load and dump work with a basic bash file.
```
maddenp-noaa, line 28: `bash` -> `sh` -- your editor's search function may turn up more cases...

src/uwtools/config/formats/ini.py
```diff
@@ -63,31 +51,25 @@ def dump(self, path: OptionalPath) -> None:
...
-        cfgdepth = depth(cfg)
-        assert cfgdepth in (1, 2)  # 1 => .sh, 2 => .ini
-        if cfgdepth == 2:
-            parser.read_dict(cfg)
-            parser.write(s, space_around_delimiters=opts.space if opts else True)
-        else:
-            for key, value in cfg.items():
-                print(f"{key}={value}", file=s)
+        parser.read_dict(cfg)
+        parser.write(s, space_around_delimiters=True)
```
christinaholtNOAA, line 72: `space_around_delimiters` is True by default, so you might simplify to `parser.write(s)` for INI formats.

src/uwtools/config/formats/sh.py
```diff
@@ -0,0 +1,70 @@
...
+        See docs for Config._load().
+
+        :param config_file: Path to config file to load.
+        """
+        cfg = configparser.ConfigParser()
+        sections = config_sections(cfg)
+        with readable(config_file) as f:
+            raw = f.read()
+            cfg.read_string("[top]\n" + raw)
+            return dict(sections.get("top"))
```
christinaholtNOAA, line 44: These two lines might also need to be outdented.

src/uwtools/config/formats/ini.py
```diff
@@ -1,10 +1,9 @@
+# pylint: disable=duplicate-code
 import configparser
 from io import StringIO
-from types import SimpleNamespace as ns
-from typing import Optional
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import depth
```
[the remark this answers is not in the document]
maddenp-noaa: It could be that `DEPTH` is still not _quite_ the right name -- maybe `REQUIRED_DEPTH` would be more explanatory. But, to answer your question: `DEPTH` is meant to define what the depth of each config format _must_ be, while `depth()` says what a given config's depth _is_. In fact, calling `depth()` probably really only makes sense for YAML configs at the moment, as that's the only format permitting variable depth. The rest are all fixed. But, if we ever needed to work with JSON, or with XML as an input instead of just as an output, `depth()` could be meaningful for them, too.

src/uwtools/config/formats/ini.py
WeirAE: So, weirdly pylint has been throwing `unnecessary-comprehension` errors but insisting on `dict(cfg[section].items())` regardless of what I do. I've disabled that check in ini/sh for now - I think its a quite valuable comprehension, pylint, thank you very much

maddenp-noaa: It's actually good advice, we should take it:
```
-        return {s: {k: v for k, v in cfg[s].items()} for s in cfg.sections()}
+        return {s: dict(cfg[s].items()) for s in cfg.sections()}
```
With this change, the `unnecessary-comprehension` complaint disappears. I should have written it that way in the first place.

Also, in general, it's better to suppress linter or typechecker complaints in the most targeted way possible, with either a `# pylint: disable=` trailing comment on the problematic line, or a `# pylint: disable-next=` comment as the line immediately preceding the problematic line. Otherwise, the directive suppresses that check for the entire scope it appears in -- which is the entire file if placed in the outermost scope of the module. That can blind us to issues that we would otherwise be warned about.

WeirAE: I was oddly having issues with the inline suppression not reading, which is why I had put it up front to see. But not an issue now - looks like pylint was right. I think I was misunderstanding the instruction, somehow I was not doing exactly what it wanted. But thank you, I've got it all fixed now.

WeirAE: Also, unsurprisingly, ChatGPT had no idea how to actually deal with it

src/uwtools/tests/config/formats/test_ini.py
```diff
@@ -1,44 +1,25 @@
...
-    assert filecmp.cmp(infile, outfile)
-    cfgobj.update({"dressing": ["ranch", "italian"]})
-    expected["dressing"] = ["ranch", "italian"]
-    assert cfgobj == expected
-
-
 def test_parse_include():
     """
-    Test that non-YAML handles include tags properly for INI with no sections.
+    Test that an INI file with no sections handles include tags properly.
```
maddenp-noaa, line 16: But `include_files.ini` _does_ have a section, `[config]`, right?

maddenp-noaa (approved):
Nice work, thanks for following through on all the ideas that came up during implementation.

PR #349, opened 2023-11-22
UW-438

maddenp-noaa:
**Description**

Create initial CLI-tracking `uwtools.api` package.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -2,7 +2,7 @@
 
 from typing import Optional
 
-from uwtools.config.atparse_to_jinja2 import convert as _a2j
+from uwtools.config.atparse_to_jinja2 import convert as _convert_atparse_to_jinja2
 from uwtools.config.tools import compare_configs as _compare
 from uwtools.config.tools import realize_config as _realize
 from uwtools.config.validator import validate_yaml as _validate
@@ -63,7 +63,7 @@ def translate(
     ???
     """
     if input_format == FORMAT.atparse and output_format == FORMAT.jinja2:
-        _a2j(input_file=input_file, output_file=output_file, dry_run=dry_run)
+        _convert_atparse_to_jinja2(input_file=input_file, output_file=output_file, dry_run=dry_run)
         return True
     return False
 
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -7,7 +7,7 @@ from uwtools.config.tools import compare_configs as _compare
 from uwtools.config.tools import realize_config as _realize
 from uwtools.config.validator import validate_yaml as _validate
 from uwtools.types import DefinitePath, OptionalPath
-from uwtools.utils.file import FORMAT
+from uwtools.utils.file import FORMAT as _FORMAT
 
 
 def compare(
@@ -62,7 +62,7 @@ def translate(
     """
     ???
     """
-    if input_format == FORMAT.atparse and output_format == FORMAT.jinja2:
+    if input_format == _FORMAT.atparse and output_format == _FORMAT.jinja2:
         _convert_atparse_to_jinja2(input_file=input_file, output_file=output_file, dry_run=dry_run)
         return True
     return False
@@ -72,6 +72,6 @@ def validate(input_file: DefinitePath, input_format: str, schema_file: DefiniteP
     """
     ???
     """
-    if input_format == FORMAT.yaml:
+    if input_format == _FORMAT.yaml:
         return _validate(config_file=input_file, schema_file=schema_file)
     return False
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -4,7 +4,8 @@ import uwtools.config.atparse_to_jinja2
 import uwtools.config.validator
 from uwtools.config.tools import compare_configs as compare
 from uwtools.config.tools import realize_config as realize
-from uwtools.utils.file import FORMAT, DefinitePath
+from uwtools.types import DefinitePath
+from uwtools.utils.file import FORMAT
 
 
 def translate(
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -1,3 +1,4 @@
 # pylint: disable=unused-import
 
 from uwtools.config.tools import compare_configs as compare
+from uwtools.config.tools import realize_config as realize
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -1,9 +1,9 @@
 # pylint: disable=unused-import
 
-import uwtools.config.atparse_to_jinja2
-import uwtools.config.validator
+from uwtools.config.atparse_to_jinja2 import convert as _a2j
 from uwtools.config.tools import compare_configs as compare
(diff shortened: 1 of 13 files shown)


src/uwtools/api/config.py
```diff
@@ -0,0 +1,77 @@
+# pylint: disable=unused-import
+
+from typing import Optional
+
+from uwtools.config.atparse_to_jinja2 import convert as _convert_atparse_to_jinja2
+from uwtools.config.tools import compare_configs as _compare
+from uwtools.config.tools import realize_config as _realize
+from uwtools.config.validator import validate_yaml as _validate
```
maddenp-noaa, line 8: Care has been taken to import code from other modules with private local names. Note that, for example, what this module makes public and what is kept private:
```
>>> import uwtools.api.config
>>> dir(uwtools.api.config)
['DefinitePath', 'FORMAT', 'Optional', 'OptionalPath', '__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_compare', '_convert_atparse_to_jinja2', '_realize', '_validate', 'compare', 'realize', 'translate', 'validate']
```
Only the API functions `compare()`, `realize()`, `translate()`, and `validate()` are public -- everything else is private. For example, users should not call `uwtools.config.atparse_to_jinja2.convert_atparse_to_jinja2()` directly, so it is imported privately in this module to avoid exposing it.

src/uwtools/api/config.py
```diff
@@ -0,0 +1,77 @@
...
+
+def compare(
+    config_a_path: DefinitePath,
+    config_a_format: str,
+    config_b_path: DefinitePath,
+    config_b_format: str,
+) -> bool:
+    """
+    ???
+    """
```
maddenp-noaa, line 21: Deferring docstring writing until [UW-443](https://jira.epic.oarcloud.noaa.gov/browse/UW-443) is worked, to ensure that the docstrings are written together to ensure a consistent style. These will be the docstrings that we will hopefully be able to use Sphinx to auto-generate API documentation from.

src/uwtools/api/config.py
```diff
@@ -0,0 +1,77 @@
...
+    return _compare(
+        config_a_path=config_a_path,
+        config_a_format=config_a_format,
+        config_b_path=config_b_path,
+        config_b_format=config_b_format,
+    )
+
+
+def realize(
+    input_file: OptionalPath,
```
maddenp-noaa, line 31: I had previously thought that the API should insist on files and not support streaming IO, but two things changed my mind:
1. Since the CLI has been updated to use the API functions as its interface to the rest of the `uwtools` library, and since the CLI needs to support both file and streaming IO, the API needs to support streams, too. Otherwise, the CLI would need to duplicate much of what's here, only with type changes for the inputs and outputs that should support streaming. That seems like it'd be an unfortunate DRY violation.
2. A user might well like to use streaming IO, too -- for example to create their own CLI based on the `uwtools` API.

So, everything in this PR continues to support the ability to use ether file or streaming IO.

src/uwtools/cli.py
```diff
@@ -191,7 +191,7 @@ def _dispatch_config_compare(args: Args) -> bool:
 
     :param args: Parsed command-line args.
     """
-    return uwtools.config.tools.compare_configs(
+    return uwtools.api.config.compare(
```
maddenp-noaa, line 194: Here are below, the CLI now accesses `uwtools` library functionality via the API.

src/uwtools/cli.py
```diff
@@ -223,16 +223,13 @@ def _dispatch_config_translate(args: Args) -> bool:
...
-    success = True
-    if args[STR.infmt] == FORMAT.atparse and args[STR.outfmt] == FORMAT.jinja2:
-        uwtools.config.atparse_to_jinja2.convert(
-            input_file=args[STR.infile],
-            output_file=args[STR.outfile],
-            dry_run=args[STR.dryrun],
-        )
-    else:
-        success = False
-    return success
```
maddenp-noaa, line 235: This logic moved into the API.

src/uwtools/config/atparse_to_jinja2.py
```diff
@@ -3,14 +3,14 @@
...
+from typing import IO, Any, Generator
 
 from uwtools.logging import log
-from uwtools.utils.file import readable, writable
+from uwtools.utils.file import OptionalPath, readable, writable
 
 
 def convert(
-    input_file: Optional[str] = None, output_file: Optional[str] = None, dry_run: bool = False
+    input_file: OptionalPath = None, output_file: OptionalPath = None, dry_run: bool = False
```
maddenp-noaa, line 13: I found a couple more `str` arguments to update to `OptionalPath` or `DefinitePath` to support either `str` or `Path` representations of filesystem paths.

src/uwtools/tests/test_cli.py
```diff
@@ -10,6 +10,11 @@
 import pytest
 from pytest import fixture, raises
 
+import uwtools.api.config
+import uwtools.api.forecast
+import uwtools.api.rocoto
+import uwtools.api.template
```
maddenp-noaa, line 16: The test of the CLI dispatch functions now essentially just check that they are wired up to the API functions correctly.

src/uwtools/cli.py
```diff
@@ -2,7 +2,7 @@
 Modal CLI.
 """
 
-import datetime
+import datetime as dt
```
NaureenBharwaniNOAA, line 5: This is the other choice I was referring to.

src/uwtools/cli.py
NaureenBharwaniNOAA: I appreciate the simplicity of moving this logic to the API. It makes this function and the others below broken up into more readable steps.

src/uwtools/tests/api/test_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@pytest.mark.parametrize(
+    "infmt,outfmt,success_expected",
+    [
+        (FORMAT.atparse, FORMAT.jinja2, True),
+        ("invalid", FORMAT.jinja2, False),
+        (FORMAT.atparse, "invalid", False),
+    ],
+)
```
NaureenBharwaniNOAA, line 46: This is great example of parametrization.

src/uwtools/api/forecast.py
```diff
@@ -0,0 +1,22 @@
+from datetime import datetime
```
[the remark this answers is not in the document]
maddenp-noaa: Thanks for noticing this -- will fix.

christinaholtNOAA (approved):
This seems like a nice, clean, straightforward start to the solution. :)

WeirAE (approved):
I've tried to find anything else I missed, but it looks solid to me

src/uwtools/api/config.py
maddenp-noaa: All of the above boils down to a simple recipe for API users:

1. Only use the `uwtools.api` package.
2. Only use the public members of that package.

NaureenBharwaniNOAA (approved):

Issue #352, opened 2023-11-30
UW-420: Forbid level-incompatible config transformations


WeirAE:
**Description**
Different config formats support varying numbers of levels. It is not necessarily possible to transform between config types with incompatible levels.

**Requirements**
This feature will expand existing depth check functionality to ensure level-incompatible config transformations are forbidden and properly logged.

**Acceptance Criteria (Definition of Done)**
1. Agreed-upon "bad" transformations fail
2. Useful error messages are presented to users
3. Tests exercise the checks and ensure that error messages are provided

**Additional context**
Agreed-upon "Bad" transformations are that when a config dict is changed, the depth limit is violated.
This ticket will not ensure .data is protected or validate key value formats.

PR #353, opened 2023-11-30
UW-420: Forbid level-incompatible config transformations

WeirAE:
**Description**

This feature will expand existing depth check functionality to ensure level-incompatible config transformations are forbidden and properly logged.

Fixes issue #352 

**Type**

- [ ] Documentation

**Impact**

Select one:

**Checklist**

Affirm:


src/uwtools/config/formats/ini.py
```diff
@@ -62,6 +63,7 @@ def dump_dict(path: OptionalPath, cfg: dict) -> None:
         # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
+        assert depth(cfg) <= INIConfig.DEPTH
```
WeirAE, line 66: I used asserts here due `config_check_depths` requiring `(Config, str)`. I tried a variations where `config_check_depths` would handle Union[Config, dict] but went with asserts here rather than an if split in the check, although it would be a fair choice

src/uwtools/config/tools.py
```diff
@@ -38,6 +38,24 @@ def compare_configs(
...
+def config_check_depths(input_obj: Config, target_format: Optional[str]) -> None:
+    """
+    Check that the depth of the input config does not exceed the target format's max.
+
+    :param input_obj: The input config.
+    :param target_format: The target config.:
+    :raises: UWConfigError on excessive input-config depth.
+    """
+    cfgclass = format_to_config(str(target_format))
+    if cfgclass.DEPTH and input_obj.depth != cfgclass.DEPTH:
```
christinaholtNOAA, line 50: I understand this was the original code, and it's just been moved into the appropriate place when it was renamed. One interesting thing that our conversations have brought up, though, is that we don't require a values file (now called supplemental files) to be the _same_ depth, but we do require that it doesn't exceed the target depth (as mentioned in the doc string above).

How should we approach a check where it can handle both cases?

christinaholtNOAA (commented):
So far, just one comment below. This looks like it's on track, but make sure to handle all the Acceptance Criteria for the ticket.

src/uwtools/config/tools.py
WeirAE: Thanks  - I wanted to make sure my work was up before I left for a break. Hmm, I reflexively wanted to make it `!<=` but that's not quite right. It will have to know 'why' it was called, either separating it into 2+ functions or by passing it a flag. I'm inclined to give it a flag while also moving the format assert versions back into here

src/uwtools/config/formats/ini.py
WeirAE: moved inside of config_check_depths for consistency

src/uwtools/config/tools.py
```diff
@@ -38,6 +38,43 @@ def compare_configs(
     return cfg_a.compare_config(cfg_b.data)
 
 
+def config_check_depths(
+    input_obj: Union[dict, Config], target_format: Optional[str], mode: Optional[str]
```
christinaholtNOAA, line 42: What happens when `target_format` is `None`?

What happens if `mode` is `None`?

This seems like another opportunity to define a strings object instead of introducing magic strings.

src/uwtools/config/formats/ini.py
```diff
@@ -63,6 +64,8 @@ def dump_dict(path: OptionalPath, cfg: dict) -> None:
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
 
+        config_check_depths(cfg, target_format="ini", mode="dump")
```
christinaholtNOAA, line 67: Formats are already defined in a data class. Consider using those here.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -127,6 +128,18 @@ def test_dereference(caplog, config):
     assert config == {"foo": 88, "a": 77, "b": {"c": 66}, "d": "{{ X }}"}
 
 
+@pytest.mark.parametrize("fmt2", [FORMAT.ini, FORMAT.nml, FORMAT.sh])
+def test_invalid_config(fmt2, tmp_path):
+    """
+    Test that invalid config files will error when attempting to dump.
+    """
+    fmt1 = FORMAT.yaml
+    outfile = tmp_path / f"test_{fmt1.lower()}to{fmt2.lower()}_dump.{fmt2}"
```
christinaholtNOAA, line 137: Does it matter if these are lowercase? Would they ever not be lowercase?

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -127,6 +128,18 @@ def test_dereference(caplog, config):
...
+    cfgin = tools.format_to_config(fmt1)(fixture_path(f"hello_workflow.{fmt1}"))
```
christinaholtNOAA, line 138:
```suggestion
    cfgin = tools.format_to_config(fmt1)(fixture_path(f"hello_workflow.yaml"))
```

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -127,6 +128,18 @@ def test_dereference(caplog, config):
...
+    with raises(UWConfigError):
+        tools.format_to_config(fmt2).dump_dict(path=outfile, cfg=cfgin.data)
```
christinaholtNOAA, line 140: Does it make sense to capture that the right log message shows up here?

src/uwtools/tests/config/test_tools.py
```diff
@@ -141,6 +141,13 @@ def test_compare_configs_bad_format(caplog):
     assert msg in str(e.value)
 
 
+@pytest.mark.parametrize("fmt", ["ini", "sh"])
```
christinaholtNOAA, line 144: Use data class here, too?

christinaholtNOAA (changes requested):

src/uwtools/tests/config/formats/test_base.py
WeirAE: Yep, got to change `raises` to matching the log message

src/uwtools/config/formats/ini.py
WeirAE: I've tried pulling from the input with `get_file_type(str(path))`

src/uwtools/config/tools.py
WeirAE: True - I've updated this to be required str, and perform a mode lookup check

src/uwtools/tests/config/formats/test_base.py
WeirAE: Now that you mention it, its just an arbitrary test object, it doesn't matter at all - I'll take these out

src/uwtools/tests/config/test_tools.py
WeirAE: Ah yes - don't even need to run both just specify it here

src/uwtools/tests/config/formats/test_base.py
WeirAE: Got it, kept both tests too

src/uwtools/config/tools.py
WeirAE: Error logging now added

src/uwtools/config/formats/ini.py
```diff
@@ -62,6 +63,8 @@ def dump_dict(path: OptionalPath, cfg: dict) -> None:
         # when an INI contains multiple sections. Unfortunately, it also adds a newline after the
         # _final_ section, resulting in an anomalous trailing newline. To avoid this, write first to
         # memory, then strip the trailing newline.
+        target_format = get_file_type(str(path))
```
maddenp-noaa, line 66: I suspect trouble if `path` is `None`, which one of the acceptable values of `OptionalPath`. In that case, we'd be streaming the output to `stdout`, but would still want to do a depth check. I think the typechecker would have complained about this without the `str()` wrapper, as the type of the input argument of `get_file_type()` is `DefinitePath`, which rules out a `None` value.

But do we even need to call `get_file_type()`? This class is `INIConfig`, so don't we _know_ the type?

WeirAE: Hmm unfortunately the input to `get_file_type()` is `DefinitePath` - I had a fixed type originally, then made it more complicated here, then kind of decided that I liked the soft requirement of a filetype suffix that this adds. Could hard-code it here but check for a suffix in `dump`?

src/uwtools/config/tools.py
```diff
@@ -38,6 +38,47 @@ def compare_configs(
...
+            )
+    if cfgclass.DEPTH and isinstance(input_obj, dict) and mode == "dump":
+        if depth(input_obj) > cfgclass.DEPTH:
+            msg = "Invalid depth-%s input for type-'%s' config" % (
+                depth(input_obj),
+                cfgclass.depth,
+            )
+    if msg:
+        log.error(msg)
+        raise UWConfigError(msg)
```
maddenp-noaa, line 79: 1. This function seems like it would be useful in many contexts, checking not just input objects, but also output objects, value-supplying objects (like those used in rendering templates), etc. What would you think of e.g. `config_obj` as a name instead of `input_obj`, to be more general?
2. Given that callers must specify a `mode` anyway, and therefore must make a choice when calling this function, how about splitting this function into `config_check_depths_dump()`, `config_check_depths_realize()`, and `config_check_depths_update()`? Then the `mode` argument disappears, the magic strings disappear, the validity check on the string disappears, and I think most of the `if` conditionals (there are seven!) disappear. If there's repeated logic between those three functions, it could be DRYed out via one or more private helper functions.

src/uwtools/config/formats/ini.py
maddenp-noaa: Combined with my other suggestion about breaking `config_check_depths()` up into mode-specific functions, could these two lines become just
```
config_check_depths_dump(config_obj=cfg, target_format=FORMAT.ini)
```
?

src/uwtools/config/tools.py
WeirAE: 1. Ah, good point! I've already generalized `output` as `target`, no need to stop there
2. Funny story, I asked ChatGPT for suggestions after I wrote it and one suggestion was to split it up. I didn't use any of its ideas, but that might be the solution

maddenp-noaa: What happens if `input_obj` is an `NMLConfig` and `mode` is `"dump"`? `NMLConfig` has a `DEPTH`, and it is not a `dict`, so wouldn't the `if` on line 60 match? But, then, `"dump"` is not handled in that block. But if I'm trying to `dump()` and `NMLConfig` whose max depth has been exceeded through user mischief, I'd want a check on that. Or have I missed something?

Please ignore this comment while considering breaking this function up into smaller, mode-specific ones -- I suspect that, if that eliminates some of the conditionals then it may be easier to see if we're missing cases we need to handle.

src/uwtools/config/formats/ini.py
WeirAE: Removing this for the data class

src/uwtools/config/tools.py
WeirAE: I ~think~ that's handled correctly in the split versions now

src/uwtools/config/formats/ini.py
WeirAE: Yep, looks nice and clean now!

src/uwtools/config/tools.py
WeirAE: Just chucked the mode entry for separate functions as recommended!

src/uwtools/config/tools.py
```diff
@@ -38,6 +37,57 @@ def compare_configs(
     return cfg_a.compare_config(cfg_b.data)
 
 
+def config_check_depths_dump(config_obj: dict, target_format: str) -> None:
+    """
+    Check that the depth of the input config does not exceed the target format's max.
+
+    :param config_obj: The reference config object.
+    :param target_format: The target format.
+    :raises: UWConfigError on excessive input-config depth.
```
maddenp-noaa, line 46: Here and below, `input-config depth` -> `reference-config depth`, given that you refer to `config_obj` above as `The reference config object`? Or just `config depth` or even just `depth`, given that the function name, `config_check_depths_dump` provides the context that we are checking depths on configs?

maddenp-noaa: I'm guessing that by `input config` in the docstring's first line you mean "input" as in "input to this function"? I think this may be confusing, as we use terms like "input" and "output" elsewhere to refer to inputs and outputs of bigger operations, like rendering a template or realizing or validating a config. Could we maybe just call it "config" or "config object" to match the terminology of the argument in question, `config_obj`? Or people sometimes say things like "..of the given config object..." where "given" (or "provided" or "specified" etc.) implies that we're talking about an argument value.

src/uwtools/config/tools.py
```diff
@@ -38,6 +37,57 @@ def compare_configs(
...
+    """
+    cfgclass = format_to_config(target_format)
+    if cfgclass.DEPTH and depth(config_obj) > cfgclass.DEPTH:
+        msg = "Invalid depth-%s input for type-'%s' config" % (
+            depth(config_obj),
+            cfgclass.depth,
```
maddenp-noaa, line 52: Should this be `target_format` instead of `cfgclass.depth`? It's being interpolated into the text `type-'%s'`, so would we want a name instead of a number here? Similarly below...

WeirAE: True - originally it was comparing depths, but this is clearer

src/uwtools/config/tools.py
WeirAE: Mmm, making it just `depth` clears a lot of that up on its own. Yes, that arg text should be updated - actually, the log message should as well since it also says 'input'

src/uwtools/config/tools.py
```diff
@@ -38,6 +37,57 @@ def compare_configs(
...
+    :param target_format: The target format.
+    :raises: UWConfigError on excessive input-config depth.
+    """
+    cfgclass = format_to_config(target_format)
+    if cfgclass.DEPTH and config_obj.depth > cfgclass.DEPTH:
+        msg = "Cannot update depth-%s input with type-'%s' values" % (
+            config_obj.depth,
+            cfgclass.depth,
+        )
+        raise log_and_error(msg)
```
maddenp-noaa, line 88: This is all looking better to me than the big block of conditionals!

I'm curious as to why one of these functions accepts a `dict` and the others accept `Config`s. Wouldn't either input be an appropriate thing to test? If so, we could take a `Union[dict, Config]` argument and then e.g.
```
config_obj = config_obj.data if isinstance(config_obj, Config) else config_obj
```
to ensure that what follows operates strictly on a `dict`. That is: Establish an invariant and then proceed with confidence.

I feel like that'd make these functions more universally useful throughout the codebase, and would eliminate an apparently (but please correct me if I'm wrong) unnecessary difference.

WeirAE: This is kind of an interesting structural question - since `dump` is a concrete static method, it takes the in-memory object and not anything else. In the first version I left the `formats` with individual `assert` statements, then the second version had everything together with that `Union` logic

WeirAE: Ah wait, I have an idea, I think this might be the way but for another reason, one sec

WeirAE: There, got it - it turns out that while we were checking `dump_dict()`, `dump()` was not being checked. While unlikely to be broken, it doesn't hurt and makes a good direct use for returning `Union`

maddenp-noaa: Looking good!

At this point, `config_check_depths_dump()` and `config_check_depths_update()` are identical except for a few characters in their respective `msg` values. Those two functions could be collapsed into a single line each that calls a common helper that takes the differing part of the `msg` as a 3rd argument. To cut down on boilerplate, I think the official docstring could live in the helper, with the two gutted functions just saying `See <whatever>()'s docstring.` where `<whatever>` is whatever the helper is called. We're honing in on the essentials now, trimming away the accidentals...

As a potential final refactor, consider that `config_check_depths_realize()` differs from the other two only in the comparison it performs (and its `msg`, but the paragraph above would generalize that). That, too, could be factored out and passed in to the common helper. For example, if `config_check_depths_realize()` defined
```
bad_depth = lambda need, have: need and have != need
```
and the other two defined
```
bad_depth = lambda need, have: need and have > need
```
and passed `bad_depth` in to the proposed helper as a 4th argument, and where, say, `action` is the 3rd argument from above making `msg` conext-appropriate, the helper could have e.g.
```
if bad_depth(cfgclass.DEPTH, config_obj.depth):
    raise log_and_error(
        "Cannot %s depth-%s config to type-'%s' config" % (
            action, config_obj.depth, target_format
        )
    )
```
(The type of a lambda/function/method argument is `Callable`, which can be imported from `typing`.)

That might be a very clean and DRY refactor.

WeirAE: Took me a minute to follow, but this is a brilliantly clean solution. I need to consider lambdas more!

src/uwtools/config/tools.py
```diff
@@ -38,6 +37,43 @@ def compare_configs(
...
 
+def config_check_depths_dump(config_obj: Union[Config, dict], target_format: str) -> None:
+    """
+    Check that the depth does not exceed the target format's max.
+
+    :param config_obj: The reference config dictionary.
+    :param target_format: The target format.
+    :raises: UWConfigError on excessive config object dictionary depth.
+    """
+    # define a function with the conditions of an invalid depth
```
maddenp-noaa, line 48: nitpick: Capitalize comment sentence and end with a period, here and below.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -127,6 +128,21 @@ def test_dereference(caplog, config):
...
+    Test that invalid config files will error when attempting to dump.
+    """
+    fmt1 = FORMAT.yaml
+    outfile = tmp_path / f"test_{fmt1}to{fmt2}_dump.{fmt2}"
+    cfgin = tools.format_to_config(fmt1)(fixture_path("hello_workflow.yaml"))
+    with raises(UWConfigError) as e:
+        tools.format_to_config(fmt2).dump_dict(path=outfile, cfg=cfgin.data)
+    msg = f"Cannot dump depth-8 config to type-'{fmt2}' config"
+    assert logged(caplog, msg)
+    assert msg in str(e.value)
```
maddenp-noaa, line 143: Nice test!

src/uwtools/config/tools.py
WeirAE: Aaaah! Thanks lol

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -127,6 +128,21 @@ def test_dereference(caplog, config):
...
+def test_invalid_config(caplog, fmt2, tmp_path):
+    """
+    Test that invalid config files will error when attempting to dump.
+    """
+    fmt1 = FORMAT.yaml
+    outfile = tmp_path / f"test_{fmt1}to{fmt2}_dump.{fmt2}"
+    cfgin = tools.format_to_config(fmt1)(fixture_path("hello_workflow.yaml"))
+    with raises(UWConfigError) as e:
+        tools.format_to_config(fmt2).dump_dict(path=outfile, cfg=cfgin.data)
+    msg = f"Cannot dump depth-8 config to type-'{fmt2}' config"
```
christinaholtNOAA, line 141: I think you might consider here and in later similar tests improving the maintainability of the messages.

If you use the tool we have to check the depth of `cfgin` and have that dynamically included in the string, the tests will be more resilient to changes in our fixture files that are irrelevant to this test. If we changed it so much that it was compatible with config it's tested against, the test would fail and we'd maybe choose a new fixture file.

christinaholtNOAA (approved):
LGTM.

src/uwtools/tests/config/formats/test_base.py
WeirAE: Good point! I've updated it to pull `depth(cfgin.data)` in both sets of tests, thats for the best - thank you!

PR #357, opened 2023-12-05
UW-439

maddenp-noaa:
**Description**

Add API support for in-memory objects, supplementing file-based operations.

**Type**

**Impact**

**Checklist**


src/uwtools/api/config.py
```diff
@@ -28,24 +33,50 @@ def compare(
...
-    values_format: Optional[str],
+    input_config: Union[dict, _Config, OptionalPath] = None,
+    input_format: Optional[str] = None,
+    output_file: OptionalPath = None,
+    output_format: Optional[str] = None,
+    values: Union[dict, _Config, OptionalPath] = None,
+    values_format: Optional[str] = None,
     values_needed: bool = False,
     dry_run: bool = False,
 ) -> bool:
```
maddenp-noaa, line 45: For the API analog to `uw config realize`, we now support providing the "base" input config as either a `dict` or some kind of `Config` object, in addition to existing support for filesystem paths and reads from `stdin`. Similarly for the values. (I'm not sure if `OptionalPath` is the correct type for the values argument, but this argument will go away soon due to [UW-421](https://jira.epic.oarcloud.noaa.gov/browse/UW-421).

The argument names changed to reflect the fact that we are not necessarily dealing with `file`s now. I would have preferred `input` over `input_config`, but that conflicts with a Python built-in. I'm open to better ideas about any names!

I also added some `None` defaults so that API users need not specify optional arguments. I didn't notice the need before because only the CLI is currently calling these API functions, and it supplies values for _all_ arguments. But API users should not have to.

src/uwtools/api/config.py
```diff
@@ -28,24 +33,50 @@ def compare(
...
+        input_config=input_config,
+        input_format=input_format,
+        output_file=output_file,
+        output_format=output_format,
+        values=values,
+        values_format=values_format,
+        values_needed=values_needed,
+        dry_run=dry_run,
+    )
+    return True
```
maddenp-noaa, line 59: The `realize()` API function calls `realize_to_str()` and just discards the returned string value. It needs to return a `bool` to interface correctly with the CLI. If errors occur in the library, past the `realize()` call, they should raise exceptions rather than signaling errors via return values.

Sorry about all the boilerplate: Much of what's happening in the API code is just piping calls from the CLI or users to the `uwtools` innards. We'll add value later to these blocks by writing good docstrings that will be turned into HTML API documentation for users.

src/uwtools/api/config.py
```diff
@@ -68,10 +98,32 @@ def translate(
...
-    if input_format == _FORMAT.yaml:
-        return _validate(config_file=input_file, schema_file=schema_file)
-    return False
+    if isinstance(config, dict):
+        cfgobj = _YAMLConfig(empty=True)
+        cfgobj.update(config)
+        return _validate_yaml_config(schema_file=schema_file, config=cfgobj)
+    if isinstance(config, _Config):
+        return validate(schema_file=schema_file, config=config.data)
+    return _validate_yaml_file(schema_file=schema_file, config_file=config)
```
maddenp-noaa, line 113: `validate()`, as extended for API use, can operate on an in-memory `dict` or `Config` objects, as well as the existing support for file/stream IO. The `YAMLConfig` class is used for validating a `dict` object.

src/uwtools/api/config.py
```diff
@@ -68,10 +98,32 @@ def translate(
...
+    config: Union[dict, _Config, OptionalPath]
+) -> Union[_Config, OptionalPath]:
+    """
+    ???
+    """
+    if isinstance(config, dict):
+        cfgobj = _YAMLConfig(empty=True)
+        cfgobj.update(config)
+        return cfgobj
+    return config
```
maddenp-noaa, line 131: A helper function to reduce arguments of `dict` / `Config` / path types to `Config` / path types.

src/uwtools/api/rocoto.py
```diff
@@ -1,17 +1,18 @@
 from uwtools.rocoto import realize_rocoto_xml as _realize
-from uwtools.rocoto import validate_rocoto_xml as _validate
+from uwtools.rocoto import validate_rocoto_xml_file as _validate
 from uwtools.types import OptionalPath
 
 
-def realize(input_file: OptionalPath, output_file: OptionalPath) -> bool:
+def realize(input_file: OptionalPath = None, output_file: OptionalPath = None) -> bool:
```
maddenp-noaa, line 6: Make optional arguments truly optional.

src/uwtools/api/template.py
```diff
@@ -1,14 +1,14 @@
...
 
 from uwtools.config.jinja2 import render as _render
 from uwtools.types import DefinitePath, OptionalPath
 
 
 def render(
-    input_file: OptionalPath,
-    output_file: OptionalPath,
-    values_file: DefinitePath,
+    values: Union[dict, DefinitePath],
```
maddenp-noaa, line 8: The API analog to `uw template render` can now accept an in-memory `dict` as a source for values to render the template, along with the existing file-based values option.

src/uwtools/cli.py
```diff
@@ -164,11 +164,7 @@ def _add_subparser_config_validate(subparsers: Subparsers) -> SubmodeChecks:
     _add_arg_schema_file(required)
     optional = _basic_setup(parser)
     _add_arg_input_file(optional)
-    _add_arg_input_format(optional, choices=[FORMAT.yaml])
-    checks = _add_args_quiet_and_verbose(optional)
-    return checks + [
-        partial(_check_file_vs_format, STR.infile, STR.infmt),
-    ]
```
maddenp-noaa, line 171: Removed some currently unneeded complexity: The only configs we can currently validate are YAML ones, so for now simply insist that configs are YAML. If we need to re-complicate this later, we'll probably be working with more knowledge and will implement a better solution.

src/uwtools/apps/srw.py
```diff
@@ -7,7 +7,7 @@
 from uwtools.drivers.facade import Facade
 from uwtools.exceptions import UWError
 from uwtools.logging import log
-from uwtools.utils.file import FORMAT, get_file_type
+from uwtools.utils.file import FORMAT, get_file_format
```
maddenp-noaa, line 10: We use `format` pretty pervasively to describe YAML, namelist, INI, etc., so use that instead of `type`.

src/uwtools/config/formats/base.py
```diff
@@ -22,15 +22,17 @@ class Config(ABC, UserDict):
     several configuration-file formats.
     """
 
-    def __init__(self, config_file: OptionalPath = None) -> None:
+    def __init__(self, config_file: OptionalPath = None, empty: bool = False) -> None:
```
maddenp-noaa, line 25: Add a flag to permit creating a config with an initially empty `data` `dict`. This is helpful for API uses cases where we have an in-memory dict and need a `Config` object based on it to conform to an interface requiring a `Config` argument, to e.g. to validate the config represented by the `dict`. It will be useful in near-future work allowing creation of `Config` objects via the API, specifically for cases where the user wants to start from an in-memory `dict`.

Similar support has been added in the subclasses of `Config`, below, where needed.

src/uwtools/config/jinja2.py
```diff
@@ -157,21 +157,22 @@ def dereference(val: _YAMLVal, context: dict, local: Optional[dict] = None) -> _
...
-    output_file: OptionalPath,
-    values_file: DefinitePath,
+    values: Union[dict, DefinitePath],
     values_format: Optional[str] = None,
+    input_file: OptionalPath = None,
+    output_file: OptionalPath = None,
     overrides: Optional[Dict[str, str]] = None,
     values_needed: bool = False,
     dry_run: bool = False,
 ) -> bool:
```
maddenp-noaa, line 167: A Jinja2 template can now be rendered via values in an in-memory `dict` as well as via the existing file/stream-based sources.

src/uwtools/api/config.py
```diff
@@ -68,10 +97,35 @@ def translate(
...
-        return _validate(config_file=input_file, schema_file=schema_file)
-    return False
+    if isinstance(config, dict):
+        cfgobj = _YAMLConfig(empty=True)
+        cfgobj.update(config)
+    elif isinstance(config, _YAMLConfig):
+        cfgobj = config
+    else:
+        cfgobj = _YAMLConfig(config_file=config)
+    return _validate_yaml_config(schema_file=schema_file, config=cfgobj)
```
maddenp-noaa: `validate()`, as extended for API use, can operate on an in-memory `dict` or `Config` objects, as well as the existing support for file/stream IO. The `YAMLConfig` class is used for validating a `dict` object.

src/uwtools/rocoto.py
```diff
@@ -21,42 +19,43 @@
 
 
 def realize_rocoto_xml(
-    config_file: OptionalPath,
-    output_file: OptionalPath = None,
-) -> bool:
+    config: Union[OptionalPath, YAMLConfig], output_file: OptionalPath = None
```
maddenp-noaa, line 22: Before our latest discussion in Slack about which in-memory support we really wanted to add at the moment, I'd already added support for realizing a Rocoto XML file from an in-memory `YAMLConfig` object in addition to a file support. I've since removed that option from the API but am hesitant to remove it from the library, in case we change our minds about this. So, it's not exposed to users, but it's latent (and unit-tested) capability in case we need it later. I can remove if there are objections.

src/uwtools/rocoto.py
```diff
@@ -21,42 +19,43 @@
...
-    return True
+    :param xml: Path to XML file (None => read stdin).
+    :return: Did the XML conform to the schema?
+    """
+    with readable(xml_file) as f:
+        return validate_rocoto_xml_string(xml=f.read())
 
 
-def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
+def validate_rocoto_xml_string(xml: str) -> bool:
```
maddenp-noaa, line 51: This function, in combination with `validate_rocoto_xml_file()`, supports validating either on-disk XML files, e.g. via the CLI `uw rocoto validate` command; or in-memory XML strings, which we have via e.g. `uw rocoto realize` and want to validate before writing them to disk. Previously, the code was writing the string to a temp file, then validating that file.

src/uwtools/api/config.py
```diff
@@ -68,10 +97,35 @@ def translate(
     return False
 
 
-def validate(input_file: DefinitePath, input_format: str, schema_file: DefinitePath) -> bool:
+def validate(
```
maddenp-noaa, line 100: `validate()`, as extended for API use, can operate on an in-memory `dict` or `Config` objects, as well as the existing support for file/stream IO. The `YAMLConfig` class is used for validating a `dict` object.

src/uwtools/config/formats/ini.py
```diff
@@ -13,16 +15,13 @@ class INIConfig(Config):
 
     DEPTH = 2
 
-    def __init__(
-        self,
-        config_file: str,
-    ):
+    def __init__(self, config: Union[dict, OptionalPath] = None):
```
maddenp-noaa, line 18: Permit creating `Config` objects with an initial `dict` in addition to the traditional external file. This is helpful for API uses cases where we have an in-memory dict and need a `Config` object based on it to conform to an interface requiring a `Config` argument, to e.g. to validate the config represented by the `dict`. It will be useful in near-future work allowing creation of `Config` objects via the API, specifically for cases where the user wants to start from an in-memory `dict`.

Similar support has been added in the subclasses of `Config`, below, where needed.

Another option would be to rename the `config_file` argument to `config` and extend its type to `Union[dict, OptionalFile]`, so that `Config(config={})` would obtain an empty `Config`, but `Config(config=d)` would initialize it with an arbitrary `dict`. Currently, some code (especially in tests) does `c = Config(empty=True)` followed by `c.update(d)`, and those could be combined in a single statement. Thoughts?

src/uwtools/cli.py
```diff
@@ -238,11 +234,7 @@ def _dispatch_config_validate(args: Args) -> bool:
 
     :param args: Parsed command-line args.
     """
-    return uwtools.api.config.validate(
-        input_file=args[STR.infile],
-        input_format=args[STR.infmt],
-        schema_file=args[STR.schemafile],
-    )
+    return uwtools.api.config.validate(schema_file=args[STR.schemafile], config=args[STR.config])
```
christinaholtNOAA, line 237: Any reason for the order of args here?

src/uwtools/config/formats/base.py
```diff
@@ -22,15 +22,19 @@ class Config(ABC, UserDict):
     several configuration-file formats.
     """
 
-    def __init__(self, config_file: OptionalPath = None) -> None:
+    def __init__(self, config: Union[dict, OptionalPath] = None) -> None:
```
christinaholtNOAA, line 25: Does it make sense for us to have a re-usable type for `Union[dict, OptionalPath]`?

src/uwtools/config/validator.py
```diff
@@ -45,6 +42,23 @@ def validate_yaml(schema_file: DefinitePath, config_file: OptionalPath = None) -
...
+
+    :param config: The config to validate.
+    :return: A dereferenced YAMLConfig object based on the input config.
+    """
+    if isinstance(config, dict):
+        cfgobj = YAMLConfig(config=config)
+    elif isinstance(config, YAMLConfig):
+        cfgobj = config
+    else:
+        cfgobj = YAMLConfig(config=config)
```
christinaholtNOAA, line 57: Am I missing an edge case where this wouldn't reduce?

```
    if isinstance(config, YAMLConfig):
        cfgobj = config
    else:
        cfgobj = YAMLConfig(config=config)
```

src/uwtools/rocoto.py
```diff
@@ -21,42 +19,43 @@
 
 
 def realize_rocoto_xml(
-    config_file: OptionalPath,
-    output_file: OptionalPath = None,
-) -> bool:
+    config: Union[YAMLConfig, OptionalPath], output_file: OptionalPath = None
```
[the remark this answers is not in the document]
christinaholtNOAA: No objections to leaving it.

src/uwtools/rocoto.py
```diff
@@ -21,42 +19,43 @@
...
+    """
+    with readable(xml_file) as f:
+        return validate_rocoto_xml_string(xml=f.read())
 
 
-def validate_rocoto_xml(input_xml: OptionalPath) -> bool:
+def validate_rocoto_xml_string(xml: str) -> bool:
     """
-    Given a rendered XML file, validate it against the Rocoto schema.
+    Validate purported Rocoto XML against its schema.
```
christinaholtNOAA, line 53: I like "purported" here. It gave me some investigative drama feels.

src/uwtools/rocoto.py
```diff
@@ -73,26 +72,30 @@ class _RocotoXML:
...
-        """
+    def __str__(self) -> str:
         # Render the tree to a string, fix mangled entities (e.g. "&amp;FOO;" -> "&FOO;"), insert
         # !DOCTYPE block, then write final XML.
         xml = etree.tostring(
             self._root, pretty_print=True, encoding="utf-8", xml_declaration=True
         ).decode()
         xml = re.sub(r"&amp;([^;]+);", r"&\1;", xml)
         xml = self._insert_doctype(xml)
+        return xml
```
christinaholtNOAA, line 89: Nice.

src/uwtools/rocoto.py
```diff
@@ -280,16 +283,16 @@ def _add_workflow_tasks(self, e: Element, config: dict) -> None:
...
 
-        :param config_file: Path to the YAML config (defaults to stdin).
+        :param config: YAMLConfig object or path to YAML file (None => read stdin).
         """
-        if not validate_yaml(
-            config_file=config_file, schema_file=resource_pathobj("rocoto.jsonschema")
-        ):
-            raise UWConfigError("YAML validation errors identified in %s" % config_file)
+        schema_file = resource_pathobj("rocoto.jsonschema")
+        ok = validate_yaml(schema_file=schema_file, config=config)
```
christinaholtNOAA, line 293: Arg ordering isn't obvious here.

src/uwtools/tests/config/formats/test_ini.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=duplicate-code
+# pylint: disable=duplicate-code,missing-function-docstring
```
christinaholtNOAA, line 1: This is interesting since it's the only addition to the file...and the ones below. Is it necessary, or just here for consistency with others?

If it's the latter, do we want to consider adding a set of rules that only apply to the tests?

christinaholtNOAA (approved):
Looks good in general. Left some comments/questions/suggestions, but all relatively minor.

src/uwtools/cli.py
maddenp-noaa: This order matches the signature of `uwtools.api.config.validate()`, which matches the signature of `uwtools.config.validator.validate_yaml()`, where the order matters because `schema_file` is required and `config` is optional. (I'd accidentally lost the `= None` default on `config` but just added it back -- so I'm glad you asked!)

src/uwtools/config/formats/base.py
maddenp-noaa: Good question. I thought about it too. But we have this one, and also `Union[dict, Config, OptionalPath]` and `Union[dict, YAMLConfig, OptionalPath]`, and `Union[dict, DefinitePath]`, and I didn't immediately come up with names that would be catch enough to not obscure the intent of these types. I thought it might be better to see the full type in the signature as a reminder about what the function/method is required to handle. I'd want that information to be obvious from any shorter type name. If anything comes to mind that you think captures that, I'm interested.

src/uwtools/config/validator.py
maddenp-noaa: I think you're right!

src/uwtools/rocoto.py
maddenp-noaa: I think I covered it in another comment, but that's how the arguments are (and have to be, as `schema_file` is required and `config` is optional) in the called function, so I've followed that ordering at the call site, too. Order isn't required for keyword arguments, of course, but if I'm looking at the call and looking at the function definition, I find it useful to be able to visually match them up in order.

maddenp-noaa:
To support the [update depth checks](https://github.com/ufs-community/workflow-tools/pull/353), I added a static `get_format()` method to `Config` and its descendants so that, if one has a reference to a `Config` object, they can get the corresponding string format name (`yaml`, `ini`, et al).

PR #365, opened 2023-12-19
UW-421: Replace uw config realize --values-file with --supplemental-configs/files, allowing for positional arguments

elcarpenterNOAA:
**Description**

We have removed `--values-file` from config realize and replaced it with `--supplemental_files` or `--supplemental_configs` where appropriate. It previously took in a single optional file, but now accepts an optional list of configs or files. 

**Type**

Select one or more:

- [ ] Documentation

**Impact**

Select one:

**Checklist**

Affirm:


src/uwtools/api/config.py
```diff
@@ -84,8 +84,7 @@ def realize(
     input_format: Optional[str] = None,
     output_file: OptionalPath = None,
     output_format: Optional[str] = None,
-    values: Union[dict, _Config, OptionalPath] = None,
-    values_format: Optional[str] = None,
+    supplemental_configs: Optional[List[Union[dict, _Config, DefinitePath]]] = None,
     values_needed: bool = False,
     dry_run: bool = False,
 ) -> bool:
```
elcarpenterNOAA, line 90: Regarding the order of arguments, I'm unsure of two things: 
1) because `supplemental_configs` takes in positional arguments, should it be placed last in this call (and all similar calls)? 
2) what is the reason behind `values_needed` being listed above `dry_run`?

src/uwtools/cli.py
```diff
@@ -722,7 +727,7 @@ def _formatter(prog: str) -> HelpFormatter:
     """
     A standard formatter for help messages.
     """
-    return HelpFormatter(prog, max_help_position=8)
+    return HelpFormatter(prog, max_help_position=3)
```
elcarpenterNOAA, line 730: This was done to make the formatting consistent, since `"PATH"` was not long enough to trigger a new line.

src/uwtools/api/config.py
```diff
@@ -97,8 +96,7 @@ def realize(
         input_format=input_format,
         output_file=output_file,
         output_format=output_format,
-        values=_ensure_config_arg_type(values),
-        values_format=values_format,
+        supplemental_configs=supplemental_configs,
```
elcarpenterNOAA, line 99: `values_format` has been removed entirely. Users will have to provide recognizable file formats.

src/uwtools/cli.py
christinaholtNOAA: Does it make sense to add more info about this seemingly magic number, especially since it has consequences I wasn't even aware of?

src/uwtools/config/tools.py
```diff
@@ -105,10 +103,10 @@ def realize_config(
         if isinstance(input_config, Config)
         else format_to_config(input_format)(config=input_config)
     )
-    input_obj.dereference()
-    input_obj = _realize_config_update(input_obj, values, values_format)
+    if supplemental_configs:
+        input_obj = _realize_config_update(input_obj, supplemental_configs)
+        input_obj.dereference()
```
christinaholtNOAA, line 108: It might not hurt to do all the updates, then dereference once after all the updates are complete.

christinaholtNOAA (commented):
This looks great and ended up being quite the collection of mods!

src/uwtools/config/tools.py
elcarpenterNOAA: I'm not sure I understand. `_realize_config_update()` is taking in the entire list of `supplemental_configs`, updating all of the values within `_realize_config_update()` and only dereferencing once, upon return to `realize_config()`. Would you prefer `dereference()` to be elsewhere in the function?

src/uwtools/cli.py
elcarpenterNOAA: I've updated it to include an in-line comment explaining `max_help_position` and also updated the number to be 6, simply because I find it more aesthetically pleasing. However, I'm not sure why 6 (or even 7) is an acceptable number for a new line on`"PATH"`, while 8 is not.

src/uwtools/api/config.py
maddenp-noaa: 1. If you mean "positional arguments" with respect to arguments given on the command line, I don't think that concept is still meaningful at this point in the code. After `argparse` parses the command-line arguments in `cli.py`, we get a `Namespace` object containing the parsed arguments, and I don't think we need to treat any of them differently than the others after that, except where it's helpful for us, or makes for a sensible function/method signature. Python only demands that mandatory parameters precede optional ones and, since all the parameters here are optional, the order is up to us. I think best practice is for the most important / meaningful parameters to come first, with less important (and, when there's a distinction, optional) parameters coming later, with those least likely to be of interest to most users coming last. So, here, I guess I see the inputs and outputs as being the "main parameters" (in fact, they're kind of mandatory, in that `stdin` and `stdout` will be used if they're not specified -- we're definitely going to do input and output whether we get values for these arguments or not), with the supplemental configs following in importance, and then the useful but one-off `values_needed` coming next, and then the fairly-fringe `dry_run` coming last.
2. About `dry_run`, another consideration is that many of our functions/methods have a `dry_run` parameter, and we've so far been pretty consistent about having it appear late-to-last in the signature. It's almost an afterthought, and doesn't have much to do with the purpose of the function. Of course it's there, but it's not in your face because it's not very important. If we let it appear in different positions for different functions, it would IMO get in the way of looking at function signatures and understanding what they do. I hope that makes some sense and is vaguely convincing. Please speak up if you disagree!

WeirAE: Interestingly the two are kind of tied together - `dry_run` was stuck at the end where it would usually sit with `quiet` and `verbose` as kind of operands. Though, shouldn't operands go before options? That would put `dry_run` and `values_needed` first and `supplemental_configs` last

src/uwtools/api/config.py
maddenp-noaa: I haven't reviewed the unit tests yet, but if there's not one already, it might be nice to have a test that demonstrates that, if a supplemental config is provided that has no or an unrecognizable suffix, that we fail in an appropriate way that will inform the user of the problem.

src/uwtools/config/tools.py
maddenp-noaa: Will dereferencing get done if there are no supplemental configs? Should we outdent the `input_obj.dereference()` line so that it executes no matter what?

src/uwtools/config/tools.py
```diff
@@ -173,32 +171,30 @@ def _print_config_section(config: dict, key_path: List[str]) -> None:
 
 def _realize_config_update(
     config_obj: Config,
-    values: Union[dict, Config, OptionalPath],
-    values_format: Optional[str] = None,
+    supplemental_configs: Optional[List[Union[dict, Config, DefinitePath]]] = None,
```
maddenp-noaa, line 174: Since we repeat this complex type several times, and since all need to match, we could factor this out as a module-level type, e.g.
```
SupplementalConfigs = Optional[List[Union[dict, Config, DefinitePath]]]
```
near the top of the module, then use `SupplementalConfigs` in place of the complex type. If we need the same across modules, it could be factored out to `uwtools.types` (where `DefinitePath` and `OptionalPath` are defined). I have mixed feelings about this because it does mean having to go look up what the type means when you've forgotten, but this one seems tricky enough to potentially justify that. Your call.

src/uwtools/tests/config/test_tools.py
```diff
@@ -374,6 +364,80 @@ def test_realize_config_simple_yaml(tmp_path):
...
+        supplemental_configs=[supplemental_path],
+        dry_run=True,
+    )
+    expected = """'1': a
+'2': b
+'3': c
+temporalis: c
+"""
+    actual = "\n".join(record.message for record in caplog.records)
+    assert actual == expected
```
maddenp-noaa, line 388: Nice test. To drive the intent home, we might add a `"deref": "d"` key to `b.yaml` to make it clear that `{{ deref }}` never actually gets rendered, but rather just gets clobbered by `"2" : "b"`.

src/uwtools/tests/config/test_tools.py
```diff
@@ -374,6 +364,80 @@ def test_realize_config_simple_yaml(tmp_path):
...
+        output_format=FORMAT.yaml,
+        supplemental_configs=None,
+        dry_run=True,
+    )
+    expected = """'1': a
+'2': '{{ deref }}'
+'3': '{{ temporalis }}'
+"""
+    actual = "\n".join(record.message for record in caplog.records)
+    assert actual == expected
```
maddenp-noaa, line 438: Great demonstrations of none/one/many correctness in these tests. A couple ideas for your consideration -- not asking for a change, just in case you find any of it helpful.

- You could omit the `=None` arguments, as they're IIRC all optional and supply `None` values themselves. Maybe there's a benefit to being explicit, but you could make these tests slightly more concise that way.

- You could reuse the variable `f` (instead of doing `f2` etc.) when doing the various file writes, because the `with` context manager closes the file beyond the block such that `f` has no useful value after the block:
```
>>> with open("/tmp/foo", "w") as f:
...   print("bar", file=f)
... 
>>> print("baz", file=f)
ValueError: I/O operation on closed file.
```
I think it's idiomatic to keep it simple and just use `f` repeatedly.

- You could use `strip()` to e.g.
```
>>> s = """
... hello
... world
... """.strip()
>>> print(s)
hello
world
```
IMO it's nice to see the `expected` output lined up nicely, without the first line of output being at the end of a line of code.

- Instead of `dry_run=True`, you could leave it blank and the output will appear on `stdout` (since you have `output_file=None` -- which you could also omit, per the first bullet point above), which you can read from `capsys` instead of `caplog`. There are examples of this in the test you can borrow from. This is probably actually safer, because technically we could decide to also log other messages alongside the ones showing the realized config, and then these tests would unexpectedly fail. But when we write the output to `stdout` (as opposed to logging to `stderr`), it will be guaranteed to be exactly the same as what would be written to a file -- i.e. it's downstream-programmatically-consumable text, not human-readable text -- and so safe to make an assertion of equality about.

maddenp-noaa (approved):
Everything here looks good to me. I made a few suggestions for your consideration.

src/uwtools/config/tools.py
elcarpenterNOAA: It did not, so I moved it outside of the if statement and have also changed the tests in a way that should catch that in the future.

src/uwtools/tests/config/test_tools.py
elcarpenterNOAA: I like it. Added.

src/uwtools/config/tools.py
elcarpenterNOAA: Since `SupplementalConfigs` requires `Config` to be imported, it creates a circular dependency. I'm going to leave it as-is for now, but I, too, have mixed feeling about the complex type.

src/uwtools/tests/config/test_tools.py
elcarpenterNOAA: I changed the `f2`s and `f3`s to `f`s, removed a bunch of the `=None` lines, and replaced `caplog` with `capsys`, but was only able to get `.strip()` working when I applied it to both `expected` and `actual`. In order to avoid having it both places, I left the lines as they were. I do appreciate all of the suggestions.

src/uwtools/config/tools.py
christinaholtNOAA: My haste in reviewing led me to read it as a loop instead of an if-statement. Sorry for the confusion.

PR #367, opened 2023-12-20
UW-454

maddenp-noaa:
**Description**

Fixes a bug whereby quoted values in `.sh` files incorrectly remain quoted when rendered into templates. For example, given `template`
```
{{ greeting }}, {{ recipient }}!
```
and `values.sh`
```
greeting="Hello there"
recipient="World"
```
the old, incorrect behavior was:
```
% uw template render --input-file template --values-file values.sh
"Hello there", "World"!
```
Given the changes in this PR:
```
% uw template render --input-file template --values-file values.sh
Hello there, World!
```
The underlying issue was that Python's built-in `ConfigParser`, meant for `.ini`-style files, was being used to parse `.sh` files. Since quotes are not required in `.ini` files, `ConfigParser` can only assume that quotes are meant to be there, and so retains them as part of the parsed values.

The updates code uses Python's built-in `shlex` to parse the `.sh` file.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -1,5 +1,6 @@
 import re
 import shlex
+from io import StringIO
 from typing import Optional, Union
 
 from uwtools.config.formats.base import Config
@@ -64,10 +65,16 @@ class SHConfig(Config):
         :param cfg: The in-memory config object to dump.
         """
 
+        # Write first to a StringIO object to avoid creating a partial file in case of problems
+        # rendering or quoting config values.
+
         config_check_depths_dump(config_obj=cfg, target_format=FORMAT.sh)
+        s = StringIO()
+        for key, value in cfg.items():
+            print("%s=%s" % (key, shlex.quote(str(value))), file=s)
         with writable(path) as f:
-            for key, value in cfg.items():
-                print("%s=%s" % (key, shlex.quote(str(value))), file=f)
+            print(s.getvalue().strip(), file=f)
+        s.close()
 
     @staticmethod
     def get_depth_threshold() -> Optional[int]:
--- a/src/uwtools/config/formats/sh.py
+++ b/src/uwtools/config/formats/sh.py
@@ -1,11 +1,10 @@
-# pylint: disable=duplicate-code
-
-import configparser
-from io import StringIO
+import re
+import shlex
 from typing import Optional, Union
 
 from uwtools.config.formats.base import Config
 from uwtools.config.tools import config_check_depths_dump
+from uwtools.logging import log
 from uwtools.utils.file import FORMAT, OptionalPath, readable, writable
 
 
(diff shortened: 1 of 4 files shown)


src/uwtools/config/formats/sh.py
```diff
@@ -27,17 +27,23 @@ def __init__(self, config: Union[dict, OptionalPath] = None):
 
     def _load(self, config_file: OptionalPath) -> dict:
         """
-        Reads and parses shell code consisting solely of key=value lines.
+        Reads and parses key=value lines from shell code.
```
maddenp-noaa, line 30: The current parsing technique should allow -- and ignore -- other valid shell code.

src/uwtools/tests/config/formats/test_sh.py
```diff
@@ -41,20 +40,17 @@ def test_parse_include():
...
     infile = fixture_path("simple.sh")
-    outfile = tmp_path / "outfile.sh"
     cfgobj = SHConfig(infile)
     expected: Dict[str, Any] = {
         **salad_base["salad"],
         "how_many": "12",
     }
     assert cfgobj == expected
-    cfgobj.dump(outfile)
-    assert filecmp.cmp(infile, outfile)
```
maddenp-noaa, line 57: An output file would no longer be identical to the input, because the input contains a comment and an executable statement that are ignored during parsing. See fixture file, below.

src/uwtools/tests/fixtures/include_files.sh
```diff
@@ -1,3 +1,3 @@
-salad_include=!INCLUDE [./fruit_config.sh]
+salad_include="!INCLUDE [./fruit_config.sh]"
```
maddenp-noaa, line 1: The `!INCLUDE` value must be quoted so that it is valid shell code. See `include_files.nml` for a parallel Fortran-namelist case.

src/uwtools/tests/fixtures/simple.sh
```diff
@@ -3,3 +3,5 @@ fruit=banana
 vegetable=tomato
 how_many=12
 dressing=balsamic
+# dressing=poison -- ensure that shlex ignores comments
+echo this will be ignored, too
```
maddenp-noaa, line 7: If `shlex` is not instructed to recognize comments, it will parse `#` as one token and `dressing=poison` as another token, and the latter will then be recognized as a key=value pair and added to the `SHConfig` object's dictionary. This comment is a tripwire to ensure that `shlex` comment recognition in the code isn't accidentally disabled.

Also note that `shlex` + the new `SHConfig` code is happy to ignore code that isn't key=value pairs. It's probably not good practice to parse real shell _scripts_ with `SHConfig`, but this demonstrates that at least some simple cases wouldn't be a problem.

NaureenBharwaniNOAA (approved):
Thanks for the comments it helped understand your exact thought process. Looks very clean to me.

elcarpenterNOAA (approved):
Short and sweet. Looks good.

src/uwtools/tests/fixtures/simple.sh
```diff
@@ -3,3 +3,5 @@ fruit=banana
 vegetable=tomato
 how_many=12
 dressing=balsamic
+# dressing=poison -- ensure that shlex ignores comments
```
christinaholtNOAA, line 6: Dark.

christinaholtNOAA (approved):
LGTM!

src/uwtools/tests/fixtures/simple.sh
maddenp-noaa: You have been warned :)

PR #373, opened 2024-01-04
UW-432

WeirAE:
**Description**

Add documentation for CLI mode config page. 
Preview docs are [here](https://unified-workflow--373.org.readthedocs.build/en/373/).

**Type**

- [x] Documentation

**Impact**

**Checklist**

Affirm:

### DIFF
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -230,15 +230,15 @@ and supplemental YAML file ``config.yaml`` with content
 
   Shell redirection via ``|``, ``>``, et al may also be used to stream output to a file, another process, etc.
 
-* Existing values can be overwritten and Jinja2 variables can be filled in by appending supplemental files to the end of the argument:
+* Values in the primary input file can be overridden via one or more supplemental files specified as positional arguments, each overriding the last; or by environment variables, which have the highest precedence.
 
   .. code:: sh
 
-    $ uw config realize --input-file values.yaml --output-format yaml config.yaml
+    $ recipient=Sun uw config realize --input-file values.yaml --output-format yaml config.yaml 
     values:
       greeting: Good Night
       recipient: Moon
-      message: Good Night MoonGood Night Moon
+      message: Good Night SunGood Night Sun
       date: 20240105
       repeat: 2
       empty: null
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -62,6 +62,7 @@ Examples
 The examples that follow use YAML file ``values.yaml`` with content
 
 .. code:: sh
+
   values:
     greeting: Hello
     recipient: World
@@ -69,6 +70,7 @@ The examples that follow use YAML file ``values.yaml`` with content
 and namelist file ``values.nml`` with content
 
 .. code:: sh
+
   &values
     greeting = "Hello"
     recipient = "World"
@@ -191,6 +193,7 @@ Examples
 The examples that follow use YAML file ``values.yaml`` with content
 
 .. code:: sh
+
   values:
     greeting: Hello
     recipient: World
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -187,7 +187,7 @@ The examples that follow use YAML file ``values.yaml`` with content
     repeat: 1
     empty: 
 
-and supplemental YAML file ``config.yaml`` with content
+supplemental YAML file ``config.yaml`` with content
 
 .. code:: sh
 
@@ -196,6 +196,14 @@ and supplemental YAML file ``config.yaml`` with content
     recipient: Moon
     date: 20240105
     repeat: 2
+
+And an additional supplemental YAML file ``supp.yaml`` with content
+
+.. code:: sh
+
+  values:
+    repeat: 5
+    empty: false
 
 * Show the values in the input config file that have unfilled Jinja2 templates or empty keys:
 
@@ -234,14 +242,14 @@ and supplemental YAML file ``config.yaml`` with content
 
   .. code:: sh
 
-    $ recipient=Sun uw config realize --input-file values.yaml --output-format yaml config.yaml 
+    $ recipient=Sun uw config realize --input-file values.yaml --output-format yaml supp.yaml config.yaml 
     values:
       greeting: Good Night
       recipient: Moon
       message: Good Night SunGood Night Sun
       date: 20240105
       repeat: 2
-      empty: null
+      empty: false
 
 * Realize the config to a file via command-line argument:
 
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -182,32 +182,38 @@ The examples that follow use YAML file ``values.yaml`` with content
   values:
     greeting: Hello
     recipient: World
+    message: '{{ (greeting + " " + recipient) * repeat }}'
     date: '{{ yyyymmdd }}'
-    empty:
+    repeat: 1
+    empty: 
 
-and supplemental YAML file ``supp.yaml`` with content
+and supplemental YAML file ``config.yaml`` with content
 
 .. code:: sh
 
   values:
     greeting: Good Night
     recipient: Moon
+    date: 
+    repeat: 2
 
 * Show the values in the input config file that have unfilled Jinja2 templates or empty keys:
 
   .. code:: sh
 
     $ uw config realize --input-file values.yaml --output-format yaml --values-needed
-    [2024-01-03T15:35:29]     INFO Keys that are complete:
-    [2024-01-03T15:35:29]     INFO     values
-    [2024-01-03T15:35:29]     INFO     values.greeting
-    [2024-01-03T15:35:29]     INFO     values.recipient
-    [2024-01-03T15:35:29]     INFO 
-    [2024-01-03T15:35:29]     INFO Keys that have unfilled Jinja2 templates:
-    [2024-01-03T15:35:29]     INFO     values.date: {{ yyyymmdd }}
-    [2024-01-03T15:35:29]     INFO 
-    [2024-01-03T15:35:29]     INFO Keys that are set to empty:
-    [2024-01-03T15:35:29]     INFO     values.empty
+    [2024-01-05T11:34:22]     INFO Keys that are complete:
+    [2024-01-05T11:34:22]     INFO     values
+    [2024-01-05T11:34:22]     INFO     values.greeting
+    [2024-01-05T11:34:22]     INFO     values.recipient
+    [2024-01-05T11:34:22]     INFO     values.message
+    [2024-01-05T11:34:22]     INFO     values.repeat
+    [2024-01-05T11:34:22]     INFO 
+    [2024-01-05T11:34:22]     INFO Keys that have unfilled Jinja2 templates:
+    [2024-01-05T11:34:22]     INFO     values.date: {{ yyyymmdd }}
+    [2024-01-05T11:34:22]     INFO 
+    [2024-01-05T11:34:22]     INFO Keys that are set to empty:
+    [2024-01-05T11:34:22]     INFO     values.empty
 
 * To realize the config to ``stdout``, a target output format must be explicitly specified:
 
@@ -217,7 +223,9 @@ and supplemental YAML file ``supp.yaml`` with content
     values:
       greeting: Hello
       recipient: World
+      message: Hello World
       date: '{{ yyyymmdd }}'
+      repeat: 1
       empty: null
 
   Shell redirection via ``|``, ``>``, et al may also be used to stream output to a file, another process, etc.
@@ -226,38 +234,45 @@ and supplemental YAML file ``supp.yaml`` with content
 
   .. code:: sh
 
-    $ uw config realize --input-file values.yaml --output-format yaml supp.yaml
+    $ uw config realize --input-file values.yaml --output-format yaml config.yaml
     values:
       greeting: Good Night
       recipient: Moon
-      date: '{{ yyyymmdd }}'
+      message: Good Night MoonGood Night Moon
+      date: 20240105
+      repeat: 2
       empty: null
 
 * Realize the config to a file via command-line argument:
 
   .. code:: sh
 
-    $ uw config realize --input-file values.yaml --output-file realized.yaml
+    $ uw config realize --input-file values.yaml --output-file realized.yaml config.yaml
 
   The content of ``realized.yaml``:
 
   .. code:: sh
 
-      values:
-        greeting: Hello
-        recipient: World
-        date: '{{ yyyymmdd }}'
-        empty: null
+    values:
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night MoonGood Night Moon
+      date: 20240105
+      repeat: 2
+      empty: null
 
 * With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
 
   .. code:: sh
 
-    $ uw config realize --input-file values.yaml --output-file realized.yaml --dry-run
-    [2024-01-03T15:39:23]     INFO values:
-      greeting: Hello
-      recipient: World
-      date: '{{ yyyymmdd }}'
+    $ uw config realize --input-file values.yaml --output-file realized.yaml --dry-run config.yaml
+    [2024-01-05T11:35:20]     INFO values:
+    values:
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night MoonGood Night Moon
+      date: 20240105
+      repeat: 2
       empty: null
 
 
@@ -265,18 +280,20 @@ and supplemental YAML file ``supp.yaml`` with content
 
   .. code:: sh
 
-    $ cat values.yaml | uw config realize --output-file realized.yaml
+    $ cat values.yaml | uw config realize --output-file realized.yaml config.yaml
     Specify --input-format when --input-file is not specified
 
 * Read the config from ``stdin`` and realize to ``stdout``:
 
   .. code:: sh
 
-    $ cat values.yaml | uw config realize --input-format yaml --output-format yaml
+    $ cat values.yaml | uw config realize --input-format yaml --output-format yaml config.yaml
     values:
-      greeting: Hello
-      recipient: World
-      date: '{{ yyyymmdd }}'
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night MoonGood Night Moon
+      date: 20240105
+      repeat: 2
       empty: null
 
 
@@ -284,18 +301,20 @@ and supplemental YAML file ``supp.yaml`` with content
 
   .. code:: sh
 
-    $ uw config realize --input-file values.txt --output-format yaml
+    $ uw config realize --input-file values.txt --output-format yaml config.yaml
     Cannot deduce format of 'values.txt' from unknown extension 'txt'
 
   In this case, the format can be explicitly specified:
 
   .. code:: sh
 
-    $ uw config realize --input-file values.txt --input-format yaml --output-format yaml
+    $ uw config realize --input-file values.txt --input-format yaml --output-format yaml config.yaml
     values:
-      greeting: Hello
-      recipient: World
-      date: '{{ yyyymmdd }}'
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night MoonGood Night Moon
+      date: 20240105
+      repeat: 2
       empty: null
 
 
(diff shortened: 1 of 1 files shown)


docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -138,7 +367,54 @@ TBD
 Examples
 ~~~~~~~~
 
-TBD
+The examples that follow use atparse-formatted template file ``atparse.txt`` with content
+
+.. code:: sh
+  @[greeting], @[recipient]!
```
maddenp-noaa, line 373: For some reason, this block is not showing up for me in the preview HTML docs.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -138,7 +367,54 @@ TBD
...
+    $ uw config translate --input-file atparse.txt --input-format atparse --output-format jinja2
+    {{greeting}}, {{recipient}}!
+
+  Shell redirection via ``|``, ``>``, et al may also be used to stream output to a file, another process, etc.
+
+* Convert the template to a file via command-line argument:
+
+  .. code:: sh
+
+    $ uw config realize --input-file values.yaml --output-file realized.yaml
```
maddenp-noaa, line 388: This is a `realize`, not `translate`, command.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -138,7 +367,54 @@ TBD
...
+
+* With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
+
+  .. code:: sh
+
+    $ uw config translate --input-file atparse.txt --input-format atparse --output-format jinja2 --dry-run
+    [2024-01-03T16:41:13]     INFO {{greeting}}, {{recipient}}!
+
+
+* If an input file is read alone from ``stdin``, ``uw`` know how to parse its content as we must always specify the formats:
```
maddenp-noaa, line 404: `know how` -> `will not know how` ? Maybe
> If the input is read from `stdin`, `uw` will not know how to parse it, so the format must be specified:

? Also, it's not necessarily a "file" that's being read -- it could be the output of another process. Though it obviously is a file in this example, it's easy enough to not mention it.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -138,7 +367,54 @@ TBD
...
+    $ cat atparse.txt | uw config translate --input-format atparse --output-format jinja2
+    {{greeting}}, {{recipient}}!
+
+
+* Request verbose log output:
+
+  .. code:: sh
+
+    $ uw config translate --input-file atparse.txt --input-format atparse --output-format jinja2 --verbose
+    {{greeting}}, {{recipient}}!
```
maddenp-noaa, line 417: Do we really not get any log lines with `--verbose`? If not, IMO we might as well remove this bullet, as it doesn't really demonstrate anything.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +445,111 @@ TBD
 Examples
 ~~~~~~~~
 
-TBD
+The examples that follow use JSON Schema file ``schema.jsonschema`` with content
+
+.. code:: sh
```
maddenp-noaa, line 450: This block is also not rendering to HTML as far as I can tell.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +445,111 @@ TBD
...
+
+  .. code:: sh
+
+    $ uw config validate --schema-file schema.jsonschema --input-file values.yaml
+    [2024-01-03T17:23:07]     INFO 0 schema-validation errors found
+
+  Shell redirection via ``|``, ``>``, et al may also be used to stream output to a file, another process, etc.
+
+
+* Read the config from ``stdin`` and render to ``stdout``:
```
maddenp-noaa, line 491: Here and below, there's no "render"ing happening -- this is validation.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +445,111 @@ TBD
...
+
+  .. code:: sh
+
+    $ cat values.yaml | uw config validate --schema-file schema.jsonschema
+    [2024-01-03T17:26:29]     INFO 0 schema-validation errors found
+
+
+* However, you cannot read the schema from ``stdin`` and render to ``stdout``:
```
maddenp-noaa, line 499: As a general rule for documentation -- or at least as a style guideline for `uwtools` -- I'd recommend avoiding pronouns like "we", "you", etc. It's more straightforward to just describe the software and what it does. There may be more room for pronouns in the breezier style of an intro page, but in the drier technical docs, I personally find it distracting. This is just my opinion, though, so I'd like to hear what others think.

For example, this could read
> However, reading the schema from `stdin` is not supported:

(I left out the "render" part, as this is "validate" mode.)

There's another "you" in the `realize` section that could go if we agree on this style point.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +445,111 @@ TBD
...
+
+  .. code:: sh
+
+    $ cat schema.jsonschema | uw config validate -input-file values.yaml
+    uw config validate: error: the following arguments are required: --schema-file
+
+* If there are differences between the config files, they will be shown with the schema. For example, with ``recipient: World`` removed from ``values.yaml``:
```
maddenp-noaa, line 506: Maybe this needs an update after copy/paste? This isn't about "differences between the config files", it's about a config failing validation.

maddenp-noaa (changes requested):

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -59,7 +59,94 @@ The ``uw`` mode for handling configs.
...
+    [2024-01-03T16:25:47]     INFO ---------------------------------------------------------------------
+    [2024-01-03T16:25:47]     INFO values:       recipient:  - None + World
+
+  Note that ``uw`` logs to ``stderr`` and writes non-log output to ``stdout``, so the streams can be redirected separately:
+
+  .. code:: sh
+
+    $ uw config compare --file-1-path values.yaml --file-2-path values.nml --verbose 2>compare.log
+
+  The content of ``realized.log``:
```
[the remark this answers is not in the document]
WeirAE: Ah, the faults of templating - thanks, got it!

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +190,149 @@ TBD
...
+
+    $ uw config realize --input-file values.yaml --output-format yaml --values-needed
+    [2024-01-03T15:35:29]     INFO Keys that are complete:
+    [2024-01-03T15:35:29]     INFO     values
+    [2024-01-03T15:35:29]     INFO     values.greeting
+    [2024-01-03T15:35:29]     INFO     values.recipient
+    [2024-01-03T15:35:29]     INFO 
+    [2024-01-03T15:35:29]     INFO Keys that have unfilled Jinja2 templates:
+    [2024-01-03T15:35:29]     INFO 
+    [2024-01-03T15:35:29]     INFO Keys that are set to empty:
```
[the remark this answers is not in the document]
WeirAE: Agreed - as below, there are definitely some changes we can add

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Ah, I thought I caught all the missing `.. code::` line breaks, thanks!

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Ah, missed that, I think I've cleaned it up

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Yeah, or with validate. Both should probably either add more info or remove quiet/verbose. The `--quiet` isn't mentioned here or there because its even more useless

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Bad conversational habit drifting in, thanks for catching that. I've fixed both entries I think

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +432,112 @@ TBD
...
+
+* Read the config from ``stdin`` and print validation results to ``stdout``:
+
+  .. code:: sh
+
+    $ cat values.yaml | uw config validate --schema-file schema.jsonschema
+    [2024-01-03T17:26:29]     INFO 0 schema-validation errors found
+
+
+* However, reading the schema from ``stdin`` is not supported:
```
christinaholtNOAA, line 487:
```suggestion
* However, reading the schema from ``stdin`` is **not** supported:
```

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +432,112 @@ TBD
...
+    [2024-01-03T17:29:46]    DEBUG Command: uw config validate --schema-file schema.jsonschema --input-file values.yaml --verbose
+    [2024-01-03T17:29:46]    DEBUG Dereferencing, initial value: {'values': {'greeting': 'Hello', 'recipient': 'World'}}
+    [2024-01-03T17:29:46]    DEBUG Rendering: {'values': {'greeting': 'Hello', 'recipient': 'World'}}
+    [2024-01-03T17:29:46]    DEBUG Rendering: {'greeting': 'Hello', 'recipient': 'World'}
+    [2024-01-03T17:29:46]    DEBUG Rendering: Hello
+    [2024-01-03T17:29:46]    DEBUG Rendering: World
+    [2024-01-03T17:29:46]    DEBUG Dereferencing, final value: {'values': {'greeting': 'Hello', 'recipient': 'World'}}
+    [2024-01-03T17:29:46]     INFO 0 schema-validation errors found
+
+  Note that ``uw`` logs to ``stderr`` and writes non-log output to ``stdout``, so the streams can be redirected separately:
```
christinaholtNOAA, line 526: This may not be relevant since `uw config validate` doesn't really print anything to `stdout`.

christinaholtNOAA (changes requested):
This is quite thorough! Nice.

docs/sections/user_guide/cli/mode_config.rst
christinaholtNOAA: Hmmmm. I typically write in a similar conversational style. In a few places in my work-in-progress, I've made a change to remove the pronouns, but what about:

> We recommend a descriptive one to make it easier to read.

Here, I feel like it makes sense to keep it a bit more personal. Thoughts?

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +190,149 @@ TBD
...
 ~~~~~~~~
 
-TBD
+The examples that follow use YAML file ``values.yaml`` with content
+
+.. code:: sh
+
+  values:
+    greeting: Hello
+    recipient: World
```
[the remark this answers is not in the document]
WeirAE: I really thought about this but Jinja templating is involved enough that it should have its own documentation. I could probably modify it to present another error type or 2 if that would help

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -138,7 +367,54 @@ TBD
...
+The examples that follow use atparse-formatted template file ``atparse.txt`` with content
+
+.. code:: sh
+  @[greeting], @[recipient]!
+
+* Convert an atparse-formatted template file to Jinja2 format:
+
+  .. code:: sh
+
+    $ uw config translate --input-file atparse.txt --input-format atparse --output-format jinja2
```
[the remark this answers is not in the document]
WeirAE: I was pondering that a bit - the cleaner UX is to default it and check at runtime, though requiring it could have an argument for reinforcing user awareness

docs/sections/user_guide/cli/mode_config.rst
maddenp-noaa: I don't think "we recommend" adds anything over
> A descriptive [thing] is easier to read.

Good advice doesn't need to be softened. We're all adults here. That "we" recommend an idea isn't what makes it good. The facts should be persuasive on their own.

Setting aside that I don't know the context to be more specific than "thing", in general that sentence sounds weak to me -- "one" and "it" are vague, even if the context clarifies what they refer to. For example, instead of writing (I'm just making this up)
> You can define a variable here to frobnicate the foo. We recommend a descriptive one to make it easier to read.

could be

> Define a variable to frobnicate the foo. A descriptive variable name will clarify its intent to readers.

A strong statement followed by clear rationale. Or even just

> Define a descriptively named variable to frobnicate the foo.

Do we need to explain why descriptive variables are desirable? Attention is a scarce resource and, when I'm the reader, I appreciate documentation that gets to the point.

That said, I think "We" can still be appropriate if substituting the "The UW Team" or "The UFS Community" or whatever is really what we mean and the context is clear. That's different, though, from e.g. "We see no output when we use the `--quiet` flag" (who is "we"?), which can be stated more strongly/directly as "The `--quiet` flag suppresses all output".

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
maddenp-noaa: Will `uw config translate` ever do anything but atparse -> Jinja2 translation? If so, I think we should keep things as they are so that we won't have to break backward-compatibility when we add another capability. It seems possible that some other kind of translation might be desired in the future, though [UW-410](https://jira.epic.oarcloud.noaa.gov/browse/UW-419) concerns might limit what we can do.

If we'll _never_ do anything but atparse -> Jinja2, maybe the submode has too general a name. I can't think of anything I like at the moment. `uw config atparse-to-jinja2` is kind of awful, if clear.

docs/sections/user_guide/cli/mode_config.rst
maddenp-noaa: Also looks good now, thanks again.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -59,7 +59,94 @@ The ``uw`` mode for handling configs.
 Examples
 ~~~~~~~~
 
-TBD
+The examples that follow use YAML file ``values.yaml`` with content
```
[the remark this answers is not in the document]
WeirAE: Good point - got it!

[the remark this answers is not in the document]
maddenp-noaa: While this specific case seems harmless, I wonder whether ([UW-419](https://jira.epic.oarcloud.noaa.gov/browse/UW-419) again, <sigh>) we even want to support this long-term. For example, how should we compare `config.yaml`
```
values:
  foo: 88
```
to `config.ini`
```
[values]
foo: 88
```
? In the YAML file, `foo` has an integer value; in the INI file, it's... a string? It has to be a string because **all** INI values are string, but is it _meant_ to represent an integer? Is it the same as or different from `foo` in the YAML file?

It might be best to avoid the issue for now and just do e.g. a YAML-to-YAML comparison (I wouldn't mind seeing a little more depth in the YAML tree, perhaps seeing how an entire missing _subtree_ is reported).

@christinaholtNOAA would a NML-to-NML comparison be useful?

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +445,111 @@ TBD
...
+  values:
+    greeting: Hello
+    recipient: World
+
+* Validate a YAML config against a given JSON schema:
+
+  .. code:: sh
+
+    $ uw config validate --schema-file schema.jsonschema --input-file values.yaml
+    [2024-01-03T17:23:07]     INFO 0 schema-validation errors found
```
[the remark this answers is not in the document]
WeirAE: Yep, thanks!

docs/sections/user_guide/cli/mode_config.rst
maddenp-noaa, line 429: This code block might render better as `json` instead of `sh`. (I've got a branch that I'll PR after this one merges that updates the language-highlighting hints in various places, so I'll get this one if you don't.)

docs/sections/user_guide/cli/mode_config.rst
christinaholtNOAA: Point taken...with a few laughs thrown in for good measure. "Frobnicate the foo"?!? 😆

maddenp-noaa (approved):
Nice work. Thanks for following the style of other pages.

docs/sections/user_guide/cli/mode_config.rst
maddenp-noaa: https://en.wiktionary.org/wiki/frobnicate

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
christinaholtNOAA: I think `realizing` a config is exactly where it should be described. That's where we fill in those jinja templates and then overwrite values provided by supplemental files. Both of those actions should be documented here in a way that shows what is actually happening. Otherwise `realizing` is only copying one file format to another, which is something we're discussing taking away.

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
christinaholtNOAA: I think it could be 6 of one....leaving it as-is would be fine. Just felt sort of verbose since that's the only thing it does and we don't have plans to expand it.

docs/sections/user_guide/cli/mode_config.rst
christinaholtNOAA: TIL.

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Indeed it does, thanks!

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -59,7 +59,94 @@ The ``uw`` mode for handling configs.
...
+  .. code:: sh
+
+    $ uw config compare --file-1-path values.yaml --file-2-path values.nml --verbose
+    [2024-01-03T16:25:47]    DEBUG Command: uw config compare --file-1-path values.yaml --file-2-path values.nml --verbose
+    [2024-01-03T16:25:47]     INFO - values.yaml
+    [2024-01-03T16:25:47]     INFO + values.nml
+    [2024-01-03T16:25:47]     INFO ---------------------------------------------------------------------
+    [2024-01-03T16:25:47]     INFO values:       recipient:  - None + World
+
+  Note that ``uw`` logs to ``stderr`` and writes non-log output to ``stdout``, so the streams can be redirected separately:
```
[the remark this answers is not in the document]
WeirAE: Thanks - I modified this to only describe `stderr` redirect, I don't think there's any value in `stdout` here

WeirAE:
Additionally, please see #375 for a quick fix that came up and I almost forgot to add!

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: Gosh I hope not - I think `translate` is sufficient, but I think we're in agreement that locking out the formats is a good refactor?

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +190,149 @@ TBD
...
+        recipient: World
+
+* With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
+
+  .. code:: sh
+
+    $ uw config realize --input-file values.yaml --output-file realized.yaml --dry-run
+    [2024-01-03T15:39:23]     INFO values:
+      greeting: Hello
+      recipient: World
```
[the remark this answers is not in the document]
WeirAE: Thanks - agreed, it could certainly use more verbosity

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: Ah thanks I misunderstood, I was thinking about `translate` for some reason - working on it!

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
maddenp-noaa: I'd vote for opening a Jira ticket describing the proposal, and we can discuss / define acceptance criteria / pull into a sprint during some near-future sprint planning.

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: Hm, this version is still kind of simplistic but a little better - let me know if you think its still too short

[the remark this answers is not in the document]
christinaholtNOAA: I'd like to see that some of those Jinja2 templates are filled in for a completely realized output. It probably makes things too SRW-specific to use this example, but check out Slide 17 in [this slide deck](https://docs.google.com/presentation/d/1GziWguL2ivZo-VlgPcZi13bxcNvJ1fwJZKMwGixYrQQ/edit#slide=id.g1a118e06cb4_0_18).

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +469,112 @@ TBD
 Examples
 ~~~~~~~~
 
-TBD
+The examples that follow use JSON Schema file ``schema.jsonschema`` with content
```
christinaholtNOAA, line 472: It would probably be good to put a link to the schema docs here.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +469,112 @@ TBD
...
+
+    $ cat schema.jsonschema | uw config validate --input-file values.yaml
+    uw config validate: error: the following arguments are required: --schema-file
+
+* If a config fails validation, differences from the schema will be displayed. For example, with ``recipient: World`` removed from ``values.yaml``:
+
+  .. code:: sh
+
+    $ uw config validate --schema-file schema.jsonschema --input-file values.yaml
+    [2024-01-03T17:31:19]    ERROR 1 schema-validation error found
```
christinaholtNOAA, line 536: Does this also change to `UW schema-validation...`? How about the rest of the ones below?

christinaholtNOAA (changes requested):

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: Sure, I can fill in some of the examples in a less-SRW-y way. You're right that it really should show full rendering

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Oops, missed that one - I'll fix it and recheck the full set

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: Opened as UW-472

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -169,4 +469,112 @@ TBD
 Examples
 ~~~~~~~~
 
-TBD
+The examples that follow use `JSON Schema <https://json-schema.org/understanding-json-schema/reference>`_ file ``schema.jsonschema`` with content
```
maddenp-noaa, line 472: We have an external link defined for JSON schema already, and can re-use it here either as a link to the main site as
```
:json-schema:`JSON Schema<>`
```
or to the specific page you're linking to here as
```
:json-schema:`JSON Schema<understanding-json-schema/reference>`
```

WeirAE: Thanks, I had a feeling I was not doing that right

docs/sections/user_guide/cli/mode_config.rst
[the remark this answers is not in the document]
WeirAE: What do you think of this:
`values.yaml`
```
values:
  greeting: Hello
  recipient: World
  message: '{{ (GREETING + RECIPIENT) * REPEAT }}'
  repeat: 1
```
`config.yaml`
```
values:
  greeting = Good Night
  recipient = Moon
  repeat: 2
```
However, its not rendering when I run `realize`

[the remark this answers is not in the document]
WeirAE: Ah, case-sensitivity fixes it - will add shortly!

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +175,253 @@ TBD
...
+      greeting: Hello
+      recipient: World
+      message: Hello World
+      date: '{{ yyyymmdd }}'
+      repeat: 1
+      empty: null
+
+  Shell redirection via ``|``, ``>``, et al may also be used to stream output to a file, another process, etc.
+
+* Existing values can be overwritten and Jinja2 variables can be filled in by appending supplemental files to the end of the argument:
```
maddenp-noaa, line 233:
```suggestion
* Values in the primary input file can be overridden via one or more supplemental files specified as positional arguments, each overriding the last; or by environment variables, which have the highest precedence.
```
Then maybe show something like
```
% recipient=Sun uw config realize --input-file values.yaml --output-format yaml config.yaml 
values:
  greeting: Good Night
  recipient: Moon
  message: Good Night SunGood Night Sun
  repeat: 2
```
That introduces the use of environment variables.

WeirAE: Ah, good point! Got it added, thank you!

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +175,253 @@ TBD
...
+* Values in the primary input file can be overridden via one or more supplemental files specified as positional arguments, each overriding the last; or by environment variables, which have the highest precedence.
+
+  .. code:: sh
+
+    $ recipient=Sun uw config realize --input-file values.yaml --output-format yaml config.yaml 
+    values:
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night SunGood Night Sun
+      date: 20240105
```
christinaholtNOAA, line 242: `date` was empty in the example `config.yaml` above, and not provided as an env var here.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +175,253 @@ TBD
 Examples
 ~~~~~~~~
 
-TBD
```
christinaholtNOAA, line 106: There seems to be a missing example of using multiple supplemental files.

christinaholtNOAA (changes requested):

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Ah thanks - this is a typo at line 197, as the date is updated by `config.yaml` elsewhere

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -103,7 +175,253 @@ TBD
...
+* With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
+
+  .. code:: sh
+
+    $ uw config realize --input-file values.yaml --output-file realized.yaml --dry-run config.yaml
+    [2024-01-05T11:35:20]     INFO values:
+    values:
+      greeting: Good Night
+      recipient: Moon
+      message: Good Night MoonGood Night Moon
```
maddenp-noaa, line 273: nitpick: I think the run-on text without a space where one would expect one between the two copies of the greeting makes this seem like a somewhat tortured example. Like, who would want that? I just tried and wasn't readily able to get a space in there -- I tried piping into `| join(" ")` but that just put a space between each character -- probably a list of two strings is needed here but I don't know how to get one and Jinja2 isn't making it easy.

I **do** like that this examples shows the use of Jinja2 beyond the simplest variable interpolation, and don't want to use that, or else I'd suggest removing the `repeat`. But since this `Good Night MoonGood Night Moon` is now repeated so many times on this page, I feel like we should try for a strong Jinja2 example that doesn't produce output that looks like a bug. I don't immediately have a suggestion, unfortunately.

docs/sections/user_guide/cli/mode_config.rst
WeirAE: OK, I think I've got it with this version, including emphasizing the precedence of overriding - thanks for keeping on me to improve the example!

docs/sections/user_guide/cli/mode_config.rst
WeirAE: I could make the `message` formula `'{{ (greeting + " " + recipient + " ") * repeat` just to add an extra space

maddenp-noaa (approved):

christinaholtNOAA (approved):

PR #378, opened 2024-01-08
Fixes to forecast driver found while testing.

christinaholtNOAA:
**Description**

I found just a couple of issues when testing the forecast driver that needed to be addressed.

**Type**

Select one or more:

**Impact**

Select one:

**Checklist**

Affirm:


src/uwtools/config/support.py
```diff
@@ -15,7 +15,7 @@ def depth(d: dict) -> int:
     :param d: The dictionary whose depth to calculate.
     :return: The length of the longest path to a value in the dictionary.
     """
-    return (max(map(depth, d.values())) + 1) if isinstance(d, dict) else 0
+    return (max(map(depth, d.values())) + 1) if d and isinstance(d, dict) else 0
```
maddenp-noaa, line 18: Out of curiosity, what does this fix? An empty `dict` that we don't want to descend into?

maddenp-cu (approved):

src/uwtools/config/support.py
christinaholtNOAA: It looks like the `(max(map(depth, d.values()))` is returning an empty sequence. I haven't dug into exactly what situation in my YAML is causing this. I can if you think it's worth it.

src/uwtools/drivers/forecast.py
```diff
@@ -282,12 +281,13 @@ def _run_via_local_execution(self) -> Tuple[bool, List[str]]:
...
             strings that make up the full command line.
         """
+        run_directory = self.prepare_directories()
         pre_run = self._mpi_env_variables(" ")
         full_cmd = f"{pre_run} {self.run_cmd()}"
         command_lines = ["Command:", *full_cmd.split("\n")]
         if self._dry_run:
             return True, command_lines
-        result = execute(cmd=full_cmd)
+        result = execute(cmd=full_cmd, cwd=run_directory, log_output=True)
```
WeirAE, line 290: Ah good fix - I would have thought it would work implicitly but there's no replacement for direct testing!

src/uwtools/config/support.py
WeirAE: As long as `d` is valid before this, is it running in to an empty dict at the end of iterating through?

maddenp-noaa: Now I'm really curious. That expression should have an `int` value -- not a sequence/collection. I'm not sure why we have the parens around it, though. It would be great to know what is triggering bad behavior. A simple analog seems to do the right thing:
```
>>> def foo(b):
...   return (max([1, 3, 2]) + 1) if b else 0
... 
>>> foo(True)
4
>>> foo(False)
0
```
But it also seems fine without the parens:
```
>>> def bar(b):
...   return max([1, 3, 2]) + 1 if b else 0
... 
>>> bar(True)
4
>>> bar(False)
0
```

christinaholtNOAA: If d is empty, then d.values() is an empty sequence. Now that Hera is back, I'm trying to dig a bit deeper into what exactly in my config is causing this.

nbharwani-cu (approved):

src/uwtools/config/support.py
christinaholtNOAA: So, I finally found the culprit this morning...a SRW-prepared namelist has this in it, leading to the empty dict issue:

```
&nam_sfcperts
/

&nam_sppperts
/

&nam_stochy
/
```

maddenp-noaa: I get it now, it's essentially
```
>>> max(map(depth, {}.values()))
ValueError: max() arg is an empty sequence
```
Another way to solve it is to use the `default` keyword on `max()` to specify a value when the sequence is empty:
```
>>> max(map(depth, {}.values()), default=0)
0
```
I'm not sure if this would have made the potential problem clearer to me. Maybe I would have seen that the problem isn't recursion into an empty `dict` per se, but that `max` might be asked to operate on an empty iterable. I guess I'm always nervous about `if d` due to the many falsey values `d` could have that would make that expression `False`; when I see `if d` I have to think about the intent. Empty collection is one, but then I wonder about the others... I think using `default=0` in the call to `max()` avoids any nuance. FWIW, please do what you think best.

maddenp-noaa:
We should probably add a unit test for this case, which fails if you temporarily hide the fix, and then passes when you expose it.

src/uwtools/config/support.py
christinaholtNOAA: I don't mind making a chance for readability/understanding. :)

christinaholtNOAA: @maddenp-noaa Sorry it's taken a while to get back to it here. I saw you make this change in PR #392, so I'll merge develop in here to pick it up and re-test.

christinaholtNOAA:
I will work on the unit test for the fix.

maddenp-noaa:
I've got a unit test for this in PR #392 -- no need to add one here, you'll get it after I merge and then you merge in `develop`.

recipe/run_test.sh
```diff
@@ -40,7 +40,7 @@ typecheck() {
   msg Running typechecker
   (
     set -eux
-    mypy --install-types --non-interactive .
+    mypy --install-types --non-interactive --warn-return-any .
```
maddenp-noaa, line 43: Probably don't want to commit this.

src/uwtools/cli.py
```diff
@@ -633,6 +633,7 @@ def _abort(msg: str) -> None:
     :param msg: The message to print.
     """
     print(msg, file=sys.stderr)
+    log.exception(msg)
```
maddenp-noaa, line 636: @WeirAE Is addressing this elsewhere, right? I think you probably just committed some debug code...

christinaholtNOAA: Ugh. I committed and pushed some things I did not mean to!

PR #388, opened 2024-01-18
tagged-scalars

maddenp-noaa:
**Synopsis**

This PR adds support for YAML tags `!float` and `!int` to support conversion of `str` values arising from Jinja2 rendering to desired Python types. For example, consider the YAML fragment
```
n: "{{ 2 + 2 }}"
```
The quotes around the Jinja2 expression `{{ 2 + 2 }}` are required to allow YAML to parse this fragment, as otherwise it would interpret the opening `{` as the start of a YAML mapping, leading to a syntax error and a failed parse. With the the quotes in place, the value of `n` will be of type `str` and, after Jinja2 rendering, the Python `dict` fragment corresponding to this input will be
```
n: "4"
```
If `n` were subsequently used to generate a Fortran namelist value where an `integer` value is expected for `n`, the namelist would instead contain e.g.
```
n = '4'
```
and the Fortran code reading the namelist would fail. So it is important that value types in Python config `dict` objects are correct. This PR adds support for the following syntax:
```
n: !int "{{ 2 + 2 }}"
```
leading to the (conceptual) transformations `n: !int "{{ 2 + 2 }}"` (original) -> `n: !int "4"` (Jinja2 rendering) -> `n: 4` (conversion via the `!int` tag). The tag `!float` provides similar conversion to `float` values.

Currently, the above example would already work because a reification step re-processes values with the YAML parser, leading to the (conceptual) transformations `n: "{{ 2 + 2 }}"` (original) -> `n: "4"` (Jinja2 rendering) -> `n: 4` (reification via `yaml.load()`). However, this creates a problem, demonstrated by the example
```
n: "48:00:00"
```
Without quotes, YAML parses the string `48:00:00` as an `int`:
```
>>> yaml.safe_load('48:00:00')
172800
```
Quoting avoids this:
```
>>> yaml.safe_load('"48:00:00"')
'48:00:00'
```
But the current reification step parses this value a _second_ time, converting it again to an int. So, reification via YAML parsing has undesirable side effects and is eliminated in this PR.

This work is a prelude to [UW-419](https://jira.epic.oarcloud.noaa.gov/browse/UW-419). Documentation will be added/updated in a separate PR as a prerequisite to completion of that ticket.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/base.py
+++ b/src/uwtools/config/formats/base.py
@@ -163,15 +163,14 @@ class Config(ABC, UserDict):
         """
         Render as much Jinja2 syntax as possible.
         """
-        log.debug("Dereferencing, initial value: %s", self.data)
         while True:
+            log.debug("Dereferencing, current value: %s", str(self))
             new = dereference(val=self.data, context={**os.environ, **self.data})
+            assert isinstance(new, dict)
             if new == self.data:
                 break
-            log.debug("Dereferencing, current value: %s", self.data)
-            assert isinstance(new, dict)
             self.data = new
-        log.debug("Dereferencing, final value: %s", self.data)
+        log.debug("Dereferencing, final value: %s", str(self))
 
     @abstractmethod
     def dump(self, path: OptionalPath) -> None:
--- a/src/uwtools/config/formats/base.py
+++ b/src/uwtools/config/formats/base.py
@@ -163,14 +163,20 @@ class Config(ABC, UserDict):
         """
         Render as much Jinja2 syntax as possible.
         """
+
+        def logstate(state: str) -> None:
+            log.debug("Dereferencing, %s value:", state)
+            for line in str(self).split("\n"):
+                log.debug("  %s", line)
+
         while True:
-            log.debug("Dereferencing, current value: %s", str(self))
+            logstate("current")
             new = dereference(val=self.data, context={**os.environ, **self.data})
             assert isinstance(new, dict)
             if new == self.data:
                 break
             self.data = new
-        log.debug("Dereferencing, final value: %s", str(self))
+        logstate("final")
 
     @abstractmethod
     def dump(self, path: OptionalPath) -> None:
--- a/src/uwtools/config/formats/base.py
+++ b/src/uwtools/config/formats/base.py
@@ -190,12 +190,12 @@ class Config(ABC, UserDict):
 
     @staticmethod
     @abstractmethod
-    def dump_dict(path: OptionalPath, cfg: dict) -> None:
+    def dump_dict(cfg: dict, path: OptionalPath = None) -> None:
         """
         Dumps a provided config dictionary to stdout or a file.
 
-        :param path: Path to dump config to.
         :param cfg: The in-memory config object to dump.
+        :param path: Path to dump config to.
         """
 
     @staticmethod
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -44,7 +44,7 @@ class YAMLConfig(Config):
         The string representation of a YAMLConfig object.
         """
         yaml.add_representer(TaggedScalar, TaggedScalar.represent)
-        return yaml.dump(self.data).strip()
+        return yaml.dump(self.data, default_flow_style=True).strip()
 
     # Private methods
 
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -43,6 +43,7 @@ class YAMLConfig(Config):
         """
         The string representation of a YAMLConfig object.
         """
+        yaml.add_representer(TaggedScalar, TaggedScalar.represent)
         return yaml.dump(self.data)
 
     # Private methods
@@ -95,22 +96,21 @@ class YAMLConfig(Config):
 
     # Public methods
 
-    def dump(self, path: OptionalPath) -> None:
+    def dump(self, path: OptionalPath = None) -> None:
         """
         Dumps the config in YAML format.
 
         :param path: Path to dump config to.
         """
-        self.dump_dict(path, self.data)
+        self.dump_dict(self.data, path)
 
     @staticmethod
-    def dump_dict(path: OptionalPath, cfg: dict) -> None:
+    def dump_dict(cfg: dict, path: OptionalPath = None) -> None:
         """
         Dumps a provided config dictionary in YAML format.
 
-        :param path: Path to dump config to.
         :param cfg: The in-memory config object to dump.
-        :param opts: Other options required by a subclass.
+        :param path: Path to dump config to.
         """
         yaml.add_representer(TaggedScalar, TaggedScalar.represent)
         with writable(path) as f:
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -1,10 +1,10 @@
 from types import SimpleNamespace as ns
-from typing import Dict, Optional, Union
+from typing import Optional
 
 import yaml
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import INCLUDE_TAG, log_and_error
+from uwtools.config.support import INCLUDE_TAG, TaggedScalar, log_and_error
 from uwtools.utils.file import FORMAT, OptionalPath, readable, writable
 
 _MSGS = ns(
@@ -128,22 +128,3 @@ class YAMLConfig(Config):
         Returns the config's format name.
         """
         return FORMAT.yaml
-
-
-class TaggedScalar:
-    """
-    PM WRITEME.
-    """
-
-    TAGS = ("!bool", "!float", "!int", "!str")
-
-    def __init__(self, _: yaml.SafeLoader, node: yaml.nodes.ScalarNode) -> None:
-        self.tag: str = node.tag
-        self.value: str = node.value
-
-    def reify(self) -> Union[bool, float, int, str]:
-        """
-        PM WRITEME.
-        """
-        converters: Dict[str, type] = dict(zip(self.TAGS, [bool, float, int, str]))
-        return converters[self.tag](self.value)
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -44,7 +44,7 @@ class YAMLConfig(Config):
         The string representation of a YAMLConfig object.
         """
         yaml.add_representer(TaggedScalar, TaggedScalar.represent)
-        return yaml.dump(self.data)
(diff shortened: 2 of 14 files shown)


src/uwtools/config/jinja2.py
```diff
@@ -204,6 +201,62 @@ def render(
 # Private functions
 
 
+def _deref_convert(val: TaggedString) -> _ConfigVal:
+    """
+    Convert a string tagged with an explicit type.
```
maddenp-noaa, line 206: This function is called during dereferencing to try to convert a `TaggedString` into its target type. If the conversion cannot be performed (e.g. the input was `!int "foo"`), a raised exception will be caught and logged, and the original, faulty value will be retained so that the user can see and potentially correct the problem.

src/uwtools/config/jinja2.py
```diff
@@ -204,6 +201,62 @@ def render(
...
+
+
+def _deref_debug(action: str, val: _ConfigVal) -> None:
+    """
+    Log a debug-level message related to dereferencing.
+
+    :param action: The dereferencing activity being performed.
+    :param val: The value being dereferenced.
+    """
+    log.debug("[dereference] %s: %s", action, val)
```
maddenp-noaa, line 231: The logging supported by this function now looks something like
```
DEBUG    root:base.py:167 Dereferencing, current value: {'foo': 88, 'a': '{{ b.c + 11 }}', 'b': {'c': '{{ N | int + 11 }}'}, 'd': '{{ X }}'}
DEBUG    root:jinja2.py:131 [dereference] Accepting: 88
DEBUG    root:jinja2.py:131 [dereference] Rendering: {{ b.c + 11 }}
DEBUG    root:jinja2.py:131 [dereference] Rendering failed: can only concatenate str (not "int") to str
DEBUG    root:jinja2.py:131 [dereference] Rendering: {{ N | int + 11 }}
DEBUG    root:jinja2.py:131 [dereference] Rendered: 66
DEBUG    root:jinja2.py:131 [dereference] Rendering: {{ X }}
DEBUG    root:jinja2.py:131 [dereference] Rendering failed: 'X' is undefined
DEBUG    root:base.py:167 Dereferencing, current value: {'foo': 88, 'a': '{{ b.c + 11 }}', 'b': {'c': '66'}, 'd': '{{ X }}'}
```

src/uwtools/config/jinja2.py
```diff
@@ -204,6 +201,62 @@ def render(
...
+
+
+def _deref_render(val: str, context: dict, local: Optional[dict] = None) -> str:
+    """
+    Render a Jinja2 variable/expression as part of dereferencing.
```
maddenp-noaa, line 236: Logic factored out of `dereference()`.

src/uwtools/config/jinja2.py
```diff
@@ -252,19 +305,6 @@ def path_join(path_components: List[str]) -> str:
...
-    """
-    Reify a string to a Python object, using YAML. Jinja2 templates will be passed as-is.
-
-    :param s: The string to reify.
-    """
-    try:
-        r = yaml.safe_load(s)
-    except yaml.YAMLError:
-        return s
-    return r
```
maddenp-noaa, line 265: No longer needed.

src/uwtools/config/support.py
```diff
@@ -46,3 +50,37 @@ def log_and_error(msg: str) -> Exception:
     """
     log.error(msg)
     return UWConfigError(msg)
+
+
+class TaggedString:
+    """
+    A class supporting custom YAML tags specifying type conversions.
```
maddenp-noaa, line 57: Hopefully self-explanatory.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -117,24 +118,33 @@ def test_depth(config):
     assert config.depth == 1
 
 
-def test_dereference(caplog, config):
+def test_dereference(caplog, tmp_path):
     # Test demonstrates that:
     #   - Config dereferencing uses environment variables.
     #   - Initially-unrenderable values do not cause errors.
     #   - Initially-unrenderable values may be rendered via iteration.
     #   - Finally-unrenderable values do not cause errors and are returned unmodified.
```
maddenp-noaa, line 126: This updated test continues to demonstrate these same properties, but now starts from a text file used to instantiate a `YAMLConfig` object, in order to have the `!int` tag processed during YAML parsing. (Alternatively, `TaggedString` objects could have been explicitly constructed and used as `dict` values -- some other new unit tests do this -- but that is somewhat convoluted and IMO obscured the purpose of the test.)

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -133,14 +130,6 @@ def test_dereference_str_filter_rendered():
     assert jinja2.dereference(val=val, context={"recipient": "world"}) == "hello, world"
 
 
-def test_dereference_str_variable_rendered_int():
-    # Due to reification, the value of a result parsable as an int is an int. The same holds for
-    # other results parsable by YAML as Python values, but this is only a representative, non-
-    # exhaustive test.
-    val = "{{ number }}"
-    assert jinja2.dereference(val=val, context={"number": "88"}) == 88
```
maddenp-noaa, line 141: The removed tests here and above are handled better by newly added tests, below.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -165,21 +154,6 @@ def test_register_filters_path_join(key):
...
-    for x in ["false", "no", "FALSE"]:
-        assert jinja2._reify_scalar_str(x) is False
-    assert jinja2._reify_scalar_str("88") == 88
-    assert jinja2._reify_scalar_str("'88'") == "88"  # quoted int not converted
-    assert jinja2._reify_scalar_str("3.14") == 3.14
-    assert jinja2._reify_scalar_str("NA") == "NA"  # no conversion
-    assert jinja2._reify_scalar_str("@[foo]") == "@[foo]"  # no conversion for YAML exceptions
-    with raises(AttributeError) as e:
-        jinja2._reify_scalar_str([1, 2, 3])  # type: ignore
-    assert "'list' object has no attribute 'read'" in str(e.value)  # Exception on unintended list
```
maddenp-noaa, line 180: The function-under-test no longer exists.

src/uwtools/tests/config/test_tools.py
```diff
@@ -240,31 +240,6 @@ def test_realize_config_field_table(tmp_path):
...
-        supplemental_configs=[cfgfile],
-    )
-    expected = NMLConfig(infile)
-    config_obj = INIConfig(cfgfile)
-    expected.update_values(config_obj)
-    expected_file = tmp_path / "expected.nml"
-    expected.dump(expected_file)
-    assert compare_files(expected_file, outfile)
-    with open(outfile, "r", encoding="utf-8") as f:
-        assert f.read()[-1] == "\n"
```
maddenp-noaa, line 265: We will no longer be supporting conversion of a namelist input to an INI output. This test has been marked `skip` for a long while, but is now removed.

src/uwtools/config/formats/yaml.py
```diff
@@ -43,7 +43,8 @@ def __repr__(self) -> str:
         """
         The string representation of a YAMLConfig object.
         """
-        return yaml.dump(self.data)
+        yaml.add_representer(TaggedString, TaggedString.represent)
+        return yaml.dump(self.data, default_flow_style=False).strip()
```
maddenp-noaa, line 47: Add a "representer" function for `TaggedString` objects, representing them as `!tag "value"`.

src/uwtools/config/formats/base.py
```diff
@@ -163,15 +163,20 @@ def dereference(self) -> None:
         """
         Render as much Jinja2 syntax as possible.
         """
-        log.debug("Dereferencing, initial value: %s", self.data)
+
+        def logstate(state: str) -> None:
+            log.debug("Dereferencing, %s value:", state)
+            for line in str(self).split("\n"):
+                log.debug("  %s", line)
```
maddenp-noaa, line 170: The change from `self.data` to `str(self)` is due to the fact that a `YAMLConfig` may now contain `TaggedString` nodes, whose default string representation is a mess. Calling `str()` on the object invokes its `__repr__()` method, which now registers a "representer" function with the YAML library to neatly serialize a `TaggedString` node, e.g. as `!int "88"`.

src/uwtools/config/jinja2.py
```diff
@@ -204,6 +201,62 @@ def render(
...
+    that can be represented as an int, the original value will be returned unchanged.
+
+    :param val: A scalar value tagged with an explicit type.
+    :return: The value translated to the specified type.
+    """
+    converted: _ConfigVal = val  # fall-back value
+    _deref_debug("Converting", val.value)
+    try:
+        converted = val.convert()
+        _deref_debug("Converted", val.value)
```
christinaholtNOAA, line 218: Does it make sense to use `converted` in the debug message here? `val.value` was reported int he previous debug message.

src/uwtools/config/jinja2.py
christinaholtNOAA: This is very useful debugging information!

src/uwtools/tests/config/test_support.py
```diff
@@ -51,3 +52,41 @@ def test_log_and_error(caplog):
...
+
+    def test_int_no(self, loader):
+        ts = support.TaggedString(loader, yaml.ScalarNode(tag="!int", value="foo"))
+        with raises(ValueError):
+            ts.convert()
+
+    def test_int_ok(self, loader):
+        ts = support.TaggedString(loader, yaml.ScalarNode(tag="!int", value="88"))
+        assert ts.convert() == 88
+        self.comp(ts, "!int '88'")
```
christinaholtNOAA, line 92: There's a weird behaving outlier that doesn't have a test, and I'm not sure what would happen. Python does not enjoy `int("3.14")`, but is fine with `int(3.14)`. I assume this might be a problem for a user at some point.

christinaholtNOAA (approved):
At first I was concerned about the change to support tagged scalers needing to touch 14 files. I appreciate the additional feature you added here for optional paths being truly optional.

WeirAE (approved):
Read back through and still looks solid to me

src/uwtools/config/jinja2.py
maddenp-noaa: Yes, I think you're right. Thanks.

src/uwtools/tests/config/test_support.py
maddenp-noaa: I'm going to go ahead and merge this PR but will happily revisit this topic in the next PR if we think there's more to be done.

Issue #390, opened 2024-01-22
UW-479: add uw --debug CLI option


WeirAE:
**Description**
Adds a --debug option to the CLI to also print the stack trace for an error.
Currently in uwtools.cli.main() only string value of exception is printed [eg Exception("fail") prints "fail"]

**Requirements**
should only update uwtools.cli

**Acceptance Criteria (Definition of Done)**
all uw modes support --debug
stacktraces from unhandled exceptions are printed to stderr
logger configured to DEBUG level
unit tests ensure
--help is updated in docs
demo in docs

**Additional context**
Discussion requests logging.exception(msg) as the logging method

WeirAE:
Resolved by #391

PR #391, opened 2024-01-22
UW-479: add uw --debug CLI option

WeirAE:
**Synopsis**

Adds a --debug option to the CLI to also print the stack trace for an error.

Fixes issue #390 

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -116,7 +116,7 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
      [2024-01-08T16:57:28]     INFO ---------------------------------------------------------------------
      [2024-01-08T16:57:28]     INFO values:       recipient:  - None + World
 
-  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
+  If additional information is needed, ``--debug`` can be used which will return the stack trace from any unhandled exception as well.
   
   Note that ``uw`` logs to ``stderr``, so the stream can be redirected:
 
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -50,6 +50,8 @@ The ``uw`` mode for handling configs.
            Format of file 1
      --file-2-format {ini,nml,sh,yaml}
            Format of file 2
+     --debug
+           Print all logging messages and the stack trace
      --quiet, -q
            Print no logging messages
      --verbose, -v
@@ -114,6 +116,8 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
      [2024-01-08T16:57:28]     INFO ---------------------------------------------------------------------
      [2024-01-08T16:57:28]     INFO values:       recipient:  - None + World
 
+  If additional information is needed, ``--debug`` can be used which will return the stack trace as well.
+  
   Note that ``uw`` logs to ``stderr``, so the stream can be redirected:
 
   .. code-block:: text
@@ -166,6 +170,8 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
           Print report of values needed to render template
     --dry-run
           Only log info, making no changes
+    --debug
+          Print all logging messages and the stack trace
     --quiet, -q
           Print no logging messages
     --verbose, -v
@@ -424,6 +430,8 @@ and additional supplemental YAML file ``values2.yaml`` with content
          Show help and exit
      --input-file PATH, -i PATH
          Path to input file (defaults to stdin)
+     --debug
+         Print all logging messages and the stack trace
      --quiet, -q
          Print no logging messages
      --verbose, -v
--- a/docs/sections/user_guide/cli/mode_config.rst
+++ b/docs/sections/user_guide/cli/mode_config.rst
@@ -51,7 +51,7 @@ The ``uw`` mode for handling configs.
      --file-2-format {ini,nml,sh,yaml}
            Format of file 2
      --debug
-           Print all logging messages and the stack trace
+           Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
      --quiet, -q
            Print no logging messages
      --verbose, -v
@@ -116,7 +116,7 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
      [2024-01-08T16:57:28]     INFO ---------------------------------------------------------------------
      [2024-01-08T16:57:28]     INFO values:       recipient:  - None + World
 
-  If additional information is needed, ``--debug`` can be used which will return the stack trace as well.
+  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
   
   Note that ``uw`` logs to ``stderr``, so the stream can be redirected:
 
@@ -171,7 +171,7 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
     --dry-run
           Only log info, making no changes
     --debug
-          Print all logging messages and the stack trace
+          Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
     --quiet, -q
           Print no logging messages
     --verbose, -v
@@ -431,7 +431,7 @@ and additional supplemental YAML file ``values2.yaml`` with content
      --input-file PATH, -i PATH
          Path to input file (defaults to stdin)
      --debug
-         Print all logging messages and the stack trace
+         Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
      --quiet, -q
          Print no logging messages
      --verbose, -v
--- a/docs/sections/user_guide/cli/mode_template.rst
+++ b/docs/sections/user_guide/cli/mode_template.rst
@@ -51,6 +51,8 @@ The ``uw`` mode for handling :jinja2:`Jinja2 templates<templates>`.
          Print report of values needed to render template
      --dry-run
          Only log info, making no changes
+     --debug
+         Print all logging messages and the stack trace
      --quiet, -q
          Print no logging messages
      --verbose, -v
@@ -173,6 +175,8 @@ and YAML file ``values.yaml`` with content
      [2023-12-18T23:25:01]    DEBUG Read initial values from values.yaml
      Hello, World!
 
+  If additional information is needed, ``--debug`` can be used which will return the stack trace as well.
+  
   Note that ``uw`` logs to ``stderr`` and writes non-log output to ``stdout``, so the streams can be redirected separately:
 
   .. code-block:: text
@@ -248,6 +252,8 @@ and YAML file ``values.yaml`` with content
          Path to output file (defaults to stdout)
      --dry-run
          Only log info, making no changes
+     --debug
+         Print all logging messages and the stack trace
      --quiet, -q
          Print no logging messages
      --verbose, -v
--- a/docs/sections/user_guide/cli/mode_template.rst
+++ b/docs/sections/user_guide/cli/mode_template.rst
@@ -52,7 +52,7 @@ The ``uw`` mode for handling :jinja2:`Jinja2 templates<templates>`.
      --dry-run
          Only log info, making no changes
      --debug
-         Print all logging messages and the stack trace
+         Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
      --quiet, -q
          Print no logging messages
      --verbose, -v
@@ -175,7 +175,7 @@ and YAML file ``values.yaml`` with content
      [2023-12-18T23:25:01]    DEBUG Read initial values from values.yaml
      Hello, World!
 
-  If additional information is needed, ``--debug`` can be used which will return the stack trace as well.
+  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
   
   Note that ``uw`` logs to ``stderr`` and writes non-log output to ``stdout``, so the streams can be redirected separately:
 
@@ -253,7 +253,7 @@ and YAML file ``values.yaml`` with content
      --dry-run
          Only log info, making no changes
      --debug
-         Print all logging messages and the stack trace
+         Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
      --quiet, -q
          Print no logging messages
      --verbose, -v
--- a/docs/sections/user_guide/cli/mode_template.rst
+++ b/docs/sections/user_guide/cli/mode_template.rst
@@ -175,7 +175,7 @@ and YAML file ``values.yaml`` with content
      [2023-12-18T23:25:01]    DEBUG Read initial values from values.yaml
      Hello, World!
 
-  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
+  If additional information is needed, ``--debug`` can be used which will return the stack trace from any unhandled exception as well.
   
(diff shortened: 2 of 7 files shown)


src/uwtools/cli.py
```diff
@@ -697,9 +709,12 @@ def _check_file_vs_format(file_arg: str, format_arg: str, args: Args) -> Args:
...
 
-def _check_quiet_vs_verbose(args) -> Args:
-    if args.get(STR.quiet) and args.get(STR.verbose):
-        _abort("Specify at most one of %s, %s" % (_switch(STR.quiet), _switch(STR.verbose)))
+def _check_verbosity(args) -> Args:
+    if sum([args.get(STR.debug, 0), args.get(STR.quiet, 0), args.get(STR.verbose, 0)]) > 1:
+        _abort(
+            "Specify at most one of %s, %s, or %s"
+            % (_switch(STR.debug), _switch(STR.quiet), _switch(STR.verbose))
+        )
```
christinaholtNOAA, line 713: Do we want to penalize both a `debug` and a `verbose` flag? Maybe I'm just the crotchety debugger that would be annoyed by this. 😬

src/uwtools/tests/test_cli.py
```diff
@@ -30,6 +30,14 @@ def test__abort(capsys):
     assert msg in capsys.readouterr().err
 
 
+def test__abort_debug_logs_stacktrace(capsys):
+    msg = "Traceback (most recent call last)"
+    with patch.object(sys, "argv", STR.debug):
+        with raises(SystemExit):
+            cli._abort(msg)
+    assert msg in capsys.readouterr().err
```
christinaholtNOAA, line 38: I'm curious if this works out as expected. I had expected that `logging.exception` would only work in a `try:` block.

Actually, the docs say "This method should only be called from an exception handler." Can we treat `_abort` as an exception handler in this case?

src/uwtools/tests/test_cli.py
```diff
@@ -385,39 +394,13 @@ def test__dispatch_template_render_yaml():
...
-    }
-    with patch.object(
-        uwtools.api.template, "_convert_atparse_to_jinja2"
-    ) as _convert_atparse_to_jinja2:
-        cli._dispatch_template_translate(args)
-    _convert_atparse_to_jinja2.assert_called_once_with(
-        input_file=None, output_file=None, dry_run=False
-    )
-
-
```
christinaholtNOAA, line 415: I'm confused about why these are removed here.

src/uwtools/cli.py
```diff
@@ -636,20 +645,23 @@ def _abort(msg: str) -> None:
...
+        log.exception(msg)
     print(msg, file=sys.stderr)
     sys.exit(1)
 
 
-def _add_args_quiet_and_verbose(group: Group) -> SubmodeChecks:
+def _add_args_verbosity(group: Group) -> SubmodeChecks:
     """
-    Add quiet and verbose arguments.
+    Add debug, quiet and verbose arguments.
```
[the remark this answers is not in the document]
WeirAE: I agree!

src/uwtools/tests/test_cli.py
WeirAE: Hm, this could be a conflict resolution error, I'll look into it

src/uwtools/tests/test_cli.py
WeirAE: I might have gotten carried away with `sys` use this time - I'll walk through the breakpoint to make sure its actually doing what I want

src/uwtools/cli.py
WeirAE: Strictly it would be redundant verbose calls, but I can see what you mean

src/uwtools/cli.py
```diff
@@ -636,20 +645,23 @@ def _abort(msg: str) -> None:
 
     :param msg: The message to print.
     """
+    if STR.debug in sys.argv:
+        log.exception(msg)
```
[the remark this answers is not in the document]
maddenp-noaa: `_switch(STR.debug)` maybe?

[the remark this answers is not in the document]
WeirAE: Oh that would be the safer use, thanks!

src/uwtools/cli.py
maddenp-noaa: I'm in favor of _not_ penalizing it. For example, if I have a git branch with unmerged commits and I say
```
git branch -d foo
```
git tells me that I have to use `-D` to force-delete it. Then I write
```
git branch -d foo -D
``
because I up-arrow to get the previous command, and I'm happy _every single time_ that I don't have to actually change `-d` but can just strengthen it with `-D`.

File: src/uwtools/cli.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (maddenp-noaa) [reply src/uwtools/cli.py:652]: 👍

File: src/uwtools/tests/test_cli.py (hunk 3 above)
Comment (WeirAE) [reply src/uwtools/tests/test_cli.py:415]: Yep, something went weird there - fixing it!

File: src/uwtools/tests/test_cli.py (hunk 2 above)
Comment (WeirAE) [reply src/uwtools/tests/test_cli.py:38]: Good catch, this test was self-defining. With this edit pushed, its not working as expected, so I'll see if the `try`: block works now

Comment (WeirAE) [reply src/uwtools/tests/test_cli.py]: OK, moved to the try block, but now I'm having issues capturing the logged trace - pushing for feedback

File: src/uwtools/cli.py (hunk 5 above)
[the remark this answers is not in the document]
Comment (WeirAE) [reply src/uwtools/cli.py]: Unresolving here since this I moved this to within the `try:` block of `main()`, ensuring the behavior is still as wanted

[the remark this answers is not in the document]
Comment (WeirAE) [reply src/uwtools/cli.py]: Re-resolved now that success is confirmed!

File: src/uwtools/cli.py [hunk 6]
```diff
@@ -693,9 +705,12 @@ def _check_file_vs_format(file_arg: str, format_arg: str, args: Args) -> Args:
...
 
-def _check_quiet_vs_verbose(args) -> Args:
-    if args.get(STR.quiet) and args.get(STR.verbose):
-        _abort("Specify at most one of %s, %s" % (_switch(STR.quiet), _switch(STR.verbose)))
+def _check_verbosity(args) -> Args:
+    if args.get(STR.quiet) and (args.get(STR.debug) or args.get(STR.verbose)):
+        _abort(
+            "Specify at most one of %s, %s, or %s"
+            % (_switch(STR.debug), _switch(STR.quiet), _switch(STR.verbose))
+        )
```
maddenp-noaa, line 713: I think the abort message could be a little more nuanced, as it's ok (if unnecessary) to use `--verbose` and `--debug` together, and the message says it isn't. Would it be more accurate if it rendered to something like
```
--quiet may not be used with --debug or --verbose
```

? I think that's more what the boolean conditional above actually says.

src/uwtools/tests/test_cli.py
```diff
@@ -165,6 +150,23 @@ def test__check_template_render_vals_args_noop_explicit_valsfmt():
...
+def test__check_verbosity_fail(capsys):
+    log.setLevel(logging.INFO)
+    args = {STR.quiet: True, STR.verbose: True}
+    with raises(SystemExit):
+        cli._check_verbosity(args)
+    assert (
+        "Specify at most one of %s, %s, or %s"
+        % (cli._switch(STR.debug), cli._switch(STR.quiet), cli._switch(STR.verbose))
+        in capsys.readouterr().err
+    )
```
maddenp-noaa, line 162: Sorry if I'm missing it, but shouldn't we also test that `--quiet` and `--debug` fails? Maybe that could be done with e.g.
```
@pytest.mark.parametrize("flag", (STR.debug, STR.verbose))
def test__check_verbosity_fail(capsys, flag):
    log.setLevel(logging.INFO)
    args = {STR.quiet: True, flag: True}
    ...
```

I think that would expand the test in to one each for `--debug` and `--verbose`.

src/uwtools/tests/test_cli.py
```diff
@@ -165,6 +150,23 @@ def test__check_template_render_vals_args_noop_explicit_valsfmt():
...
+
+
+def test__check_verbosity_ok():
+    args = {"foo": 88}
+    assert cli._check_verbosity(args) == args
```
maddenp-noaa, line 167: I think this could also be parametrized to check the combinations we care about (and apologies for the initial not-great test, which I probably wrote...), maybe something like
```
@pytest.mark.parametrize("flags", ([STR.debug], [STR.quiet], [STR.verbose], [STR.debug, STR.verbose]))
def test__check_verbosity_ok(flags):
    args = {flag: True for flag in flags}
    assert cli._check_verbosity(args) == args
```

i.e. It's ok to use each flag on its own, it's ok to use `--debug` and `--verbose` together, but it's not ok to use `--quiet` with anything else.

src/uwtools/tests/test_cli.py
```diff
@@ -413,11 +415,24 @@ def test__dispatch_template_translate_no_optional():
...
 
 
+def test_main_debug_logs_stacktrace(caplog):
+    log.setLevel(logging.DEBUG)
+    msg = "Test failed intentionally"
+    with patch.object(cli, "_parse_args", side_effect=Exception(msg)):
+        with patch.object(sys, "argv", cli._switch(STR.debug)):
+            with raises(SystemExit):
+                cli.main()
+                assert logged(caplog, "Traceback (most recent call last):")
```
maddenp-noaa, line 425: Nice test.

src/uwtools/tests/test_cli.py
```diff
@@ -413,11 +415,24 @@ def test__dispatch_template_translate_no_optional():
...
+
+
+@pytest.mark.parametrize("debug", [False])
 @pytest.mark.parametrize("quiet", [True])
 @pytest.mark.parametrize("verbose", [False])
-def test_main_fail_checks(capsys, quiet, verbose):
+def test_main_fail_checks(capsys, debug, quiet, verbose):
```
maddenp-noaa, line 431: I might be missing/forgetting something here, but I'm not sure what benefit we're getting from the parametrization here (and I see that this isn't new, just expanded): Since each parameter only provides a single value, it seems like those could just be set inside the test function, and the decorators and function arguments could be removed. And, then, the `if` statements seem unnecessary because we already know the outcome of everything. Maybe the parametrizations were originally put in place with the thought that they might be extended? I'm not sure. Do you see something I don't see?

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -166,6 +170,8 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
           Print report of values needed to render template
     --dry-run
           Only log info, making no changes
+    --debug
+          Print all logging messages and the stack trace
```
maddenp-noaa, line 174: These `.rst` docs will probably merge with conflicts with PR #393, which changes the indentation of a lot of these code blocks. If this PR merges first I'm happy to take care of that on my branch, but thought I should warn you in case mine merges first.

src/uwtools/tests/test_cli.py
WeirAE: Ah thats true, and that may fix the later test too

src/uwtools/tests/test_cli.py
WeirAE: Honestly I added these for consistency of options rather than the test consistency itself - lemme think about a better way to run the test from the start

christinaholtNOAA: When you parameterize like this, don't you get all the possible combos? It looks like the logic of the test is set up to handle that. For example, if each of the parameters was set up for both a True and False value, it might give us some combos we we forgot to check now that there are 3 flags.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -114,6 +116,8 @@ The examples that follow use namelist files ``values1.nml`` and ``values2.nml``,
      [2024-01-08T16:57:28]     INFO ---------------------------------------------------------------------
      [2024-01-08T16:57:28]     INFO values:       recipient:  - None + World
 
+  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
```
maddenp-noaa, line 119:
```suggestion
  If additional information is needed, ``--debug`` can be used which will return the stack trace from any unhandled exception as well.
```

docs/sections/user_guide/cli/mode_template.rst
```diff
@@ -173,6 +175,8 @@ and YAML file ``values.yaml`` with content
      [2023-12-18T23:25:01]    DEBUG Read initial values from values.yaml
      Hello, World!
 
+  If additional information is needed, ``--debug`` can be used which will return the the stack trace from any unhandled exception as well.
```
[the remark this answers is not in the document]
WeirAE: Added separately to not have to go back for 'retur'

src/uwtools/cli.py
```diff
@@ -693,9 +707,9 @@ def _check_file_vs_format(file_arg: str, format_arg: str, args: Args) -> Args:
     return args
 
 
-def _check_quiet_vs_verbose(args) -> Args:
-    if args.get(STR.quiet) and args.get(STR.verbose):
-        _abort("Specify at most one of %s, %s" % (_switch(STR.quiet), _switch(STR.verbose)))
+def _check_verbosity(args) -> Args:
+    if args.get(STR.quiet) and (args.get(STR.debug) or args.get(STR.verbose)):
+        _abort("--quiet may not be used with --debug or --verbose")
```
maddenp-noaa, line 712: I appreciate this updated message, but it would still be a good idea to use `_switch(STR.quiet)` etc. in place of the literal versions, to keep strings DRY.

docs/sections/user_guide/cli/mode_config.rst
WeirAE: It wasn't too bad, should all be corrected

src/uwtools/cli.py
WeirAE: Oh I missed that one, thanks!

src/uwtools/cli.py
```diff
@@ -464,6 +466,17 @@ def _add_arg_cycle(group: Group) -> None:
     )
 
 
+def _add_arg_debug(group: Group) -> None:
+    group.add_argument(
+        _switch(STR.debug),
+        "--debug",
```
maddenp-noaa, line 472: I don't think this line is needed. For switches with short forms (e.g. `-v` for `--verbose`, this line would give the short form. But this one doesn't, and the long form is handled by the preceding line, which expands to the same thing.

src/uwtools/tests/test_cli.py
maddenp-noaa: I like the way this turned out, with the cartesian product of all the switches handled.

src/uwtools/cli.py
```diff
@@ -636,16 +649,17 @@ def _abort(msg: str) -> None:
     sys.exit(1)
 
 
-def _add_args_quiet_and_verbose(group: Group) -> ActionChecks:
+def _add_args_verbosity(group: Group) -> SubmodeChecks:
```
maddenp-noaa, line 652: Is `SubmodeChecks` actually still defined somewhere? I thought it was eliminated in my previous PR...

src/uwtools/cli.py
WeirAE: Ah good point - got this one cleaned out

maddenp-noaa (approved):
Nice work, thanks!

elcarpenterNOAA (approved):
Nice work, Brian.

christinaholtNOAA (approved):
Nice work!

PR #394, opened 2024-01-24
typechecking-updates

maddenp-noaa:
**Synopsis**

Various updates related to strengthening typechecking settings and dealing with issues reported due to extra strictness. Don't Panic: There are a lot of repetitions of the same kind of change, and only a handful of interesting things worth attention.

Many updates involve adding explicit typehints where `mypy` was not able to deduce them. I've added inline comments for more interesting changes.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -25,6 +25,7 @@ check_untyped_defs = true
 follow_imports = "skip"
 ignore_missing_imports = true
 pretty = true
+warn_return_any = true
 
 [tool.pylint.messages_control]
 disable = [
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -22,8 +22,6 @@ profile = "black"
 
 [tool.mypy]
 check_untyped_defs = true
-follow_imports = "skip"
-ignore_missing_imports = true
 pretty = true
 warn_return_any = true
 
--- a/src/uwtools/apps/srw.py
+++ b/src/uwtools/apps/srw.py
@@ -40,8 +40,8 @@ class SRW210(Facade):
                     ">%s" % f,
                 ]
                 cmd = " ".join(cmd_components)
-                result = execute(cmd=cmd)
-                if not result.success:
+                _, success = execute(cmd=cmd)
+                if not success:
                     raise UWError(f"Command failed: {cmd}")
         elif file_type == FORMAT.yaml:
             shutil.copy2(config_file, "config.yaml")
@@ -64,8 +64,8 @@ class SRW210(Facade):
         """
         # Note: This is a temporary path until parsing the SRW directory is implemented.
         cmd = "python generate_FV3LAM_wflow.py"
-        result = execute(cmd=cmd)
-        if not result.success:
+        _, success = execute(cmd=cmd)
+        if not success:
             raise UWError(f"Command failed: {cmd}")
 
     def create_manager_files(self) -> None:
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -27,8 +27,8 @@ FORMATS = list(FORMAT.formats().keys())
 TITLE_REQ_ARG = "Required arguments"
 
 Args = Dict[str, Any]
-SubmodeChecks = List[Callable[[Args], Args]]
-ModeChecks = Dict[str, SubmodeChecks]
+ActionChecks = List[Callable[[Args], Args]]
+ModeChecks = Dict[str, ActionChecks]
 Checks = Dict[str, ModeChecks]
 
 
@@ -44,7 +44,7 @@ def main() -> None:
     setup_logging(quiet=True)
     try:
         args, checks = _parse_args(sys.argv[1:])
-        for check in checks[args[STR.mode]][args[STR.submode]]:
+        for check in checks[args[STR.mode]][args[STR.action]]:
             check(args)
         setup_logging(quiet=args[STR.quiet], verbose=args[STR.verbose])
         log.debug("Command: %s %s", Path(sys.argv[0]).name, " ".join(sys.argv[1:]))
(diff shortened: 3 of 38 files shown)


src/pyproject.toml
```diff
@@ -22,9 +22,8 @@ profile = "black"
 
 [tool.mypy]
 check_untyped_defs = true
-follow_imports = "skip"
-ignore_missing_imports = true
```
maddenp-noaa, line 26: These weakened `mypy`.

src/pyproject.toml
```diff
@@ -22,9 +22,8 @@ profile = "black"
...
 pretty = true
+warn_return_any = true
```
maddenp-noaa, line 26: This strengthens `mypy`.

src/uwtools/config/formats/nml.py
```diff
@@ -1,7 +1,7 @@
 from collections import OrderedDict
 from typing import Optional, Union
 
-import f90nml
+import f90nml  # type: ignore
```
maddenp-noaa, line 4: The `f90nml` library does not yet provide type-hint stubs for `mypy`. This suppresses a warning from `mypy` about that issue. In the future, we could choose to write our own type stubs for the parts of `f90nml` that we use, if the maintainer does not provide stubs.

src/uwtools/apps/srw.py
```diff
@@ -40,8 +40,8 @@ def load_config(self, config_file: str) -> None:
                     ">%s" % f,
                 ]
                 cmd = " ".join(cmd_components)
-                result = execute(cmd=cmd)
-                if not result.success:
+                _, success = execute(cmd=cmd)
```
maddenp-noaa, line 43: Previously, `execute()` returned a `NameSpace` object with `.output` and `.success` attributes, but these did not carry any useful type information. That functions has been updated to return a `Tuple[bool, str]`-typed value that _does_ carry useful type information, and is now unpacked into two local variables wherever it is called.

src/uwtools/config/formats/yaml.py
```diff
@@ -59,7 +59,9 @@ def _load(self, config_file: OptionalPath) -> dict:
         loader = self._yaml_loader
         with readable(config_file) as f:
             try:
-                return yaml.load(f.read(), Loader=loader)
+                config = yaml.load(f.read(), Loader=loader)
+                assert isinstance(config, dict)
+                return config
```
maddenp-noaa, line 64: In this case, it's actually possible for `yaml.load` to return a number of things other than a `dict`, but we expect (so far, at least) YAML that defines a top-level dictionary. So I've added an assertion to that effect, which I hope is better than lying and saying `config: dict = yaml.load(...)`, which will placate `mypy` but cause problems later when the expected `dict` object is a `list`, `int`, etc. We could alternatively raise an exception if a `dict` isn't received, but I'm not sure, at the moment, what's best. Maybe we'll see.

src/uwtools/rocoto.py
```diff
@@ -8,7 +8,7 @@
 from typing import Any, List, Optional, Tuple, Union
 
 from lxml import etree
-from lxml.etree import Element, SubElement
+from lxml.etree import Element, SubElement, _Element
```
maddenp-noaa, line 11: It turns out that `_Element` is the name of the _type_ defined my `lxml` for XML element objects, while `Element` is a function that creates elements:
```
>>> import lxml.etree
>>> type(lxml.etree._Element)
<class 'type'>
>>> type(lxml.etree.Element)
<class 'cython_function_or_method'>
```
So, `_Element` needs to be used in type hints.

src/uwtools/tests/test_cli.py
```diff
@@ -110,7 +110,7 @@ def test__check_file_vs_format_pass_explicit():
     assert args[STR.infmt] == fmt
 
 
-@pytest.mark.parametrize("fmt", vars(FORMAT).keys())
+@pytest.mark.parametrize("fmt", FORMAT.formats())
```
maddenp-noaa, line 113: I added a static method on the `FORMAT` dataclass to return a map from format variant names (e.g. `yaml` and `yml`) to their canonical names (e.g. `yaml`) and use it here to avoid getting unwanted attribute names from `vars(FORMAT)`.

src/uwtools/cli.py
```diff
@@ -23,7 +23,7 @@
 from uwtools.logging import log, setup_logging
 from uwtools.utils.file import FORMAT, get_file_format
 
-FORMATS = [FORMAT.ini, FORMAT.nml, FORMAT.sh, FORMAT.yaml]
+FORMATS = list(FORMAT.formats().keys())
```
NaureenBharwaniNOAA, line 26: Great update

elcarpenterNOAA (approved):
A nice collection of updates. Looks good!

src/uwtools/config/formats/yaml.py
WeirAE: Hrm, my first preference would be to enforce at load, but barring that I think I like raising an exception. However, this is already better than before knowing that

src/uwtools/config/formats/nml.py
WeirAE: We'll probably have to do this at some point to be honest

WeirAE (approved):
A few namespaces fixed and types adjusted but it really cleans it up!

src/uwtools/config/formats/nml.py
```diff
@@ -34,7 +34,8 @@ def _load(self, config_file: OptionalPath) -> dict:
         :param config_file: Path to config file to load.
         """
         with readable(config_file) as f:
-            return f90nml.read(f)
+            config: dict = f90nml.read(f)
+            return config
```
christinaholtNOAA, line 38: This type of change is a little unfortunate. Bleh.

christinaholtNOAA (approved):
I'm sorry I did this! I really appreciate you fixing it all so quickly. There are some really nice updates here.

src/uwtools/config/formats/nml.py
maddenp-noaa: Yeah. The problem is that the type of `f90nml.read()` is `Any` because `f90nml` has no type stubs. An unfortunate workaround for now.

maddenp-noaa:
@christinaholtNOAA Don't be sorry! This was a good thing to get done. We'd hidden some things from `mypy` and our software-quality checks are stronger now.

PR #405, opened 2024-01-31
fv3-schema-update

maddenp-noaa:
**Synopsis**

Updates to the FV3Forecast JSON Schema to try to track current code and documentation. Issues mentioned in [UW-495](https://jira.epic.oarcloud.noaa.gov/browse/UW-495) may still need to be addressed in the context of this PR.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/uw_yaml/forecast_yaml.rst
+++ b/docs/sections/user_guide/uw_yaml/forecast_yaml.rst
@@ -198,7 +198,7 @@ The MPI command used to run the model executable. Typical options are ``srun``, 
 
 ``scheduler:``
 ^^^^^^^^^^^^^^
-The name of the batch system. Supported options are ``lfs``, ``pbs``, and ``slurm``.
+The name of the batch system. Supported options are ``lsf``, ``pbs``, and ``slurm``.
 
 The ``preprocessing:`` block
 ----------------------------
--- a/src/uwtools/resources/FV3Forecast.jsonschema
+++ b/src/uwtools/resources/FV3Forecast.jsonschema
@@ -4,17 +4,7 @@
       "minProperties": 1,
       "patternProperties": {
         "^.*$": {
-          "oneOf": [
-            {
-              "type": "string"
-            },
-            {
-              "items": {
-                "type": "string"
-              },
-              "type": "array"
-            }
-          ]
+          "type": "string"
         }
       },
       "type": "object"
--- a/src/uwtools/resources/FV3Forecast.jsonschema
+++ b/src/uwtools/resources/FV3Forecast.jsonschema
@@ -121,6 +121,18 @@
         },
         "model_configure": {
           "additionalProperties": false,
+          "anyOf": [
+            {
+              "required": [
+                "base_file"
+              ]
+            },
+            {
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
           "properties": {
             "base_file": {
               "type": "string"
@@ -139,13 +151,22 @@
               "type": "object"
             }
           },
-          "required": [
-            "base_file"
-          ],
           "type": "object"
         },
         "namelist": {
           "additionalProperties": false,
+          "anyOf": [
+            {
+              "required": [
+                "base_file"
+              ]
+            },
+            {
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
           "properties": {
             "base_file": {
               "type": "string"
@@ -173,9 +194,6 @@
               "type": "object"
             }
           },
-          "required": [
-            "base_file"
-          ],
           "type": "object"
         },
         "run_dir": {
--- a/src/uwtools/resources/FV3Forecast.jsonschema
+++ b/src/uwtools/resources/FV3Forecast.jsonschema
@@ -1,38 +1,33 @@
 {
   "$defs": {
-    "updatable_config": {
-      "additionalProperties": false,
-      "properties": {
-        "properties": {
-          "base_file": {
-            "format": "uri",
-            "type": "string"
-          },
-          "update_values": {
-            "type": "object"
-          }
-        },
-        "required": [
-          "base_file"
-        ],
-        "type": "object"
-      }
+    "filesToStage": {
+      "minProperties": 1,
+      "patternProperties": {
+        "^.*$": {
+          "oneOf": [
+            {
+              "type": "string"
+            },
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            }
+          ]
+        }
+      },
+      "type": "object"
     }
   },
-  "description": "This document is to validate user-defined FV3 forecast config files",
   "properties": {
     "forecast": {
       "additionalProperties": false,
-      "description": "parameters of the forecast",
       "properties": {
-        "cycle_dependent|static": {
-          "propertyNames": {
-            "type": "string"
-          },
-          "type": "object"
+        "cycle_dependent": {
+          "$ref": "#/$defs/filesToStage"
         },
         "diag_table": {
-          "format": "uri",
           "type": "string"
         },
         "domain": {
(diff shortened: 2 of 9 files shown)


docs/sections/user_guide/uw_yaml/forecast_yaml.rst
```diff
@@ -198,7 +198,7 @@ The MPI command used to run the model executable. Typical options are ``srun``,
 
 ``scheduler:``
 ^^^^^^^^^^^^^^
-The name of the batch system. Supported options are ``lfs``, ``pbs``, and ``slurm``.
+The name of the batch system. Supported options are ``lsf``, ``pbs``, and ``slurm``.
```
maddenp-noaa, line 201: Right?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -43,88 +38,237 @@
...
+            },
+            "update_values": {
+              "minProperties": 1,
+              "patternProperties": {
+                "^.*$": {
+                  "type": [
+                    "boolean",
+                    "number",
+                    "string"
+                  ]
```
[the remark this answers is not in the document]
christinaholtNOAA: We are going to have problems with `model_configure` as a YAML. I'm almost ready to consider proposing that we update the model code to use a YAML parser for this file. I am not there just yet.

I know that we already have one problem with standard model_configure files -- they use the file type entry with too many quotes like this:

```
filename_base:           'dyn''phy'
```

I think it would be a helpful upgrade to the model to make this "almost YAML" file an actual YAML file. Thoughts?

This is a lot of risk at the moment. I would be fine if we wanted to treat this file as a flat text file template until we can ensure it's going to be YAML compliant.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -43,44 +38,180 @@
           "type": "string"
         },
         "executable": {
-          "format": "uri",
           "type": "string"
         },
-        "fd_ufs": {
-          "$ref": "#/$defs/updatable_config"
-        },
```
christinaholtNOAA, line 51: I think it might be good to keep this one around, but given that we don't know a ton about it, I believe it can go into the vat of static files for now. If anyone requests the ability to update it, we can add it to the configurable file list at that time.

How does that sound?

A more general question: how do we want to handle the user staging those required files that require a specific name? Are they required entries in the `static` block?

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -43,44 +38,180 @@
...
+                  "type": "object"
+                }
+              },
+              "type": "object"
+            }
+          },
+          "required": [
+            "base_file"
+          ],
+          "type": "object"
```
christinaholtNOAA, line 116: A `field_table` file is going to be required either here or in the static files.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -90,41 +221,54 @@
           "type": "string"
         }
       },
+      "required": [
+        "scheduler"
+      ],
```
christinaholtNOAA, line 226: Is a scheduler required at the YAML level? I don't think we would need that if we're running without the `--batch-script` flag. We do always need the `mpicmd`, though.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -90,41 +221,54 @@
...
+            "interval_hours",
+            "offset",
+            "output_file_path"
+          ],
           "type": "object"
         }
       },
+      "required": [
+        "lateral_boundary_conditions"
+      ],
```
christinaholtNOAA, line 257: This would only be required for a regional run. I'm not sure if we want to worry about that just yet since the whole thing is pretty hard-coded for that case at the moment.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -90,41 +221,54 @@
...
       "type": "object"
     },
     "user": {
+      "additionalProperties": false,
```
christinaholtNOAA, line 261: I think we will want the `user` section to be able to set anything. That's where they might do anything they like, such as tossing in re-usable variables that can be referenced in a `realize` step.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -90,41 +221,54 @@
...
+      "required": [
+        "scheduler"
+      ],
       "type": "object"
     },
     "preprocessing": {
+      "additionalProperties": false,
       "properties": {
         "lateral_boundary_conditions": {
+          "additionalProperties": false,
```
christinaholtNOAA, line 233: When we get to the point of having a driver for creating the lateral boundary conditions instead of using staged files, we'll likely want to have more information in this section.

christinaholtNOAA (commented):
I'm realizing with this PR that we will likely want to brainstorm soon-ish how to deal with the interaction of the requirements of various drivers with each other just as a forward thinking exercise.

Here, it becomes evident in the parts that limit entries of a given section like the `preprocessing: lateral_boundary_conditions:` block. We will likely want more than that to be able to communicate the needs of that driver, but these are the bare minimum needed to complete a regional atmospheric uncoupled forecast. I tried to point out those bits in particular that are related to this concept.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Let me see if I understand: Are you suggesting that we remove explicit support for `fd_ufs` as a standalone entry in UW FV3Forecast YAML and just let users specify it, for now at least, in the `static:` block? If so, that sounds good to me.

For the more general question, I think you're asking whether we should explicitly require certain static files that we know must be present. That brings more questions to mind for me:

1. Is there a meaningful difference, from a driver's perspective, between "cycle specific" and "static" files, excep that the former may contain e.g. Jinja2 expressions related to the cycle that need to be rendered to select the correct files? Since a driver (or, at least, the FV3 driver we have so far) is specific to a given cycle, could we simplify and just have e.g. `files_to_link:`, which might contain both plain path strings (static files) and path strings with Jinja2 expressions, which may be cycle-specific (or could be based on something else in the config). Is the distinction useful enough to justify the complexity?
2. `uwtools` currently allows, for these file lists, either key-value pairs where the key is the symlink name relative to the run directory and the value is the target path, OR pairs where the key is a directory relative to the run directory, and the value is a list of target files. If we want to enforce that some files are _required_, it's definitely going to lead to a lot of duplication (and schema complexity) to enforce the same requirements for both styles of value -- and ideally it would be forbidden to specify the same run-directory file two different ways in the same config. I'm not even sure it's possible to enforce required elements in the second, list (JSON Schema `array`) style: If so, it may be through so-called "tuple validation", which allows for required and addition entries, but probably enforces _order_ of entries, too, and it might be a big ask from users to not only specify all these files, but keep them in a specific order. It would be a lot simpler to just have the first key-value style, as in a JSON Schema it's simple (not to mention DRY) to specify required keys and then permit additional, unknown, free-form key-value pairs. So: Is this list feature worth the complexity?
3. Given the above, we could potentially simplify and also support required link-into-run-directory files as well as optional, user-specified ones. How realistic is the use case where a user has changed the code such that some file `uwtools` thinks is required no longer is, but the schema forces the user to specify it anyway? I do think that it could be a big help to most users to tell them when they're not providing all the files we know they're going to need. Would we just help my hypothetical power user tweak their `FV3Forecast.jsonschema` to let them avoid specifying files they don't want? Suggest they just provide the file anyway, even though they don't need it?

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Not according to the code, in `create_field_table():
```
config_values=self._config.get("field_table", {})
```

Should that become
```
config_values=self._config("field_table")
```

? Then it would be required. I can make that update if so.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: It doesn't look like `scheduler` is ever accessed unless `--batch_script` (CLI) or `batch_script` (API) is set. Will fix.

`mpicmd` is optional in the code:
```
self._platform_config.get("mpicmd")
```

Should that change to
```
self._platform_config["mpicmd"]
```

?

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: I _think_ this is currently mandatory from the code's perspective: `_define_boundary_files()` has
```
lbcs_config = self._experiment_config["preprocessing"]["lateral_boundary_conditions"]
```

and that method is unconditionally called by `prepare_directories()`, which is called by both `_run_via_batch_submission()` and `_run_via_local_execution()`, _one of which_ is called by `run()`, all in class `FV3Forecast`. So, I think the code will crash, given the current control flow, if this key is not present.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Having a place for free-form user config sounds like a good idea, but I don't think that should be mixed with schema-controlled content. Should `account` perhaps move next to `scheduler`, as they are both related to the batch system?

src/uwtools/resources/FV3Forecast.jsonschema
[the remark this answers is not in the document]
maddenp-noaa: `model_configure` as-is is the worst of both worlds, a YAML/namelist chimera. IMO UFS should pick one or the other and use it. If it is being read by something other than Fortran, probably YAML is the right language, but if there's an easier path to implementation (e.g. a library for a different config language is already in UFS, or is easier to add than a YAML library), probably that would be fine, too.

I think templating would be a safe alternative in the short term, though unfortunate in that the app or user will have to define everything in advance.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: I proactively made `mpicmd` required in both schema and code, but please let me know if that's not right.

src/uwtools/resources/FV3Forecast.jsonschema
christinaholtNOAA: So, FV3 requires a `field_table` file to run. The user might want to be able to generate it from a YAML config, but may instead provide a fully formed one that needs no modification. In the first case, there would be a `field_table:` YAML, entry. In the second case, it would be provided as a static file entry.

src/uwtools/resources/FV3Forecast.jsonschema
christinaholtNOAA: Having `mpicmd` required both in code and in the schema is probably the right move. I don't think FV3 could run without some MPI command.

src/uwtools/resources/FV3Forecast.jsonschema
christinaholtNOAA: Let's leave it as-is for now and remember that we have work to do to move to supporting global configurations.

src/uwtools/resources/FV3Forecast.jsonschema
christinaholtNOAA: I like that idea. Let's move it.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: OK, so it sounds like `field_table` requires _at least one of_ `base_file` and `update_values`.

maddenp-noaa: I'll push what I think is a correct implementation of the above momentarily.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: `mpicmd` is now required, in schema and in code.

src/uwtools/resources/FV3Forecast.jsonschema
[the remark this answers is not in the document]
maddenp-noaa: @christinaholtNOAA What do we want to do with this in the short term? If FV3 _requires_ a `model_configure` file, then we need to require a `model_configure` YAML entry with some combination of `base_file` and/or `update_values`. Should we treat it the same as `field_table`, i.e. at-least-one handling of `base_file` and `update_values` so that the user can 1. Provide only a file, 2. Provide only values, 3. Provide and file and values to update the file with?

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Done. Please un-resolve this or create a new comment if there appear to be issues.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: Moved `account` to `platform:` block alongside `mpicmd` and `scheduler`. Made it required iff `scheduler` is defined.

src/uwtools/tests/drivers/test_forecast.py
```diff
@@ -404,3 +404,22 @@ def test_FV3Forecast__run_via_local_execution(fv3_run_assets):
             assert success is True
             assert lines[0] == "Command:"
             execute.assert_called_once_with(cmd=ANY, cwd=ANY, log_output=True)
+
+
+# Schema tests
```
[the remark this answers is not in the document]
maddenp-noaa: I have now added more granular schema tests but will add more before this merges, once we settle on open questions.

src/uwtools/resources/FV3Forecast.jsonschema
maddenp-noaa: I think we sorted this out in group today: We'll now have `files_to_copy` and `files_to_link` YAML blocks and, in all cases, these blocks will contain `str` -> `str` key-value pairs where the keys are filenames/paths relative to the run directory, and values are absolute paths to files to copy/link. Keys and values may contain Python template syntax, which we'll render in the driver when copying/linking the files. (This is more likely to apply to to-be-linked files, which are more likely to be cycle-specific, and the formatting is likely to involve cycle-specific data.)

I think the schema is appropriately updated here. The copying/linking work will be handled in a different PR involving the graph-based driver work.

maddenp-noaa:
[config.txt](https://github.com/ufs-community/uwtools/files/14185779/config.txt)
Here's a YAML config that passes validation given the schema as of this writing. Please let me know if you see any issues.

src/uwtools/tests/fixtures/forecast.yaml
maddenp-noaa, line 1: Sorted for easier comparison to schema, test config files, etc., and changed `cycle_dependent` -> `files_to_copy`, `static` -> `files_to_link`.

src/uwtools/resources/FV3Forecast.jsonschema
```diff
@@ -43,44 +25,199 @@
...
+            }
+          },
+          "type": "object"
         }
       },
+      "required": [
+        "field_table",
+        "run_dir",
+        "runtime_info"
+      ],
```
maddenp-noaa, line 204: @christinaholtNOAA I suspect that this list is incomplete. Aren't `domain`, `executable`, `length`, `namelist`, possible others also actually required? Although their contents might be variable, isn't at least _one_ file to link and one file to copy required for a run, such that the (new) `files_to_copy` and `files_to_link` blocks would also be required?

christinaholtNOAA: After our discussion yesterday, I think it might make sense for us to handle only those entries that are not file-related here in the schema: `domain`, `executable`, `length`, `run_dir`, `runtime_info`.

All others, I think we can leave up to the user whether they'd prefer to have them staged or handled via a config entry for either template or config handling. The task would enforce their existence before run time.

christinaholtNOAA (approved):
LGTM. Thanks for sticking with me through several iterations, and owning getting this into decent shape.

PR #408, opened 2024-02-11
UW-501 / UW-502 FV3 graph-based driver implementation

maddenp-noaa:
**Synopsis**

Rework `forecast` driver as `fv3` driver, using task-graph techniques based on the `iotaa` library.

~Many of the unit tests are currently disabled and need to be updated, which I'll take care of before marking this PR ready to review, but the main library code is worth reviewing.~ Unit tests are back at 100%.

There's a comment with at attached FV3 YAML config, and a path on `hera` where a sample run can be inspected.

Documentation updates will follow in a separate PR.

**Type**

**Impact**

**Checklist**


src/uwtools/resources/fv3.jsonschema
```diff
@@ -11,7 +11,7 @@
     }
   },
   "properties": {
-    "forecast": {
+    "fv3": {
```
maddenp-noaa, line 14: File [config.txt](https://github.com/ufs-community/uwtools/files/14231292/config.txt) (actually contains YAML) validates against this schema and should give a good idea of the current state of the FV3 config. On `hera`, the run in `/home/Paul.Madden/run` was created with this `config.yaml` and the command `uw fv3 run --config-file config.yaml --cycle 2019-07-01T00 --batch`. That run gets as far and fails in the same way as the reference run `/scratch2/BMC/wrfruc/cholt/forecast_driver_test/expt/2019070100`, so I think the code in this PR at least reproduces the capabilities of the previous version. (Which is not to say it doesn't still need some work.)

src/uwtools/resources/platform.jsonschema
```diff
@@ -0,0 +1,29 @@
+{
```
maddenp-noaa, line 1: I factored out the `platform` schema so that it can be used (and validated) by multiple drivers. The FV3 driver validates both the `fv3` and `platofrm` schema blocks, as it uses both of them.

src/uwtools/scheduler.py
```diff
@@ -1,346 +1,372 @@
 """
-Job Scheduling.
+Support for HPC schedulers.
```
maddenp-noaa, line 2: This module has been updated to cooperate with the `uwtools.drivers.driver` and `uwtools.drivers.fv3` modules.

Makefile
```diff
@@ -6,7 +6,7 @@ TARGETS     = clean-devenv devshell docs env format lint meta package test typec
 
 export RECIPE_DIR := $(shell cd ./recipe && pwd)
 
-clean = $(shell $(CONDA_EXE) env remove -n DEV-$(call val,name))
+clean = $(info $(shell $(CONDA_EXE) env remove -n DEV-$(call val,name)))
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I was going to bring this up in our meeting today, last Friday I tried to remove and remake the `devshell` env. This was the error message I received `make: *** No rule to make target 'clean-devenv'.  Stop.` Does this update fix the faulty behavior you're mentioning above?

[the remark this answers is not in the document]
maddenp-noaa: This fixes a different issue. Essentially, `make` tries to execute any text it sees, and when the underlying `conda env remove` command emitted a `Removing...` message, `make` tried to execute that message as if it were a command. The `$(info)` wrapper prevents that from happening.

I'm not sure why you would have seen the error message you show above. If it happens again, let me know and I can look at it with you.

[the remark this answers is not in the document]
NaureenBharwaniNOAA: I tried it again this morning to replicate the error to post the comment, so I did see it again this morning.

recipe/channels
```diff
@@ -1 +1,2 @@
 conda-forge
+maddenp
```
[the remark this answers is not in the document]
WeirAE: I was thinking about this a bit, but it would probably be best to capture some of those options here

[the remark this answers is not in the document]
maddenp-noaa: Three options occur to me:
1. Copy the core `iotaa` module from its module as `iotaa.py` into this repo. This isn't great because it would complicate syncing changes and risk updates being made in `uwtools` that aren't properly upstreamed to the `iotaa` repo, leading to divergence.
2. Fork the `iotaa` repo and set it up to publish packages to the `ufs-community` anaconda.org channel. This probably makes keeping the repos in sync somewhat easier, though not completely because their CI configurations will diverge.
3. Get `iotaa` published to `conda-forge`. This would eliminate use of the `maddenp` channel, which I assume would be the source of any complaints. I don't know how much work this would take, but probably not much, assuming that whatever policies conda-forge has for inclusion could be met. This would probably mean the least amount of long-term maintenance.

maddenp-noaa:
I went ahead and did the documentation updates in this PR. Preview docs are [here](https://uwtools--408.org.readthedocs.build/en/408/).

src/uwtools/config/formats/base.py
```diff
@@ -159,7 +159,7 @@ def depth(self) -> int:
         """
         return depth(self.data)
 
-    def dereference(self) -> None:
+    def dereference(self, context: Optional[dict] = None) -> None:
```
maddenp-noaa, line 162: I added some support here for doing a custom dereferencing run with special values, and a unit test to shore it up. The immediate use case is that I think we need to support the ability of users to have cycle-specific values in FV3 UW YAML (and probably in other drivers' UW YAML later). For example, we should be able to have, say,
```
run_dir: /some/path/fv3/{{ cycle.strftime('%Y%m%d%H') }}
```
that would become
```
run_dir: /some/path/fv3/2024021212
```
for today's 12Z cycle, e.g. if `--cycle 2024-02-12T12` were specified on the command line. But `dereference()` as it was wouldn't help, because `cycle` is not part of the YAML file, but is something we obtain from the CLI or API. But, with this, I was able to add an extra line to the constructor of class `FV3` to get this kind of Jinja2 `cycle` reference rendered correctly.

After I did this I noticed a line in the old docs saying that this was not supported yet. Well, it is now!

src/uwtools/config/jinja2.py
```diff
@@ -127,7 +127,7 @@ def dereference(val: _ConfigVal, context: dict, local: Optional[dict] = None) ->
     """
     rendered: _ConfigVal = val  # fall-back value
     if isinstance(val, dict):
-        return {k: dereference(v, context, local=val) for k, v in val.items()}
+        return {dereference(k, context): dereference(v, context, local=val) for k, v in val.items()}
```
maddenp-noaa, line 130: Keys in UW YAML may also contain e.g. `cycle` references that need to be dereferenced, so that's done here now. A use case is the keys in the FV3 driver's `files_to_copy` and `files_to_link` entries, which represent filenames relative to the FV3 run directory and which may need to be cycle-specific.

src/uwtools/drivers/fv3.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        The FV3 driver.
+
+        :param config_file: Path to config file.
+        :param cycle: The forecast cycle.
+        :param dry_run: Run in dry-run mode?
+        :param batch: Run component via the batch system?
+        """
+        super().__init__(config_file=config_file, dry_run=dry_run, batch=batch)
+        self._config.dereference(context={"cycle": cycle})
```
maddenp-noaa, line 41: Here's the new line that renders `cycle` Jinja2 references in the input YAML config.

src/uwtools/scheduler.py
```diff
@@ -1,346 +1,364 @@
...
+        for key, value in self._processed_props.items():
+            if key in self._managed_directives:
+                switch = self._managed_directives[key]
+                ds.append(
+                    "%s %s" % (pre, switch(value))
+                    if callable(switch)
+                    else "%s %s%s%s" % (pre, switch, sep, value)
+                )
+            else:
+                ds.append("%s %s%s%s" % (pre, key, sep, value))
```
maddenp-noaa, line 47: The two `if` conditionals deal with the cases

1. The directive is managed by this module (recognized by a canonical short name, like `jobname` instead of ``-J`` for LSF, `-N` for PBS, `--job-name` for Slurm), and its value is a `lambda` that needs to be called with the supplied value to calculate the full directive text;
2. The directive is managed but its value can be directly used; or
3. The directive is unmanaged.

In case 3, the user will need to supply the full, scheduler-specific switch and value, like `--nice` and `100` for Slurm. This is a power-user feature, and we should consider extending support for managed directives if it seems like users would benefit. But there are *lots* of directives, and trying to add comprehensive support for obscure ones might not be worth the effort.

docs/sections/user_guide/cli/mode_fv3.rst
```diff
@@ -0,0 +1,103 @@
...
+
+  The driver creates a ``runscript`` file in the directory specified by ``run_dir`` in the config and submits it to the batch system. Running with ``--batch`` requires a correctly configured ``platform`` block in ``config,yaml``, as well as appropriate settings in the ``execution`` block under ``fv3``.
+
+* Specifying the ``--dry-run`` flag results in the driver logging messages about actions it would have taken, without actually taking any.
+
+  .. code-block:: text
+
+     $ uw fv3 run --config-file config.yaml --cycle 2024-02-11T12 --batch --dry-run
+
+* The ``run`` task depends on the other available tasks and executes them as prerequisites. It is possible to execute any task directly, which entails execution of any of *its* depdendencies. For example, to create an FV3 run directory provisioned with all the files, directories, symlinks, etc. required per the configuration file:
```
elcarpenterNOAA, line 99: depdendencies -> dependencies

src/uwtools/drivers/fv3.py
```diff
@@ -0,0 +1,321 @@
...
+        """
+        yield "File %s" % path
+        yield asset(path, path.is_file)
+
+    @task
+    def _filecopy(self, src: Path, dst: Path):
+        """
+        A copy of an existing file.
+
+        :param src: Path to the soruce file.
```
elcarpenterNOAA, line 234: soruce -> source

docs/sections/user_guide/cli/mode_fv3.rst
maddenp-noaa: Fixing, thank you.

src/uwtools/drivers/driver.py
```diff
@@ -1,177 +1,146 @@
...
 
-    @abstractmethod
-    def batch_script(self) -> BatchScript:
+    @staticmethod
+    def _create_user_updated_config(
+        config_class: Type[Config], config_values: dict, path: Path
+    ) -> None:
         """
-        Create a script for submission to the batch scheduler.
+        Create a config from a base file, user-provided values, of a combination of the two.
```
NaureenBharwaniNOAA, line 43: I think you might have meant to say `Create a config from a base file, user-provided values, of a combination of the two.` --> `Create a config from a base file, user-provided values, *or* a combination of the two.`

maddenp-noaa: Yes, will fix, thanks!

src/uwtools/drivers/fv3.py
```diff
@@ -0,0 +1,321 @@
...
+        The FV3 lateral boundary-condition files.
+        """
+        yield self._taskname("lateral boundary condition files")
+        lbcs = self._driver_config["lateral_boundary_conditions"]
+        offset = abs(lbcs["offset"])
+        endhour = self._driver_config["length"] + offset + 1
+        interval = lbcs["interval_hours"]
+        symlinks = {}
+        for n in [7] if self._driver_config["domain"] == "global" else range(1, 7):
+            for bndyhour in range(offset, endhour, interval):
```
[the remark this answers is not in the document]
maddenp-noaa: Yes, since you had to ask, I see that the longer version would be better! Will change. Thanks.

elcarpenterNOAA (approved):
Nice work, Paul.

NaureenBharwaniNOAA (approved):

PR #410, opened 2024-02-15
UW-506 sfc_climo_gen driver

maddenp-noaa:
**Synopsis**

- Add a driver for the `sfc_climo_gen` component.
- Wired up driver to API and CLI.
- Factored some common task logic out of `uwtools.drivers.fv3` to a new `uwtools.utils.tasks` module for reuse.
- Moved resources to new `jsonschema/` and `rocoto/` subdirectories of `src/uwtools/resources`, as appropriate.
- Factored shared schemas out into individual files and added logic to load these as needed via `$ref` entries.
- Colocated all granular JSON Schema tests in a single module, including new tests.
- Factored formatting logic from `Makefile` out into a `format` script.
- Added API module to docs.
- Further documentation will follow in a separate PR.

See `/home/Paul.Madden/run_sfc_climo_gen` for an example run via the driver. The (actually YAML) config file [config-scg.txt](https://github.com/ufs-community/uwtools/files/14301148/config-scg.txt) was used to drive that run, and validates with
```
uw config validate --schema-file src/uwtools/resources/jsonschema/sfc-climo-gen.jsonschema --input-file config-scg.txt
```

**Type**

**Impact**

**Checklist**


docs/sections/user_guide/index.rst
```diff
@@ -7,4 +7,4 @@ User Guide
    installation
    cli/index
    api/index
-   uw_yaml/index
+   yaml/index
```
maddenp-noaa, line 10: Simplified name. I think the `uw` part goes without saying.

format
```diff
@@ -0,0 +1,15 @@
...
+echo "=> Running isort"
+isort -q src
+
+echo "=> Running docformatter"
+(cd src && docformatter . || test $$? -eq 3)
+
+echo "=> Running jq"
+for a in $(find src -type f -name "*.jsonschema"); do
+  b=$(jq -S . $a) && echo "$b" >$a || (echo "    in $a"; false)
+done
```
maddenp-noaa, line 15: In addition to factoring existing bash logic out of `Makefile`, the `jq` formatting step will now report _which file_ had a syntax error. This was fairly obvious before when we only had one or two files, but it's crucial now that we have a growing collection.

Having this logic in its own scripts also means that editors can do better syntax highlighting than when it lived in a `Makefile`.

recipe/run_test.sh
```diff
@@ -56,7 +56,6 @@ unittest() {
 }
 
 test "${CONDA_BUILD:-}" = 1 && cd ../test_files || cd $(dirname $0)/../src
-msg Running in $PWD
```
maddenp-noaa, line 59: While working on this PR, I iterated on the tests a lot and was annoyed by this not-very-useful message, so removed it.

src/pyproject.toml
```diff
@@ -28,6 +28,7 @@ warn_return_any = true
 [tool.pylint.messages_control]
 disable = [
   "consider-using-f-string",
+  "duplicate-code",
```
maddenp-noaa, line 31: We already had a number of in-code suppressors of this `pylint` complaint, and I think that a growing collection of component drivers are going to have graph tasks that look "duplicate" from `pylint`'s perspective, so it might be less annoying to suppress at this level. I suspect that "we're all adults here" and will DRY our code whenever and wherever we can.

src/uwtools/api/config.py
```diff
@@ -1,3 +1,6 @@
+"""
+API access to uwtools configuration management tools.
+"""
```
maddenp-noaa, line 3: Added module docstrings to `uwtools.api.*` modules.

src/uwtools/api/sfc_climo_gen.py
```diff
@@ -0,0 +1,42 @@
+"""
+API access to the uwtools sfc_climo_gen driver.
+"""
```
maddenp-noaa, line 3: The API interface to the `sfc_climo_gen` driver.

src/uwtools/cli.py
```diff
@@ -327,6 +329,58 @@ def _dispatch_rocoto_validate(args: Args) -> bool:
     return uwtools.api.rocoto.validate(xml_file=args[STR.infile])
 
 
+# Mode sfc_climo_gen
```
maddenp-noaa, line 332: The CLI interface to the `sfc_climo_gen` driver.

src/uwtools/cli.py
```diff
@@ -442,7 +496,7 @@ def _add_arg_config_file(group: Group) -> None:
         help="Path to config file",
         metavar="PATH",
         required=True,
-        type=str,
+        type=Path,
```
maddenp-noaa, line 499: Let `argparse` convert string values to `Path` objects as early as possible.

src/uwtools/config/validator.py
```diff
@@ -57,6 +58,19 @@ def _prep_config(config: Union[dict, YAMLConfig, Optional[Path]]) -> YAMLConfig:
...
+
+    # See https://github.com/python-jsonschema/referencing/issues/61 about typing issues.
+
+    def retrieve(uri: str) -> Resource:
+        name = uri.split(":")[-1]
+        with open(resource_path(f"jsonschema/{name}.jsonschema"), "r", encoding="utf-8") as f:
+            return Resource(contents=json.load(f), specification=DRAFT202012)  # type: ignore
+
+    registry = Registry(retrieve=retrieve)  # type: ignore
+    validator = jsonschema.Draft202012Validator(schema, registry=registry)
```
maddenp-noaa, line 75: The updated logic in this function configures `jsonschema` to dynamically load appropriate `.jsonschema` files from `uwtools.resources.jsonschema` when they are referenced by another schema, supporting DRYing out common elements between schemas -- for example, blocks describing component-execution properties (`execution.jsonschema`) or block describing namelists and their sets of names and values (`namelist.jsonschema`). `$ref` entries in schemas can now refer to other schemas as `urn:uwtools:<name>` (e.g. `urn:uwtools:execution`) to access the appropriate schema. See below for examples.

src/uwtools/drivers/fv3.py
```diff
@@ -100,29 +100,29 @@ def field_table(self):
     @tasks
     def files_copied(self):
         """
-        Files copied for FV3 run.
+        Files copied for run.
```
maddenp-noaa, line 103: We're probably going to copy-paste some tasks between drivers, so I tried to make their docstrings less likely to have too-specific terms that we'll inevitably forget to fix. Of course, we should factor _identical_ tasks out where it makes sense (e.g. to the new `uwtools.utils.tasks` module), but my guess is that there will be some copy-paste-lightly-customize going on.

src/uwtools/drivers/fv3.py
```diff
@@ -217,33 +217,10 @@ def runscript(self):
...
-        """
-        A copy of an existing file.
-
-        :param src: Path to the source file.
-        :param dst: Path to the destination file to create.
-        """
-        yield "Copy %s -> %s" % (src, dst)
-        yield asset(dst, dst.is_file)
-        yield self._file(src)
-        copy(src, dst)
```
maddenp-noaa, line 241: These two private (task) methods, along with `_symlink()` below, now live in `uwtools.utils.tasks` as public functions.

src/uwtools/drivers/sfc_climo_gen.py
```diff
@@ -0,0 +1,162 @@
+"""
+A driver for sfc_climo_gen.
+"""
```
maddenp-noaa, line 3: This follows the pattern set by the FV3 driver: Instantiate the driver object, which entails reading in and validating the config file against its schema, then provide graph-task methods that can be called from the CLI or API.

src/uwtools/drivers/sfc_climo_gen.py
```diff
@@ -0,0 +1,162 @@
...
+        self._rundir = Path(self._driver_config["run_dir"])
+
+    # Workflow tasks
+
+    @task
+    def namelist_file(self):
+        """
+        The namelist file.
+        """
+        fn = "fort.41"
```
maddenp-noaa, line 43: Sad but true.

src/uwtools/drivers/sfc_climo_gen.py
```diff
@@ -0,0 +1,162 @@
...
+        yield self._taskname(f"namelist file {fn}")
+        path = self._rundir / fn
+        yield asset(path, path.is_file)
+        vals = self._driver_config["namelist"]["update_values"]["config"]
+        input_paths = [Path(v) for k, v in vals.items() if k.startswith("input_")]
+        input_paths += [Path(vals["mosaic_file_mdl"])]
+        input_paths += [Path(vals["orog_dir_mdl"]) / fn for fn in vals["orog_files_mdl"]]
+        yield [file(input_path) for input_path in input_paths]
```
maddenp-noaa, line 51: In the current implementation, the `namelist_file` task collects the names of all the input files needed by `sfc_climo_gen`, as specified in its `config` namelist, and yields them as `@external` tasks, to ensure that the namelist file is not created if any of the files it specifies do not exist. This could potentially be done instead in the `_run_via_batch_submission()` method, so that `uw` would decline to submit a batch job that is (probably?) doomed to fail, which would be a waste of time and other resources. But doing the check here essentially tells the user "Your namelist is incorrect", which might be useful.

src/uwtools/drivers/sfc_climo_gen.py
```diff
@@ -0,0 +1,162 @@
...
+        yield asset(path, path.is_file)
+        yield None
+        envcmds = self._driver_config.get("execution", {}).get("envcmds", [])
+        execution = [self._runcmd, "test $? -eq 0 && touch %s/done" % self._rundir]
+        scheduler = self._scheduler if self._batch else None
+        path.parent.mkdir(parents=True, exist_ok=True)
+        rs = self._runscript(envcmds=envcmds, execution=execution, scheduler=scheduler)
+        with open(path, "w", encoding="utf-8") as f:
+            print(rs, file=f)
+        os.chmod(path, os.stat(path).st_mode | stat.S_IEXEC)
```
maddenp-noaa, line 93: We'll probably want to do even more DRY refactoring once we have a third driver and see yet more commonalities. For example, while the arguments to `self._runscript()` vary between this driver and the `fv3` one, lines 89-93 will probably be mostly identical between drivers and could be factored out to a common method, with the arguments to `self._runscript()` being passed to that common method. In particular, the `chmod` command to make the script executable is ugly/weird and should probably not be repeatedly duplicated.

I propose that, if we don't find unexpected unique behavior in our third driver, that we should do a refactoring sweep and see what can be factored out either to the `uwtools.drivers.driver.Driver` class or to the `uwtools.utils.tasks` module.

src/uwtools/resources/jsonschema/execution.jsonschema
```diff
@@ -0,0 +1,34 @@
...
+    "threads": {
+      "minimum": 0,
+      "type": "integer"
+    }
+  },
+  "required": [
+    "executable"
+  ],
+  "type": "object"
+}
```
maddenp-noaa, line 34: The schema for the `execution` block common to both the `fv3` and `sfc_climo_gen` configs.

src/uwtools/resources/jsonschema/files-to-stage.jsonschema
```diff
@@ -0,0 +1,7 @@
+{
+  "additionalProperties": {
+    "type": "string"
+  },
+  "minProperties": 1,
+  "type": "object"
+}
```
maddenp-noaa, line 7: Currently only used by `fv3` because `sfc_climo_gen` can refer to all its input files via absolute paths in its `config` namelist, but certain to be used by other drivers.

src/uwtools/resources/jsonschema/fv3.jsonschema
```diff
@@ -1,15 +1,4 @@
...
-    "filesToStage": {
-      "minProperties": 1,
-      "patternProperties": {
-        "^.*$": {
-          "type": "string"
-        }
-      },
-      "type": "object"
-    }
-  },
```
maddenp-noaa, line 12: Factored out to its own schema file.

src/uwtools/resources/jsonschema/fv3.jsonschema
```diff
@@ -25,38 +14,7 @@
...
-            },
-            "threads": {
-              "minimum": 0,
-              "type": "integer"
-            }
-          },
-          "required": [
-            "executable"
-          ],
-          "type": "object"
```
maddenp-noaa, line 59: Factored out to `execution.jsonschema`.

src/uwtools/resources/jsonschema/fv3.jsonschema
```diff
@@ -141,10 +99,10 @@
           "type": "object"
         },
         "files_to_copy": {
-          "$ref": "#/$defs/filesToStage"
+          "$ref": "urn:uwtools:files-to-stage"
```
maddenp-noaa, line 102: This is how a reference to another schemas now looks.

src/uwtools/resources/jsonschema/fv3.jsonschema
```diff
@@ -225,26 +183,7 @@
...
-                        "number",
-                        "string"
-                      ]
-                    },
-                    "type": "object"
-                  },
-                  "type": "object"
-                }
-              },
-              "type": "object"
```
maddenp-noaa, line 247: Factored out to `namelist.jsonschema`.

src/uwtools/resources/jsonschema/fv3.jsonschema
```diff
@@ -261,9 +200,6 @@
         "run_dir"
       ],
       "type": "object"
-    },
-    "user": {
-      "type": "object"
```
maddenp-noaa, line 266: I don't think we need this. Our drivers will validate each top-level config block that they use (e.g. the FV3 driver validates the `fv3` and `platform` blocks), and anything else in the YAML config is fair game for users. See e.g. the example config file attached to to PR and how it DRYs out `sfc_climo_gen` input filenames with Jinja2 references.

src/uwtools/resources/jsonschema/sfc-climo-gen.jsonschema
```diff
@@ -0,0 +1,137 @@
...
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
+          "properties": {
+            "base_file": {
+              "type": "string"
+            },
+            "update_values": {
```
maddenp-noaa, line 27: For `sfc_climo_gen`, I explicitly enumerated the possible variables supported in the `config` namelist, per [the source code](https://ufs-community.github.io/UFS_UTILS/sfc_climo_gen/program__setup_8f90_source.html) and their data types as, hopefully, a convenient sanity check for users. Additional properties are also permitted, just in case a developer is working on the `sfc_climo_gen` code itself and needs to add new variables: Those may be any appropriate type, but the types of the _known_ variables can be explicitly checked.

This more explicit checking is why `namelist.jsonschema` is not used via a `$ref` here: That schema is more permissive.

src/uwtools/tests/config/test_validator.py
```diff
@@ -13,7 +13,7 @@
 from uwtools.config import validator
 from uwtools.config.formats.yaml import YAMLConfig
 from uwtools.logging import log
-from uwtools.utils.file import resource_pathobj
+from uwtools.utils.file import resource_path
```
maddenp-noaa, line 16: Since we've now eliminated (I think) `str` paths from the codebase, paths are always `Path` objects, so `pathobj` is a somewhat redundant name. Simplified.

src/uwtools/tests/drivers/test_fv3.py
```diff
@@ -45,64 +44,64 @@ def config(tmp_path):
...
-    path = tmp_path / "fv3.yaml"
+    path = tmp_path / "config.yaml"
     with open(path, "w", encoding="utf-8") as f:
         yaml.dump(config, f)
     return path
 
 
 @fixture
-def fv3obj(config_file, cycle):
+def driverobj(config_file, cycle):
```
maddenp-noaa, line 54: Here, too, since I think we'll be doing some copy-paste-tweak in the future, I tried to make names a little more generic so that we'll need to edit them less when we borrow from existing code.

src/uwtools/tests/drivers/test_fv3.py
```diff
@@ -217,371 +216,57 @@ def test_FV3_runscript(fv3obj):
...
-    assert fv3obj._taskname("foo") == "20240201 18Z FV3 foo"
+def test_FV3__taskanme(driverobj):
+    assert driverobj._taskname("foo") == "20240201 18Z FV3 foo"
 
 
-def test_FV3__validate(fv3obj):
-    fv3obj._validate()
-
-
-# Schema fixtures
```
maddenp-noaa, line 276: All granular schema tests and their fixtures now live together in `uwtools.tests.test_schema`. They have not changed in any meaningful way, just moved.

src/uwtools/tests/support.py
```diff
@@ -94,16 +94,18 @@ def regex_logged(caplog: LogCaptureFixture, msg: str) -> bool:
     return any(pattern.search(record.message) for record in caplog.records)
 
 
-def validator(schema_fn: str, *args: Any) -> Callable:
+def schema_validator(schema_name: str, *args: Any) -> Callable:
```
maddenp-noaa, line 97: Use a more explicit anme to avoid confusion with the name of the `uwtools.config.validator` module.

src/uwtools/tests/test_cli.py
```diff
@@ -69,6 +71,31 @@ def test__add_subparser_fv3(subparsers):
...
+
+
+def test__add_subparser_rocoto_realize(subparsers):
+    cli._add_subparser_rocoto_realize(subparsers)
+    assert subparsers.choices[STR.realize]
+
+
+def test__add_subparser_rocoto_validate(subparsers):
+    cli._add_subparser_rocoto_validate(subparsers)
+    assert subparsers.choices[STR.validate]
```
maddenp-noaa, line 86: There were a handful of missing (though fairly trivial) `rocoto`-related CLI tests. Added here.

src/uwtools/tests/test_rocoto.py
```diff
@@ -433,86 +433,3 @@ def test_dump(self, instance, tmp_path):
         path = tmp_path / "out.xml"
         instance.dump(path=path)
         assert rocoto.validate_rocoto_xml_file(path)
-
-
-# Schema tests
```
maddenp-noaa, line 438: Moved to `uwtools.tests.test_schema`.

src/uwtools/api/sfc_climo_gen.py
```diff
@@ -0,0 +1,42 @@
...
+    dry_run: bool = False,
+) -> bool:
+    """
+    Execute an sfc_climo_gen task.
+
+    If ``batch`` is specified, a runscript will be written and submitted to the batch system.
+    Otherwise, the forecast will be run directly on the current system.
+
+    :param task: The task to execute
+    :param config_file: Path to UW YAML config file
```
NaureenBharwaniNOAA, line 25: I saw you removed a lot of the `uw yaml` language. In case you wanted to change this to just `YAML` here as well.

src/uwtools/resources/jsonschema/fv3.jsonschema
NaureenBharwaniNOAA: Could you explain what `urn` is for? I was looking this up in the JSONSchema documentation and didn't find a hit minus this is how you reference an external schema. Assuming the `utwools:files-to-stage` part of `urn:uwtools:files-to-stage` shows a path of where this schema now lives based on the new file created within resources/jsonschema

src/uwtools/api/sfc_climo_gen.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def execute(
+    task: str,
+    config_file: Path,
+    batch: bool = False,
+    dry_run: bool = False,
+) -> bool:
+    """
+    Execute an sfc_climo_gen task.
```
elcarpenterNOAA, line 19: SRW documentation refers to the `make_sfc_climo` task, while `sfc_climo_gen` refers to the executable. I see we have the `chgres_cube` driver coming next, which is an executable created by the build process as well. For future drivers, should we only refer to the executable name?

src/uwtools/api/sfc_climo_gen.py
maddenp-noaa: Yes, thank you, it kind of goes without saying. Fixing.

src/uwtools/resources/jsonschema/fv3.jsonschema
maddenp-noaa: As background, the `jsonschema` package now handles inter-schema references via the [`referencing`](https://referencing.readthedocs.io/en/stable/intro/) library, and has provided [a guide](https://python-jsonschema.readthedocs.io/en/latest/referencing/#migrating-from-refresolver) for migrating away from the previous mechanism.

URNs are conceptually related to URLs and URIs -- see e.g. https://pitt.libanswers.com/faq/213722. They are simple, and defined by [RFC 2141](https://www.ietf.org/rfc/rfc2141.txt). The only really interesting thing about them for `uwtools` is that `referencing` -- and, thus, `jsonschema` -- recognized and accepts them as the values of `$ref` values in schema documents. (Search for `urn:` in [this page](https://referencing.readthedocs.io/en/stable/intro/) to see use cases from the `referencing` docs.) Our code takes care of mapping these URN values to schema files, loading those, and adding them to a "registry" object that  `jsonschema` then uses to resolve the `$ref`s.

Our other option for `$ref` values would be to use a URI scheme. You can see lots of examples of this on the internet, where people might use, in a case like the one you cite above, like `https://uwtools.com/files-to-stage`. I find this uncomfortable because, as I understand it, it really has nothing to do with HTTPS and no network access is done to resolve the references. All the URI scheme is really providing here is _hierarchy_ -- a way to identify resources based on the slash characters separating components. The `uwtools.com` domain doesn't have to exist, and could probably be any text as long as it agreed with other URIs, at least in some cases. Alternatively, we could use something like `file://uwtools/files-to-stage`, but I find this misleading, too. I think that any URI scheme (and there are [lots of official ones](https://referencing.readthedocs.io/en/stable/intro/)) would do.

I find the URNs conceptually cleaner. There is a potential downside in that, as far as I can tell, URNs don't provide a way to "drill down" into a referenced schema the way URIs can. For example, if I had a big `common.jsonschema` document with lots of schemas, I might write `file://uwtools/common#/properties/namelist` to reference a sub-schema in that document. I think with URNs we'll need to have a one-to-one mapping between URNs a schema files, but that might be ok.

src/uwtools/api/sfc_climo_gen.py
maddenp-noaa: I think that _drivers_ should use the name of the executable. Well, I mean, you can technically name an executable program whatever you want, and there's not necessarily a canonical name for it. I've seen, for example, FV3 compiled to an executable program called just `fv3`, or `fv3.exe`, or `ufs_model`. But calling it "FV3" / `fv3` seems sensible to me. Hard to say about `sfc_climo_gen`: The [Fortran source code](https://ufs-community.github.io/UFS_UTILS/sfc_climo_gen/driver_8F90_source.html) says that the `program` is called `driver`, which is pretty generic. So, here, I feel like a community-recognized name for the program itself is probably the right identifier to use for the driver.

If there are other actions that need to be taken to "make surface climatology" (which I guess is what the name `make_sfc_climo` is getting at, then I can imagine a higher-level task called `make_sfc_climo` that would have a constituent task that runs the `sfc_climo_gen` component via the driver implemented in this PR, but possibly take other actions either based on the `sfc_climo_gen` executable's output or other actions.

So, I guess I'm thinking that `sfc_climo_gen` is an appropriate name for a _component driver_, and `make_sfc_climo` might be an appropriate name for a TDB _experiment task_ that does additional work beyond running this driver. I hope that makes sense, and I'm happy to hear other ideas if it sounds wrong!

christinaholtNOAA (approved):
There's a lot of content here. It looks very clean and clear. I can't seem to find anything wrong with the approach at this point. :)

src/uwtools/drivers/fv3.py
```diff
@@ -217,33 +217,10 @@ def runscript(self):
...
-
     @task
     def _run_via_batch_submission(self):
         """
-        FV3 run Execution via the batch system.
+        Run execution via the batch system.
```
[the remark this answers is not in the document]
maddenp-noaa: It should be a noun, not a verb: Tasks are stand-ins for what they create (asset(s)), and this task creates "a run". Would
> A run executed via the batch system.

be better? I think it's important to avoid thinking of tasks as actions, or using language that obscures their asset-orientation. Focusing on nouns, the question "What exactly IS a 'run'?" comes to mind, and it's a crucial question to have a good answer to. Some of the outputs/assets wrapped by tasks will necessarily be somewhat conceptual, but there will have to be some defined, observable state -- maybe an enumerated set of expected output files, or even just a sentinel file written at the end of the run.

src/uwtools/drivers/sfc_climo_gen.py
maddenp-noaa: We saw it just now in the live demo but, for `@external` tasks like the ones representing the files names as inputs in the namelist, `iotaa` will emit `WARNING` messages about those, e.g.
```
[2024-02-21T18:58:45]  WARNING File /scratch1/NCEPDEV/nems/role.epic/UFS_SRW_data/develop/fix/fix_sfc_climo/facsf.1.0.ncx: State: Pending (EXTERNAL)
```

in the case of a file that doesn't exist (I changed `.nc` to `.ncx` in the filename in the YAML config to make this break) vs e.g.
```
[2024-02-21T18:58:45]     INFO File /scratch1/NCEPDEV/nems/role.epic/UFS_SRW_data/develop/fix/fix_sfc_climo/maximum_snow_albedo.0.05.nc: State: Ready
```

for an `@external` file that exists. These `WARNING` message stand out well in the `uw` output IMO, and users will quickly get used to paying attention to them.

src/uwtools/drivers/fv3.py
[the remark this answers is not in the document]
maddenp-noaa: I went ahead and updated these task descriptions. We can certainly revisit if there are better ideas later.

Issue #413, opened 2024-02-23
UW-500 - Allow forced conversion between formats


WeirAE:
**Description**
User story: Can you use config realize to translate a config file from one format to another? E.g., I wanted to take an nml file and convert it to .yaml. Could I do that via uw config realize -i foo.nml --input-format nml -o foo.yaml --output-format yaml?

**Requirements**
YAML should always be accepted as input or output format. Documentation and tests must be updated appropriately

**Acceptance Criteria (Definition of Done)**
ability to convert supported formats to yaml format is restored
--force flag for overriding the check that allows YAML to convert to other file formats
documentation updated
tests updated

**Additional context**
Allowing YAML as always accepted supplemental file will be covered in UW-504

WeirAE:
Marked as 'refactor' as this behavior was reasonably expected rather than necessarily 'new'

WeirAE:
Initial testing shows that YAML is already accepted, tested and documented as universally allowed input, so perhaps --force is not needed, only improved documentation

WeirAE:
After further re-checking requirements and testing, YAML as output has some issues in creating a valid YAML from sh/ini or nml dicts. This may again require --force to allow this behavior.

WeirAE:
Further discussion in group: we will allow YAML as output always, for much the same reason it is a valid input. However, this will be exclusive to YAML explicitly.

WeirAE:
Completed by #414

PR #414, opened 2024-02-23
UW 500 - Allow YAML in all conversions

WeirAE:
**Synopsis**

YAML should always be accepted as input or output format. Documentation and tests must be updated appropriately

Completes issue #413 

**Type**

- [ ] Documentation

**Impact**

**Checklist**


src/uwtools/config/tools.py
```diff
@@ -251,8 +251,11 @@ def _validate_format_output(input_fmt: str, output_fmt: str) -> None:
     :param output_fmt: Output format.
     :raises: UWError if output format is incompatible.
     """
-    if not input_fmt in (FORMAT.yaml, output_fmt):
-        raise UWError("Output format %s must match input format %s" % (output_fmt, input_fmt))
+    if FORMAT.yaml not in (input_fmt, output_fmt) and input_fmt != output_fmt:
```
WeirAE, line 254: This test already loops all formats, so only the condition needs to be changed

src/uwtools/tests/config/test_tools.py
```diff
@@ -645,12 +645,15 @@ def test__realize_config_values_needed_negative_results(caplog, tmp_path):
 @pytest.mark.parametrize("output_fmt", FORMAT.extensions())
 def test__validate_format_output(input_fmt, output_fmt):
     call = lambda: tools._validate_format_output(input_fmt=input_fmt, output_fmt=output_fmt)
-    if input_fmt in (FORMAT.yaml, output_fmt):
+    if FORMAT.yaml in (input_fmt, output_fmt) or input_fmt == output_fmt:
```
WeirAE, line 648: Comments offline note that this does not ensure that the actual rendering is correct or expected. Particularly, namelist dicts are understood by YAML with !! notation but are not rendered as expected.

docs/sections/user_guide/cli/mode_config.rst
```diff
@@ -434,16 +434,16 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
...
    In all cases, any supplemental configs must be in the same format as the input config and must have recognized extensions.
 
    ``uw`` considers invalid combination requests errors:
 
    .. code-block:: text
 
-      $ uw config realize --input-file b.nml --output-file a.yaml
-      Output format yaml must match input format nml
+      $ uw config realize --input-file b.nml --output-file a.ini
+      Output format ini must match input format nml unless either is YAML
```
maddenp-noaa, line 446: That's a weird message: Obviously neither is `YAML` if one is `ini` and the other is `nml`. There's bound to be a better way to state this.

maddenp-noaa (changes requested):

docs/sections/user_guide/cli/mode_config.rst
WeirAE: The user guide messages need to be reinstated, they were lost in resolving conflicts, but I definitely was not happy with this phrasing, so I'm working on it

src/uwtools/tests/config/test_tools.py
WeirAE: This issue is now covered by the additional YAML representer

src/uwtools/config/support.py
```diff
@@ -88,3 +89,17 @@ def represent(dumper: yaml.Dumper, data: TaggedString) -> yaml.nodes.ScalarNode:
...
+
+def represent_namelist(dumper: yaml.Dumper, data) -> yaml.nodes.MappingNode:
+    """
+    Serialize f90nml tags.
+    """
+    # Convert the f90nml Namelist object to a dictionary
+    namelist_dict = data.todict()
+    for key, value in namelist_dict.items():
+        if isinstance(value, OrderedDict):
+            namelist_dict[key] = dict(value)
```
christinaholtNOAA, line 102: I was (perhaps naively) expecting this mod to have been in the `nml` module. Perhaps even the logic in the `__repr__` method of a nml config object, and the dumper function in the module. I'd be interested in hearing other thoughts on that.

christinaholtNOAA (commented):
This is the behavior I was expecting! Thanks for working out the kinks.

docs/sections/user_guide/cli/mode_config.rst
WeirAE: Possibly better phrasing has been updated

src/uwtools/config/support.py
WeirAE: Hmm, my feeling was to try to keep `yaml` behavior in `yaml` methods, but I can see where you were going with that. `__repr__` might be the best place if we were to do that, I'll have to think about that one

maddenp-noaa: The current PR logic seems to me like a solution to the problem "How do I deal with unrepresentable `OrderedDict` objects when I dump a config to YAML?" It is independent of the fact that this situation is, at the moment, arising for us only in the context of `f90nml` `Namelist` objects -- it seems like it would handle/replace `OrderedDict` objects from any source at YAML dump time. That seems fine to me, if possibly more than what's strictly needed at the moment.

I think that maybe this routine belongs in `uwtools.config.formats.yaml` instead of in this module, which strikes me as too general -- would some code other than `uwtoolsconfig.formats.yaml` need this?

Also, does this correctly operate with any amount of nesting? Old code for dealing with `OrderedDict` objects in `f90nml` objects (see [here](https://github.com/ufs-community/uwtools/blob/965d8ddd3347777f5a5f7af8e8e22839ca79b8e5/src/uwtools/config/formats/nml.py#L36-L42)) operated recursively -- do we need to care about that here?

I don't immediately see why this would be related to `NMLConfig`'s `__repr__()` method: That would be a way to get a string representation of the object, but that's not necessarily YAML. In fact, shouldn't it be, well, the string representation of a namelist file, like `&nl n=88 /`, etc.? I could be missing something...

src/uwtools/config/support.py
```diff
@@ -88,3 +89,17 @@ def represent(dumper: yaml.Dumper, data: TaggedString) -> yaml.nodes.ScalarNode:
         documentation for details.
         """
         return dumper.represent_scalar(data.tag, data.value)
+
+
+def represent_namelist(dumper: yaml.Dumper, data) -> yaml.nodes.MappingNode:
+    """
+    Serialize f90nml tags.
+    """
+    # Convert the f90nml Namelist object to a dictionary
```
maddenp-noaa, line 98: The command says `Namelist` object (which is a real type from the `f90nml` library), but the `isinstance()` call checks for `OrderedDict`.

maddenp-noaa:
Given `a.nml`
```
&a i4(1:2)=88,88 /
```
the code on `main` can
```
% uw config realize --input-file a.nml --output-format nml
&a
    i4(1:2) = 88, 88
/
```
Are things ok on this branch if that changes to `--output-format yaml`?

src/uwtools/config/support.py
WeirAE: Whats interesting is that afaik, `ini`'s implementation should create `OrderedDict`s, but `yaml` doesn't seem to mind. Testing with the other example `ini` files, something in the parser seems to step correctly instead of creating nested `OrderedDict`s. My wariness of putting it too high up was whether there was a condition that the native parser would handle correctly on its own but would generate errors if forced through. 

I don't think this comes up in any circumstance except when feeding non-`YAML` data directly to `yaml.dump()` so we could pre-parse it by default there?

src/uwtools/config/support.py
WeirAE: Oof thanks, I was rushing through that

src/uwtools/config/support.py
maddenp-noaa: It sounds like they once used `OrderedDict` in `configparser` but eliminated it since all currently-support Python versions have an ordered `dict` by default: https://bugs.python.org/issue33504. Are you still seeing `OrderedDict` objects in values returned by `configparser`?

WeirAE: So then the trick then would be to use configparser instead of the functions within f90nml. I had tried to do other handling within the yaml formatting, but ended up walking back to the dumper. The benefit should be in correctly parsing without having to use a conditional. 
The f90 Namelist object is a subclass of OrderedDict, so its handling was what I had been working through

maddenp-noaa: > So then the trick then would be to use configparser instead of the functions within f90nml.

No, sorry, that's not what I was getting at. I was just pointing out that `f90nml` may be the only place where `OrderedDict` object appear for us. Once upon a time that might have been an issue for `configparser` representations of INI data, but not these days.

> The benefit should be in correctly parsing without having to use a conditional.

Yes. I think I agree with your approach. It lets `YAMLConfig` do the right thing with an `OrderedDict` object _no matter how it got introduced_, with no conditional logic for handling `NMLConfig` objects separately. My only concerns were 1. Is the code in the right module? 2. Does it handle nesting/recursion correctly? 3. Are there any types other than `OrderedDict` that we need to be concerned with immediately? My best guess answers are: No, Yes, No.

I was chatting about this with @christinaholtNOAA in Slack, too, but since it's late in the day, maybe we can revisit this in DSM or Group tomorrow when we all have clear heads?

WeirAE: AH, got it up on my big screen and I think I'm following properly now - I'll run a few tests through in the morning to ensure we have no spurious `OrderedDict` elsewhere (just to be sure), scoot this over to under `formats.yaml`, and fix that other namelist condition you suggested that's not working. Thanks a ton for the support!

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+    """
+    Convert ordered dict objects to dictionaries and serialize.
+
+    :param dumper: The YAML dumper.
+    :param data: The ordereddict to serialize.
+    """
+    # Convert the Ordered Dict to a dictionary
+    for key, value in data.items():
+        if isinstance(value, OrderedDict):
+            data[key] = dict(value)
```
christinaholtNOAA, line 82: In this more generalized approach, we will want the recursive version.

christinaholtNOAA (changes requested):

src/uwtools/config/formats/yaml.py
```diff
@@ -45,6 +53,8 @@ def __repr__(self) -> str:
         The string representation of a YAMLConfig object.
         """
         yaml.add_representer(TaggedString, TaggedString.represent)
+        yaml.add_representer(f90nml.namelist.Namelist, represent_namelist)
```
maddenp-noaa, line 56: It might look cleaner to
```
from f90nml import Namelist
```
and just
```
yaml.add_representer(Namelist, represent_namelist)
```
here.

But I also see references to both `f90nml.Namelist` and `f90nml.namelist.Namelist` -- is there a difference, or can we use just one? I'm not sure this is accurate but FWIW:
```
>>> f90nml.Namelist is f90nml.namelist.Namelist
True
```
Shorter is better IMO if that's accurate.

src/uwtools/config/support.py
WeirAE: Strictly, moving it to a representer makes it automatically recursive within YAML, but its just better behavior to loop it here

src/uwtools/config/formats/yaml.py
```diff
@@ -116,6 +126,8 @@ def dump_dict(cfg: dict, path: Optional[Path] = None) -> None:
         :param path: Path to dump config to.
         """
         yaml.add_representer(TaggedString, TaggedString.represent)
+        yaml.add_representer(f90nml.namelist.Namelist, represent_namelist)
+        yaml.add_representer(OrderedDict, represent_ordereddict)
```
maddenp-noaa, line 130: I think we have the same 3 `add_representer()` calls now, both here and in `__repr__()`, if I'm not mistaken. Maybe it's time to factor those out into a method that can be called from both call sites, so that these are kept in-sync, assuming that all 3 should always be called in both cases? We'd need a direct unit test of that new method, but maybe there's logic to factor out of existing tests to create that one; then the ones that (hopefully) currently make assertions about `add_representer()` having been called could mock the new method and assert that it was called.

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
     return UWConfigError(msg)
 
 
+def represent_namelist(dumper: yaml.Dumper, data: f90nml.Namelist) -> yaml.nodes.MappingNode:
+    """
+    Convert f90 Namelist objects to ordered dicts and serialize.
```
maddenp-noaa, line 60:
```suggestion
    Convert f90nml Namelist to OrderedDict and serialize.
```

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+
+    :param dumper: The YAML dumper.
+    :param data: The f90nml Namelist to serialize.
+    """
+    # Convert the f90nml Namelist to an ordered dictionary
```
maddenp-noaa, line 65:
```suggestion
    # Convert the f90nml Namelist to an OrderedDict.
```

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+
+
+def represent_ordereddict(dumper: yaml.Dumper, data: OrderedDict) -> yaml.nodes.MappingNode:
+    """
+    Convert ordered dict objects to dictionaries and serialize.
+
+    :param dumper: The YAML dumper.
+    :param data: The ordereddict to serialize.
+    """
+    # Convert the Ordered Dict to a dictionary
```
maddenp-noaa, line 79:
```suggestion
    # Convert the OrderedDict to a dict.
```

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+    # Represent the dictionary as a YAML mapping
+    return dumper.represent_mapping("tag:yaml.org,2002:map", namelist_dict)
+
+
+def represent_ordereddict(dumper: yaml.Dumper, data: OrderedDict) -> yaml.nodes.MappingNode:
+    """
+    Convert ordered dict objects to dictionaries and serialize.
+
+    :param dumper: The YAML dumper.
+    :param data: The ordereddict to serialize.
```
maddenp-noaa, line 77:
```suggestion
    :param data: The OrderedDict to serialize.
```

src/uwtools/config/support.py
```diff
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from collections import OrderedDict
 from importlib import import_module
 from typing import Dict, Type, Union
 
+import f90nml  # type: ignore
```
maddenp-noaa, line 7: Here, too, maybe `from f90nml import Namelist` unless there are other `f90nml` items needed in this module.

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+    """
+    Convert f90 Namelist objects to ordered dicts and serialize.
+
+    :param dumper: The YAML dumper.
+    :param data: The f90nml Namelist to serialize.
+    """
+    # Convert the f90nml Namelist to an ordered dictionary
+    namelist_dict = data.todict()
+
+    # Represent the dictionary as a YAML mapping
```
maddenp-noaa, line 68:
```suggestion
    # Represent the OrderedDict as a YAML mapping.
```

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+    return dumper.represent_mapping("tag:yaml.org,2002:map", namelist_dict)
+
+
+def represent_ordereddict(dumper: yaml.Dumper, data: OrderedDict) -> yaml.nodes.MappingNode:
+    """
+    Convert ordered dict objects to dictionaries and serialize.
```
maddenp-noaa, line 74:
```suggestion
    Convert OrderedDict to dict and serialize.
```

src/uwtools/config/support.py
maddenp-noaa: If I'm not mistaken, this method is called by the YAML library to operate on a single value, in which case plural terminology would not be accurate.

src/uwtools/config/support.py
```diff
@@ -53,6 +55,36 @@ def log_and_error(msg: str) -> Exception:
...
+
+    :param dumper: The YAML dumper.
+    :param data: The ordereddict to serialize.
+    """
+    # Convert the Ordered Dict to a dictionary
+    for key, value in data.items():
+        if isinstance(value, OrderedDict):
+            data[key] = dict(value)
+
+    # Represent the dictionary as a YAML mapping
```
maddenp-noaa, line 83:
```suggestion
    # Represent the dict as a YAML mapping.
```

src/uwtools/config/formats/yaml.py
WeirAE: I had been leaving the full title so as to provide context, but honestly its unnecessary - I'll clean this up

src/uwtools/config/formats/yaml.py
WeirAE: I'm still learning this depth of YAML manipulation, but I believe there's a way to set up a group and call that - I was thinking about it again once they started stacking up, I'll look more into it

src/uwtools/config/formats/yaml.py
maddenp-noaa: There's definitely a place for keeping the explicit namespaces for context and clarity. My gut feeling is that we don't need that much specificity here, but I think you're right to keep it in mind, especially in places where there might be ambiguity.

src/uwtools/config/formats/yaml.py
WeirAE: Hm. It looks like `yaml.add_multi_representer()` doesn't do what I'd like, and a few other options just clutter space with tuples instead. I think factoring it out to another method is the answer, barring any new novel approaches

src/uwtools/config/support.py
```diff
@@ -12,6 +14,20 @@
...
 
+# Public functions
+def add_representers():
+    """
+    Add representers to the YAML dumper for custom types.
+
+    This function is called by the YAMLConfig class to add representers for the TaggedString,
+    Namelist, and OrderedDict types to the YAML dumper. The representers are used to serialize these
+    types to YAML.
+    """
```
maddenp-noaa, line 25: I'd recommend not stating in the docstring what code might call this code -- any code could call it. The docstring should only describe the behavior of _this_ function.

src/uwtools/config/support.py
```diff
@@ -12,6 +14,20 @@
 INCLUDE_TAG = "!INCLUDE"
 
 
+# Public functions
+def add_representers():
```
maddenp-noaa, line 18:
```suggestion
def add_representers() -> None:
```

src/uwtools/tests/config/test_support.py
```diff
@@ -56,6 +66,18 @@ def test_log_and_error(caplog):
...
+def test_represent_namelist():
+    namelist = reads("&namelist\n key = value\n/\n")
+    assert yaml.dump(namelist, default_flow_style=True).strip() == "{namelist: {key: value}}"
+
+
+def test_represent_ordereddict():
+    ordereddict_values = OrderedDict([("example", OrderedDict([("key", "value")]))])
+    assert (
+        yaml.dump(ordereddict_values, default_flow_style=True).strip() == "{example: {key: value}}"
+    )
```
maddenp-noaa, line 78: Nice tests.

maddenp-noaa (approved):

christinaholtNOAA (approved):
LGTM. 

Just needs a format push.

PR #424, opened 2024-02-29
Omnibus 2024-03-01

maddenp-noaa:
**Synopsis**

Several minor updates:

- Remove the `--debug` CLI switch and do not try to catch raised exceptions.
- DRY out some graph-related logic in two drivers to a support module, and update tests.
- Rename `add_representers()` -> `add_yaml_representers()` as the YAML part may not be clear from context.
- Tweak some whitespace and comments.

**Type**

**Impact**

i.e. the `--debug` switch is removed, but the behavior it provided is now the default.

**Checklist**

### DIFF
--- a/src/uwtools/api/fv3.py
+++ b/src/uwtools/api/fv3.py
@@ -5,8 +5,7 @@ import datetime as dt
 from pathlib import Path
 from typing import Dict, Optional
 
-import iotaa as _iotaa
-
+from uwtools.drivers import support
 from uwtools.drivers.fv3 import FV3
 
 
@@ -30,7 +29,7 @@ def execute(
     :param batch: Submit run to the batch system
     :param dry_run: Do not run forecast, just report what would have been done
     :param graph_file: Write Graphviz DOT output here
-    :return: True if task completes without raising an exception
+    :return: ``True`` if task completes without raising an exception
     """
     obj = FV3(config_file=config_file, cycle=cycle, batch=batch, dry_run=dry_run)
     getattr(obj, task)()
@@ -44,13 +43,11 @@ def graph() -> str:
     """
     Returns Graphviz DOT code for the most recently executed task.
     """
-    return _iotaa.graph()
+    return support.graph()
 
 
 def tasks() -> Dict[str, str]:
     """
     Returns a mapping from task names to their one-line descriptions.
     """
-    return {
-        task: getattr(FV3, task).__doc__.strip().split("\n")[0] for task in _iotaa.tasknames(FV3)
-    }
+    return support.tasks(FV3)
--- a/src/uwtools/cli.py
+++ b/src/uwtools/cli.py
@@ -43,24 +43,19 @@ def main() -> None:
     # are known, then dispatch to the [sub]mode handler.
 
     setup_logging(quiet=True)
-    try:
-        args, checks = _parse_args(sys.argv[1:])
-        for check in checks[args[STR.mode]][args[STR.action]]:
-            check(args)
-        setup_logging(quiet=args[STR.quiet], verbose=args[STR.verbose])
-        log.debug("Command: %s %s", Path(sys.argv[0]).name, " ".join(sys.argv[1:]))
-        modes = {
-            STR.config: _dispatch_config,
-            STR.fv3: _dispatch_fv3,
-            STR.rocoto: _dispatch_rocoto,
-            STR.sfcclimogen: _dispatch_sfc_climo_gen,
-            STR.template: _dispatch_template,
-        }
-        sys.exit(0 if modes[args[STR.mode]](args) else 1)
-    except Exception as e:  # pylint: disable=broad-exception-caught
-        if _switch(STR.debug) in sys.argv:
-            log.exception(str(e))
-        _abort(str(e))
+    args, checks = _parse_args(sys.argv[1:])
+    for check in checks[args[STR.mode]][args[STR.action]]:
+        check(args)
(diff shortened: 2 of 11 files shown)


src/uwtools/api/fv3.py
```diff
@@ -30,7 +29,7 @@ def execute(
     :param batch: Submit run to the batch system
     :param dry_run: Do not run forecast, just report what would have been done
     :param graph_file: Write Graphviz DOT output here
-    :return: True if task completes without raising an exception
+    :return: ``True`` if task completes without raising an exception
```
maddenp-noaa, line 32: Looks nicer in the Sphinx docs.

src/uwtools/cli.py
```diff
@@ -43,24 +43,19 @@ def main() -> None:
...
+    setup_logging(quiet=args[STR.quiet], verbose=args[STR.verbose])
+    log.debug("Command: %s %s", Path(sys.argv[0]).name, " ".join(sys.argv[1:]))
+    modes = {
+        STR.config: _dispatch_config,
+        STR.fv3: _dispatch_fv3,
+        STR.rocoto: _dispatch_rocoto,
+        STR.sfcclimogen: _dispatch_sfc_climo_gen,
+        STR.template: _dispatch_template,
+    }
+    sys.exit(0 if modes[args[STR.mode]](args) else 1)
```
maddenp-noaa, line 58: An unhandled exception will now propagate all the way up. If we find there are cases we can deal with better than this default behavior, we can catch and handle.

src/uwtools/config/support.py
```diff
@@ -15,7 +15,9 @@
 
 
 # Public functions
-def add_representers() -> None:
+
+
+def add_yaml_representers() -> None:
```
maddenp-noaa, line 20: Renamed for clarity, since this module is about more than just YAML.

src/uwtools/tests/api/test_fv3.py
```diff
@@ -25,30 +23,12 @@ def test_execute(tmp_path):
...
-    @external
-    def ready():
-        yield "ready"
-        yield asset("ready", lambda: True)
-
-    ready()
-    assert fv3.graph().startswith("digraph")
+    with patch.object(fv3.support, "graph") as graph:
+        fv3.graph()
+    graph.assert_called_once_with()
```
maddenp-noaa, line 28: This is just a plumbing test now -- the real test is in `uwtools.tests.drivers.test_support`.

src/uwtools/tests/api/test_fv3.py
```diff
@@ -25,30 +23,12 @@ def test_execute(tmp_path):
...
-        "@tasks t3"
-
-    with patch.object(fv3, "FV3") as FV3:
-        FV3.t1 = t1
-        FV3.t2 = t2
-        FV3.t3 = t3
-        assert fv3.tasks() == {"t2": "@task t2", "t3": "@tasks t3", "t1": "@external t1"}
+    with patch.object(fv3.support, "tasks") as _tasks:
+        fv3.tasks()
+    _tasks.assert_called_once_with(fv3.FV3)
```
maddenp-noaa, line 34: Similarly, just a plumbing test now. Ditto in `test_sfc_climo_gen.py`, below.

src/uwtools/tests/config/test_support.py
```diff
@@ -67,7 +68,7 @@ def test_log_and_error(caplog):
 
 
 def test_represent_namelist():
-    namelist = reads("&namelist\n key = value\n/\n")
+    namelist = f90nml.reads("&namelist\n key = value\n/\n")
```
maddenp-noaa, line 71: I think `Namelist` says enough on its own not to be `f90nml.Namelist`, but `reads` seemed generic enough to warrant an explicit namespace for clarity.

WeirAE: Thats fair, no complaints!

src/uwtools/config/support.py
```diff
@@ -66,17 +68,19 @@ def log_and_error(msg: str) -> Exception:
...
     Convert f90nml Namelist to OrderedDict and serialize.
 
     :param dumper: The YAML dumper.
     :param data: The f90nml Namelist to serialize.
     """
-    # Convert the f90nml Namelist to an OrderedDict.
-    namelist_dict = data.todict()
 
-    # Represent the OrderedDict as a YAML mapping.
+    # Convert the f90nml Namelist to an OrderedDict, then represent as a YAML mapping.
```
christinaholtNOAA, line 81: I'm curious if this could be the docstring info, and we don't need an additional comment since it's only a 2-liner.

src/uwtools/config/support.py
```diff
@@ -88,11 +92,11 @@ def _represent_ordereddict(dumper: yaml.Dumper, data: OrderedDict) -> yaml.nodes
     :param data: The OrderedDict to serialize.
     """
 
-    # Convert the OrderedDict to a dict.
+    # Recursrively convert an OrderedDict to a dict, then represent as a YAML mapping.
```
christinaholtNOAA, line 95: Same here.

WeirAE (approved):
So far all nice, clean changes

christinaholtNOAA (approved):
Looks good. Minor suggestions, but I don't feel too strongly about them.

src/uwtools/config/support.py
maddenp-noaa: Excellent idea, will do.

src/uwtools/config/support.py
maddenp-noaa: Done.

PR #425, opened 2024-03-02
UW-513 Add environment-variable support for template render

maddenp-noaa:
**Synopsis**

Previously, `uw template render` only made use of environment variables if no values file was provided, to avoid applying a potentially large number of obscure values to the render. This PR adds support for environment variables via an optional `--env` switch.

For example, given `values.yaml` with contents
```
recipient: World
```
one may now
```
% echo "{{ greeting }}{{ sep }} {{ recipient }}" | greeting=Hello uw template render --values-file values.yaml --env sep=,
Hello, World
```
So, values may now be taken simultaneously from file, CLI key=value pairs, with variables from the environment overriding those from the CLI, and those from the CLI overriding those from file.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/tools/mode_template.rst
```diff
@@ -47,14 +47,14 @@ The ``uw`` mode for handling :jinja2:`Jinja2 templates<templates>`.
...
+     --env
+         Use environment variables
      --values-needed
          Print report of values needed to render template
      --partial
          Permit partial template rendering
      --dry-run
          Only log info, making no changes
-     --debug
-         Print all log messages, plus any unhandled exception's stack trace (implies --verbose)
```
maddenp-noaa, line 57: This should have been updated in the PR that removed the `--debug` switch.

docs/sections/user_guide/cli/tools/mode_template.rst
```diff
@@ -136,35 +136,6 @@ and a YAML file called ``values.yaml`` with the following contents:
      $ uw template render --input-file template --values-file values.txt --values-format yaml
      Hello, World!
 
-* It is an error to render a template without providing all needed values. For example, with ``recipient: World`` removed from ``values.yaml``:
```
maddenp-noaa, line 139: This block moved down in the doc, under a **NB** heading that makes it more obvious that several tests that follow use a modified `values.yaml` file with a missing value.

docs/sections/user_guide/cli/tools/mode_template.rst
```diff
@@ -215,6 +186,46 @@ and a YAML file called ``values.yaml`` with the following contents:
...
+**NB**: The following examples are based on a ``values.yaml`` file with ``recipient: World`` removed.
+
+* It is an error to render a template without providing all needed values.
+
+  .. code-block:: text
+
+   $ uw template render --input-file template --values-file values.yaml
+   [2024-03-02T16:42:48]    ERROR Required value(s) not provided:
+   [2024-03-02T16:42:48]    ERROR   recipient
+   [2024-03-02T16:42:48]    ERROR Template could not be rendered.
```
maddenp-noaa, line 198: The code change appears later, but I indented the missing values to stand out better from the surrounding log messages.

docs/sections/user_guide/cli/tools/mode_template.rst
```diff
@@ -215,6 +186,46 @@ and a YAML file called ``values.yaml`` with the following contents:
...
+     $ uw template render --input-file template --values-file values.yaml recipient=Reader
+     Hello, Reader!
+
+  The optional ``-env`` switch allows environment variables to be used to supply values:
+
+  .. code-block:: text
+
+     $ export recipient=You
+     $ uw template render --input-file template --values-file values.yaml --env
+     Hello, You!
```
maddenp-noaa, line 220: Here's the new block about the `--env` switch. This example shows an explicitly `export`ed environment variable; the example below shows the use case where the environment variable is set on the command line.

docs/sections/user_guide/cli/tools/mode_template.rst
```diff
@@ -237,8 +248,6 @@ and a YAML file called ``values.yaml`` with the following contents:
      $ uw template render --input-file template --values-file values.nml
      Hello, World!
 
-  Note that ``ini`` and ``nml`` configs are, by definition, depth-2 configs, while ``sh`` configs are depth-1, and ``yaml`` configs have arbitrary depth.
```
maddenp-noaa, line 240: This line seemed out of place to me -- there's no other discussion of depth here, and it wasn't clear to me what the reader was supposed to make of this statement.

src/uwtools/api/template.py
```diff
@@ -11,11 +11,12 @@
 
 
 def render(
-    values: Union[dict, Path],
+    values_src: Optional[Union[dict, Path]] = None,
```
maddenp-noaa, line 14: I couldn't see a reason why users shouldn't be able to provide any combination of `dict`- or file-based values and/or CLI overrides and/or environment-variable values. In the extreme case, they could provide none of these, in which case rendering a template with Jinja2 expressions would be an error *except* if `--partial` is supplied, in which case the template passes through unmodified -- maybe not very useful, but who knows?

I also renamed `values` to `values_src` because I wanted to use `values` to represent the final set of values from the `dict`/file source plus overrides plus environment variables.

src/uwtools/cli.py
```diff
@@ -42,20 +42,27 @@ def main() -> None:
...
+            STR.config: _dispatch_config,
+            STR.fv3: _dispatch_fv3,
+            STR.rocoto: _dispatch_rocoto,
+            STR.sfcclimogen: _dispatch_sfc_climo_gen,
+            STR.template: _dispatch_template,
+        }
+        sys.exit(0 if modes[args[STR.mode]](args) else 1)
+    except UWError as e:
+        log.error(str(e))
+        sys.exit(1)
```
maddenp-noaa, line 65: We removed the `--debug` switch in a recent PR with the understanding that we'd handle exceptions that we recognize and can do something about, letting only exceptions from obvious causes (e.g. `ZeroDivisionError`) or obscure causes (that we might later decide to write handlers for) percolate up.

I realized while implementing this PR's `--env` feature that there are already some cases where we're raising `UWError` exceptions with helpful messages for users, which we can catch here and report without the accompanying stacktrace. For example, on `main` at the moment:
```
% echo "{{ x }}" | uw template render --values-file values.txt
  File "/home/maddenp/git/uwtools/src/uwtools/utils/file.py", line 103, in get_file_format
    return FORMAT.formats()[suffix]
           ~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'txt'

The above exception was the direct cause of the following exception:

  File "/home/maddenp/sw/conda/envs/DEV-uwtools/bin/uw", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/maddenp/git/uwtools/src/uwtools/cli.py", line 48, in main
    check(args)
  File "/home/maddenp/git/uwtools/src/uwtools/cli.py", line 784, in _check_template_render_vals_args
    args[STR.valsfmt] = get_file_format(args[STR.valsfile])
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maddenp/git/uwtools/src/uwtools/utils/file.py", line 107, in get_file_format
    raise UWError(msg) from e
uwtools.exceptions.UWError: Cannot deduce format of 'values.txt' from unknown extension 'txt'
```
Kind of a mess. But since this is _our_ error, and we know we're providing a useful text value in the exception, we'd probably prefer (as implemented in this PR):
```
% echo "{{ x }}" | uw template render --values-file values.txt
Cannot deduce format of 'values.txt' from unknown extension 'txt'
```
So, it seems to me that a good convention might be to catch and handle `UWError` exceptions (which includes subclasses, though we might sometimes catch and handle those further down the call stack) here, letting non-UW exceptions through.

src/uwtools/cli.py
```diff
@@ -464,18 +472,20 @@ def _dispatch_template_render(args: Args) -> bool:
...
+            env=args[STR.env],
             values_needed=args[STR.valsneeded],
             partial=args[STR.partial],
             dry_run=args[STR.dryrun],
         )
     except UWTemplateRenderError:
-        msg = "Template could not be rendered. Try with %s for details." % _switch(STR.valsneeded)
-        log.error(msg)
+        if args[STR.valsneeded]:
+            return True
```
maddenp-noaa, line 487: Currently on `main`:
```
% echo "{{ x }}" | uw template render --values-needed
[2024-03-02T17:35:56]     INFO Value(s) needed to render this template are:
[2024-03-02T17:35:56]     INFO x
[2024-03-02T17:35:56]    ERROR Template could not be rendered. Try with --values-needed for details.
% echo $?
1
```
But this is not an error: The user specified `--values-needed` to get a report and is not asking for rendering to happen.

As of this PR:
```
% echo "{{ x }}" | uw template render --values-needed
[2024-03-02T17:39:21]     INFO Value(s) needed to render this template are:
[2024-03-02T17:39:21]     INFO   x
% echo $?
0
```

src/uwtools/cli.py
```diff
@@ -704,7 +722,7 @@ def _add_arg_verbose(group: Group) -> None:
 # Support
 
 
-def _abort(msg: str) -> None:
+def _abort(msg: str) -> NoReturn:
```
maddenp-noaa, line 725: The `NoReturn` type signals to typecheckers that this function will not return at all. It prevents e.g. "may not be defined" warnings about variables accessed after the call to this function: The typechecker knows that it doesn't matter that they may not be defined, because they will never be references, because the call to this function will never return.

src/uwtools/config/jinja2.py
```diff
@@ -327,32 +328,35 @@ def _report(args: dict) -> None:
...
-        values = dict(os.environ)  # Do not modify os.environ: Make a copy.
-        log.debug("Initial values taken from environment")
+        values = values_src or {}
     if overrides:
         values.update(overrides)
-        log.debug("Updated values with overrides: %s", " ".join(overrides))
+        log.debug("Supplemented values with overrides: %s", " ".join(overrides))
+    if env:
+        values.update(os.environ)
+        log.debug("Supplemented values with environment variables")
```
maddenp-noaa, line 359: Here's the updated logic to compose `dict`- or file-based values with CLI overrides and/or environment variables. The priority of each source derives from the order of the `update()` calls here.

elcarpenterNOAA (approved):

docs/sections/user_guide/cli/tools/mode_template.rst
WeirAE: We do need to explain something about the depth limits in the documentation, but it is awkward here. I think we still need it, but it should move elsewhere and rephrased

docs/sections/user_guide/cli/tools/mode_template.rst
WeirAE: Oof, thanks for the fix

WeirAE (approved):
Agreed with all the additional changes, great work

PR #427, opened 2024-03-08
UW 517: Adding a chgres_cube driver and schema

christinaholtNOAA:
**Synopsis**

Adds a driver for chgres_cube, along with a JSON schema files and tests.

Additional work includes:
* Several common-use methods were factored up into the base class. 
* Strings from the cli and file format were factored into a strings module for more common use.
* Run scripts were renamed to include the driver name to help with concurrency when tasks use the same run directory.

Documentation will come in a separate PR.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


src/uwtools/strings.py
```diff
@@ -0,0 +1,105 @@
...
+@dataclass(frozen=True)
+class STR:
+    """
+    A lookup map for CLI-related strings.
+    """
+
+    action: str = "action"
+    batch: str = "batch"
+    cfgfile: str = "config_file"
+    chgrescube: str = "chgres_cube"
```
christinaholtNOAA, line 67: This is the only one added. Everything else was copied from the original location.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
```diff
@@ -0,0 +1,205 @@
...
+      },
+      "required": [
+        "execution",
+        "run_dir"
+      ],
+      "type": "object"
+    }
+  },
+  "type": "object"
+}
```
maddenp-noaa, line 205: Maybe it's still WIP, or maybe happening in @WeirAE 's branch/PR, but it would be valuable to have some granular test for this schema in `uwtools.tests.test_schemas`. Without those it's IMO hard to tell if the schema really does what's intended, as it's hard to test so many possible combinations via unit tests of other code. It might not need much, but I see now that the one I wrote for  `sfc_climo_gen` could probably have somewhat better coverage: There's probably no sense in testing that something like
```
"input_vegetation_type_file": {
  "type": "string"                                                                                                                                                                                                   
}
```
doesn't accept, say, an `int` value, but possibly something like
```
maximum_snow_albedo_method": {
  "enum": [                                                                                                                                                                                                          
    "bilinear",                                                                                                                                                                                                      
    "conservative"                                                                                                                                                                                                   
  ]                                                                                                                                                                                                                  
}
```
could use tests against misspellings. It's hard to know where to draw the line, but today I feel like I could have added a few more tests of `sfc_climo_gen.jsonschema`. `chgres-cube.jsonschema` probably deserves at least a block in `test_schemas.py` to make it obvious that we didn't just overlook it.

src/uwtools/api/chgres_cube.py
```diff
@@ -0,0 +1,57 @@
+"""
+API access to the uwtools chgres_cube driver.
+"""
+import datetime as dt
+from pathlib import Path
+from typing import Dict, Optional
+
+import iotaa as _iotaa
+
+from uwtools.drivers.chgres_cube import ChgresCube
```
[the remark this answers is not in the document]
christinaholtNOAA: I could handle that here. It's worth doing for privacy and consistency.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -0,0 +1,133 @@
...
+        yield asset(path, path.is_file)
+        vals = self._driver_config["namelist"]["update_values"]["config"]
+        inputs = [
+            ("data_dir_input_grid", "atm_files_input_grid"),
+            ("data_dir_input_grid", "grib2_file_input_grid"),
+            ("data_dir_input_grid", "sfc_files_input_grid"),
+            "mosaic_file_target_grid",
+            "varmap_file",
+            "vcoord_file_target_grid",
+        ]
```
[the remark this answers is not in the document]
christinaholtNOAA: I double-checked with the Fortran namelist interface and saw that these may actually be defined as arrays:
```
 character(len=500), public      :: sfc_files_input_grid(6) = "NULL" !<  File names containing input surface data. Not used for 'grib2' input type.
```

They are not treated as such in SRW, with only a single string provided. Is that a bug in SRW? Is Fortran okay with that?? Here's one example of how the value is set from SRW:

```
fn_sfc="gfs.t${hh}z.sfcanl.nemsio"
...
config:
 sfc_files_input_grid: ${fn_sfc},
```
Now, that could also be my rusty Fortran memory for how to define a character array. Perhaps this "lazy" syntax is fine given how Fortran manages memory.

Either way, this is going to cause an issue with the way I've prepared the JSON Schema. I believe I've used strings for all of them. I'll need to update to "string or array" if Fortran is actually okay with both.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -0,0 +1,133 @@
...
+        input_paths = []
+        for item in inputs:
+            if isinstance(item, str):
+                input_paths += [Path(vals[item])]
+            else:
+                input_paths += [Path(vals[item[0]]) / vals[item[1]]]
```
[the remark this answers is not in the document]
christinaholtNOAA: I am giving `config_files` a try here since we're really looking for just those config settings. Otherwise, I think `update_vals` might be okay. Thoughts?

src/uwtools/drivers/chgres_cube.py
```diff
@@ -0,0 +1,133 @@
...
+        envcmds = self._driver_config.get("execution", {}).get("envcmds", [])
+        execution = [self._runcmd, "test $? -eq 0 && touch %s/done" % self._rundir]
+        scheduler = self._scheduler if self._batch else None
+        path.parent.mkdir(parents=True, exist_ok=True)
+        rs = self._runscript(
+            envcmds=envcmds, envvars=envvars, execution=execution, scheduler=scheduler
+        )
+        with open(path, "w", encoding="utf-8") as f:
+            print(rs, file=f)
+        os.chmod(path, os.stat(path).st_mode | stat.S_IEXEC)
```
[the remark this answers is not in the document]
christinaholtNOAA: I think it's likely a good idea, but I'd still like to be able to clearly understand what's going into the run script.

src/uwtools/drivers/chgres_cube.py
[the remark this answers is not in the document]
maddenp-noaa: I like `config_files`.

src/uwtools/drivers/chgres_cube.py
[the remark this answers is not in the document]
maddenp-noaa: I think Fortran would expect a namelist entry like
```
sfc_files_input_grid = "file1", "file2" ! etc.
```
but
```
sfc_files_input_grid = "file1"
```
should be fine. (I'm not sure whether the comma you show is required.) I think it would be an error to have
```
sfc_files_input_grid = "file1,file2"
```
so probably our schema and config should accommodate a YAML sequence / Python `dict` so that `f90nml` can serialize it correctly.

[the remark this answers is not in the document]
christinaholtNOAA: Oh! No, that comma is related to how the config file is prepared, and not the output in the namelist.

And just to clarify, did you mean "a YAML sequence / Python `list`"? You have `dict` there.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -0,0 +1,213 @@
...
+            "atm_files_input_grid": src,
+            "grib2_file_input_grid": src,
+            "sfc_files_input_grid": src,
+            "mosaic_file_target_grid": src,
+            "varmap_file": src,
+            "vcoord_file_target_grid": src,
+        }
+        driverobj.namelist_file()
+    assert dst.is_file()
+    assert isinstance(f90nml.read(dst), f90nml.Namelist)
```
maddenp-noaa, line 117: I'm wondering why this test defines `src` to be an (almost?) real file and modifies the driver object's in-memory config to refer to that file as an input file, given that it also mocks the `file` task to always return an asset that is ready. I looked back at the similar test for `sfc_climo_gen` and see that it doesn't modify the config vs what's set in the fixture, which is fine because the `@external` `file` tasks that the driver wraps around each input file are mocked to always be ready, so the `namelist_file` task believes it has all its requirements satisfied and so can create the namelist file, which the final two `assert` statements test. So, I don't immediately see why we need to define `src` or modify the config if we're mocking the `file` task.

Alternatively -- and it would be a nice and strong test -- we could actually create `src` (we might then also need to receive a `tmp_path` fixture and set `driverobj._rundir` to that before `.touch()`ing `src`) and then *not* mock `file`, which would let all those `file` `@external` tasks actually verify the existence of `src`.

Please let me know if I'm missing something, but at first glance it seems to me that there's no reason to interfere with the config since those original fixture `/path/to/file` values should be fine given that they're not _really_ checked.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def test_ChgresCube_provisioned_run_directory(driverobj):
+    with patch.multiple(
+        driverobj,
+        namelist_file=D,
+        runscript=D,
```
maddenp-noaa, line 124: nitpick: I think if we remove the trailing comma on this line, `make format` will turn it into a single line again. It doesn't look long enough to need wrapping. `black` will fix it if that comma goes away.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -0,0 +1,213 @@
...
+    runscript = driverobj._runscript_path
+    with patch.object(driverobj, "provisioned_run_directory") as prd:
+        with patch.object(
+            chgres_cube.ChgresCube, "_scheduler", new_callable=PropertyMock
+        ) as scheduler:
+            driverobj._run_via_batch_submission()
+            scheduler().submit_job.assert_called_once_with(
+                runscript=runscript, submit_file=Path(f"{runscript}.submit")
+            )
+        prd.assert_called_once_with()
```
maddenp-noaa, line 171: Now that `_run_via_batch_submission()` and `_run_via_local_execution()` (below) are part of the `Driver` base class, could we consider having just one copy of the test methods, in `uwtools.tests.drivers.test_driver`, and remove the (I think) duplicates in each of the `Driver` subclass test modules? Since the `ConcreteDriver` class defined in `test_driver.py` inherits those `_run_*()` methods, it seems like it should be possible to test them there, once and for all `Driver`s. Then we could potentially remove any duplicates.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
maddenp-noaa, line 205: I'd still very much like to see at least a minimal (perhaps following the one for `sfc-climo-gen.jsonschema`) test or set of tests for this schema, whether in this PR or in a more-or-less-immediate follow-on.

src/uwtools/tests/drivers/test_chgres_cube.py
WeirAE: Hmm, I made this to mirror the test in `sfc_climo_gen`, which didn't mind, but` chgres_cube` was checking for these entries and failing, thats the only reason I added them. I'm using `src` as a generic placeholder here like in other tests, but that would be a stronger test

WeirAE: Gotcha - I've changed them to `/path/to/file` and I'd like to update both `test_chgres_cube` and `test_sfc_climo_gen` with `file` `@external` task tests along with the granular tests in `test_schemas` as a followup

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -147,6 +161,29 @@ def test_Driver__runscript_execution_only(driver_good):
...
+def test_Driver__run_via_local_execution(driver_good):
+    with patch.object(driver_good, "provisioned_run_directory") as prd:
+        with patch.object(driver, "execute") as execute:
+            driver_good._run_via_local_execution()
+            execute.assert_called_once_with(
+                cmd="{x} >{x}.out 2>&1".format(x=driver_good._runscript_path),
+                cwd=driver_good._rundir,
+                log_output=True,
+            )
+        prd.assert_called_once_with()
```
maddenp-noaa, line 184: Thanks for factoring these out!

src/uwtools/tests/drivers/test_chgres_cube.py
maddenp-noaa: I checked out your branch, moved _only_ the `sfc_files_input_grid`, `varmap_file`, and `vcoord_file_target_grid` key-value pairs from the `dict` contained in lines 104-112 to the top-level definition of `config` starting on line 16, then removed the 104-112 block and all the test pass. The rest of the keys in the 104-112 block seem to have already been defined in the top-level `config`, which I believe is what `driverobj` is instantiated with. That seems DRYer. Alternatively, if there were some reason the three missing keys couldn't be relocated to the top-level `config`, we could call a `.update()` on the `driverobj` config to just add those keys without repetition. I hope I'm not missing something about the intent of the test.

src/uwtools/drivers/chgres_cube.py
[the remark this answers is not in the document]
maddenp-noaa: I fixed my comment, thanks.

Given that YAML, I'd expect to see e.g.
```
&config
    sfc_files_input_grid = "gfs.t12z.sfcanl.nemsio,"
```
in the namelist, if that's how that YAML relates to namelist content. That looks wrong, if so. Is that not what appears? I'd expect that comma to be part of the `str` value of `sfc_files_input_grid` in any case.

src/uwtools/tests/drivers/test_chgres_cube.py
WeirAE: Fair, I think I may have just tunnel-visioned on this one - I've gotten those moved and thats a better way for sure.

maddenp-cu (approved):
I think I understood that we'll do the schema unit tests in a follow-on PR and, if that's right, this looks good-to-go to me.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
WeirAE: Yep, I'm going to just do both schema tests as a follow-on

src/uwtools/drivers/chgres_cube.py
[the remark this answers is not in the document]
christinaholtNOAA: The comma does not show up in namelist comment, so I think it should accept a `string` or `array`.

PR #429, opened 2024-03-14
UW-515 File Copy/Link Tool

maddenp-noaa:
**Synopsis**

Add a file copy/link tool to `uwtools`.

**Type**

**Impact**

**Checklist**


docs/sections/user_guide/cli/drivers/fv3.rst
```diff
@@ -1,5 +1,5 @@
-Mode ``fv3``
-============
+``fv3``
```
maddenp-noaa, line 1: I think the "mode" part looks a little odd and unnecessary now that we've split the docs into `Drivers` and `Tools` sections.

docs/sections/user_guide/cli/tools/file.rst
```diff
@@ -0,0 +1,174 @@
+``file``
```
maddenp-noaa, line 1: Docs for the new tool.

docs/sections/user_guide/yaml/components/fv3.rst
```diff
@@ -84,7 +84,7 @@ Supports ``base_file:`` and ``update_values:`` blocks (see the :ref:`updating_va
 files_to_copy:
 ^^^^^^^^^^^^^^
 
-Defines files to be copied to the run directory. Keys in the ``files_to_copy:`` YAML map specify destination paths relative to the run directory, and values specify source paths. Both keys and values may contain Jinja2 variables/expressions using a ``cycle`` variable, which is a Python ``datetime`` object corresponding to the FV3 cycle being run. This supports specification of cycle-specific filenames/paths. For example, a key-value pair
+See :ref:`this page <files_yaml>` for details. For the ``fv3`` driver, both keys and values may contain Jinja2 variables/expressions using a ``cycle`` variable, which is a Python ``datetime`` object corresponding to the FV3 cycle being run. This supports specification of cycle-specific filenames/paths. For example, a key-value pair
```
maddenp-noaa, line 87: Since YAML blocks with these semantics are now used both by the `fv3` driver and by the `file` tool (and will undoubtedly be used for other purposes in the future), I've factored the description out into its own file and have linked to it where needed.

format
```diff
@@ -7,7 +7,7 @@ echo "=> Running isort"
 isort -q src
 
 echo "=> Running docformatter"
-(cd src && docformatter . || test $$? -eq 3)
+(cd src && docformatter . || test $? -eq 3)
```
maddenp-noaa, line 10: Fixed a latent bug from when this code lived in the `Makefile`.

src/uwtools/config/validator.py
```diff
@@ -11,12 +11,31 @@
...
+    :param config: The config to validate.
+    :param schema_name: Name of uwtools schema to validate the config against.
+    :raises: UWConfigError if config fails validation.
+    """
+
+    log.info("Validating config against internal schema %s", schema_name)
+    schema_file = resource_path("jsonschema") / f"{schema_name}.jsonschema"
+    log.debug("Using schema file: %s", schema_file)
+    if not validate_yaml(config=config, schema_file=schema_file):
+        raise UWConfigError("YAML validation errors")
```
maddenp-noaa, line 36: I factored this out of `uwtools.drivers.driver` (it was `_validate_one()` there), as it is now also needed for the `file` tool, to validate file dst/src blocks.

src/uwtools/drivers/driver.py
```diff
@@ -208,18 +206,4 @@ def _validate(self) -> None:
...
-
-        :param schema_name: Name of uwtools schema to validate the config against.
-        :raises: UWConfigError if config fails validation.
-        """
-
-        log.info("Validating config per schema %s", schema_name)
-        schema_file = resource_path("jsonschema") / f"{schema_name}.jsonschema"
-        log.debug("Using schema file: %s", schema_file)
-        if not validator.validate_yaml(config=self._config, schema_file=schema_file):
-            raise UWConfigError("YAML validation errors")
```
maddenp-noaa, line 225: Moved (as `validate_internal()` to `uwtools.config.validator`.

src/uwtools/tests/test_cli.py
```diff
@@ -588,19 +652,3 @@ def test__parse_args():
...
-    # Return actions (named subparsers) belonging to the given parser.
-    if actions := [x for x in parser._actions if isinstance(x, _SubParsersAction)]:
-        return list(actions[0].choices.keys())
-    return []
-
-
-@fixture
-def subparsers():
-    # Create and return a subparsers test object.
-    return Parser().add_subparsers()
```
maddenp-noaa, line 606: Moved to top of module and split into `Helpers` and `Fixtures` sections, for consistency with other test modules.

docs/sections/user_guide/api/index.rst
```diff
@@ -3,6 +3,7 @@ API
```
christinaholtNOAA, line 3: Does it make sense to organize this section the same as the CLI, using headings for Drivers and Tools?

docs/sections/user_guide/cli/tools/file.rst
```diff
@@ -0,0 +1,174 @@
...
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/foo -> /tmp/target/foo: Checking requirements
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/foo -> /tmp/target/foo: Requirement(s) ready
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/foo -> /tmp/target/foo: Executing
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/foo -> /tmp/target/foo: Final state: Ready
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/bar -> /tmp/target/subdir/bar: Initial state: Pending
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/bar -> /tmp/target/subdir/bar: Checking requirements
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/bar -> /tmp/target/subdir/bar: Requirement(s) ready
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/bar -> /tmp/target/subdir/bar: Executing
+   [2024-03-14T19:00:02]     INFO Copy /tmp/source/bar -> /tmp/target/subdir/bar: Final state: Ready
+   [2024-03-14T19:00:02]     INFO File copies: Final state: Ready
```
christinaholtNOAA, line 86: This is such nice logging! I'm enjoying every new tool that has this feature.

src/uwtools/api/file.py
```diff
@@ -0,0 +1,46 @@
...
+    Copy files.
+
+    :param target_dir: Path to target directory
+    :param config_file: Path to YAML config file (defaults to stdin)
+    :param keys: YAML keys leading to file dst/src block
+    :param dry_run: Do not copy files
+    :return: True if no exception is raised
+    """
+    _FileCopier(target_dir=target_dir, config_file=config_file, keys=keys, dry_run=dry_run).go()
+    return True
```
christinaholtNOAA, line 27: Do you think it might be worth adding a dict as an input option for the API? If so, I think it would make sense for the caller to pass the appropriate block. This could also be added as a feature later.

src/uwtools/cli.py
```diff
@@ -269,6 +272,99 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+    """
+    parser = _add_subparser(subparsers, STR.link, "Link files")
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_target_dir(required, required=True)
+    optional = _basic_setup(parser)
+    _add_arg_config_file(group=optional, required=False)
+    _add_arg_dry_run(optional)
+    checks = _add_args_verbosity(optional)
+    _add_arg_keys(optional)
+    return checks
```
christinaholtNOAA, line 324: The previous two functions feel pretty WET.

christinaholtNOAA (commented):
This looks great. I have a few questions/suggestions below.

I was also a little surprised not to have seen any edits to the fv3 driver code. Do you think it may be worthwhile at some point to call the new `FileCopier` and `FileLinker` tasks from the drivers? Perhaps we want to wait until there are 3 instances again, like we did with the driver development refactoring? Maybe there's something I'm missing about why we _wouldn't_ want to use these standalone tasks from the drivers?

docs/sections/user_guide/cli/tools/file.rst
maddenp-noaa: We get it for free when we take the graph approach, so any work that fits that approach will benefit.

src/uwtools/api/file.py
maddenp-noaa: Good idea. I'll either implement it in this PR or add a backlog task for it.

src/uwtools/cli.py
maddenp-noaa: I'll DRY those out a bit.

maddenp-noaa:
@christinaholtNOAA `fv3` didn't change here because there is already a lot of DRY reuse between `fv3` and `file`. Both use the same schema for validating their file blocks, and both delegate to the `filecopy` and `symlink` tasks from `uwtools.utils.tasks`. Due to the task reuse, methods `FV3.files_copied()` and `FV3.files_linked()` each comprise only two Python statements, so there's essentially nothing to factor out -- especially given that one of those two statements is a "branded" logging function, i.e. one that contains information about the fact that this is the FV3 driver, and that includes cycle information. The task names (and therefore log messages) in the `file` tool have to be more generic, since we don't know what they're being used for. So, given that we're getting IMO helpful specificity in log messages and task names for FV3, and since there's basically no duplication, I don't think this is a good candidate for combination: I think a refactor would be more lines of code, and would involve passing extra arguments to try configure something generic.

christinaholtNOAA:
Thanks for your thoughts on the reuse. I agree with where you landed.

src/uwtools/api/file.py
christinaholtNOAA: Either way is fine by me.

docs/sections/user_guide/api/index.rst
christinaholtNOAA: We might not be on the same page here. I was thinking it might look nice to have something like this show up on the index page:


```
Drivers
=====
.. toctree::
chgres_cube
fv3
...

Tools
====
.. toctree::
config
file
logging
...
```

maddenp-noaa: Sorry, you're right, I was thinking of the `yaml` section.

I think maybe, if we want a hierarchy, it should first be present in the package namespaces -- e.g. `uwtools.api.config` -> `uwtools.api.tools.config` and `uwtools.api.fv3` -> `uwtools.api.drivers.fv3` -- after which we may or may not need different doc sections if items are already grouped together usefully (which I agree that at the moment they are not) and not overwhelming in number (which they are not). Maybe we should refine the `uwtools.api` namespace first? I'd be happy to do that in a quick follow-on PR.

christinaholtNOAA: I think that sounds like a solid path forward. Thanks for humoring me on this one. :)

christinaholtNOAA (approved):
LGTM!

docs/sections/user_guide/api/index.rst
maddenp-noaa: I do see two potential downsides:
1. It will be a breaking change for API users. We probably have only one at the moment, so do you think it will be feasible to update the SRW `uwtools` `import` statements easily when it eventually updates to use a newer `uwtools` version?
2. This would also break what has been a nice 1:1 mapping between CLI modes and API modules (`uw config` -> `uwtools.api.config`, `uw fv3` -> `uwtools.api.fv3`, etc.). But it might also be that the CLI will become similarly crowded, with tools and drivers intermixed in an unhelpful way. Would we ever consider an additional CLI level, like `uw tool config ...`, `uw driver fv3 ...`, etc.? I don't love the way that looks, but looking at `uw --help` today (and thinking of all the drivers we're going to add) isn't necessarily great, either.
Can we table this for the moment and discuss it in our group session tomorrow?

src/uwtools/api/file.py
maddenp-noaa: I implemented the `dict` option for configuring the file copy/link tool via the API in this PR.

PR #432, opened 2024-03-19
UW-518: Add schema tests for new drivers

WeirAE:
**Synopsis**

Continuation of [UW-518](https://github.com/christinaholtNOAA/workflow-tools/pull/1)
Adding further testing of sfc_climo_gen and chgres_cube schemas. Tests kept consistent with those of fv3

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/test_schemas.py
+++ b/src/uwtools/tests/test_schemas.py
@@ -307,7 +307,7 @@ def test_schema_fv3_namelist_update_values(fv3_fcstprop):
     )
     # At least one namelist entry is required:
     assert "does not have enough properties" in errors({})
-    # At least one val/var pair are required:
+    # At least one val/var pair is required:
     assert "does not have enough properties" in errors({"nml": {}})
 
 
--- a/src/uwtools/tests/test_schemas.py
+++ b/src/uwtools/tests/test_schemas.py
@@ -9,6 +9,97 @@ from pytest import fixture
 
 from uwtools.tests.support import schema_validator, with_del, with_set
 
+# fixtures
+
+
+@fixture
+def chgres_cube_prop():
+    return partial(schema_validator, "chgres-cube", "properties", "chgres_cube", "properties")
+
+
+@fixture
+def fv3_field_table_vals():
+    return (
+        {
+            "foo": {
+                "longname": "foofoo",
+                "profile_type": {"name": "fixed", "surface_value": 1},
+                "units": "cubits",
+            }
+        },
+        {
+            "bar": {
+                "longname": "barbar",
+                "profile_type": {"name": "profile", "surface_value": 2, "top_value": 3},
+                "units": "rods",
+            }
+        },
+    )
+
+
+@fixture
+def fv3_prop():
+    return partial(schema_validator, "fv3", "properties", "fv3", "properties")
+
+
+@fixture
+def sfc_climo_gen_prop():
+    return partial(schema_validator, "sfc-climo-gen", "properties", "sfc_climo_gen", "properties")
+
+
+# chgres-cube
+
+
+def test_schema_chgres_cube():
+    d = {
+        "execution": {"executable": "chgres_cube"},
+        "run_dir": "/tmp",
+    }
+    errors = schema_validator("chgres-cube", "properties", "chgres_cube")
+    # Basic correctness:
+    assert not errors(d)
+    # Some top-level keys are required:
+    for key in ("execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(d, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**d, "foo": "bar"})
+
+
+def test_schema_chgres_cube_namelist(chgres_cube_prop):
+    base_file = {"base_file": "/some/path"}
+    update_values = {"update_values": {"config": {"var": "val"}}}
+    errors = chgres_cube_prop("namelist")
+    # Just base_file is ok:
+    assert not errors(base_file)
+    # base_file must be a string:
+    assert "88 is not of type 'string'" in errors({"base_file": 88})
+    # Just update_values is ok:
+    assert not errors(update_values)
+    # config is required with update_values:
+    assert "'config' is a required property" in errors({"update_values": {}})
+    # A combination of base_file and update_values is ok:
+    assert not errors({**base_file, **update_values})
+    # At least one is required:
+    assert "is not valid" in errors({})
+
+
+def test_schema_chgres_cube_namelist_update_values(chgres_cube_prop):
+    errors = chgres_cube_prop("namelist", "properties", "update_values", "properties", "config")
+    # array, boolean, number, and string values are ok:
+    assert not errors({"array": [1, 2, 3], "bool": True, "int": 88, "float": 3.14, "string": "foo"})
+    # Other types are not, e.g.:
+    assert "None is not of type 'array', 'boolean', 'number', 'string'" in errors({"null": None})
+    # No minimum number of entries is required:
+    assert not errors({})
+
+
+def test_schema_chgres_cube_run_dir(chgres_cube_prop):
+    errors = chgres_cube_prop("run_dir")
+    # Must be a string:
+    assert not errors("/some/path")
+    assert "88 is not of type 'string'" in errors(88)
+
+
 # execution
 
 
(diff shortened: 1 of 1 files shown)


src/uwtools/tests/test_schemas.py
```diff
@@ -456,10 +455,116 @@ def test_schema_rocoto_workflow_cycledef():
 # sfc-climo-gen
 
 
+@fixture
+def sfc_climo_gen_fcstprop():
```
maddenp-noaa, line 459: I'm not sure about the `fcstproop` part of this name: I'm guessing it came from the `fv3_fcstprop` fixture, but I'm trying to recall why even that was a good name: Why not just `fv3_prop`, for example? I wonder if that name evolved from when the `fv3` driver was just called `forecast`... Anyway, unless it makes some sense to you, I don't see why we'd apply the term `fcst` to `sfc_climo_gen`, which has nothing to do with forecasting.

Similarly for `chgres_cube_fcstprop()`.

src/uwtools/tests/test_schemas.py
```diff
@@ -456,10 +455,116 @@ def test_schema_rocoto_workflow_cycledef():
...
+
+
+def test_schema_sfc_climo_gen_run_dir(sfc_climo_gen_fcstprop):
+    errors = sfc_climo_gen_fcstprop("run_dir")
+    # Must be a string:
+    assert not errors("/some/path")
+    assert "88 is not of type 'string'" in errors(88)
+
+
+# chgres-cube
```
maddenp-noaa, line 515: It might be nice to have this block nearer the top, as the existing blocks are all currently in order by component name.

src/uwtools/tests/test_schemas.py
WeirAE: You know, thats a good point - I should be making a habit of questioning existing objects. I'll ponder if there's still a better name

src/uwtools/tests/test_schemas.py
WeirAE: Ah, I left them grouped for my work efficiency and never put them back up!

src/uwtools/tests/test_schemas.py
```diff
@@ -456,10 +455,116 @@ def test_schema_rocoto_workflow_cycledef():
 # sfc-climo-gen
 
 
+@fixture
+def sfc_climo_gen_fcstprop():
+    return partial(schema_validator, "sfc-climo-gen", "properties", "sfc_climo_gen", "properties")
```
NaureenBharwaniNOAA, line 460: I'm confused here by "sfc-climo-gen" and "properties" showing up twice when being passed to `partial`. 

I looked up `partial` as a python function that accepts arguments and it looks to me through some articles online that not all the arguments being passed to the call have a fixed result yet (still in the process of collecting the value for the variable)?

I think these values are potential placeholders for the call to `schema_validator` and just wanted some clarification on it.

src/uwtools/tests/test_schemas.py
```diff
@@ -456,10 +455,116 @@ def test_schema_rocoto_workflow_cycledef():
...
+
+def test_schema_sfc_climo_gen_namelist_update_values(sfc_climo_gen_fcstprop):
+    errors = sfc_climo_gen_fcstprop(
+        "namelist", "properties", "update_values", "properties", "config"
+    )
+    # array, boolean, number, and string values are ok:
+    assert not errors({"array": [1, 2, 3], "bool": True, "int": 88, "float": 3.14, "string": "foo"})
+    # Other types are not, e.g.:
+    assert "None is not of type 'array', 'boolean', 'number', 'string'" in errors({"null": None})
+    # No minimum number of entries is required:
```
WeirAE, line 504: I did change this test from the fv3 version as both sfc_climo_gen and chgres_cube schemas have no minimum number of entries. Additionally, allowing additional_properties, while necessary, makes validating known entries for spelling and case...involved

src/uwtools/tests/test_schemas.py
WeirAE: That was my understanding of `partial`, basically leaving "properties" as a preset for later additions to not have to specify it

src/uwtools/tests/test_schemas.py
WeirAE: Fixed component name sorting and moved all fixtures to the top

src/uwtools/tests/test_schemas.py
WeirAE: I went and changed them all to `prop` , no need to make it any more complicated than necessary

src/uwtools/tests/test_schemas.py
maddenp-noaa: Sounds right. With additional properties enabled, the only thing I thought of to do with known entries is to test that a value of the wrong type is rejected, and that the error message includes the correct type(s).

src/uwtools/tests/test_schemas.py
maddenp-noaa: Python's `partial()` is an implementation of the functional programming concept called [partial application](https://en.wikipedia.org/wiki/Partial_application), which turns a function of n arguments into a function of n-m arguments by providing a fixed value for the m arguments. The benefit is that you get a new function with a (hopefully) more specific name that you can call without repeatedly supplying the same fixed argument.

The arguments to `partial()` are 1. The name of the n-argument function that is being simplified, and 2. The m arguments that we want to supply fixed values for.

In this case, we're partially applying (also sometimes called [currying](https://en.wikipedia.org/wiki/Currying) after the mathematician Haskell Curry -- after whom the programming language Haskell is also named -- though I think there are some subtle technical differences between partial application and currying) _all_ the arguments of the `schema_validator` function: First we supply the base name of the `.jsonschema` file to use for validation (which is why we see `sfc-climo-gen` with hyphens here, which will be used to locate the file `sfc-climo-gen.jsonschema`), then we supply a series of JSON keys used to navigate from the top of the schema to the block that we're interested in testing -- that's why we see `sfc_climo_gen` with underscores -- that's the first JSON key in `sfc-climo-gen.jsonschema`).

So, this whole function returns a _new function_ that is a schema validator for a specific part of the whole `sfc_climo_gen` schema. It might help to have a look at the `schema_validator()` function, too.

src/uwtools/tests/test_schemas.py
```diff
@@ -7,9 +7,99 @@
 
 from pytest import fixture
 
-from uwtools.tests.drivers import test_sfc_climo_gen
 from uwtools.tests.support import schema_validator, with_del, with_set
 
+# fixtures
```
maddenp-noaa, line 12:
```suggestion
# Fixtures
```

src/uwtools/tests/test_schemas.py
NaureenBharwaniNOAA: Thanks for the information, this definitely helps clarify the use of `partial` and what exactly is happening on this line.

I previously took a look at the `schema_validator()` function which helped with some understanding. I think the big connection I was missing here was the subtle difference between `sfc-climo-gen` and `sfc_climo_gen`. This makes more sense that it's looking for a specific key and navigating further into the schema with the underscores vs. the other use case with hyphens.

src/uwtools/tests/test_schemas.py
WeirAE: @christinaholtNOAA the individual drivers also run validation against example dicts similar to the YAML in the docs. I can store those YAML examples in `fixtures` and change the tests to that if you would prefer

maddenp-noaa:
@christinaholtNOAA I don't think we should be testing big blocks of YAML in these tests. The goal is to drill down to very small sub-schemas and test them with the most specific good and bad values feasible.

One thing I think we potentially can do is targeted tests of known namelist values. For example, `atm_core_files_input_grid` is declared in the schema as taking either array or string values, so that could be tested. If we tried to provide, say, an integer value with for that key, we should get an error message stating that it requires a string or array, which we could make an assertion about.

maddenp-noaa (approved):
Nice work.

christinaholtNOAA (commented):
@maddenp-noaa I think that is getting pretty close to testing JSON Schema itself. 

I do think that if we can run our docs YAML through tests though, it would at least check that we haven't done something inconsistent with docs and Schema updates.

christinaholtNOAA (approved):
Sounds good. Thanks for the clarification and future plans.

PR #433, opened 2024-03-20
UW-532: Ungrib driver

elcarpenterNOAA:
**Synopsis**
`ungrib` driver for MPAS app. 

The driver ran successfully on Jet and the run directory can be found here: `/lfs4/BMC/wrfruc/elcarpenter/uwtools/workflow-tools/run`. The `config.yaml` file can be found in `workflow-tools` which has contains the paths to `ungrib.exe`, and the GFS and Vtable files. 

Documentation and the schema still need to be completed, but the JSONSchema has been started. 

**Type**

- [ ] Documentation

**Impact**

**Checklist**


docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
+ungrib:
+  execution:
+    mpicmd: srun
```
maddenp-noaa, line 3: I recall that when we tested with a `null` value here, the schema validator rightly complained. But it would be interesting to know if we can omit `mpicmd` entirely, and maybe reconsider our schema for the `execution` block (and potentially our Python code that accesses this value), since not all components we write drivers for will need -- or IMO should have to specify if they don't -- a command to run under MPI, because they're not MPI codes. `ungrib` is not an MPI code, for example.

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
...
+    batchargs:
+      nodes: 1
```
maddenp-noaa, line 5: I'd suggest trying `cores: 1` instead of `nodes: 1` and keeping that if it works. Since `ungrib` is a serial (i.e. not parallel, not MPI, not OpenMPI) code, it only needs a single core to run. On some systems, users will have no choice but to run an `ungrib` job on a full node, but many HPCs (like NOAA RDHPCS) allow requesting that your job run on a node where other jobs might be running, too: On an n-core node, if you ask for 1 core, you get than and then 1/n of the node's memory. This can save both time (because your job may not queue as long) and money (on a, say, 96-core node, you'll pay only 1/96 what you would have paid for using the whole node!). Reflecting those facts in our example config might give users helpful information about what the codes requirements really are -- although there's more authoritative and comprehensive documentation for that elsewhere.

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
...
+      walltime: 00:05:00
```
maddenp-noaa, line 6: IIRC `00:01:00` might be enough for `ungrib`. As above, this could be a useful hint to users about what `ungrib`'s actual requirements are.

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
...
+    executable: "path/to/ungrib.exe"
```
maddenp-noaa, line 7:
```suggestion
    executable: "/path/to/ungrib.exe"
```

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
...
+  gfs_file: "/path/to/dir/gfs.2024.03.18.12"
```
maddenp-noaa, line 8:
```suggestion
  gfs_file: "/path/to/dir/gfs.t12z.pgrb2.0p25.f000"
```
Readers might better recognize the filename that NCEP uses, which in this case would be `gfs.t12z.pgrb2.0p25.f000`, than the name we used during development and testing.

(FWIW, NCEP only encodes the hour part of the cycle information (`t12z` for the 12 UTC cycle) in the filename; the year/month/day information is encoded in their directory structure. The GRIB data inside the file contains cycle information, too, though, so it's not as if paths and filenames are the only way to get that.)

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,13 @@
...
+  run_dir: "/path/to/dir/run"
+  vtable: "/path/to/Vtable.GFS"
+platform:
+  account: wrfruc
+  scheduler: slurm
```
maddenp-noaa, line 13: We probably don't want the `platform` block in this file, as it has its own documentation.

src/uwtools/api/ungrib.py
```diff
@@ -0,0 +1,53 @@
+"""
+API access to the uwtools ungrib driver.
```
maddenp-noaa, line 2:
```suggestion
API access to the ``uwtools`` ``ungrib`` driver.
```
Looks nice when RST renders to HTML.

src/uwtools/api/ungrib.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+import datetime as dt
+from pathlib import Path
+from typing import Dict, Optional
+
+from uwtools.drivers import support
```
maddenp-noaa, line 8:
```suggestion
from uwtools.drivers import _support
```
We'll want this to be marked private so that API users don't think that `support` is something we intend to be used directly. You'll also need to update the matching reference(s) to `support` below.

src/uwtools/api/ungrib.py
```diff
@@ -0,0 +1,53 @@
...
+from uwtools.drivers.ungrib import Ungrib
```
maddenp-noaa, line 9:
```suggestion
from uwtools.drivers.ungrib import _Ungrib
```
Ditto.

src/uwtools/api/ungrib.py
```diff
@@ -0,0 +1,53 @@
...
+def execute(
+    task: str,
+    config_file: Path,
+    cycle: dt.datetime,
+    batch: bool = False,
+    dry_run: bool = False,
+    graph_file: Optional[Path] = None,
+) -> bool:
+    """
+    Execute an Ungrib task.
```
maddenp-noaa, line 21:
```suggestion
    Execute an ``ungrib`` task.
```

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
+"""
+A driver for the Ungrib model.
```
maddenp-noaa, line 2:
```suggestion
A driver for the ungrib component.
```

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+
+from uwtools.config.formats.nml import NMLConfig
+from uwtools.drivers.driver import Driver
+from uwtools.strings import STR
+from uwtools.utils.tasks import file
+
+
+class Ungrib(Driver):
+    """
+    A driver for Ungrib.
```
maddenp-noaa, line 21:
```suggestion
    A driver for ungrib.
```
We use an initial capital for the class name because that's Python convention, but a quick google for "wrf ungrib" indicates that the program/component itself is generally referred to in lower case.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+                    "max_dom": 1,
+                    "start_date": self._cycle.strftime("%Y-%m-%d_%H:00:00"),
+                    "wrf_core": "ARW",
+                },
+                "ungrib": {
+                    "out_format": "WPS",
+                    "prefix": "FILE",
+                },
+            }
+        }
```
maddenp-noaa, line 77: I suggested to @elcarpenterNOAA during pairing that the namelist file for `ungrib` is simple enough that we could consider just creating it from scratch. The cycle-related values here we probably always want to supply. If there's a reason why, at least in the short term where we'll be using this strictly for MPAS, any of the other values shouldn't be hardcoded and have to be exposed for user configuration in UW YAML, that can be done. Can we YAGNI this for now?

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+        yield asset(path, path.is_file)
+        yield None
+        envcmds = self._driver_config.get("execution", {}).get("envcmds", [])
+        execution = [self._runcmd, "test $? -eq 0 && touch %s/done" % self._rundir]
+        scheduler = self._scheduler if self._batch else None
+        path.parent.mkdir(parents=True, exist_ok=True)
+        rs = self._runscript(envcmds=envcmds, execution=execution, scheduler=scheduler)
+        with open(path, "w", encoding="utf-8") as f:
+            print(rs, file=f)
+        os.chmod(path, os.stat(path).st_mode | stat.S_IEXEC)
```
maddenp-noaa, line 126: I think most of the logic in this method was lately factored out to `Driver._write_runscript()`. Please have a look at the other drivers, which should be updated (on `main` at least) to use that refactoring.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+            "scheduler": self._config["platform"]["scheduler"],
+            **self._driver_config.get("execution", {}).get("batchargs", {}),
+        }
+
+    @property
+    def _runscript_path(self) -> Path:
+        """
+        Returns the path to the runscript.
+        """
+        return self._rundir / "runscript"
```
maddenp-noaa, line 167: I think there's also been an update, available on `main`, that includes the name of the component in the runscript name.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def _taskname(self, suffix: str) -> str:
+        """
+        Returns a common tag for graph-task log messages.
+
+        :param suffix: Log-string suffix.
+        """
+        return "%s Ungrib %s" % (self._cycle.strftime("%Y%m%d %HZ"), suffix)
```
maddenp-noaa, line 175:
```suggestion
        return "%s ungrib %s" % (self._cycle.strftime("%Y%m%d %HZ"), suffix)
```

maddenp-noaa (commented):
I'm still reviewing the tests but thought I'd share these comments now.

src/uwtools/tests/api/test_ungrib.py
```diff
@@ -0,0 +1,34 @@
...
+def test_execute(tmp_path):
+    dot = tmp_path / "graph.dot"
+    args: dict = {
+        "config_file": "config.yaml",
+        "cycle": dt.datetime.utcnow(),
+        "batch": False,
+        "dry_run": True,
+        "graph_file": dot,
+    }
+    with patch.object(ungrib, "Ungrib") as Ungrib:
```
maddenp-noaa, line 18:
```suggestion
    with patch.object(ungrib, "_Ungrib") as Ungrib:
```
You'll need this when you change `Ungrib` to `_Ungrib` to mark it private in the `uwtools.api.ungrib` imports.

src/uwtools/tests/api/test_ungrib.py
```diff
@@ -0,0 +1,34 @@
...
+def test_graph():
+    with patch.object(ungrib.support, "graph") as graph:
+        ungrib.graph()
+    graph.assert_called_once_with()
+
+
+def test_tasks():
+    with patch.object(ungrib.support, "tasks") as _tasks:
+        ungrib.tasks()
+    _tasks.assert_called_once_with(ungrib.Ungrib)
```
maddenp-noaa, line 34:
```suggestion
    _tasks.assert_called_once_with(ungrib._Ungrib)
```
I think this'll also be needed with the `as Ungrib` -> `as _Ungrib` import change.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@fixture
+def config(tmp_path):
+    return {
+        "ungrib": {
+            "execution": {
+                "mpicmd": "srun",
+                "batchargs": {
+                    "nodes": 1,
```
maddenp-noaa, line 32:
```suggestion
                    "cores": 1,
```
...if this works and `docs/shared/ungrib.yaml` likewise changes.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+                    "walltime": "00:05:00",
```
maddenp-noaa, line 33:
```suggestion
                    "walltime": "00:01:00",
```
if `docs/shared/ungrib.yaml` likewise changes.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+                },
+                "executable": str(tmp_path / "ungrib.exe"),
+            },
+            "gfs_file": str(tmp_path / "gfs.2024.03.18.12"),
```
maddenp-noaa, line 37:
```suggestion
            "gfs_file": str(tmp_path / "gfs.t12z.pgrb2.0p25.f000"),
```

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def test_Ungrib_gribfile_aaa(driverobj):
+    src = driverobj._rundir / "GRIBFILE.AAA.in"
+    src.touch()
+    driverobj._driver_config["gfs_file"] = src
+    dst = driverobj._rundir / "GRIBFILE.AAA"
+    assert not dst.is_symlink()
+    driverobj.gribfile_aaa()
+    assert dst.is_symlink()
```
maddenp-noaa, line 82: Nice.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+    with open(dst, "r", encoding="utf-8") as f:
+        lines = f.read().split("\n")
+    # Check directives:
+    assert "#SBATCH --account=me" in lines
+    assert "#SBATCH --time=01:10:00" in lines
+    # Check environment commands:
+    assert "cmd1" in lines
+    assert "cmd2" in lines
+    # Check execution:
+    assert "test $? -eq 0 && touch %s/done" % driverobj._rundir
```
maddenp-noaa, line 144: I think that, since the logic for creating the runscript has been factored out to `Driver`, probably this test and the corresponding ones for the other drivers need to be simplified to just assert that they call `Driver._write_runscript()` correctly. I've added that to my TODO list. I think this method is fine for now, though.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def test_Ungrib__driver_config(driverobj):
+    assert driverobj._driver_config == driverobj._config["ungrib"]
+
+
+def test_Ungrib__resources(driverobj):
+    account = "me"
+    scheduler = "slurm"
+    walltime = "01:10:00"
```
maddenp-noaa, line 164:
```suggestion
    walltime = "00:01:00"
```
or similar, to be realistic.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -0,0 +1,184 @@
...
+    assert driverobj._resources == {
+        "account": account,
+        "rundir": driverobj._rundir,
+        "scheduler": scheduler,
+        "walltime": walltime,
+    }
+
+
+def test_Ungrib__runscript_path(driverobj):
+    assert driverobj._runscript_path == driverobj._rundir / "runscript"
```
maddenp-noaa, line 176:
```suggestion
    assert driverobj._runscript_path == driverobj._rundir / "runscript.ungrib"
```
if the driver code gets updated to follow the newer convention from `main` wherein runscript files have their component name as a suffix.

maddenp-noaa (commented):
Comments on unit tests...

docs/shared/ungrib.yaml
```diff
@@ -0,0 +1,10 @@
+ungrib:
+  execution:
+    mpicmd: srun
+    batchargs:
+      nodes: 1
+      walltime: 00:01:00
```
christinaholtNOAA, line 6: Is this a realistic time for this application?

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+from uwtools.strings import STR
+from uwtools.utils.tasks import file
+
+
+class Ungrib(Driver):
+    """
+    A driver for ungrib.
+    """
+
+    _driver_name = STR.ungrib
```
christinaholtNOAA, line 24: This ended up as a private method with a `@property` decorator in a recent omnibus PR, IIR.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        The gribfile.
+        """
+        path = self._rundir / "GRIBFILE.AAA"
+        yield self._taskname(str(path))
+        yield asset(path, path.is_symlink)
+        infile = Path(self._driver_config["gfs_file"])
+        yield file(path=infile)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.symlink_to(infile)
```
christinaholtNOAA, line 56: This will definitely work for preparing initial conditions. I think (and I might be wrong) that you can do a whole batch of GFS files to handle all the times for ICs and LBCs in one iteration. In that case links would look something like:

```
GRIBFILE.AAA -> "/path/to/dir/gfs.t12z.pgrb2.0p25.f000"
GRIBFILE.AAB -> "/path/to/dir/gfs.t12z.pgrb2.0p25.f006"
GRIBFILE.AAC -> "/path/to/dir/gfs.t12z.pgrb2.0p25.f012"
```

This does not need to be modified now, but would need to be extended for limited-area in this way at some point if I'm remembering correctly from my WRF days.

src/uwtools/drivers/ungrib.py
christinaholtNOAA: So we won't be allowing users to update any of these settings like we do with the other namelists? I'm not sure how I feel about that as the general pattern, but it should be fine here.

This end date will likely need to change along with changes to support LBCs.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+            self.vtable(),
+        ]
+
+    @tasks
+    def run(self):
+        """
+        A run.
+        """
+        yield self._taskname("run")
+        yield (self._run_via_batch_submission() if self._batch else self._run_via_local_execution())
```
christinaholtNOAA, line 108: I was thinking this moved to the base class, so it might not be necessary here.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Private helper methods
+
+    @property
+    def _driver_config(self) -> Dict[str, Any]:
+        """
+        Returns the config block specific to this driver.
+        """
+        driver_config: Dict[str, Any] = self._config["ungrib"]
+        return driver_config
```
christinaholtNOAA, line 148: This may also have moved to the base class along with `_runscript_path` below.

src/uwtools/resources/jsonschema/ungrib.jsonschema
```diff
@@ -0,0 +1,29 @@
+{
```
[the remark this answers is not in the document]
christinaholtNOAA: This seems good at this point. Am I missing something, or is that why you marked the comment resolved?

src/uwtools/tests/drivers/test_ungrib.py
christinaholtNOAA: +1

christinaholtNOAA (commented):
I have just a few suggestions below that have changed in `main` since you started this work, but it all looks great!

docs/shared/ungrib.yaml
elcarpenterNOAA: It was previously `00:05:00` and Paul suggested `00:01:00`. So far I've seen it run within `00:00:05` and `00:00:07`. Would you suggest increasing it?

src/uwtools/drivers/ungrib.py
maddenp-noaa: This is the only component namelist I've seen so far that seemed like it might not need user interaction. It'll be easy enough to add if/when we need it.

It'll be interesting to see how we decide to handle LBCs in the UW YAML: It would be _nice_ to just take a start time and some kind of time delta and calculate the rest, but there's no telling what the GFS GRIB files will be called. I don't think we can safely assume the NCEP filenames, so we'll probably have to accept either an explicit list of filenames, or some kind of template from which the filenames can be constructed.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+        path = self._rundir / "GRIBFILE.AAA"
+        yield self._taskname(str(path))
+        yield asset(path, path.is_symlink)
+        infile = Path(self._driver_config["gfs_file"])
+        yield file(path=infile)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.symlink_to(infile)
+
+    @task
+    def namelist_wps(self):
```
maddenp-noaa, line 59: We may want to call just just `namelist()` to be consistent with other drivers.

src/uwtools/drivers/ungrib.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        return self._rundir / "runscript.ungrib"
+
+    def _taskname(self, suffix: str) -> str:
+        """
+        Returns a common tag for graph-task log messages.
+
+        :param suffix: Log-string suffix.
+        """
+        return "%s ungrib %s" % (self._cycle.strftime("%Y%m%d %HZ"), suffix)
```
maddenp-noaa, line 175: To summarize and extend several previous comments, several of the methods defined here were recently factored out to the `Driver` base class, so they can be removed or simplified here. I think, but please verify:
- `run()` can be deleted, as its's present in the base class.
- `runscript()` can be simplified by calling `self._write_runscript()`.
- `_driver_config()` can be deleted, as it's present in the base class.
- `_driver_name()` should be implemented as a `@property` method here.
- `_runscript_path()` can be deleted, as it's present in the base class.

elcarpenterNOAA: This was my list as well and we're still running successfully on Jet after doing the above.

src/uwtools/tests/drivers/test_ungrib.py
maddenp-noaa: I see this was marked resolved, but the value didn't change. IMO we should sync it to whatever we decide [here](https://github.com/ufs-community/uwtools/pull/433/files#diff-65ceadb1fe0fb3a73a3c4e33d05a1bd5ab88cd66770f92b1847a47f1977c1f59R6).

docs/shared/ungrib.yaml
maddenp-noaa: Quick follow-up on this one: Could you try running with this line removed and see what happens? It looks to me like it's not required by the schema, but the `Driver` code appears to require it. If that's the case, I propose that we leave this as-is for now, but I'll put a TODO on my list to make this more flexible, so that non-MPI codes don't need to specify an `mpicmd`. @christinaholtNOAA Any reason you can think of not to go this way?

src/uwtools/tests/drivers/test_ungrib.py
elcarpenterNOAA: I've changed it to match the others for now.

docs/shared/ungrib.yaml
elcarpenterNOAA: When I ran it without `mpicmd` I got a `KeyError` because it's needed in `_runcmd()`.

maddenp-noaa: I thought so. Thanks for testing. I think you can resolve this comment, and I'll take a TODO to make this flexible unless @christinaholtNOAA dissuades me.

maddenp-noaa (approved):
Nice work. I recorded a couple of small items for a later fixup, but have no concerns about moving forward with this.

christinaholtNOAA (approved):
lgtm!

PR #438, opened 2024-03-22
omnibus-2024-03-22

maddenp-noaa:
**Synopsis**

- Make `mpicmd` entry for driver configs optional and update `ungrib` driver to not use it.
- Let drivers receive their YAML config on `stdin` as well as via an on-disk file.
- Do not use shell environment when realizing configs (a template-rendering pre-processing step can be used if needed).
- Factor `_resources()` method out from individual driver classes to `Driver` base class and simplify unit tests.
- Prefix all component invocations with `time` to record execution time.
- Simplify `_runscript()` test methods by testing `Driver` method and mocking it in individual drivers' tests.

**Type**

**Impact**

**Checklist**


src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -130,40 +131,18 @@ def test_ChgresCube_run_local(driverobj):
...
-    assert "cmd2" in lines
-    # Check execution:
-    assert "srun --export=ALL --ntasks $SLURM_CPUS_ON_NODE /path/to/chgres_cube" in lines
-    assert "test $? -eq 0 && touch %s/done" % driverobj._rundir
+    with patch.object(driverobj, "_runscript") as runscript:
+        driverobj.runscript()
+        runscript.assert_called_once()
+        args = ("envcmds", "envvars", "execution", "scheduler")
+        types = [list, dict, list, Slurm]
+        assert [type(runscript.call_args.kwargs[x]) for x in args] == types
```
maddenp-noaa, line 139: This test (and the corresponding ones for other drivers) can be simpler now, as the bulk of the logic in the `Driver` class has its own test.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -55,39 +46,53 @@ def write(path, s):
 
 
 @fixture
-def a_driver(tmp_path):
+def driverobj(tmp_path):
```
maddenp-noaa, line 49: Made the name of this fixture consistent with that in other driver test modules.

src/uwtools/tests/test_cli.py
```diff
@@ -276,16 +276,19 @@ def test__dict_from_key_eq_val_strings():
 
 
 def test__dispatch_chgres_cube():
+    cycle = (dt.datetime.now(),)
```
christinaholtNOAA, line 279: I'm confused by this tuple and the next one.

christinaholtNOAA (approved):
LGTM.

src/uwtools/tests/test_cli.py
maddenp-noaa: I think there was a rogue comma, which made the value technically a tuple, which `black` then dutifully wrapped in parens to make that explicit. I'll clean it up. Thanks.

elcarpenterNOAA (approved):

PR #444, opened 2024-03-27
minibus 03-27-2024

christinaholtNOAA:
**Synopsis**

Include support for jobname rendering in a task config, and compound strings in environment variables.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/rocoto.py
+++ b/src/uwtools/rocoto.py
@@ -31,10 +31,10 @@ def realize_rocoto_xml(
     """
     rxml = _RocotoXML(config)
     xml = str(rxml).strip()
-    with writable(output_file) as f:
-        print(xml, file=f)
     if not validate_rocoto_xml_string(xml):
         raise UWError("Internal error: Invalid Rocoto XML")
+    with writable(output_file) as f:
+        print(xml, file=f)
     return xml
 
 
@@ -381,6 +381,7 @@ class _RocotoXML:
 
         :param config: Configuration data for this element.
         :param taskname: The name of the task being defined.
+        :return: The updated dict with all jobname entries rendered.
         """
         if STR.jobname not in config:
             config[STR.jobname] = taskname
@@ -403,6 +404,7 @@ class _RocotoXML:
         Return the tag and metadata extracted from a metadata-bearing key.
 
         :param key: A string of the form "<tag>_<metadata>" (or simply STR.<tag>).
+        :return: Tag and name of key.
         """
         # For example, key "task_foo_bar" will be split into tag "task" and name "foo_bar".
         parts = key.split("_")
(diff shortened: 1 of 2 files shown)


src/uwtools/rocoto.py
```diff
@@ -31,10 +31,10 @@ def realize_rocoto_xml(
     """
     rxml = _RocotoXML(config)
     xml = str(rxml).strip()
-    if not validate_rocoto_xml_string(xml):
-        raise UWError("Internal error: Invalid Rocoto XML")
     with writable(output_file) as f:
         print(xml, file=f)
+    if not validate_rocoto_xml_string(xml):
+        raise UWError("Internal error: Invalid Rocoto XML")
```
christinaholtNOAA, line 37: Adding this here gives a tangible but incorrect file that is easier for debugging.

src/uwtools/rocoto.py
```diff
@@ -383,6 +384,9 @@ def _set_and_render_jobname(self, config: dict, taskname: str) -> None:
         """
         if STR.jobname not in config:
             config[STR.jobname] = taskname
+        cfg = YAMLConfig(config=config)
+        cfg.dereference()
+        return cfg.data
```
christinaholtNOAA, line 389: Since any of the entries might have values that include a `{{jobname}}` template will need to be updated, it can't be done in place as easily as it can with returning a new dict.

WeirAE (approved):
Ah, nice quick fixes! Thanks for the explanations, LGTM

src/uwtools/rocoto.py
```diff
@@ -374,7 +375,7 @@ def _insert_doctype(self, xml: str) -> str:
             lines.insert(1, doctype)
         return "\n".join(lines)
 
-    def _set_and_render_jobname(self, config: dict, taskname: str) -> None:
+    def _set_and_render_jobname(self, config: dict, taskname: str) -> dict:
```
maddenp-noaa, line 378: Needs a `:return:` entry in the docstring, or the first docstring line updated to state `Return ...`.

src/uwtools/rocoto.py
WeirAE: Hmm, thats a fair point, I think the `.bad`/`,err`/etc would be good, or we could dump to `stdout` in the conditional too

maddenp-noaa: Only output consumable by downstream tools (including `uw` via pipes) should be written to `stdout`. Our logging is configured to write to `stderr`, so iterating over the lines and logging them would be fine as an alternative to writing to a file with an extension.

christinaholtNOAA: I will take this change out for now and may revisit it later when the time crunch is reduced a bit.

maddenp-cu (approved):

PR #446, opened 2024-03-28
UW-522 Improve handling of missing variables in template render mode

maddenp-noaa:
**Synopsis**

Given `template.yaml`
```
roses: "{{ colors.color1 }}"
violets: "{{ colors.color2 }}"
```
and `values.yaml`
```
colors:
  color1: red
```
Old behavior:
```
% uw template render --input-file template.yaml --values-file values.yaml
roses: "red"
violets: ""
```
New behavior:
```
% uw template render --input-file template.yaml --values-file values.yaml
[2024-03-28T18:49:30]    ERROR Render failed with error: 'dict object' has no attribute 'color2'
[2024-03-28T18:49:30]    ERROR Template could not be rendered
```

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/config/test_jinja2.py
+++ b/src/uwtools/tests/config/test_jinja2.py
@@ -64,10 +64,9 @@ cannot:
 # Helpers
 
 
-def render_helper(input_file, values_file, **kwargs):
+def render_helper(input_file, **kwargs):
     return jinja2.render(
         input_file=input_file,
-        values_src=values_file,
         **kwargs,
     )
 
@@ -174,7 +173,7 @@ def test_register_filters_path_join(key):
 def test_render(values_file, template_file, tmp_path):
     outfile = str(tmp_path / "out.txt")
     expected = "roses are red, violets are blue"
-    result = render_helper(input_file=template_file, values_file=values_file, output_file=outfile)
+    result = render_helper(input_file=template_file, values_src=values_file, output_file=outfile)
     assert result == expected
     with open(outfile, "r", encoding="utf-8") as f:
         assert f.read().strip() == expected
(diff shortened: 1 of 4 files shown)


src/uwtools/tests/config/test_jinja2.py
```diff
@@ -171,25 +170,25 @@ def test_register_filters_path_join(key):
             template.render(**context)  # path_join filter fails
 
 
-def test_render(values_file, template, tmp_path):
+def test_render(values_file, template_file, tmp_path):
     outfile = str(tmp_path / "out.txt")
     expected = "roses are red, violets are blue"
-    result = render_helper(input_file=template, values_file=values_file, output_file=outfile)
+    result = render_helper(input_file=template_file, values_src=values_file, output_file=outfile)
```
NaureenBharwaniNOAA, line 176: Looking through this updated file with `template` updated to `template_file`, it became more obvious that we're renaming `values_src=values_file` in the second argument. I was wondering why we chose to place it into a new variable name? I see that the main instances of this occurring are when we are calling the `render_helper` function. 

Could we simply use `values_file=values_file`? Or does `src` has more of an appropriate meaning when being passed to the `render_helper` function?

maddenp-noaa: This is related to the change on line 67, which removes the mandatory `values_file` argument from the `render_helper()` function. Since there is no named `values_file` argument in that function after the change, the keyword argument needs to match the ultimate argument name that `render_helper()` passes to `jinja2.render()`, which is `values_src`.

That said, I made that change because I thought I had a use case where I did not want to supply a `values_src` argument to `jinja2.render()`, and it looks like that use case did not materialize. So I'm going to make `values_file` mandatory again in `render_helper()` and fix the calls.

More philosophically, though: `jinja2.render()` specifies a formal argument called `values_src` because the actual argument may be a `Path` to a file, or a `dict`, or nothing (in which case `stdin` will be read). So the "source of the values" is somewhat generic while, in the _test_, the actual argument is always a file, such that a call like `jinja2.render(values_src=values_file, ...)` makes sense: The "source of the values" is a "values file".

elcarpenterNOAA (approved):
LGTM!

WeirAE (approved):
Gave it another morning look and LGTM

PR #447, opened 2024-03-29
UW-533 & UW-534 MPAS driver for init_atmosphere and atmosphere forecast 

elcarpenterNOAA:
**Synopsis**

- Added drivers for `init_atmosphere` and `atmosphere`
- Wired up API and CLI 
- Added JSONSchema
- Documentation will be a separate PR. 

`init_atmosphere` was successfully run on Jet. See `/mnt/lfs4/BMC/wrfruc/elcarpenter/uwtools/workflow-tools/rundir` for an example run. 

**Type**

**Impact**

**Checklist**


docs/shared/mpas_init.yaml
```diff
@@ -0,0 +1,62 @@
+user:
+  mpas_app: /path/to/mpas_app
+platform:
+  account: me
+  scheduler: slurm
+mpas_init:
+  boundary_conditions:
+    interval_hours: 6
+    length: 6
```
elcarpenterNOAA, line 9: Sample `yaml` that can be used in the documentation later.

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,179 @@
...
+        d = {
+            "nhyd_model": {
+                "config_start_time": self._cycle.strftime("%Y-%m-%d_%H:00:00"),
+                "config_stop_time": stop_time.strftime("%Y-%m-%d_%H:00:00"),
+            }
+        }
+        namelist = self._driver_config.get("namelist", {})
+        values = YAMLConfig(d)
+        values.update_values(namelist.get("update_values", {}))
+        namelist["update_values"] = values.data
```
elcarpenterNOAA, line 103: Need to update `namelist.init_atmosphere` with cycle info.

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,179 @@
...
+        fn = "streams.init_atmosphere"
+        yield self._taskname(fn)
+        path = self._rundir / fn
+        yield asset(path, path.is_file)
+        yield file(path=Path(self._driver_config["streams"]["path"]))
+        render(
+            input_file=Path(self._driver_config["streams"]["path"]),
+            output_file=path,
+            values_src=self._driver_config["streams"]["values"],
+        )
```
elcarpenterNOAA, line 150: Need to fill in `streams.init_atmosphere` file with appropriate values from the `yaml`.

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __init__(self, config: Path, cycle: datetime, dry_run: bool = False, batch: bool = False):
+        """
+        The driver.
+
+        :param config_file: Path to config file.
+        :param cycle: The forecast cycle.
+        :param dry_run: Run in dry-run mode?
+        :param batch: Run component via the batch system?
+        """
```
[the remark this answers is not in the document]
elcarpenterNOAA: Fixed it in one place, but not the other. Thanks!

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,179 @@
...
+            **self._driver_config.get("execution", {}).get("batchargs", {}),
+        }
+
+    def _taskname(self, suffix: str) -> str:
+        """
+        Returns a common tag for graph-task log messages.
+
+        :param suffix: Log-string suffix.
+        """
+        return "%s mpas-init %s" % (self._cycle.strftime("%Y%m%d %HZ"), suffix)
```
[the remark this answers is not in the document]
elcarpenterNOAA: Nice catch. Updated here and in `drivers/ungrib.py` to be consistent with the others.

src/uwtools/api/mpas_init.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+from typing import Dict, Optional
+
+import uwtools.drivers.support as _support
+from uwtools.drivers.mpas_init import MPASInit as _MPASInit
+
+
+def execute(
+    task: str,
+    config: Path,
```
NaureenBharwaniNOAA, line 15: Change `config: Path`  --> `config: Optional[Path] = None`

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,178 @@
...
+    def _resources(self) -> Dict[str, Any]:
+        """
+        Returns configuration data for the runscript.
+        """
+        return {
+            "account": self._config["platform"]["account"],
+            "rundir": self._rundir,
+            "scheduler": self._config["platform"]["scheduler"],
+            **self._driver_config.get("execution", {}).get("batchargs", {}),
+        }
```
NaureenBharwaniNOAA, line 170: General question, I don't see a `_resources` private helper method implemented for any other driver anymore. I think it's just in the base class driver. I wasn't sure if this was an abstract method we needed to implement for each specific driver. I have it implemented in my jedi driver currently as well.

@maddenp-noaa

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,178 @@
...
+        endhour = lbcs["length"]
+        interval = lbcs["interval_hours"]
+        symlinks = {}
+        boundary_filepath = lbcs["path"]
+        for boundary_hour in range(0, endhour + 1, interval):
+            file_date = self._cycle + timedelta(hours=boundary_hour)
+            target = Path(boundary_filepath) / f"FILE:{file_date.strftime('%Y-%m-%d_%H')}"
+            linkname = self._rundir / f"FILE:{file_date.strftime('%Y-%m-%d_%H')}"
+            symlinks[target] = linkname
+        yield [symlink(target=t, linkname=l) for t, l in symlinks.items()]
```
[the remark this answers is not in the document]
WeirAE: I'm partial to the `fn` use in the #448 , but we should update the fv3 driver with it as well. I'd almost want to add `offset=0` for consistency too.

src/uwtools/drivers/mpas_init.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        d = {
+            "nhyd_model": {
+                "config_start_time": self._cycle.strftime("%Y-%m-%d_%H:00:00"),
+                "config_stop_time": stop_time.strftime("%Y-%m-%d_%H:00:00"),
+            }
+        }
+        namelist = self._driver_config.get("namelist", {})
+        values = YAMLConfig(d)
+        values.update_values(namelist.get("update_values", {}))
```
maddenp-noaa, line 101: Given this order of operations, it looks like we'd start with the known-correct cycle info in `d` and then potentially clobber that with possibly-wrong values taken from `update_values`. One would hope the user wouldn't provide bad cycle info in `update_values`, but I feel like the information provided by the `--cycle` switch (or equivalent API argument) should always take precedence. If that's right, how about something like
```
update_values = namelist.get("update_values", {})
update_values.setdefault("nhyd_model", {}).update({"config_start_time": self._cycle.strftime("%Y-%m-%d_%H:00:00"), "config_stop_time": stop_time.strftime("%Y-%m-%d_%H:00:00")})
```
(It'll look nicer when `black` wraps it.)

Here's a demo (I've set the `self._cycle` and `stop_time` variables):
```
>>> update_values = {}
>>> update_values.setdefault("nhyd_model", {}).update({"config_start_time": self._cycle.strftime("%Y-%m-%d_%H:00:00"), "config_stop_time": stop_time.strftime("%Y-%m-%d_%H:00:00")})
>>> update_values
{'nhyd_model': {'config_start_time': '2024-03-29_12:00:00', 'config_stop_time': '2024-03-30_12:00:00'}}
```
That's if `update_values` isn't provided in YAML, but it should work just as well if it is:
```
>>> update_values = {"foo": "bar", "nhyd_model": {"baz": "qux", "config_start_time": "bad-value", "config_stop_time": "bad-value"}}
>>> update_values.setdefault("nhyd_model", {}).update({"config_start_time": self._cycle.strftime("%Y-%m-%d_%H:00:00"), "config_stop_time": stop_time.strftime("%Y-%m-%d_%H:00:00")})
>>> update_values
{'foo': 'bar', 'nhyd_model': {'baz': 'qux', 'config_start_time': '2024-03-29_12:00:00', 'config_stop_time': '2024-03-30_12:00:00'}}
```
The `.setdefault()` call creates the `nhyd_model` `dict` if it doesn't exist yet, and otherwise just uses it.

src/uwtools/drivers/mpas_init.py
maddenp-noaa: Are we assuming that the streams file is a Jinja2 template? IIRC that's not how it's supplied with MPAS. It's XML, right? If so, I guess I would have expected use of `lxml` (and probably its XPath support) to edit the file. I wouldn't hold up this PR for that, but I'd like to discuss this and potentially create a ticket for using `lxml` to edit the file that ships with MPAS. Apologies if I've misunderstood what's happening here or any assumptions we're making.

maddenp-noaa (commented):
Initial comments, will review the tests shortly.

src/uwtools/resources/jsonschema/mpas-init.jsonschema
NaureenBharwaniNOAA, line 1: Differences between MPAS and MPAS-init:

- The name of the boundary condition variable, i.e. `boundary_conditions` vs. `lateral_boundary_conditions`.
- Where `files_to_copy`, `files_to_link` and `execution` live -- one is inside boundary conditions the other is outside.
- Where required properties include for `path`, and `values` within `streams` and the other doesn't.
- `streams` is a required property and the other streams isn't.

docs/shared/mpas_init.yaml
```diff
@@ -0,0 +1,62 @@
...
+  execution:
+    executable: "{{ user.mpas_app }}/exec/init_atmosphere_model"
+    batchargs:
+      walltime: 01:30:00
+      cores: 4
```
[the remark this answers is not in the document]
christinaholtNOAA: We were testing with Janet's small use case. She already had the 4-proc decomposition file, so we followed her lead. Once we nail down what the GSL group thinks is useful, we can circle back to this update.

src/uwtools/api/mpas.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Dict, Optional
+
+import uwtools.drivers.support as _support
+from uwtools.drivers.mpas import MPAS as _MPAS
+
+
+def execute(
+    task: str,
+    cycle: dt.datetime,
+    config: Optional[Path] = None,
```
NaureenBharwaniNOAA, line 16: Seems like everywhere else, `config` comes before `cycle`

src/uwtools/resources/jsonschema/mpas.jsonschema
```diff
@@ -0,0 +1,87 @@
...
+            },
+            "path": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "interval_hours",
+            "offset",
+            "path"
+          ]
```
NaureenBharwaniNOAA, line 34: Right below this I think we should have
```
          "type": "object"
```

src/uwtools/drivers/mpas_init.py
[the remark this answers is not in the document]
christinaholtNOAA: I agree that the `fn` might be one way to DRY the mpas-init code, but otherwise I think these two are doing what they should for their given drivers. 

For mpas_init, it prepares boundary conditions, both lateral and boundary. In mpas, all the files that are needed are named `lbc.*.nc`, so it seemed more consistent to name it specifically `lateral_boundary_conditions`.

For mpas_init, it needs to know how many boundary conditions to prepare, so we included `length` as a setting for the boundary conditions. For mpas, the length is of the forecast itself, so we chose to stay consistent with the `fv3` driver and have `length` live under the driver config.

src/uwtools/tests/api/test_mpas.py
```diff
@@ -0,0 +1,35 @@
...
+def test_execute(tmp_path):
+    cycle = dt.datetime.now()
+    dot = tmp_path / "graph.dot"
+    args: dict = {
+        "batch": False,
+        "config": "config.yaml",
+        "cycle": cycle,
+        "dry_run": True,
+        "graph_file": dot,
+    }
```
NaureenBharwaniNOAA, line 18: What it looks like for mpas-init
```
    args: dict = {
        "config": "config.yaml",
        "cycle": dt.datetime.now(),
        "batch": False,
        "dry_run": True,
        "graph_file": dot,
    }
```

src/uwtools/tests/drivers/test_mpas_init.py
```diff
@@ -0,0 +1,233 @@
...
+
+from uwtools.drivers import mpas_init
+from uwtools.tests.support import fixture_path
+
+# Fixtures
+
+
+@fixture
+def cycle():
+    return dt.datetime(2024, 2, 1, 18)
```
NaureenBharwaniNOAA, line 23: Do we want to update the `datetime`?  
```
    return dt.datetime(2024, 3, 22, 6)
```

src/uwtools/drivers/mpas_init.py
christinaholtNOAA: I wouldn't be opposed to this approach for updating the XML that MPAS produces at build time. Right now, we've taken the one that's built with the mpas_app and turned it into a template. I think this would be relatively straight-forward to modify for support.

In general, it sounds like this could open us up to adding general XML to our config languages we support with YAML?

maddenp-noaa:
We might update the PR title for posterity, now that both init and atm are represented here.

src/uwtools/tests/drivers/test_mpas_init.py
```diff
@@ -0,0 +1,233 @@
...
+    assert "#SBATCH --account=me" in lines
+    assert "#SBATCH --time=00:01:00" in lines
+    # Check environment commands:
+    assert "cmd1" in lines
+    assert "cmd2" in lines
+    # Check execution:
+    assert "test $? -eq 0 && touch %s/done" % driverobj._rundir
+
+
+def test_MPASInit_streams_file(driverobj):
```
NaureenBharwaniNOAA, line 197: MPAS naming for test --> `def test_MPAS_streams(driverobj):`,  remove `file`

src/uwtools/drivers/mpas_init.py
christinaholtNOAA: I factored out `_resources` to the drive base class with some light refactoring when the `chgres_cube` driver was added. It's no longer an abstract class, but a concrete one in the base class.

src/uwtools/resources/jsonschema/mpas-init.jsonschema
```diff
@@ -0,0 +1,93 @@
...
+            "path": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "length",
+            "interval_hours",
+            "offset",
+            "path"
+          ]
```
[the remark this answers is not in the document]
christinaholtNOAA: I can make these edits. We definitely don't want so many things nested under `boundary_conditions`.

src/uwtools/drivers/mpas_init.py
maddenp-noaa: General XML support could be added to the roadmap, but I don't think we'd need to go that far to directly support the MPAS XML files. I'd prefer to support MPAS in the short term, and consider general XML handling in the longer therm.

src/uwtools/cli.py
```diff
@@ -425,6 +429,114 @@ def _dispatch_fv3(args: Args) -> bool:
...
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    :param task: The task to add a subparser for.
+    :param helpmsg: Help message for task.
+    """
+    parser = _add_subparser(subparsers, task, helpmsg.rstrip("."))
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_config_file(group=required, required=True)
+    _add_arg_cycle(required)
+    optional = _basic_setup(parser)
```
elcarpenterNOAA, line 516: @maddenp-noaa I've updated `config` to be `Optional` within the two `mpas_init.py` files, but I'm curious why it doesn't cause an issue here. I will update it, I just found it to be oddly inconsistent.

src/uwtools/resources/jsonschema/mpas-init.jsonschema
christinaholtNOAA: @NaureenBharwaniNOAA I think it makes sense to leave the difference between `boundary_conditions` and `lateral_boundary_conditions` given the difference between the two components. I'm happy to hear others' thoughts, though.

I do not see that the `files_to_copy`, `files_to_link`, and `execution` sections are inside the boundary_conditions section in either of the schemas. Maybe I'm reading that wrong. Would you mind double-checking on that? I do see that they are ordered differently related to their different block name.

I've updated both schemas for consistency on the last two bullets above.

src/uwtools/tests/drivers/test_mpas_init.py
elcarpenterNOAA: I'm going to keep `_file` in the test name to keep it consistent with `test_MPAS_namelist_file()` and the function names in `drivers/mpas_init.py`.

src/uwtools/resources/jsonschema/mpas-init.jsonschema
NaureenBharwaniNOAA: Nope, you're right they're not included in the `boundary_conditions`. Just ordered differently.

src/uwtools/drivers/mpas_init.py
christinaholtNOAA: I will apply this suggested approach to the mpas driver.

src/uwtools/tests/api/test_mpas.py
elcarpenterNOAA: I've updated this in mine.

src/uwtools/cli.py
maddenp-noaa: I think `--config-file` should be optional here, too, as for all (I think) the other drivers: The config can be read from `stdin` as well as from a file.

src/uwtools/drivers/mpas_init.py
christinaholtNOAA: Sounds good to me.

src/uwtools/strings.py
```diff
@@ -86,6 +86,7 @@ class STR:
     link: str = "link"
     mode: str = "mode"
     model: str = "model"
+    mpasinit: str = "mpas_init"
```
christinaholtNOAA, line 90: This is weird. I have an mpas entry in my local copy, but not here. I'm not sure how to resolve that...digging in.

src/uwtools/cli.py
```diff
@@ -425,6 +429,114 @@ def _dispatch_fv3(args: Args) -> bool:
...
+    """
+    Subparser for mode: mpas <task>
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    :param task: The task to add a subparser for.
+    :param helpmsg: Help message for task.
+    """
+    parser = _add_subparser(subparsers, task, helpmsg.rstrip("."))
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_config_file(group=required, required=True)
```
maddenp-noaa, line 460: The config file can also be provided on `stdin`, so `--config-file` should be optional. This should be the case for the existing drivers, too.

src/uwtools/tests/drivers/test_mpas_init.py
christinaholtNOAA: I don't think there's a need for that. I chose a different date and cycle in the mpas test just so we weren't always testing with the same cycle assumptions.

src/uwtools/drivers/mpas.py
```diff
@@ -0,0 +1,167 @@
...
+        self._cycle = cycle
+
+    # Workflow tasks
+
+    @task
+    def boundary_files(self):
+        """
+        Boundary condition files.
+        """
+        yield self._taskname("boundary files")
```
maddenp-noaa, line 50: Slight disagreement between method name, docstring (which appears in CLI `--help`) and task name as yielded to `iotaa`. Maybe remove `condition` from the docstring for consistency? Whatever you think is best, though.

src/uwtools/strings.py
christinaholtNOAA: Apparently, it was a brief fluke. I see it here now.

src/uwtools/drivers/mpas.py
maddenp-noaa: Same complain for the init version.

src/uwtools/drivers/mpas_init.py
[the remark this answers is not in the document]
maddenp-noaa: I'm also bothered by the apparent duplication here, though I believe you (but haven't read closely enough to fully understand) that there are important differences. Still, I wonder if some generalization could be done later (not in this PR) to DRY this out. If I'm not just wrong, can we at least create a ticket to consider refactoring this later?

docs/shared/mpas_init.yaml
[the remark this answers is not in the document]
maddenp-noaa: OK, I think I recall now that, when we were running MPAS at Spire, we used two layouts / MPI task counts: one for init and one for forecast. We found that the efficiency of init topped out at a task count much lower than forecast, but got good speedups up to that point. It might have been something like 256 tasks for init and a much larger number for forecast.

src/uwtools/drivers/mpas_init.py
maddenp-noaa: I created [UW-556](https://jira.epic.oarcloud.noaa.gov/browse/UW-556) to address this.

src/uwtools/drivers/mpas.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        The namelist file.
+        """
+        path = self._rundir / "namelist.atmosphere"
+        yield self._taskname(str(path))
+        yield asset(path, path.is_file)
+        yield None
+        duration = timedelta(hours=self._driver_config["length"])
+        str_duration = str(duration).replace(" days, ", "")
+        namelist = self._driver_config.get("namelist", {})
```
maddenp-noaa, line 95: I'm reviving this comment, which fell into "outdated" status and was hidden before we had a chance to discuss...

Thinking out loud: What is the case in which `namelist` is really optional here? Here are the cases I can think of:
1. User provides `namelist` with `updates_values` but no `base_file`. While possible, this may be unlikely because the `update_values` would have to completely specify the namelist content, which might mean a lot of YAML. But fine: We would still be able to add the cycle info and create the file.
2. User provides `namelist` with `base_file` but no `update_values`. Fine: We create an empty `update_values`, two lines down, populate it with the cycle info, and write the file.
3. User provides `namelist` with both `base_file` and `update_values`. Fine: We modify `update_values` with cycle info, and write the file.
4. User provides no `namelist`. That's permitted by the schema, but means that they must provide a complete, ready-to-go namelist file. If they do that, we'll never reach this code because `yield asset(path, path.is_file)`, above, will have located that file and this task will be deemed complete already.

Based on that case analysis, I wonder if we should use `self._driver_config["namelist"]` in a `try` / `except` block (or not provide the `{}` default value here and check for a `None` value) and raise a `UWConfigError` describing the situation. It seems to me like the `namelist` block is only optional in one case -- a case that should prevent execution of this code. I think we can detect if the user has done that incorrectly and help them here.

Please let me know if you see problems with that argument. But if it's right, we might need to open a ticket to make the same improvement in other drivers.

EDIT: What I'm proposing (here and similarly in the init driver) would be something like
```
try:
    namelist = self._driver_config("namelist")
except KeyError:
    raise UWConfigError("Provide either a 'namelist' YAML block or the %s file" % path)
```

src/uwtools/drivers/mpas_init.py
maddenp-noaa: I also created [UW-557](https://jira.epic.oarcloud.noaa.gov/browse/UW-557) for factoring out some of the commonalities between the two MPAS drivers.

src/uwtools/drivers/mpas.py
elcarpenterNOAA: Just pushed this change in the two drivers. I still need to update the schema, though.

src/uwtools/cli.py
```diff
@@ -425,6 +429,114 @@ def _dispatch_fv3(args: Args) -> bool:
...
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    :param task: The task to add a subparser for.
+    :param helpmsg: Help message for task.
+    """
+    parser = _add_subparser(subparsers, task, helpmsg.rstrip("."))
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_cycle(required)
+    optional = _basic_setup(parser)
+    _add_arg_config_file(group=optional, required=False)
```
elcarpenterNOAA, line 462: This has now been updated to reflect that `config` is optional in the `MPAS` driver.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -0,0 +1,234 @@
...
+    assert dst.is_file()
+    assert isinstance(f90nml.read(dst), f90nml.Namelist)
+
+
+def test_MPAS_namelist_missing(driverobj):
+    path = driverobj._rundir / "namelist.atmosphere"
+    del driverobj._driver_config["namelist"]
+    with raises(UWConfigError) as e:
+        assert driverobj.namelist_file()
+    assert str(e.value) == ("Provide either a 'namelist' YAML block or the %s file" % path)
```
elcarpenterNOAA, line 141: New tests for the `try`/`except` blocks.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -0,0 +1,234 @@
...
+    with open(dst, "r", encoding="utf-8") as f:
+        lines = f.read().split("\n")
+    # Check directives:
+    assert "#SBATCH --account=me" in lines
+    assert "#SBATCH --time=00:01:00" in lines
+    # Check environment commands:
+    assert "cmd1" in lines
+    assert "cmd2" in lines
+    # Check execution:
+    assert "test $? -eq 0 && touch %s/done" % driverobj._rundir
```
maddenp-noaa, line 194: IMO, for methods that have been factored out to `Driver` -- either just with the bulk of their logic moved, or the entire method moved there -- it's enough in these per-driver unit tests to mock that `Driver` method or function and then `assert` here that it was called with the expected arguments. Then, we can concentrate our efforts on vigorously testing those core, shared method/functions in the `Driver` unit tests, and all drivers will benefit, without duplication of effort. Here, I'm not sure if the tests of `runscript()`, `driver_config()`, or `resources()` add much over what's already tested in the `Driver` tests. Please let me know if I'm wrong!

src/uwtools/drivers/mpas.py
maddenp-noaa: Exactly. The schema needs to be permissive, but in the driver we know all the facts and can check things that we couldn't check at schema-validation time. (We don't explicitly check that the namelist file exists, but it's implied by the fact that the task method makes it past the `yield` statements and into the imperative code.)

src/uwtools/tests/drivers/test_mpas.py
elcarpenterNOAA: Without `test_MPAS_runscript()` we don't have 100% test coverage. The `driver_config()` and `resources()` tests do seem unnecessary, though.

maddenp-noaa: I didn't mean that this necessarily has to change now. Just that it might not have been necessary to write quite so much test code, given that the MPAS drivers call `Driver` in cases that are already well-tested. Duplicating similar tests across test modules does mean that more code may have to be updated if something changes that effects all of them.

maddenp-noaa: You do need a method to exercise it, but it can be simpler. See the `runscript()` tests in the other drivers.

maddenp-noaa (approved):
This looks great, nice work. I created two Jira tickets for refactoring follow-ups, as mentioned. I didn't have time to read everything, but support merging this if you'd like to move ahead. If it doesn't merge before Monday I'll probably spend a little more time reading the unit tests, though they look solid overall.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -0,0 +1,233 @@
...
+        driverobj = mpas.MPAS(config=config_file, cycle=cycle, batch=True, dry_run=True)
+    assert driverobj._dry_run is True
+    dryrun.assert_called_once_with()
+
+
+@pytest.mark.parametrize(
+    "key,task,test",
+    [("files_to_copy", "files_copied", "is_file"), ("files_to_link", "files_linked", "is_symlink")],
+)
+def test_MPAS_files_copied(config, cycle, key, task, test, tmp_path):
```
maddenp-noaa, line 110: This method tests both copying and linking files, so its name should probably be more general. Ditto for the corresponding test in the `MPASInit` tests. And ditto for the FV3 test that these are based on, which should have had a better name to begin with (my fault). Maybe even just more explicit like `test_*_files_copied_and_linked`.

maddenp-noaa (approved):
I think I had one outstanding comment about simplifying the unit tests where good `Driver` tests are already providing coverage, and made one small suggestion here about renaming a few test methods. I think that's it for my review. Everything looks good.

src/uwtools/tests/drivers/test_fv3.py
```diff
@@ -120,7 +120,7 @@ def test_FV3_field_table(driverobj):
     "key,task,test",
     [("files_to_copy", "files_copied", "is_file"), ("files_to_link", "files_linked", "is_symlink")],
 )
-def test_FV3_files_copied(config, cycle, key, task, test, tmp_path):
+def test_FV3_files_copied_and_linked(config, cycle, key, task, test, tmp_path):
```
elcarpenterNOAA, line 123: Updated test names for `FV3`, `MPAS`, and `MPASInit`.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -0,0 +1,209 @@
...
+
+@fixture
+def config(tmp_path):
+    return {
+        "mpas": {
+            "execution": {
+                "executable": "atmosphere_model",
+                "batchargs": {
+                    "walltime": "01:30:00",
+                },
```
elcarpenterNOAA, line 39: Added and `walltime` and `platform` blocks to `MPAS` and `MPASInit` configs in order to simplify their `runscript()` tests.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -0,0 +1,209 @@
...
+    func.assert_called_once_with()
+
+
+def test_MPAS_runscript(driverobj):
+    with patch.object(driverobj, "_runscript") as runscript:
+        driverobj.runscript()
+        runscript.assert_called_once()
+        args = ("envcmds", "envvars", "execution", "scheduler")
+        types = [list, dict, list, Slurm]
+        assert [type(runscript.call_args.kwargs[x]) for x in args] == types
```
elcarpenterNOAA, line 188: `runscript()` tests now match the test for the `FV3` driver.

NaureenBharwaniNOAA (approved):
Nice work on the drivers. The PR looks great to me!

WeirAE (approved):
Nicely merged and sorted, looks great

PR #449, opened 2024-03-30
UW-475 Support for YAML !remove tag

maddenp-noaa:
**Synopsis**

Add support for a YAML `!remove` tag.

Given `config.yaml`
```
fruit:
  a: apple
  b: banana
  c: cherry
```
and `remove.yaml`
```
fruit:
  b: !remove
```
we can
```
% uw config realize --input-file config.yaml --output-format yaml remove.yaml
fruit:
  a: apple
  c: cherry
```
Similarly with `config.nml`
```
&fruit
  a = "apple"
  b = "banana"
  c = "cherry"
/
```
we can
```
% uw config realize --input-file config.nml --output-format nml remove.yaml
&fruit
    a = 'apple'
    c = 'cherry'
/
```

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/base.py
+++ b/src/uwtools/config/formats/base.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-import json
 import os
 import re
 from abc import ABC, abstractmethod
 from collections import UserDict
 from copy import deepcopy
+from io import StringIO
 from pathlib import Path
 from typing import List, Optional, Tuple, Union
 
@@ -46,7 +46,9 @@ class Config(ABC, UserDict):
         """
         Returns the string representation of a Config object.
         """
-        return json.dumps(self.data)
+        s = StringIO()
+        yaml.dump(self.data, s)
+        return s.getvalue()
 
     # Private methods
 
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -5,7 +5,13 @@ from typing import Optional
 import yaml
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import INCLUDE_TAG, TaggedString, add_yaml_representers, log_and_error
+from uwtools.config.support import (
+    INCLUDE_TAG,
+    UWYAMLConvert,
+    UWYAMLRemove,
+    add_yaml_representers,
+    log_and_error,
+)
 from uwtools.strings import FORMAT
 from uwtools.utils.file import readable, writable
 
@@ -94,8 +100,9 @@ class YAMLConfig(Config):
         """
         loader = yaml.SafeLoader
         loader.add_constructor(INCLUDE_TAG, self._yaml_include)
-        for tag in TaggedString.TAGS:
-            loader.add_constructor(tag, TaggedString)
+        for tag_class in (UWYAMLConvert, UWYAMLRemove):
+            for tag in getattr(tag_class, "TAGS"):
+                loader.add_constructor(tag, tag_class)
         return loader
 
     # Public methods
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -133,10 +133,14 @@ def dereference(
     """
     rendered: _ConfigVal = val  # fall-back value
     if isinstance(val, dict):
-        return {
-            dereference(k, context): dereference(v, context, local=val, keys=[*(keys or []), k])
-            for k, v in val.items()
-        }
+        keys = keys or []
+        new = {}
+        for k, v in val.items():
+            if isinstance(v, UWYAMLRemove):
+                _deref_debug("Removing value at", " > ".join([*keys, k]))
+            else:
+                new[dereference(k, context)] = dereference(v, context, local=val, keys=[*keys, k])
+        return new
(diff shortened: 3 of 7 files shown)


src/uwtools/config/formats/base.py
```diff
@@ -46,7 +46,9 @@ def __repr__(self) -> str:
         """
         Returns the string representation of a Config object.
         """
-        return json.dumps(self.data)
+        s = StringIO()
+        yaml.dump(self.data, s)
+        return s.getvalue()
```
maddenp-noaa, line 51: `json.dumps()` was handy because of PyYAML's (as far as I know) lack of a dump-to-string function, but `json` has no idea how to handle a `UWYAMLRemove` object and raises an exception when it encounters it. `yaml`, however, has a representer registered for that type, so using `yaml.dump()` with a `StringIO` object does the trick.

src/uwtools/config/formats/yaml.py
```diff
@@ -94,8 +100,9 @@ def _yaml_loader(self) -> type[yaml.SafeLoader]:
         """
         loader = yaml.SafeLoader
         loader.add_constructor(INCLUDE_TAG, self._yaml_include)
-        for tag in TaggedString.TAGS:
-            loader.add_constructor(tag, TaggedString)
+        for tag_class in (UWYAMLConvert, UWYAMLRemove):
+            for tag in getattr(tag_class, "TAGS"):
+                loader.add_constructor(tag, tag_class)
```
maddenp-noaa, line 105: Add constructors for both the existing `UWYAMLConvert` type (renamed from `TaggedString`), and the new `UWYAMLRemove`.

src/uwtools/config/jinja2.py
```diff
@@ -109,7 +109,9 @@ def undeclared_variables(self) -> Set[str]:
 # Public functions
 
 
-def dereference(val: _ConfigVal, context: dict, local: Optional[dict] = None) -> _ConfigVal:
+def dereference(
+    val: _ConfigVal, context: dict, local: Optional[dict] = None, keys: Optional[List[str]] = None
+) -> _ConfigVal:
```
maddenp-noaa, line 114: The new `keys` argument allows us to keep track of the path of keys that led to a value, in particular to a value that is then removed. Verbose log output would now include a message like
```
[2024-03-30T19:12:44]    DEBUG [dereference] Removing value at: a > b > c
```
given YAML containing
```
a:
  b:
    c: !remove
```

src/uwtools/config/jinja2.py
```diff
@@ -126,17 +128,25 @@ def dereference(val: _ConfigVal, context: dict, local: Optional[dict] = None) ->
...
     if isinstance(val, dict):
-        return {dereference(k, context): dereference(v, context, local=val) for k, v in val.items()}
+        keys = keys or []
+        new = {}
+        for k, v in val.items():
+            if isinstance(v, UWYAMLRemove):
+                _deref_debug("Removing value at", " > ".join([*keys, k]))
+            else:
+                new[dereference(k, context)] = dereference(v, context, local=val, keys=[*keys, k])
+        return new
```
maddenp-noaa, line 143: When dereferencing, if a `UWYAMLRemove` value is encountered, simply decline to add it to the new `dict` of dereferenced keys and values, and debug-log a message noting that it is being removed.

src/uwtools/config/formats/yaml.py
```diff
@@ -5,7 +5,13 @@
 import yaml
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import INCLUDE_TAG, TaggedString, add_yaml_representers, log_and_error
+from uwtools.config.support import (
```
WeirAE, line 8: I always get wary when a `from-include` block starts to need a new line. I suspect if we have to pack more things in `config.support`, we'll be better off just importing the whole thing or splitting it up further

src/uwtools/tests/config/test_tools.py
```diff
@@ -296,6 +298,55 @@ def test_realize_config_output_file_conversion(tmp_path):
...
+    loader = yaml.Loader(dedent(s).strip())
+    loader.add_constructor("!remove", UWYAMLRemove)
+    path = tmp_path / "config.nml"
+    assert not path.is_file()
+    tools.realize_config(
+        input_config=base,
+        output_file=path,
+        supplemental_configs=[YAMLConfig(loader.get_data())],
+    )
+    assert f90nml.read(path) == {"constants": {"pi": 3.141}}
```
christinaholtNOAA, line 316: Just making sure here...the end user would not need to add their own UW constructors to use the api version of the `realize` tool, correct? This is just setting up an isolated case to test the new feature directly?

christinaholtNOAA (commented):
This looks great, just one question below.

src/uwtools/tests/config/test_tools.py
maddenp-noaa: Yes. This test confuses that, though. I'm going to update it to try to make that clearer.

christinaholtNOAA (approved):
Thanks for the clarification!

WeirAE (approved):
LGTM!

PR #453, opened 2024-04-03
UW-489 JEDI driver

NaureenBharwaniNOAA:
**Synopsis**

Completed:
- Added driver for JEDI.
- API and CLI are working and tested.

To be completed in a future ticket:
- Adding JSONSchema and unit tests for the schema.
- Documentation for the JEDI driver.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


src/uwtools/tests/test_cli.py
```diff
@@ -480,6 +495,21 @@ def test__dispatch_fv3():
...
+    args: dict = {
+        "batch": True,
+        "config_file": "config.yaml",
+        "cycle": cycle,
+        "dry_run": False,
+    }
+    with patch.object(uwtools.api.jedi, "execute") as execute:
+        cli._dispatch_jedi({**args, "action": "foo"})
+    execute.assert_called_once_with(
+        task="foo", config_file="config.yaml", cycle=cycle, batch=True, dry_run=False
```
NaureenBharwaniNOAA, line 509: I'm trying to standardize the way `config` vs. `config_file` is used for the jedi driver to match other drivers. This was also the expected output vs. the actual output. If anyone has any suggestions to get this to conform better let me know. It wanted `config_file` here and not simply `config`.

My actual output the tests didn't like:
```
batch=True, config="config.yaml", cycle=cycle, dry_run=False, graph_file=None, task="foo"
```

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,165 @@
...
+                    "stdout": "/path/to/file",
+                    "walltime": "00:02:00",
+                },
+                "envcmds": [
+                    "cmd1",
+                    "cmd2",
+                    "module load right-modules",
+                    "module load jedi-modules",
+                ],
+                "executable": "/scratch2/BMC/zrtrr/Naureen.Bharwani/build/bin/qg_forecast.x",
```
NaureenBharwaniNOAA, line 46: Change executable value to "/path/to/qg_forecast.x"

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,165 @@
...
+    logging.getLogger().setLevel(logging.INFO)
+    # with patch.object(jedi, "is_symlink") as is_symlink:
+
+
+def test_JEDI_provisioned_run_directory(driverobj):
+    with patch.multiple(
+        driverobj,
+        files_copied=D,
+        files_linked=D,
+        runscript=D,
```
NaureenBharwaniNOAA, line 108: Does `validate_only` need to be apart of this provisioned run dir. Should this be an optional argument if supplie dby the user?

i.e. 
```
if validate_only supplied:
    add validate_only in mocks

```

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,165 @@
...
+
+def test_JEDI_validate_only(caplog, config_file, cycle, driverobj):
+    logging.getLogger().setLevel(logging.INFO)
+    with patch.object(jedi, "run") as run:
+        result = Mock(output="", success=True)
+        run.return_value = result
+        driverobj.validate_only()
+        run.assert_called_once_with(
+            "20240201 18Z jedi validate_only",
+            "cmd1 && cmd2 && module load right-modules && module load jedi-modules && time /scratch2/BMC/zrtrr/Naureen.Bharwani/build/bin/qg_forecast.x --validate-only /scratch2/BMC/zrtrr/Naureen.Bharwani/uwtools/src/uwtools/tests/fixtures/jedi.yaml 2>&1",
```
NaureenBharwaniNOAA, line 146: Make more generic for the driver.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,165 @@
...
+        )
+    assert regex_logged(caplog, "Config is valid")
+
+
+def test_JEDI_yaml_file(driverobj):
```
NaureenBharwaniNOAA, line 151: Need to finish implementation on test case.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,207 @@
...
+        driverobj._driver_config["run_dir"] = "/path/to/run"
+        driverobj.files_copied()
+        assert filecopy.call_count == 2
+        assert (
+            call(src=Path("/path/to/baz"), dst=Path("/path/to/run/bar/baz"))
+            in filecopy.call_args_list
+        )
+        assert (
+            call(src=Path("/path/to/foo"), dst=Path("/path/to/run/foo")) in filecopy.call_args_list
+        )
```
elcarpenterNOAA, line 109: Can you explain why the `files_copied()` and `files_linked()` tests are different for JEDI than the other drivers?

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,207 @@
...
+    cfgfile = Path(driverobj._driver_config["run_dir"]) / "input.yaml"
+    assert not cfgfile.is_file()
+    driverobj.configuration_file()
+    assert cfgfile.is_file()
+    newcfg = YAMLConfig(config=cfgfile)
+    assert newcfg == {**basecfg, "baz": "qux"}
+
+
+def test_JEDI__driver_config(driverobj):
+    assert driverobj._driver_config == driverobj._config["jedi"]
```
elcarpenterNOAA, line 199: Is this already being adequately tested in the `Driver` tests?

src/uwtools/cli.py
```diff
@@ -431,6 +433,57 @@ def _dispatch_fv3(args: Args) -> bool:
...
 
+# Mode jedi
+
+
+def _add_subparser_jedi(subparsers: Subparsers) -> ModeChecks:
+    """
+    Subparser for mode: jedi
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.jedi, "Execute Jedi tasks")
```
maddenp-noaa, line 444:
```suggestion
    parser = _add_subparser(subparsers, STR.jedi, "Execute JEDI tasks")
```

src/uwtools/cli.py
```diff
@@ -494,7 +548,7 @@ def _add_subparser_mpas_init(subparsers: Subparsers) -> ModeChecks:
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.mpasinit, "Execute MPAS Init tasks")
+    parser = _add_subparser(subparsers, STR.mpasinit, "Execute MPASInit tasks")
```
maddenp-noaa, line 551:
```suggestion
    parser = _add_subparser(subparsers, STR.mpasinit, "Execute MPAS Init tasks")
```

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+    def provisioned_run_directory(self):
+        """
+        Run directory provisioned with all required content.
+        """
+        yield self._taskname("provisioned run directory")
+        yield [
+            self.files_copied(),
+            self.files_linked(),
+            self.configuration_file(),
+            self.runscript(),
```
maddenp-noaa, line 70:
```suggestion
            self.configuration_file(),
            self.files_copied(),
            self.files_linked(),
            self.runscript(),
```

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        path = self._runscript_path
+        yield self._taskname(path.name)
+        yield asset(path, path.is_file)
+        yield None
+        self._write_runscript(path=path, envvars={})
+
+    @task
+    def validate_only(self):
+        """
+        Validate config.
```
maddenp-noaa, line 87: I think users might benefit from seeing a slightly more thorough one-line description of this task.

Also, at the moment, I think this task would

1. Only be executed when explicitly requested by the user, and
2. Only validate the `base_file`, before any `update_values` are applied to it.

I'm not sure, but is our idea that `base_file` is likely to be something supplied by JEDI, or perhaps a version of that file modified by an app? If that's the case, it seems likely that this file will usually be valid, and that it might be useful to also validate the file created by the `configuration_file` task, after applying the `update_values` supplied by the user. If I'm not on the wrong track here, maybe it'd be useful to make this task depend on `configuration_file()` instead of `file(base_file)` (ok, it actually says `file(path)` here now, but I'll add a suggestion to change `path` to `base_file` to make it clearer what that file is) and add it as a requirement of `provisioned_run_directory` in place of `configuration_file` (since it transitively requires `configuration_file`)... and/or make it a requirement of `run` so that the config file is always validated before a run is performed.

docs/shared/jedi.yaml
```diff
@@ -0,0 +1,8 @@
+jedi:
+  execution:
+    mpicmd: srun
+    batchargs:
+      nodes: 1
+      walltime: 08:00:00
+    executable: "/path/to/jedi.exe"
+  run_dir: "/path/to/dir/run"
```
christinaholtNOAA, line 8: I would have expected this example to be somewhat similar to the one used to test the driver in a real-world scenario.

EDIT: The one you use for test configuration for the driver would be a nice one to model.

src/uwtools/api/jedi.py
```diff
@@ -0,0 +1,54 @@
...
+def execute(
+    task: str,
+    config: Path,
+    cycle: dt.datetime,
+    batch: bool = False,
+    dry_run: bool = False,
+    graph_file: Optional[Path] = None,
+) -> bool:
+    """
+    Execute a jedi task.
```
christinaholtNOAA, line 22:
```suggestion
    Execute a JEDI task.
```

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        taskname = self._taskname("validate_only")
+        yield taskname
+        a = asset(None, lambda: False)
+        yield a
+        path = Path(self._driver_config["configuration_file"]["base_file"])
```
maddenp-noaa, line 93:
```suggestion
        base_file = Path(self._driver_config["configuration_file"]["base_file"])
```
Other lines will have to be updated to match. Since `executable` is also a path, I find it odd to find this one called `path`. This may also change depending on what happens with the comment I made above about validating the `configuration_file` file instead of the `base_file`.

src/uwtools/cli.py
christinaholtNOAA, line 444:
```suggestion
    parser = _add_subparser(subparsers, STR.jedi, "Execute JEDI tasks")
```

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        yield [file(executable), file(path)]
+        env = " && ".join(self._driver_config["execution"]["envcmds"])
+        cmd = "{env} && time {x} --validate-only {p} 2>&1".format(env=env, x=executable, p=path)
+        result = run(taskname, cmd)
+        if result.success:
+            logging.info("%s: Config is valid", taskname)
+            a.ready = lambda: True
+
+    @task
+    def configuration_file(self):
```
maddenp-noaa, line 104: Since the name of this method changed, it could be moved up to preserve the alphabetical ordering of the method names in the class.

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        The configuration file.
+        """
+        fn = "input.yaml"
```
maddenp-noaa, line 108: Just out of curiosity, does `input.yaml` conform to an existing convention for naming JEDI configuration YAML files? Maybe there is no convention, or maybe this is it -- I have no idea.

src/uwtools/cli.py
```diff
@@ -431,6 +433,57 @@ def _dispatch_fv3(args: Args) -> bool:
...
+    Dispatch logic for jedi mode.
+
+    :param args: Parsed command-line args.
+    """
+    return uwtools.api.jedi.execute(
+        task=args[STR.action],
+        config=args[STR.cfgfile],
+        cycle=args[STR.cycle],
+        batch=args[STR.batch],
+        dry_run=args[STR.dryrun],
```
christinaholtNOAA, line 483: This doesn't have a graph file argument.

src/uwtools/cli.py
```diff
@@ -431,6 +433,57 @@ def _dispatch_fv3(args: Args) -> bool:
...
+    :param task: The task to add a subparser for.
+    :param helpmsg: Help message for task.
+    """
+    parser = _add_subparser(subparsers, task, helpmsg.rstrip("."))
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_config_file(group=required, required=True)
+    _add_arg_cycle(required)
+    optional = _basic_setup(parser)
+    _add_arg_batch(optional)
+    _add_arg_dry_run(optional)
```
christinaholtNOAA, line 467: There's no graph arg here.

src/uwtools/drivers/jedi.py
christinaholtNOAA, line 87:
```suggestion
        Validate JEDI config YAML.
```

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        yield self._taskname(fn)
+        path = self._rundir / fn
+        yield asset(path, path.is_file)
+        yield None
+        self._create_user_updated_config(
+            config_class=YAMLConfig,
+            config_values=self._driver_config.get("configuration_file", {}),
```
maddenp-noaa, line 115: Using `.get()` with a `{}` default implies that it's OK for this task to produce an empty YAML file as `input.yaml`. Is that really ok? If the driver needs _some_ file, then this should perhaps be
```
config_values=self._driver_config["configuration_file"],
```
so that the code will at least fail if we somehow make it this far without specifying either `base_file` or `update_values`. (When the schema is fleshed out, that should catch such an error even earlier by making `configuration_file` required; and the `validate_only` task, if made a requirement of other task methods, like `provisioned_run_directory` or `run` as suggested above, might also reject an empty YAML file. But the code here should also reflect the intent that we _must have_ a non-empty YAML file, if that's accurate.)

maddenp-noaa (commented):
Here are some initial comments, stopping just before the tests, which I'll review ASAP.

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,134 @@
...
+        taskname = self._taskname("validate_only")
+        yield taskname
+        a = asset(None, lambda: False)
+        yield a
+        path = Path(self._driver_config["configuration_file"]["base_file"])
+        executable = Path(self._driver_config["execution"]["executable"])
+        yield [file(executable), file(path)]
+        env = " && ".join(self._driver_config["execution"]["envcmds"])
+        cmd = "{env} && time {x} --validate-only {p} 2>&1".format(env=env, x=executable, p=path)
+        result = run(taskname, cmd)
```
christinaholtNOAA, line 98: Why do we use `iotaa`'s `run` instead of `uwtool`'s `execute` here?

src/uwtools/drivers/jedi.py
christinaholtNOAA, line 108: Could we name this one `config.yaml`? It's more consistent with the file name JCSDA uses in the [JEDI documentation](https://jointcenterforsatellitedataassimilation-jedi-docs.readthedocs-hosted.com/en/latest/using/building_and_running/config_content.html) and `input.nml` is a very FV3-specific name.

Another option is `jedi.yaml`.

Yet another...we could use the name of the file provided as the base file. (Not sure I love this one given we're modifying it)

src/uwtools/tests/drivers/test_jedi.py
maddenp-noaa: Since @NaureenBharwaniNOAA and I paired on this, and I recommended these simpler tests, I can try to answer.

Honestly, I'm asking myself why the other tests are so complex! I think I probably wrote the one for FV3, from which MPAS and MPAS Init probably borrowed, and I'm wondering why I didn't use mocks there like we've done here, since the `filecopy` and `symlink` task methods from `uwtools.utils.tasks`, which perform the actual copy/link actions, have their own tests.

I wonder now if I wrote the FV3 tests before `filecopy` and `symlink` appeared in `uwtools.utils.tasks`, such that the FV3 tests _had to_ test the real filesystem operations, because the FV3 driver was initially performing them directly rather than delegating to `filecopy` and `symlink`.

I'm looking at the FV3 tests and haven't convinced myself yet that these couldn't be simplified with mocks. What do you think?

src/uwtools/tests/drivers/test_jedi.py
christinaholtNOAA: It could be a nice way of ensuring that the files we're generating for configuration are consistent with the executable before wasting compute resources.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,207 @@
...
+            result = Mock(output="", success=True)
+            run.return_value = result
+            driverobj._driver_config["configuration_file"]["base_file"] = "/path/to/jedi.yaml"
+            driverobj.validate_only()
+            cmds = [
+                "module load some-module",
+                "module load jedi-module",
+                "time /path/to/qg_forecast.x --validate-only /path/to/jedi.yaml 2>&1",
+            ]
+            run.assert_called_once_with("20240201 18Z jedi validate_only", " && ".join(cmds))
```
christinaholtNOAA, line 181: Does the string `"20240201 18Z jedi validate_only"` actually get passed as an argument to `run()`? Or is that the `iotaa` `run()` that I commented on above?

src/uwtools/tests/fixtures/jedi.yaml
```diff
@@ -0,0 +1,19 @@
+forecast length: P2D
```
christinaholtNOAA, line 1: This file isn't used for tests as far as I can tell. Perhaps we can remove it.

christinaholtNOAA (commented):
@NaureenBharwaniNOAA Thanks for your work on this!

src/uwtools/tests/fixtures/jedi.yaml
NaureenBharwaniNOAA: I was thinking the same thing. Perhaps move it to be included under what already exists for the shared/docs/jedi.yaml file

src/uwtools/api/jedi.py
maddenp-noaa: I think @christinaholtNOAA and I were reviewing simultaneously and made some similar comments. No pile-on intended!

src/uwtools/drivers/jedi.py
NaureenBharwaniNOAA: I agree, I think picking either config.yaml or jedi.yaml, and using that across the board will keep it consistent. Since JCSDAs documentation uses config.yaml. I think I’ll go with that one.

src/uwtools/drivers/jedi.py
maddenp-noaa: @christinaholtNOAA I have an answer, from pairing: The `iotaa` version accepts and logs a task name, which is nice in this context vs `uwtools` `execute()`, which is generic and doesn't know that it's part of a task.

src/uwtools/tests/drivers/test_jedi.py
elcarpenterNOAA: That would make sense to me. To be honest, I had previously chalked these up to be "Paul loves to `parametrize` his unittests," but perhaps I could have put some extra thought into that before copying them over to other drivers. I have a strong preference for these tests. We should carry this change over to the other drivers at some point.

src/uwtools/api/jedi.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+from typing import Dict, Optional
+
+import uwtools.drivers.support as _support
+from uwtools.drivers.jedi import JEDI as _JEDI
+
+
+def execute(
+    task: str,
+    config: Path,
```
elcarpenterNOAA, line 15: What is the reason why `config` is not an `Optional[Path]` here?

src/uwtools/drivers/jedi.py
elcarpenterNOAA: @NaureenBharwaniNOAA, there's an example of this in the MPAS and MPAS Init drivers in their respective `namelist_file()` methods that use a `try`/`except` block. There's also unit tests that check for empty YAML files.

src/uwtools/tests/drivers/test_ungrib.py
```diff
@@ -165,7 +165,7 @@ def test_Ungrib__runscript_path(driverobj):
     assert driverobj._runscript_path == driverobj._rundir / "runscript.ungrib"
 
 
-def test_Ungrib__taskanme(driverobj):
+def test_Ungrib__taskname(driverobj):
```
elcarpenterNOAA, line 168: Nice. Only took 4 PRs to catch it!

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,207 @@
...
+                "update_values": {"baz": "qux"},
+            },
+            "files_to_copy": {
+                "foo": "/path/to/foo",
+                "bar/baz": "/path/to/baz",
+            },
+            "files_to_link": {
+                "foo": "/path/to/foo",
+                "bar/baz": "/path/to/baz",
+            },
```
maddenp-noaa, line 61: For testing purposes, it's ok if the `files_to_link` and `files_to_copy` blocks appear to deal with the same files, but if this gets copied to the `docs/shared/jedi.yaml` file to flesh that out, it would probably be a better example for users if the filenames differ so that it doesn't look like we're linking and copying the same files to the run directory.

src/uwtools/tests/drivers/test_jedi.py
maddenp-noaa: I might still parametrize these two, given how similar they are, though I'm not asking for that as a reviewer. Even parametrized, they'd still be much simpler than the existing FV3/MPAS/MPASInit versions.

src/uwtools/tests/drivers/test_jedi.py
maddenp-noaa: I made some related comments above about where and when the `validate_only` task should run.

src/uwtools/tests/drivers/test_jedi.py
maddenp-noaa: It's the `iotaa` `run()` that does take a `taskname` argument to make its logging more explicit.

src/uwtools/tests/fixtures/jedi.yaml
maddenp-noaa: As we use more inline `pytest` fixtures, we should probably periodically do a sweep through the external fixture files and see if there are any that are no longer used by any unit tests. I've done it once or twice in the past and will try to do another check someday soon.

src/uwtools/drivers/jedi.py
NaureenBharwaniNOAA: A couple questions on this suggestion. 

There seems to be a couple parts to this change:
1. Add a dependency for `validate_only` to depend on `configuration_file`
2. Change `path` to `base_file` where appropriate
3. Add the requirement of `base_file` to `provisioned_run_directory`


Questions:
1. In regards to the first change, I was trying to find some documentation on iotaa to create a dependency like this. Does this require a separate `@tasks` method that handles this dependency? Can this dependency be addressed by `validate_only` calling `configuration_file` and waiting for the contents back in a `yield` statement? Here I was thinking we are calling `self.configuration_file()` instead of `self._driver_config["configuration_file"]["base_file"]` and storing this value in the `base_file` variable. I was thinking about accessing the values within the returned config from the method, but it seems like it is not a scriptable and without any value it gives an error of `no file or directory to "/path/to/jedi.yaml"` which is in our tests. 
Trying to access this correctly could be the issue i.e., `base_file = Path(self.configuration_file["configuration_file"]["base_file"])` vs. `base_file = Path(self.configuration_file()["configuration_file"]["base_file"])`

```
    @task
    def validate_only(self):
        """
        Validate the JEDI driver config
        """
        taskname = self._taskname("validate_only")
        yield taskname
        a = asset(None, lambda: False)
        yield a
        base_file = Path(self.configuration_file["configuration_file"]["base_file"])
        executable = Path(self._driver_config["execution"]["executable"])
        yield [file(executable), file(base_file)]
        env = " && ".join(self._driver_config["execution"]["envcmds"])
        cmd = "{env} && time {x} --validate-only {p} 2>&1".format(env=env, x=executable, p=path) 
        result = run(taskname, cmd)
        if result.success:
            logging.info("%s: Config is valid", taskname)
            a.ready = lambda: True
```

3. Add the requirement of `base_file` to `provisioned_run_directory`. To me this only makes sense when `base_file` is a `@task` itself. Unless we are trying to use that dependency with `configuration_file` again to do something like `self.configuration_file()["base_file"]

src/uwtools/drivers/jedi.py
NaureenBharwaniNOAA: I think `config.yaml` would be a better option after discussion last week.

docs/shared/jedi.yaml
NaureenBharwaniNOAA: I updated this accordingly. I made the update to `files_to_copy` and `files_to_link` more generic for now. I'm not sure if we have good examples of what users might see for jedi.

src/uwtools/api/jedi.py
maddenp-noaa: If #460 merges before this PR, I think this should become `Optional[Union[Path, str]] = None`, and a call made to `_ensure_data_source()` on the value -- see that PR. `graph_file` also needs to support a `str` value; there, too, see that PR. If this merges first, I'll take care of it on my branch for that PR.

docs/shared/jedi.yaml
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I updated most of the file contents. I'm not sure exactly what the files copied and files linked examples could look like. I started with a generic change to begin with:
```
    files_to_copy:
      copy_foo: /path/to/foo
      copy_bar/baz: /path/to/baz
    files_to_link:
      link_foo: /path/to/foo
      link_bar/baz: /path/to/baz
```

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,203 @@
...
+            run.assert_called_once_with("20240201 18Z jedi validate_only", " && ".join(cmds))
+    assert regex_logged(caplog, "Config is valid")
+
+
+def test_JEDI_configuration_file(driverobj):
+    basecfg = {"foo": "bar"}
+    basefile = Path(driverobj._driver_config["configuration_file"]["base_file"])
+    with open(basefile, "w", encoding="utf-8") as f:
+        yaml.dump(basecfg, f)
+    cfgfile = Path(driverobj._driver_config["run_dir"]) / "input.yaml"
```
NaureenBharwaniNOAA, line 190: Interesting, I changed this from "input.yaml" to "config.yaml" and replaced the same thing in the `jedi.py` ` configuration_file` task. After doing this it breaks with an error message of:

```
driverobj = <uwtools.drivers.jedi.JEDI object at 0x7faf0c333aa0>

    def test_JEDI_configuration_file(driverobj):
        basecfg = {"foo": "bar"}
        basefile = Path(driverobj._driver_config["configuration_file"]["base_file"])
        with open(basefile, "w", encoding="utf-8") as f:
            yaml.dump(basecfg, f)
        cfgfile = Path(driverobj._driver_config["run_dir"]) / "config.yaml"
>       assert not cfgfile.is_file()
E       AssertionError: assert not True
E        +  where True = <bound method Path.is_file of PosixPath('/tmp/pytest-of-Naureen.Bharwani/pytest-19/test_JEDI_configuration_file0/config.yaml')>()
E        +    where <bound method Path.is_file of PosixPath('/tmp/pytest-of-Naureen.Bharwani/pytest-19/test_JEDI_configuration_file0/config.yaml')> = PosixPath('/tmp/pytest-of-Naureen.Bharwani/pytest-19/test_JEDI_configuration_file0/config.yaml').is_file

src/uwtools/tests/drivers/test_jedi.py:191: AssertionError
```

Once I change the name back it's fine and I haven't found another place where `input.yaml` is used.

maddenp-noaa: Do `basefile` and `cfgfile` end up being the same file here? If so, is the `yaml.dump()` creating that file, such that the `assert` is failing because the file _does_ exist when it is expected not to yet?

docs/shared/jedi.yaml
```diff
@@ -0,0 +1,27 @@
...
+    mpicmd: srun
+    configuration_file: 
+      base_file: path/to/config.yaml
+      update_values: {"baz": "qux"}
+    files_to_copy:
+      copy_foo: /path/to/foo
+      copy_bar/baz: /path/to/baz
+    files_to_link:
+      link_foo: /path/to/foo
+      link_bar/baz: /path/to/baz
```
maddenp-noaa, line 26:
```suggestion
    files_to_copy:
      f1: /path/to/f1
      d/f2: /path/to/f2
    files_to_link:
      f3: /path/to/f3
      d/f4: /path/to/f4
```
I was thinking something more like ^^^, as it I would find it surprising to see the same two files (`foo` and `baz`) bot copied _and_ symlinked into a run directory, even if the copies/links in the run directory don't conflict with one another. I think it would be more likely to see four different files, some of them copied, some of them symlinked. That's more like what we see with components' run directories. This example would represent one file and one symlink in the root of the run directory, and then a different file and a different symlink in the `d/` subdirectory of the run directory.

src/uwtools/tests/drivers/test_jedi.py
christinaholtNOAA: Thanks for the clarification.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -0,0 +1,203 @@
...
+                "envcmds": [
+                    "module load some-module",
+                    "module load jedi-module",
+                ],
+                "executable": "/path/to/qg_forecast.x",
+                "mpiargs": ["--export=ALL", "--ntasks $SLURM_CPUS_ON_NODE"],
+                "mpicmd": "srun",
+            },
+            "configuration_file": {
+                "base_file": str(tmp_path / "config.yaml"),
```
christinaholtNOAA, line 51: I am sorry to do this...we've gone round and round about the name of this file, but I'm just noticing one other potential issue. The base file is named `config.yaml` as is the fixture on line 71. Both are in the same place. I'm not sure what sort of debugging issue this may present in the future, but it might be good to go ahead and fix that ambiguity now.

I suggest changing this instance to`base.yaml`.

christinaholtNOAA (approved):
@NaureenBharwaniNOAA Great work here.

I have one last suggestion below, but otherwise it looks good to me.

WeirAE (approved):
Fantastic work going through this tricky driver and all of everyone's feedback!

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,135 @@
...
+        taskname = self._taskname("validate_only")
+        yield taskname
+        a = asset(None, lambda: False)
+        yield a
+        config_dep = self.configuration_file()
+        path = refs(config_dep)
+        executable = Path(self._driver_config["execution"]["executable"])
+        yield [file(executable), config_dep]
+        env = " && ".join(self._driver_config["execution"]["envcmds"])
+        cmd = "{env} && time {x} --validate-only {p} 2>&1".format(env=env, x=executable, p=path)
```
maddenp-noaa, line 114:
```suggestion
        executable = file(Path(self._driver_config["execution"]["executable"]))
        config = self.configuration_file()
        yield [executable, config]
        cmd = "%s && time %s --validate-only %s 2>&1" % (
            " && ".join(self._driver_config["execution"]["envcmds"]),
            executable,
            refs(config)
        )
```
Nice work here -- I think this is now properly validating the config file in the run directory, which may be modified by `update_values` in the UW YAML. I think the proposed edit is equivalent and eliminates a couple of temporary variables.

src/uwtools/drivers/jedi.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Run directory provisioned with all required content.
+        """
+        yield self._taskname("provisioned run directory")
+        yield [
+            self.configuration_file(),
+            self.files_copied(),
+            self.files_linked(),
+            self.runscript(),
+        ]
```
maddenp-noaa, line 87: I think that, if we add `self.validate_only()` to this list, we'll get validation of the config file automatically when we run the `provisioned_run_directory` task.

src/uwtools/drivers/jedi.py
NaureenBharwaniNOAA: @maddenp-noaa Based on your suggested change for the jedi driver `validate_only` task. I updated the test case for `validate_only` to use the `%s` syntax instead of the f string it originally had. 

In addition, the expected and actual results vary between what is set in the test case and what is returned from the jedi driver. The actual value shows the `MagicMock` value and I don't think this is what we are looking for here since it seems to vary run to run. The error message is below:

```
E           Expected: run('20240201 18Z jedi validate_only', 'module load some-module && module load jedi-module && time /path/to/qg_forecast.x --validate-only /tmp/pytest-of-Naureen.Bharwani/pytest-31/test_JEDI_validate_only0/jedi.yaml  2>&1')
E             Actual: run('20240201 18Z jedi validate_only', "module load some-module && module load jedi-module && time <MagicMock name='file()' id='140405575998192'> --validate-only /tmp/pytest-of-Naureen.Bharwani/pytest-31/test_JEDI_validate_only0/jedi.yaml 2>&1")
```

I modified your suggestion slightly to be:
```
        executable = Path(self._driver_config["execution"]["executable"])
        config = self.configuration_file()
        yield [file(executable), config]
        env = " && ".join(self._driver_config["execution"]["envcmds"])
        cmd = "%s && time %s --validate-only %s 2>&1" % ( 
            env,
            executable,
            refs(config),
        )   
```

When passing executable with `file()` wrapped around it to construct the command, it was passing in the MagicMock object and not the executable path itself. Let me know if this looks good.

src/uwtools/drivers/jedi.py
NaureenBharwaniNOAA: Question on this, the assert in the test cases call:
```
def test_JEDI_provisioned_run_directory(driverobj):
    with patch.multiple(
        driverobj,
        files_copied=D,
        files_linked=D,
        runscript=D,
        configuration_file=D,
    ) as mocks:
        driverobj.provisioned_run_directory()
    for m in mocks:
        mocks[m].assert_called_once_with()
```

Here now that the `validate_only` task calls `configuration_file` task. We are actually calling `configuration_file` task twice. I was looking into the `unittest.mock` references to `assert_called_...` and there are a couple options [here](https://docs.python.org/3/library/unittest.mock.html). Should we use `assert_has_calls` for the `configuration_file` task and if so I'm wondering what these would look like.

src/uwtools/drivers/jedi.py
maddenp-noaa: I opened https://github.com/NaureenBharwaniNOAA/workflow-tools/pull/5 to try to help with this. Please feel free to merge it if you're ok with it.

maddenp-noaa (approved):
Nice work, looks good.

PR #455, opened 2024-04-04
UW-565: Reduce a config to one of its subsection

christinaholtNOAA:
**Synopsis**

The ``--output-block`` flag was added to config realize to indicate that only a particular subsection of an input config file should be used to prepare output.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/cli/tools/config.rst
+++ b/docs/sections/user_guide/cli/tools/config.rst
@@ -154,10 +154,11 @@ In ``uw`` terminology, to realize a configuration file is to transform it from i
 
 .. code-block:: text
 
-   $ uw config realize --help
+   $ uw config realize -h
    usage: uw config realize [-h] [--version] [--input-file PATH] [--input-format {ini,nml,sh,yaml}]
                             [--output-file PATH] [--output-format {ini,nml,sh,yaml}]
-                            [--values-needed] [--total] [--dry-run] [--quiet] [--verbose]
+                            [--output-block KEY[:KEY:...]] [--values-needed] [--total] [--dry-run]
+                            [--quiet] [--verbose]
                             [PATH ...]
 
    Realize config
@@ -175,6 +176,8 @@ In ``uw`` terminology, to realize a configuration file is to transform it from i
          Path to output file (defaults to stdout)
      --output-format {ini,nml,sh,yaml}
          Output format
+     --output-block KEY[:KEY:...]
+         Colon-separated path of keys to the block to be output
      --values-needed
          Print report of values needed to render template
      --total
@@ -356,6 +359,18 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
        message: 'Good Night Moon Good Night Moon '
        recipient: Moon
        repeat: 2
+
+* To read the config from ``stdin`` and realize a subsection to ``stdout`` in a different format:
+
+  .. code-block:: text
+
+     $ cat config.yaml | uw config realize --input-format yaml --output-format sh --output-block values values1.yaml
+     date=20240105
+     empty=None
+     greeting='Good Night'
+     message='Good Night Moon Good Night Moon '
+     recipient=Moon
+     repeat=2
 
 * If the config file has an unrecognized (or no) extension, ``uw`` will not know how to parse its contents:
 
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -94,6 +94,7 @@ def get_yaml_config(config: Union[dict, Optional[Path]] = None) -> _YAMLConfig:
 def realize(
     input_config: Union[dict, _Config, Optional[Path]] = None,
     input_format: Optional[str] = None,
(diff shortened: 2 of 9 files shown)


NaureenBharwaniNOAA (approved):
This looks good to me! I took a look at the documentation [here](https://uwtools--455.org.readthedocs.build/en/455/sections/user_guide/cli/tools/config.html) as well.

WeirAE (approved):
Useful change, thanks!

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -154,10 +154,11 @@ In ``uw`` terminology, to realize a configuration file is to transform it from i
 
 .. code-block:: text
 
-   $ uw config realize --help
+   $ uw config realize -h
```
maddenp-noaa, line 157: All the other docs use `--help` here, FWIW.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -175,6 +176,8 @@ In ``uw`` terminology, to realize a configuration file is to transform it from i
          Path to output file (defaults to stdout)
      --output-format {ini,nml,sh,yaml}
          Output format
+     --output-block KEY[:KEY:...]
```
maddenp-noaa, line 179: I wonder if `.` might be a better separator for this use case, since it would match Jinja2's syntax, e.g. `{{ a.b.c }}`. I think colons make sense as a path separator since they're used in the shell `PATH` environment variable. This was mentioned in the AC for [UW-559](https://jira.epic.oarcloud.noaa.gov/browse/UW-559), which is a similar use case, so maybe the use of colons here is an intentional change.

src/uwtools/api/config.py
```diff
@@ -94,6 +94,7 @@ def get_yaml_config(config: Union[dict, Optional[Path]] = None) -> _YAMLConfig:
 def realize(
     input_config: Union[dict, _Config, Optional[Path]] = None,
     input_format: Optional[str] = None,
+    output_block: Optional[List[Union[str, int]]] = None,
```
maddenp-noaa, line 97: Per the YAML spec (https://yaml.org/spec/1.2.2/#mapping, section 10.1.1.1)
> YAML places no restrictions on the type of [keys](https://yaml.org/spec/1.2.2/#nodes); in particular, they are not restricted to being [scalars](https://yaml.org/spec/1.2.2/#scalars).

So we might consider expanding the type of this argument. The `uwtools.config.jinja2` module has a `_ConfigVal` custom type that could maybe be copied and pared down (removing the `UWYAMLConvert` and `UWYAMLRemove` types) for this purpose.

src/uwtools/api/config.py
```diff
@@ -107,6 +108,7 @@ def realize(
     _realize(
         input_config=_ensure_config_arg_type(input_config),
         input_format=input_format,
+        output_block=output_block,
```
maddenp-noaa, line 111: The docstrings for `realize()` and `realize_to_dict()` need to be updated to include the new `output_block` argument. At the moment, this argument is undocumented in the API docs.

src/uwtools/config/tools.py
```diff
@@ -79,6 +79,7 @@ def config_check_depths_update(config_obj: Union[Config, dict], target_format: s
 def realize_config(
     input_config: Union[Config, Optional[Path]] = None,
     input_format: Optional[str] = None,
+    output_block: Optional[List[Union[str, int]]] = None,
```
maddenp-noaa, line 82: The docstring for this function needs to be updated, too.

maddenp-noaa (commented):
Sorry I wasn't available for the initial review, but I've added a few comments here. The one regarding the API docs probably needs to be addressed, whatever we decide about the others.

docs/sections/user_guide/cli/tools/config.rst
christinaholtNOAA: Wow! We literally just discussed that one and I missed the dot-separation. I like it better and will update.

src/uwtools/api/config.py
christinaholtNOAA: I'm not even entirely sure I still support my initial call to include `int` in the options here. Despite the wide open options that YAML has, I'm not entirely sure UW will want to support all of those. We found that allowing `int` keys leads to problems when we try to expand the dict's `a` and `b` with this sort of syntax:

```
{**a, **b}
```

Namespace becomes a real issue with boolean values. We won't be able to (I don't think) navigate to a list/dict entry in the YAML from the command line, and I'm not entirely sure what that even means to have those objects as keys. 

Perhaps we do want to limit UW YAML to scalar strings?

maddenp-noaa: All good points. I think that string-only keys would be a good convention to try for. If something comes up that makes that infeasible, I'm sure we can figure something out, but it's a sane starting position.

PR #467, opened 2024-04-22
UW-573 Add regional_esg_grid driver.

NaureenBharwaniNOAA:
**Synopsis**

- Added drivers for regional_esg_grid
- Added regional_esg_grid introduced in CLI and API
- Added JSON Schema validation
- Added Documentation for CLI, API, and landing page
- Added shared documentation on regional_esg_grid yaml

Rendered docs are available for review here: https://uwtools--467.org.readthedocs.build/en/467/
**Type**

**Impact**

**Checklist**


docs/sections/user_guide/yaml/components/esg_grid.rst
```diff
@@ -0,0 +1,28 @@
...
+------------------------------------
+
+execution:
+^^^^^^^^^^
+
+See :ref:`here <execution_yaml>` for details.
+
+namelist:
+^^^^^^^^^
+Supports ``base_file:`` and ``update_values:`` blocks (see the :ref:`updating_values` for details). Namelist options are described :ufs-utils:`here<id54>`.
```
NaureenBharwaniNOAA, line 23: Still working on getting the link to work correctly.

docs/conf.py
```diff
@@ -48,6 +48,7 @@
     "noaa": ("https://www.noaa.gov/%s", "%s"),
     "pylint": ("https://pylint.readthedocs.io/en/stable/%s", "%s"),
     "pytest": ("https://docs.pytest.org/en/7.4.x/%s", "%s"),
+    "regional-esg-grid": ("https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#%s", "%s"),
```
NaureenBharwaniNOAA, line 51: the driver itself is called `esg_grid`, but the documentation is referenced as `regional_esg_grid`.

docs/index.rst
```diff
@@ -109,6 +109,12 @@ Drivers for UFS
 
 To prepare a complete forecast, drivers would typically be run in the order shown here (along with additional drivers still in development).
 
+esg_grid
```
NaureenBharwaniNOAA, line 112: I believe the `esg_grid` component should come before the `sfc_climo_gen` and also before Christina's open PR #469.

docs/sections/user_guide/cli/drivers/esg_grid.rst
```diff
@@ -0,0 +1,97 @@
+``esg_grid``
+============
+
+The ``uw`` mode for configuring and running the :ufs-utils:`regional_esg_grid<regional-esg-grid>` component.
```
NaureenBharwaniNOAA, line 4: I'm working on this link for the `esg_grid`. The naming convention we went with for this driver is `esg_grid`, and the UFS Doucmentation for the driver is called under `regional_esg_grid` causing problems with linking. I haven't been able to figure this one out and it's causing a broken link if I use `esg_grid` here instead of `regional_esg_grid`, but after looking at the documentation it still doesn't link correctly.

The broken link address: https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#esg-grid
The correct link address: https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#regional-esg-grid

docs/sections/user_guide/yaml/components/esg_grid.rst
```diff
@@ -0,0 +1,28 @@
...
+-----------------------------------------
+
+execution:
+^^^^^^^^^^
+
+See :ref:`here <execution_yaml>` for details.
+
+namelist:
+^^^^^^^^^
+Supports ``base_file:`` and ``update_values:`` blocks (see the :ref:`updating_values` for details). Namelist options are described :ufs-utils:`regional_esg_grid<regional-esg-grid>`.
```
NaureenBharwaniNOAA, line 23: Same here with the link not working as desired.

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -8,7 +8,7 @@ The ``uw`` mode for configuring and running the :ufs-utils:`chgres_cube<chgres-c
    $ uw chgres_cube --help
    usage: uw chgres_cube [-h] [--version] TASK ...
 
-   Execute chgres_cube tasks
+   Execute Chgres Cube tasks
```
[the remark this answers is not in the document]
maddenp-noaa: I appreciate the commitment to consistency, but am not sure about this change, or the related change w.r.t. `sfc_climo_gen`. For some drivers, there's some kind of identity that we can refer to at a higher level than the actual name of the executable, for example "FV3" and "JEDI". In fact, for the FV3 driver, in a particular UFS application, the executable name might be something like `ufs_model`, which is definitely not the name we'd want to name the whole driver.

For `chgres_cube`, though, I don't think there's a better name. If you google "Chgres Cube", you won't find that actual text anywhere on the internet, and only a couple instances of `chgres cube` that may be typos. But googling "chgres_cube" turns up lots of hits, as this is how that program is actually referred to in its documentation. Similarly for "Sfc Climo Gen" vs "sfc_climo_gen".

So I'd probably vote for some small inconsistencies in our internal naming in order to be more consistent with the names of these components vs their own chosen names and documentation. I'd be curious to know how others feel about this, though.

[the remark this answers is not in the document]
christinaholtNOAA: I agree with Paul's assessment here.

docs/sections/user_guide/yaml/components/esg_grid.rst
maddenp-noaa: The link looks fine to me when I build and view the docs locally. The docs built by RtD, accessible via the PR, look ok to me to. Maybe I'm missing what the problem is.

src/uwtools/api/esg_grid.py
```diff
@@ -0,0 +1,56 @@
...
+def execute(
+    task: str,
+    config: Optional[Union[Path, str]] = None,
+    batch: bool = False,
+    dry_run: bool = False,
+    graph_file: Optional[Union[Path, str]] = None,
+    stdin_ok: bool = False,
+) -> bool:
+    """
+    Execute a ``esg_grid`` task.
```
maddenp-noaa, line 24:
```suggestion
    Execute an ``esg_grid`` task.
```

src/uwtools/cli.py
```diff
@@ -87,7 +89,7 @@ def _add_subparser_chgres_cube(subparsers: Subparsers) -> ModeChecks:
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.chgrescube, "Execute chgres_cube tasks")
+    parser = _add_subparser(subparsers, STR.chgrescube, "Execute chgres_cube  tasks")
```
maddenp-noaa, line 92:
```suggestion
    parser = _add_subparser(subparsers, STR.chgrescube, "Execute chgres_cube tasks")
```

src/uwtools/cli.py
```diff
@@ -290,6 +292,57 @@ def _dispatch_config_validate(args: Args) -> bool:
     )
 
 
+# Mode esg grid
```
maddenp-noaa, line 295:
```suggestion
# Mode esg_grid
```

src/uwtools/cli.py
```diff
@@ -290,6 +292,57 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+
+
+def _add_subparser_esg_grid(subparsers: Subparsers) -> ModeChecks:
+    """
+    Subparser for mode: esg_grid
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.esggrid, "Execute Esg Grid tasks")
```
maddenp-noaa, line 303:
```suggestion
    parser = _add_subparser(subparsers, STR.esggrid, "Execute esg_grid tasks")
```

src/uwtools/cli.py
```diff
@@ -290,6 +292,57 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+def _add_subparser_esg_grid_task(subparsers: Subparsers, task: str, helpmsg: str) -> ActionChecks:
+    """
+    Subparser for mode: esg_grid <task>
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    :param task: The task to add a subparser for.
+    :param helpmsg: Help message for task.
+    """
+    parser = _add_subparser(subparsers, task, helpmsg.rstrip("."))
+    required = parser.add_argument_group(TITLE_REQ_ARG)
+    _add_arg_config_file(group=required, required=True)
```
maddenp-noaa, line 321: Check the other drivers, but I think `--config-file` should be optional, with a read from `stdin` performed if no value is supplied.

src/uwtools/cli.py
```diff
@@ -290,6 +292,57 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+    _add_arg_batch(optional)
+    _add_arg_dry_run(optional)
+    _add_arg_graph_file(optional)
+    checks = _add_args_verbosity(optional)
+    return checks
+
+
+def _dispatch_esg_grid(args: Args) -> bool:
+    """
+    Dispatch logic for esg grid mode.
```
maddenp-noaa, line 332:
```suggestion
    Dispatch logic for esg_grid mode.
```

docs/shared/esg_grid.yaml
```diff
@@ -0,0 +1,22 @@
+esg_grid:
+  execution:
+    batchargs:
+      nodes: 1
+      stdout: /path/to/runscript.out
+      walltime: "08:00:00"
+    executable: /path/esg_grid
+    mpicmd: srun
```
maddenp-noaa, line 8: I think the notes doc for this component said that this is a serial (non-MPI) component, in which case we probably shouldn't provide a `mpicmd` in this example YAML.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
```diff
@@ -0,0 +1,84 @@
...
+              "properties": {
+                "regional_grid_nml": {
+                  "additionalProperties": {
+                    "type": [
+                      "array",
+                      "boolean",
+                      "number",
+                      "string"
+                    ]
+                  },
```
maddenp-noaa, line 37:
```suggestion
                  "additionalProperties": false
```
The notes doc indicated that we didn't think we'd need additional properties for this component. If not, we could simplify this.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
```diff
@@ -0,0 +1,84 @@
...
+                    "ly": {
+                      "type": "number"
+                    },
+                    "pazi": {
+                      "type": "number"
+                    },
+                    "plat": {
+                      "type": "number"
+                    },
+                    "plon": {
```
maddenp-noaa, line 57: I don't know which is correct, but this key is listed as `plot` in the sample config under `docs/shared`. If this is correct, and if the sample config was tested with `uw esg_grid validate` and passed, then the permissive `additionalProperties` above would be why. But I guess this is probably `plot` given the ordering...

We might also consider making these values required if they're required to be set in the namelist.

src/uwtools/tests/drivers/test_esg_grid.py
```diff
@@ -0,0 +1,140 @@
...
+            "namelist": {
+                "update_values": {
+                    "regional_grid_nml": {
+                        "delx": 0.11,
+                        "dely": 0.11,
+                        "lx": -214,
+                        "ly": -128,
+                        "pazi": 0.0,
+                        "plat": 38.5,
+                        "plon": -97.5,
```
maddenp-noaa, line 46: Here, too, is this supposed to be `plon` or `plot`? If it's `plon` it's out-of-order; if it's `plot` it's misspelled :)

NaureenBharwaniNOAA: The YAML options for the namelist should be: delx, dely, lx, ly, pazi, plat, and plon. I'll update the YAML in `docs/shared` for this entry.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
NaureenBharwaniNOAA: Based on the example code, they should all be required. I'll add a section for this and `plon` is the correct key name based on the [documentation](https://github.com/ufs-community/UFS_UTILS/blob/ef3a1d2c39d0620692ec1bb8026d587cb40dca4f/sorc/grid_tools.fd/regional_esg_grid.fd/pesg.f90#L1820).

src/uwtools/tests/drivers/test_esg_grid.py
maddenp-noaa: My mistake, `plon` is not out-of-order!

docs/sections/user_guide/cli/drivers/esg_grid.rst
NaureenBharwaniNOAA: When I looked earlier they still weren't working properly/ Seems like they are now. Possibly just had to wait for them to build again.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
NaureenBharwaniNOAA: Wondering a bit more on 

> We might also consider making these values required if they're required to be set in the namelist.

When we want them to be required, we'd like them to be required in the `namelist` object, but they don't necessarily need to be supplied by the user every time they want to update their base files. So in the `namelist` object we should be adding a `required` object and in addition to having the `namelist` options in the `update_values` object as optional, we should also be adding them in the `namelist` object as required.

Did I understand your suggestion right? Or do we simply make these required in the `update_values` object and update the schema tests accordingly. One test might look like:

src/uwtools/resources/jsonschema/esg-grid.jsonschema
```diff
@@ -0,0 +1,119 @@
+{
+  "allOf": [
+    {
+      "if": {
+        "not": {
+          "required": [
+            "namelist.base_file"
+          ]
+        }
+      },
```
NaureenBharwaniNOAA, line 10: Added in an IF - THEN conditional to confirm that when the `base_file` is not supplied, we will require all the `namelist` options be required in the `uodates_values` object.

src/uwtools/cli.py
NaureenBharwaniNOAA: The other drivers all list this as optional. I'll update this.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
```diff
@@ -0,0 +1,119 @@
...
+      "then": {
+        "properties": {
+          "namelist": {
+            "properties": {
+              "update": {
```
maddenp-noaa, line 15: Should this be `update_values`? If so, I'm guessing that we need schema unit tests to back up what the schema says here.

src/uwtools/tests/drivers/test_esg_grid.py
```diff
@@ -0,0 +1,140 @@
...
+                    "stdout": "/path/to/file",
+                    "walltime": "00:02:00",
+                },
+                "envcmds": [
+                    "module load some-module",
+                    "module load esg-grid-module",
+                ],
+                "executable": "/path/to/esg_grid",
+                "mpiargs": ["--export=ALL", "--ntasks $SLURM_CPUS_ON_NODE"],
+                "mpicmd": "srun",
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I implemented the second suggestion by duplicating what @christinaholtNOAA had in her PR.

src/uwtools/resources/jsonschema/execution-serial.jsonschema
NaureenBharwaniNOAA, line 1: @maddenp-noaa Added this new schema for drivers that do not need to support mpi

src/uwtools/tests/test_schemas.py
```diff
@@ -90,6 +95,87 @@ def test_schema_chgres_cube_run_dir(chgres_cube_prop):
...
+    assert not errors(config)
+    # Some top-level keys are required:
+    for key in ("execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(config, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+    # Additional mpi support is not allowed:
+    assert "Additional properties are not allowed ('mpicmd' was unexpected)" in errors(
+        {"execution": {"mpicmd": "srun"}}
+    )
```
NaureenBharwaniNOAA, line 117: @christinaholtNOAA @maddenp-noaa Added this test case to confirm `mpi` support is not allowed.

src/uwtools/tests/drivers/test_esg_grid.py
```diff
@@ -0,0 +1,134 @@
...
+def config(tmp_path):
+    return {
+        "esg_grid": {
+            "execution": {
+                "batchargs": {
+                    "export": "NONE",
+                    "nodes": 1,
+                    "stdout": "/path/to/file",
+                    "walltime": "00:02:00",
+                },
```
NaureenBharwaniNOAA, line 28: Based on Christina's open PR, some of them might not be needed in the `execution` block. Specifically `export` and `stdout`. I also have `nodes` here instead of `cores`. Not sure which values are appropriate.

src/uwtools/resources/jsonschema/execution-serial.jsonschema
maddenp-noaa: This is looking very good. I think we could still use some additional schema unit tests demonstrating various correct and incorrect `namelist` block content, e.g.
1. `base_file` with no `update_values` is ok.
2. `base_file` with some subset of supported `update_values` is ok.
3. `update_values` with no `base_file` is ok if `update_values` is complete.
4. `update_values` with no `base_file` fails validation if any key/value pair is missing.
5. Any `update_values` value that isn't the right data type (e.g. a `str` value instead of a `number` fails.
6. It looks like our current schema demands at least one of `base_file` and `update_values`, so that edge case we discussed of a user placing a namelist file in the run directory before running the driver is probably not currently supported -- I mean, it'll still work, but config validation will still complain if both `base_file` and `update_values` are missing. I think. It'd be good to make sure the schema unit tests check this and have a comment explaining why it is the way it is.

Some of this might already be covered, but I'm pretty sure not all of it is. And you might think of something else. Sorry for piling up so much rigor here, but I think your work here is nailing down some important behavioral invariants for the drivers. Thank you for your persistence!

src/uwtools/tests/test_schemas.py
```diff
@@ -90,6 +95,87 @@ def test_schema_chgres_cube_run_dir(chgres_cube_prop):
...
+    }
+    errors = schema_validator("esg-grid", "properties", "esg_grid")
+    # Basic correctness:
+    assert not errors(config)
+    # Some top-level keys are required:
+    for key in ("execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(config, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+    # Additional mpi support is not allowed:
```
maddenp-noaa, line 114:
```suggestion
    # Additional MPI support is not allowed:
```

src/uwtools/tests/test_schemas.py
maddenp-noaa: Excellent.

src/uwtools/tests/drivers/test_esg_grid.py
maddenp-noaa: Per [our docs](https://uwtools.readthedocs.io/en/main/sections/user_guide/yaml/execution.html#batchargs), the only required entry is `walltime`, but I just tested with the `docs/shared/ungrib.yaml` and schema validation passes without it. There might be a runtime error later, but I'm not immediately seeing where it would come from. Probably not for this PR, but we should consider making `walltime` schema-required if we mean what we say in the docs.

maddenp-noaa: Oh, maybe `_validate_props()` in `uwtools.scheduler` enforces this. It looks like that logic also requires `account`, but that is provided in the `platform:` block, which has its own schema that does require `account` if `scheduler` is specified. So, maybe our `execution` `batchargs` subschema should require `walltime`... But maybe we should make a note of this rather than hold up this PR for it.

src/uwtools/resources/jsonschema/esg-grid.jsonschema
```diff
@@ -0,0 +1,119 @@
...
+              "required": [
+                "base_file"
+              ]
+            },
+            {
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
```
maddenp-noaa, line 55: I see now that it *is* ok for neither `base_file` or `update_values` to be provided: Just don't provide the `namelist` block at all! But if it _is_ provided, then at least one of `base_file` or `update_values` is required. That makes sense, and means that users can place their own namelist file in the run directory and not mention `namelist` at all in their UW YAML. Sorry for the confusion.

src/uwtools/resources/jsonschema/execution-serial.jsonschema
NaureenBharwaniNOAA: I worked on these and added almost all test case suggestion. Working on two test cases that are a bit trickier.

src/uwtools/tests/test_schemas.py
```diff
@@ -90,6 +96,117 @@ def test_schema_chgres_cube_run_dir(chgres_cube_prop):
...
+    #  )
+    # update_values values must be a number:
+    assert "'/some/str' is not of type 'number'" in errors(
+        {"update_values": {"regional_grid_nml": {"delx": "/some/str"}}}
+    )
+    # regional_grid_nml is required with update_values:
+    assert "'regional_grid_nml' is a required property" in errors({"update_values": {}})
+    # At least one is required:
+    assert "is not valid" in errors({})
+    # Just namelist file is not sufficient:
```
NaureenBharwaniNOAA, line 178: @maddenp-noaa I'm working on this suggestion below

> I see now that it *is* ok for neither `base_file` or `update_values` to be provided: Just don't provide the `namelist` block at all! But if it _is_ provided, then at least one of `base_file` or `update_values` is required. That makes sense, and means that users can place their own namelist file in the run directory and not mention `namelist` at all in their UW YAML. Sorry for the confusion.

> It looks like our current schema demands at least one of `base_file` and `update_values`, so that edge case we discussed of a user placing a namelist file in the run directory before running the driver is probably not currently supported -- I mean, it'll work, but config validation will still complain if both `base_file` and `update_values` are missing. I think. It'd be good to make sure the schema unit tests check this and have a comment explaining why it is the way it is.

I added a YAML above called `config` and I am getting an error of:
```
E       TypeError: schema_validator.<locals>.<lambda>() got an unexpected keyword argument 'regional_grid_nml'
```

I thought I can record an exception here for a TypeError and then log "Just namelist file is not sufficient please supply `base_file` or `update_values`. But re-reading your comment maybe this isn't what you were looking for? Maybe we can discuss more on this together in UW Group.

src/uwtools/tests/test_schemas.py
```diff
@@ -90,6 +96,117 @@ def test_schema_chgres_cube_run_dir(chgres_cube_prop):
...
+        {"update_values": {"regional_grid_nml": {"delx": 0.11, "lx": -180, "plat": 38.0}}}
+    )
+
+    # Subsection of update_values is ok if base_file is supplied:
+    #  assert not errors(
+    #      {
+    #          **base_file,
+    #          "update_values": {"regional_grid_nml": {"delx": 0.11, "lx": -180, "plat": 38.0}},
+    #      }
+    #  )
```
NaureenBharwaniNOAA, line 169: I thought providing `base_file` and a subsection of `update_values` would pass the schema, but it's currently failing unless I supply all the key/value pairs for `update_values`. I thought this assertion should pass the schema, but now i'm starting to second guess if it actually correctly presenting an error. 

Thoughts?

src/uwtools/resources/jsonschema/execution-serial.jsonschema
NaureenBharwaniNOAA: all the suggestions above were added to the `test_schema_esg_grid_namelist` [test case](2bf3204d5e478d61c3f59d7dfbc09d9577fec1c65724e35c5134e2a3b5d4ea4fR121):

src/uwtools/tests/drivers/test_esg_grid.py
NaureenBharwaniNOAA: Sounds great, would the best way to document this to be to open an issue and tag this discussion?

src/uwtools/resources/jsonschema/esg-grid.jsonschema
maddenp-noaa: It's not that `base_file` is optional, it's that all of `namelist` is optional. We've previously discussed leaving open the possibility that a user might place their own namelist file in a run directory. Of course, they could also automate that by using `base_file`, but they might do it completely outside the driver. It looks like this is the choice we've made in all the driver schemas so far: `namelist` is optional, but if you provide it, you have to provide at least one of `base_file` and/or `update_values`. If `namelist` isn't provided, and if the driver doesn't find it in the run directory already, we'll get `WARNING` messages about the unfulfilled requirement. (Well, I see that in `chgres_cube` we `["namelist"]` instead of `.get("namelist")`, so we should fix that and check the other drivers for agreement between the schema and the `dict` access method used for optional items.)

So... I think we should decide what we really want and then double check that that's the case. I'm happy to do the check-and-fix part, but a design decision has to come first.

src/uwtools/tests/test_schemas.py
maddenp-noaa: I don't think I completely understand what you were hoping to do, and see that `cfg` isn't actually used, but based on some other, more recent comments, I think the only thing we need to test -- if it's not already tested, and it might be -- is that *if* `namelist` is present (and it is currently optional), then at least one of `base_file` and/or `update_values` must also be present. That's what, if I'm not mistaken, the schema is written to express.

src/uwtools/tests/test_schemas.py
maddenp-noaa: I think the schema might not be constructed properly: The single item in the `allOf` list at the very top has an `if` item that refers to something called `namelist.base_file`, but I wouldn't think, at that level of the JSON structure, that would refer to what we'd hope: The `base_file` block, from this top level, is found by following the keys `properties` -> `esg_grid` -> `properties` -> `namelist` -> `properties` -> `base_file`. Maybe JSON Schema understand a simplified version of that that elides the `properties` steps/keys, but at the very least I assume the `esg_grid` level would have to be taken into account for a relative reference. I don't think I've done anything in JSON Schema with anything but immediate-neighbor properties, so this is a gray area for me.

Also, the `then` item of that block just refers to `namelist`, and I can't see how JSON Schema could connect that to the more deeply nested `namelist` block that we want to talk about.

I feel like the definition of which kind of `update_values` block is expected should be happening deeper in the schema, closer to the `base_file` and `update_values` blocks themselves.

docs/sections/user_guide/cli/drivers/esg_grid.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. code-block:: text
+
+   $ uw esg_grid --help
+    usage: uw esg_grid [-h] [--version] TASK ...
+    
+    Execute Esg Grid tasks
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I feel like I saw a comment although I can't find it now to update this to `esg_grid tasks`. Is that right?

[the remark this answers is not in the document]
maddenp-noaa: This should be whatever `uw esg_grid --help` shows, and I think you already updated the code.

docs/shared/esg_grid.yaml
```diff
@@ -0,0 +1,21 @@
+esg_grid:
+  execution:
+    batchargs:
+      nodes: 1
+      stdout: /path/to/runscript.out
+      walltime: "08:00:00"
+    executable: /path/esg_grid
+  namelist:
+    update_values:
```
NaureenBharwaniNOAA, line 9: Looking at this again, I don't have a `base_file` entry here and I think it should be added based on our previous conversations.

maddenp-noaa: Since it's used for documentation, I agree with you: Showing as complete an example as possible probably helps readers.

src/uwtools/drivers/esg_grid.py
```diff
@@ -0,0 +1,100 @@
...
+from typing import Optional
+
+from iotaa import asset, dryrun, task, tasks
+
+from uwtools.config.formats.nml import NMLConfig
+from uwtools.drivers.driver import Driver
+from uwtools.strings import STR
+
+
+class EsgGrid(Driver):
```
maddenp-noaa, line 15: I'm sorry to make this comment so late in the game, but I think that maybe `ESGGrid` would be the commonly preferred form for this name, given that "ESG" ("Extended Schmidt Gnomonic" -- I just looked it up) is an acronym. Similarly, our JEDI driver is called `JEDI` and not `Jedi`, because it's an acronym; or `FV3` and not `Fv3`. I don't think we should change things like `esg_grid` or `esg-grid`, but probably `ESGGrid` as a class name and `ESG Grid` as a human-text descriptor when referring to the tool and not the Python class, would be appropriate.

NaureenBharwaniNOAA: I'll work on switching this over. Feel free to point out any I miss in my later commits.

src/uwtools/tests/test_schemas.py
```diff
@@ -90,6 +95,140 @@ def test_schema_chgres_cube_run_dir(chgres_cube_prop):
...
+    # A combination of base_file and update_values is ok:
+    assert not errors({**base_file, **update_values})
+    # All key/value pairs in update_values must be present if base_file is not supplied:
+    # pylint: disable=line-too-long
+    assert (
+        "{'update_values': {'regional_grid_nml': {'delx': 0.11, 'lx': -180, 'plat': 38.0}}} is not valid under any of the given schemas"
+        in errors(
+            {"update_values": {"regional_grid_nml": {"delx": 0.11, "lx": -180, "plat": 38.0}}}
+        )
+    )
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I did run into trouble trying to use `with_del`. When I tried to do
`"{'update_values': {}} is not valid under any of the given schemas in {'base_file': 'some/str', 'update_values': {}} is not valid under any of the given schemas" in "{'base_file': 'some/str', 'update_values': {}} is not valid under any of the given schemas`

I was having a problem with this second `in` without the assert statement because it was showing up as a special character. So I went about it this way to avoid that.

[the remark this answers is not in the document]
NaureenBharwaniNOAA: I tried your suggestion of:

> I also think that a check for just is not valid under any of the given schemas would be fine here, without the detail

That worked great so I'll simplify that!

christinaholtNOAA (approved):
Thanks for your extra work here that applies to both our drivers. This looks great.

PR #469, opened 2024-04-23
UW 574: Add global_equiv_resol driver

christinaholtNOAA:
**Synopsis**

Adds the global_equiv_resol driver, along with its tests, tested JSONschema, and documentation.

Rendered docs are here: https://uwtools--469.org.readthedocs.build/en/469/
**Type**

- [ ] Bug fix (corrects a known issue)
- [ ] Code maintenance (refactoring, etc. without behavior change)
- [ ] Documentation
- [x] Enhancement (adds a new functionality)
- [ ] Tooling (CI, code-quality, packaging, revision-control, etc.)

**Impact**

- [ ] This is a breaking change (changes existing functionality)
- [x] This is a non-breaking change (existing functionality continues to work as expected)

**Checklist**

- [x] I have added myself and any co-authors to the PR's _Assignees_ list.
- [x] I have reviewed the documentation and have made any updates necessitated by this change.

PR #473, opened 2024-04-24
UW 572: Add make_hgrid driver

WeirAE:
**Synopsis**

Adds the make_hgrid driver, along with its tests, tested JSONschema, and documentation.

Rendered docs are here: https://uwtools--473.org.readthedocs.build/en/473/

**Type**

- [ ] Documentation

**Impact**

**Checklist**


docs/shared/make_hgrid.yaml
```diff
@@ -0,0 +1,26 @@
+make_hgrid:
+  execution: 
+    batchargs: 
+      nodes: 1
+      stdout: /path/to/runscript.out
+      walltime: "08:00:00"
+    executable: /path/make_hgrid
+  config:
+    grid_type: gnomonic_ed
```
WeirAE, line 9: Example config here remaps the settings from SRW. Several of these are not currently exposed to the user, so I debated configuring special cases, but for now I'm leaving them open

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
```diff
@@ -0,0 +1,119 @@
...
+            "jstart_nest": { "type": "array", "items": { "type": "integer" } },
+            "jend_nest": { "type": "array", "items": { "type": "integer" } },
+            "great_circle_algorithm": { "type": "boolean" },
+            "out_halo": { "type": "integer" },
+            "angular_midpoint": { "type": "boolean" },
+            "non_length_angle": { "type": "boolean" },
+            "rotate_poly": { "type": "boolean" },
+            "verbose": { "type": "boolean" }
+          },
+          "required": ["grid_type"],
```
WeirAE, line 59: I'm currently having issues with the formatter removing my conditionals here, I'm working on a solution but have not solved it yet.

src/uwtools/drivers/make_hgrid.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @property
+    def _driver_name(self) -> str:
+        """
+        Returns the name of this driver.
+        """
+        return STR.makehgrid
+
+    @property
+    def _runcmd(self):
```
WeirAE, line 79: runcmd override has not yet been updated to match the changes in #467 / #469

WeirAE: Checked and confirmed

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,58 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    assert not errors(config)
+    # All top-level keys are required:
+    for key in ("execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(config, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+    # config needs at least one entry:
+    assert "{} should be non-empty" in errors({"config": {}})
+    # the following config keys are allowed:
+    assert not errors(
```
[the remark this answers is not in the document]
WeirAE: Test complete

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
WeirAE: Formatter fixed by fixing some issues with my conditionals!

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,24 @@
...
+
+make_hgrid
+==========
+
+Structured YAML to run :ufs-utils:`make_hgrid<make-hgrid>` is validated by JSON Schema and requires the ``make_hgrid:`` block, described below. If ``make_hgrid`` is to be run via a batch system, the ``platform:`` block, described :ref:`here <platform_yaml>`, is also required.
+
+Here is a prototype UW YAML ``make_hgrid:`` block, explained in detail below:
+
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/make_hgrid.yaml
```
WeirAE, line 11: This may be a good place to show a comparison of the SRW YAML parameters to the UFS_UTILS parameters

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    # config needs at least the grid_type key:
+    assert "'grid_type' is a required property" in errors({"config": {}})
+
+
+def test_schema_make_hgrid_grid_type():
+    # Get errors function from schema_validator
+    errors = schema_validator("make-hgrid", "properties", "make_hgrid")
+
+    # If grid_type is "from_file", my_grid_file is required
+    config = {
```
WeirAE, line 395: I regenerated `config` here each time due to issues in remapping consistently

src/uwtools/drivers/make_hgrid.py
```diff
@@ -0,0 +1,80 @@
+"""
+A driver for chgres_cube.
```
maddenp-noaa, line 2:
```suggestion
A driver for make_hgrid.
```

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
```diff
@@ -0,0 +1,301 @@
...
+      "required": [
+        "execution",
+        "run_dir"
+      ]
+    }
+  },
+  "required": [
+    "make_hgrid"
+  ]
+}
```
maddenp-noaa, line 301: Great thorough work on this!

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,24 @@
...
+
+execution:
+^^^^^^^^^^
+
+See :ref:`here <execution_yaml>` for details.
+
+run_dir:
+^^^^^^^^
+
+The path to the directory where ``make_hgrid`` will write its outputs.
```
maddenp-noaa, line 24: Sorry if I missed it, or if it's still a work in progress, but: The schema seems to permit (and conditionally require) many more keys than just these two. I don't think we should re-document them here if there's documentation we can point to, but maybe we should explicitly say that more is possible than what's documented here, and point to the authoritative docs. If our key names differ from what's in the authoritative docs, we could perhaps give a table here matching each UW YAML key name to the name used in those docs so that the user could understand the meaning of the setting and the kinds of values expected.

src/uwtools/tests/drivers/test_make_hgrid.py
```diff
@@ -0,0 +1,109 @@
+# pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
+"""
+global_equiv_resol driver tests.
```
maddenp-noaa, line 3:
```suggestion
make_hgrid driver tests.
```

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
```diff
@@ -0,0 +1,301 @@
...
+              },
+              "type": "array"
+            }
+          },
+          "required": [
+            "grid_type"
+          ]
+        },
+        "execution": {
+          "$ref": "urn:uwtools:execution"
```
maddenp-noaa, line 286: If this is a serial / non-MPI executable (and since the prototype config in `docs/shared` doesn't include MPI info, I'm guessing it is), we should use `urn:uwtools:execution-serial` here, which is on `main` now (but was probably not when you wrote this).

src/uwtools/drivers/make_hgrid.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        return STR.makehgrid
+
+    @property
+    def _runcmd(self):
+        """
+        Returns the full command-line component invocation.
+        """
+        executable = self._driver_config["execution"]["executable"]
+        return "%s" % (executable)
```
maddenp-noaa, line 72: It turns out that we probably don't need this custom implementation: The one inherited from `Driver` should be fine. The one @christinaholtNOAA wrote for `global_equiv_resolv` needed special treatment, but the one @NaureenBharwaniNOAA added for `esg_grid` turned out not to be necessary, and #475 removes it. If you remove this, all your tests should still pass with no other changes -- I'll be very curious to hear if they don't.

src/uwtools/tests/drivers/test_make_hgrid.py
```diff
@@ -0,0 +1,109 @@
...
+def test_MakeHgrid__runscript_path(driverobj):
+    assert driverobj._runscript_path == driverobj._rundir / "runscript.make_hgrid"
+
+
+def test_MakeHgrid__taskname(driverobj):
+    assert driverobj._taskname("foo") == "make_hgrid foo"
+
+
+def test_MakeHgrid__validate(driverobj):
+    driverobj._validate()  # pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
```
maddenp-noaa, line 109:
```suggestion
    driverobj._validate()
```
The `pylint` suppressors at the top of the module seem to cover this already.

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    for key in ("execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(config, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+    # config needs at least the grid_type key:
+    assert "'grid_type' is a required property" in errors({"config": {}})
+
+
+def test_schema_make_hgrid_grid_type():
+    # Get errors function from schema_validator
```
maddenp-noaa, line 391:
```suggestion
    # Get errors function from schema_validator.
```
I'm going to nitpick-suggest some punctuation on the comments here...

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    errors = schema_validator("make-hgrid", "properties", "make_hgrid")
+
+    # If grid_type is "from_file", my_grid_file is required
```
maddenp-noaa, line 394:
```suggestion
    # If grid_type is "from_file", my_grid_file is required.
```

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    config = {
+        "execution": {"executable": "make_hgrid"},
+        "run_dir": "/tmp",
+        "config": {"grid_type": "from_file"},
+    }
+    assert "'my_grid_file' is a required property" in errors(config)
+
+    # If grid_type is "tripolar_grid" or "regular_lonlat_grid",
+    # nxbnds, nybnds, xbnds, and ybnds are required
```
maddenp-noaa, line 403:
```suggestion
    # nxbnds, nybnds, xbnds, and ybnds are required.
```

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+        config = {
+            "execution": {"executable": "make_hgrid"},
+            "run_dir": "/tmp",
+            "config": {"grid_type": grid_type},
+        }
+        for prop in ("nxbnds", "nybnds", "xbnds", "ybnds"):
+            assert f"'{prop}' is a required property" in errors(config)
+
+    # If grid_type is "simple_cartesian_grid",
+    # nxbnds, nybnds, xbnds, ybnds, simple_dx, and simple_dy are required
```
maddenp-noaa, line 414:
```suggestion
    # nxbnds, nybnds, xbnds, ybnds, simple_dx, and simple_dy are required.
```

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    config = {
+        "execution": {"executable": "make_hgrid"},
+        "run_dir": "/tmp",
+        "config": {"grid_type": "simple_cartesian_grid"},
+    }
+    for prop in ("nxbnds", "nybnds", "xbnds", "ybnds", "simple_dx", "simple_dy"):
+        assert f"'{prop}' is a required property" in errors(config)
+
+    # If grid_type is "f_plane_grid" or "beta_plane_grid", f_plane_latitude is required
```
maddenp-noaa, line 423:
```suggestion
    # If grid_type is "f_plane_grid" or "beta_plane_grid", f_plane_latitude is required.
```

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    for grid_type in ("f_plane_grid", "beta_plane_grid"):
+        config = {
+            "execution": {"executable": "make_hgrid"},
+            "run_dir": "/tmp",
+            "config": {"grid_type": grid_type},
+        }
+        assert "'f_plane_latitude' is a required property" in errors(config)
+
+    # If grid_type is "gnomonic_ed" and nest_grids is present, halo is required
```
maddenp-noaa, line 432:
```suggestion
    # If grid_type is "gnomonic_ed" and nest_grids is present, halo is required.
```

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    config = {
+        "execution": {"executable": "make_hgrid"},
+        "run_dir": "/tmp",
+        "config": {"grid_type": "gnomonic_ed", "nest_grids": 1},
+    }
+    assert "'halo' is a required property" in errors(config)
+
+    # If do_schmidt and do_cube_transform are present,
+    # stretch_factor, target_lat, and target_lon are required
```
maddenp-noaa, line 441:
```suggestion
    # stretch_factor, target_lat, and target_lon are required.
```

src/uwtools/tests/test_schemas.py
maddenp-noaa: I'm curious about what issue you were seeing. It should be possible to DRY out these `config` definitions. If you've got a base `config`, you should be able to test an augmented version with `errors({**config, key: val})` and a diminished version with `errors(with_del(config, key[, key, ...]))`. If it doesn't work out, though, no worries -- but I might take a crack at it in #475.

src/uwtools/tests/test_schemas.py
```diff
@@ -362,6 +367,94 @@ def test_schema_jedi_run_dir(jedi_prop):
...
+    config = {
+        "execution": {"executable": "make_hgrid"},
+        "run_dir": "/tmp",
+        "config": {"grid_type": "gnomonic_ed", "do_schmidt": True, "do_cube_transform": True},
+    }
+    for prop in ("stretch_factor", "target_lat", "target_lon"):
+        assert f"'{prop}' is a required property" in errors(config)
```
maddenp-noaa, line 448: These thorough tests of a complex condition schema look great.

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Yeah, thats what I was thinking in addition to adding comparisons for the existing YAMLs that we're replacing, working on it

src/uwtools/drivers/make_hgrid.py
WeirAE: I knew I'd miss a few copied names - thanks

src/uwtools/drivers/make_hgrid.py
WeirAE: Ah, thanks, this was from before that was updated - I'll make sure to test through to make sure,

src/uwtools/tests/test_schemas.py
WeirAE: I was getting schema mismatches, although it might also have been a typo that tripped me up - I'll have another go but if I get stuck you're welcome to have at it!

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
WeirAE: Yep, I forgot to leave a note that it was pending that update to `main` - I'll add it, thanks!

src/uwtools/tests/test_schemas.py
WeirAE: Cleaned up by defining the `grid_type` in `errors()` and bypassing the base schema entirely

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
```diff
@@ -0,0 +1,301 @@
...
+              },
+              "then": {
+                "required": [
+                  "stretch_factor",
+                  "target_lat",
+                  "target_lon"
+                ]
+              }
+            }
+          ],
```
maddenp-noaa, line 115: Should we add `"minProperties": 3`? We require `execution` and `run_dir`, so that's 2, but then would it ever make sense to run this program with zero arguments? If not, such a check might at least clue the user to the fact that they need to supply more configuration information.

maddenp-noaa:
@WeirAE #475 includes a fix for the `src/fv3` test problem -- you can safely ignore it in your PR.

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
WeirAE: We could also just require Config here, since from_file is included there, though I'll have to tweak the tests

maddenp-noaa: That's an even better idea. There's really no point in running the program without configuring it, right?

WeirAE: Hrm. When I require `config` and try to update the example to have `config: {"grid_type": "from_file", "my_grid_file": "/path/to/my_grid_file"}` it is failing the schema trying to require the last conditional entries `["stretch_factor", "target_lat", "target_lon"]` and I've had no success modifying it to stop this. And its weird because it seemed to work correctly in the schema test before!

maddenp-noaa: I checked out your branch, made `"config"` required in the schema, then added a `"config": {"grid_type": "from_file", "my_grid_file": "/some/file"}` entry in the `config` fixture in `test_make_hgrid.py`. I then get the error you note above. I notice that if I remove _only_ the last of the `allOf` items -- the one that requires `stretch_factor`, `target_lat`, and `target_lon` -- the errors disappear. Maybe it's something about this specific item, though I don't immediately see what it could be. I suppose it's not just that the _last_ item, because, of course, when this one is removed the previous one becomes the "last item", and it's not causing a problem. I don't see yet what's happening...

WeirAE: So I added `allOf` to ensure the multi-parameter conditionals worked, but it may prove to be unnecessary - I can hope, at least

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Added quite a lot more key details here, let me know what you think of the formatting

maddenp-noaa:
`uw -h` does not show `make_hgrid` as a mode -- I think one more line is needed in `cli.py` to wire it up -- in `_parse_args()` I think.

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
WeirAE: Still not successful - I'm going to call in @NaureenBharwaniNOAA for help since we did all those crazy conditions in the rocoto schema. I thought that should be right, but we might have managed to get away with some `anyOf` trickery in the end

maddenp-noaa: OK, I have a theory. If I change the schema item to
```
{
  "if": {
    "properties": {
      "do_cube_transform": {
        "type": "boolean"
      },
      "do_schmidt": {
        "type": "boolean"
      }
    },
    "required": ["do_cube_transform", "do_schmidt"]
  },
  "then": {
    "required": [
      "stretch_factor",
      "target_lat",
      "target_lon"
    ]
  }
}
```
the test passes. I think the issue is that the `if` in the original version is always true because `do_cube_transform` and `do_schmidt` are optional in the min-schema being tested, because those items are optional. I believe the `if` would become false if one of these values were provided with a non-`boolean` value in the mini-`dict` being tested, because the min-schema is explicit about that fact. The update to the min-schema above makes the _presence_ of those keys mandatory in the hypothetical scenario being tested.

The other min-schemas in the `allOf` sequence are different because they test not the _presence_ of their keys, but specific `const` or `enum` values.

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+
+config:
+^^^^^^^
+
+Describes the required parameters to run a ``make_hgrid`` configuration. 
+
+
+grid_type:
+""""""""""
```
maddenp-noaa, line 33: It's not a problem specific to this PR, but I think the visual difference between
```
config:
^^^^^^
```
and
```
grid_type:
"""""""""
```
is negligible to non-existent. Looking at the rendered HTML, it's not at all clear to me that `grid_type` et al. are subordinate to `config`. Possibly we should reconsider our formatting to make this clearer. Maybe we could discuss this in an upcoming DSM or Group.

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+
+See :ref:`here <execution_yaml>` for details.
+
+run_dir:
+^^^^^^^^
+
+The path to the directory where ``make_hgrid`` will write its outputs.
+
+config:
+^^^^^^^
```
maddenp-noaa, line 27: Any reason not to have the items under `config:` and in the corresponding `make_hgrid.yaml` file sorted? Maybe there's a reason for the current order, but I couldn't guess it.

WeirAE: I've listed them in order of the `make_hgrid` function docs, but I can resort them to something actually reasonable!

docs/shared/make_hgrid.yaml
```diff
@@ -0,0 +1,26 @@
+make_hgrid:
+  execution: 
+    batchargs: 
+      nodes: 1
+      stdout: /path/to/runscript.out
+      walltime: "08:00:00"
+    executable: /path/make_hgrid
```
NaureenBharwaniNOAA, line 7:
```suggestion
    executable: /path/to/make_hgrid
```

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+
+Set to calculate polar polygon areas by calculating the area of a copy of the polygon, with the copy  being rotated far away from the pole.
+
+verbose:
+""""""""
+
+Will print out running time message when this is set. Otherwise the run will be silent when there is no error.
+
+Example UW YAML With SRW YAML References
+----------------------------------------
```
christinaholtNOAA, line 229: I would prefer we leave this section out. The point of a portion of our work is to refactor how SRW does this, in which case none of these environment variable will exist, and none of our tools support the syntax used here.

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+Specify type of topography. See the following options for more details:
+
+* from_file:              --my_grid_file must be specified.
+* spectral_grid:          no other optional or required arguments.
+* regular_lonlat_grid:    --nxbnds, --nybnds --xbnds, --ybnds, must be specified to define the grid bounds.
+* tripolar_grid:          --nxbnds, --nybnds, --xbnds, --ybnds, must be specified to define the grid bounds.
+* conformal_cubic_grid:   --nratio is optional argument.
+* gnomonic_ed:            equal distance gnomonic cubic grid.
+* simple_cartesian_grid:  --xbnds, --ybnds must be specified to define the grid bounds location and grid size. --simple_dx and --simple_dy must be specified to specify uniform cell length.
+* f_plane_grid:           For setting geometric fractors according f-plane. f_plane_latitude need to be specified
```
NaureenBharwaniNOAA, line 44:
```suggestion
* f_plane_grid:           For setting geometric fractors according f-plane. f_plane_latitude must be specified
```

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+config:
+^^^^^^^
+
+Describes the required parameters to run a ``make_hgrid`` configuration. 
+
+
+grid_type:
+""""""""""
+
+Specify type of topography. See the following options for more details:
```
christinaholtNOAA, line 35:
```suggestion
Specify grid type with one of the following recognized key words:
```

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+
+* from_file:              --my_grid_file must be specified.
+* spectral_grid:          no other optional or required arguments.
+* regular_lonlat_grid:    --nxbnds, --nybnds --xbnds, --ybnds, must be specified to define the grid bounds.
+* tripolar_grid:          --nxbnds, --nybnds, --xbnds, --ybnds, must be specified to define the grid bounds.
+* conformal_cubic_grid:   --nratio is optional argument.
+* gnomonic_ed:            equal distance gnomonic cubic grid.
+* simple_cartesian_grid:  --xbnds, --ybnds must be specified to define the grid bounds location and grid size. --simple_dx and --simple_dy must be specified to specify uniform cell length.
+* f_plane_grid:           For setting geometric fractors according f-plane. f_plane_latitude need to be specified
+* beta_plane_grid:        For setting geometric fractors according to beta plane. f_plane_latitude need to be specified
```
christinaholtNOAA, line 45: The white space in this section is not preserved when rendered. You might consider a table. Please also use the double backtick highlighting for any of the flags. The use of colons here also makes me think these should be entries in the YAML instead of punctuation.

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: I'm somewhat inclined to put grid_type first as it defines the subsequent parameters, and maintaining the similar parameter groupings like nx/ny - then most of the alphabetical order is already broken, but I can improve some of it

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+"""""""""""""""""
+
+When specified, will not output length(dx,dy) and angle (angle_dx, angle_dy).
+
+rotate_poly:
+""""""""""""
+
+Set to calculate polar polygon areas by calculating the area of a copy of the polygon, with the copy  being rotated far away from the pole.
+
+verbose:
```
NaureenBharwaniNOAA, line 223: Is this an option that a user can add in the `config` block as a key? I didn't see it as an option in the example block provided, and wonder if it is does this add confusion to a user who will be using uwtools and see our verbose flag?  If it is an option for the `config` block should we add it in the YAML.

src/uwtools/resources/jsonschema/make-hgrid.jsonschema
WeirAE: Confirmed and updated, thank you!

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
+.. _make_hgrid_yaml:
+
+make_hgrid
+==========
+
+Structured YAML to run :ufs-utils:`make_hgrid<make-hgrid>` is validated by JSON Schema and requires the ``make_hgrid:`` block, described below. If ``make_hgrid`` is to be run via a batch system, the ``platform:`` block, described :ref:`here <platform_yaml>`, is also required.
+
```
christinaholtNOAA, line 7: Please reference the source code where this documentation was taken from. We should probably be explicit that a lot of it is taken verbatim.

docs/sections/user_guide/yaml/components/make_hgrid.rst
maddenp-noaa: I think that'd be good if accompanied by a little additional text stating e.g. that `grid_type` is kind of a primary choice, and that the options that follow depend on that choice. In fact, is `grid_type` perhaps even required? Does it make sense not to have it?

Maybe we could present a table visually mapping `grid_type` values to the options that then become required for and are related to them.

christinaholtNOAA (changes requested):
This is great work toward getting all of this captured in the schema and docs. 

I do not yet see how all those command line values get included in the runscript.

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Its a parameter for `make_hgrid`, so I put it as a key here, although its not required. Thats a valid concern, tho, I'm not exactly sure how to combine the two

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Thats fair - I was torn between that, and wanting to help users with the remapping thats not necessarily clear without deeper references

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+out_halo:
+"""""""""
+
+Extra halo size data to be written out.
+
+angular_midpoint:
+"""""""""""""""""
+
+If specified, when grid_type is from_file and the input is a NetCDF file, then the supergrid face midpoint coordinates are simply and independely (the lat independently from the lon) calculated as simple angular midpoints from the model grid coordiantes.
```
elcarpenterNOAA, line 211:
```suggestion
If specified, when grid_type is from_file and the input is a NetCDF file, then the supergrid face midpoint coordinates are simply and independently (the lat independently from the lon) calculated as simple angular midpoints from the model grid coordinates.
```

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+
+shift_fac:
+""""""""""
+
+Shift west by 180/shift_fac. Default value is 18.
+
+do_schmidt:
+"""""""""""
+
+Set to do Schmidt transformation to create astretched grid. When do_schmidt is set, the following must be set: --stretch_factor, --target_lon and --target_lat.
```
elcarpenterNOAA, line 136:
```suggestion
Set to do Schmidt transformation to create a stretched grid. When do_schmidt is set, the following must be set: --stretch_factor, --target_lon and --target_lat.
```

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,252 @@
...
+Specify grid type with one of the following recognized key words:
+
+* from_file:              --my_grid_file must be specified.
+* spectral_grid:          no other optional or required arguments.
+* regular_lonlat_grid:    --nxbnds, --nybnds --xbnds, --ybnds, must be specified to define the grid bounds.
+* tripolar_grid:          --nxbnds, --nybnds, --xbnds, --ybnds, must be specified to define the grid bounds.
+* conformal_cubic_grid:   --nratio is optional argument.
+* gnomonic_ed:            equal distance gnomonic cubic grid.
+* simple_cartesian_grid:  --xbnds, --ybnds must be specified to define the grid bounds location and grid size. --simple_dx and --simple_dy must be specified to specify uniform cell length.
+* f_plane_grid:           For setting geometric fractors according f-plane. f_plane_latitude must be specified
```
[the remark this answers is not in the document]
WeirAE: Thanks for catching the rest of the 'interesting' spelling left from the source!

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: I've kept it as a note if needed later and removed it from here

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Thanks - the table format looks much better!

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Gotcha, text and link added

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: @maddenp-noaa I'm pondering the use case here, I'm not sure this flag is ever used by any of the apps. Would it be worth it to remove this as a YAML key and update the cli to take `--verbose` and driver itself to add `--verbose` to the runscript vars if used?

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: As reluctant as I am to update the jsonschema and its test again, I think that might be the way to go.

docs/sections/user_guide/yaml/components/make_hgrid.rst
maddenp-noaa: I think that verbose behavior or `uw` and verbose behavior of `make_hgrid` should be kept orthogonal so that users can choose which (or both, or neither) they want. `uw --verbose` controls the former, so I'd assume that a `verbose` UW YAML key for `make_hgrid` would control the latter. If `make_hgrid` takes a verbose flag, I think they YAML entry would be the correct way to control that.

src/uwtools/drivers/make_hgrid.py
```diff
@@ -0,0 +1,72 @@
...
+    @task
+    def runscript(self):
+        """
+        The runscript.
+        """
+        path = self._runscript_path
+        yield self._taskname(path.name)
+        yield asset(path, path.is_file)
+        yield None
+        self._write_runscript(path=path, envvars={})
```
WeirAE, line 55: @christinaholtNOAA From your question, this ~should~ be converting the YAML keys to vars to write, but it doesn't appear to be actually doing that, let me retest this

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,240 @@
...
+Specify grid type with one of the following recognized key words:
+
+.. list-table:: Grid Types and Requirements
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Grid Type
+     - Requirements
+   * - from_file
+     - --my_grid_file must be specified.
```
maddenp-noaa, line 46:
```suggestion
   * - ``from_file``
     - ``my_grid_file:`` must be specified.
```
I think the `--` prefixes here are confusing, as they refer to CLI switches that UW users are not directly exposed to. I think the suggested formatting would help the user connect this advice to the `my_grid_file:` section in the docs below.

Same suggestion applies to all the `--` CLI switches mentioned in this table.

Any reason to also not sort these?

WeirAE: I let both of these slip through when fighting with the table format, I'll clear this up

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: AH, it is already required - I'm starting to lose track of all the conditions! I'll think about a visual map

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Fixed it, it should be ok here

src/uwtools/drivers/make_hgrid.py
WeirAE: Update: It was in the` run_cmd` override we forgot and removed earlier, thank you! Also updated to correctly handle the keys and their conditions

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,241 @@
+.. _make_hgrid_yaml:
+
+make_hgrid
+==========
+
+Structured YAML to run :ufs-utils:`make_hgrid<make-hgrid>` is validated by JSON Schema and requires the ``make_hgrid:`` block, described below. If ``make_hgrid`` is to be run via a batch system, the ``platform:`` block, described :ref:`here <platform_yaml>`, is also required.
+
+YAML keys and their descriptions are taken largely verbatim from the ``make_hgrid`` tool source code, available `here <https://github.com/NOAA-GFDL/FRE-NCtools/blob/65c34e76cfa2ff567f70da83f44cd6f991315f08/tools/make_hgrid/make_hgrid.c>`_.
```
maddenp-noaa, line 8: Is that very-specific URL preferable to the simpler https://github.com/NOAA-GFDL/FRE-NCtools/blob/master/tools/make_hgrid/make_hgrid.c ? Maybe so, if it somehow ties the docs to the specific version of `make_hgrid` being used by... well, SRW? This leads into the fraught topic of versioning drivers, docs, etc. based on the versions of apps and components. Maybe it would be better to used the unversioned URL until we know more about version-pinning requirements?

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: I've made sure all the key names have double-backstrokes and have used the table for `grid_type`, this helps a lot without adding further diagrams here

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: Good point, thats safer for sure. As long as we're just doing references and not pulling full text, this should be ok

docs/sections/user_guide/yaml/components/make_hgrid.rst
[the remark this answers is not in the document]
christinaholtNOAA: Also, should that be factor or fractor. Seems like maybe a typo (inherited from the code)?

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,241 @@
...
+
+``grid_type`` is the primary required key that defines the grid and additional required keys. It is specified with one of the following recognized key words:
+
+.. list-table:: Grid Types and Requirements
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Grid Type
+     - Requirements
+   * - beta_plane_grid
```
maddenp-noaa, line 45:
```suggestion
   * - ``beta_plane_grid``
```
Suggest formatting the entries in this column as code, too.

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,241 @@
...
+config:
+^^^^^^^
+
+Describes the required parameters to run a ``make_hgrid`` configuration. 
+
+
+grid_type:
+""""""""""
+
+``grid_type`` is the primary required key that defines the grid and additional required keys. It is specified with one of the following recognized key words:
```
maddenp-noaa, line 37: It's so awful that I hesitate to mention it, but: Our schema currently ensures that, when `grid_type` is set to some value, that certain other `config` items are defined as well. But it does not, I think, ensure that _other_ values are unset -- and, if they are, I think they'll get translated into `--cli_flags` too, possibly resulting in invalid run commands. Maybe not: Maybe the `make_hgrid` program is resilient against this and ignores irrelevant flags. Hopefully! It might be possible to model this stricter use case with `"additionalProperties": false` lines in the `"then"` clauses we already have, or perhaps using a `oneOf` clause instead of `allOf`, to make sure that exactly one valid set of arguments is specified. It might be interesting to do a quick manual check to see how `make_hgrid` deal with an extra, inapplicable argument, so set our minds (or at least mine) at ease in case it behaves graceully. Otherwise, we could revisit this topic if/when a user runs into difficulty.

maddenp-noaa (approved):
Nice work. I left a final comment/misgiving, but unless someone else feels strongly about it, I think we should move forward and revisit later if necessary.

docs/sections/user_guide/yaml/components/make_hgrid.rst
WeirAE: To make this more unnerving, its also not especially clear which parameters are not "required" but are still necessary anyway... unfortunately, it may be up to the users to give us more examples

WeirAE: I've parsed it to pseudocode and added that to my summary doc, working on converting that to even more conditionals

WeirAE: Saving further details for future work

docs/sections/user_guide/yaml/components/make_hgrid.rst
```diff
@@ -0,0 +1,242 @@
+.. _make_hgrid_yaml:
+
+make_hgrid
+==========
+
+Structured YAML to run :ufs-utils:`make_hgrid<make-hgrid>` is validated by JSON Schema and requires the ``make_hgrid:`` block, described below. If ``make_hgrid`` is to be run via a batch system, the ``platform:`` block, described :ref:`here <platform_yaml>`, is also required.
+
+.. important::
```
christinaholtNOAA, line 8: I like the green more!

src/uwtools/drivers/make_hgrid.py
```diff
@@ -0,0 +1,89 @@
...
+        config = self._driver_config["config"]
+        flags = []
+        for k, v in config.items():
+            if isinstance(v, bool):
+                flags.append("--%s" % k)
+            elif isinstance(v, list):
+                flags.append("--%s %s" % (k, ",".join(map(str, v))))
+            else:
+                flags.append("--%s %s" % (k, v))
+        return f"{executable} " + " ".join(flags)
```
christinaholtNOAA, line 81: Totally being picky, but why not:

```
f"{executable} {' '.join(flags)}"
```

src/uwtools/tests/drivers/test_make_hgrid.py
```diff
@@ -0,0 +1,116 @@
...
+def test_MakeHgrid__runscript_path(driverobj):
+    assert driverobj._runscript_path == driverobj._rundir / "runscript.make_hgrid"
+
+
+def test_MakeHgrid__taskname(driverobj):
+    assert driverobj._taskname("foo") == "make_hgrid foo"
+
+
+def test_MakeHgrid__validate(driverobj):
+    driverobj._validate()
```
christinaholtNOAA, line 116: Did I miss a test for the additional runcmd method?

christinaholtNOAA (changes requested):
Looks great. Just needs one more test for the runcmd method on the driver.

NaureenBharwaniNOAA (approved):
Looks great! Great job on this!

christinaholtNOAA (approved):
LGTM! Thanks for the quick fix.

src/uwtools/tests/drivers/test_make_hgrid.py
WeirAE: Thanks for the catch, got it in

PR #474, opened 2024-04-26
Reduce driver API boilerplate

maddenp-noaa:
# NOT READY FOR REVIEW

**Synopsis**

Programmatically generate driver API modules, DRYing out boilerplate code and tests.

**Type**

**Impact**

**Checklist**


src/uwtools/drivers/support.py
```diff
@@ -15,8 +19,10 @@ def graph() -> str:
 def tasks(driver_class: type[Driver]) -> Dict[str, str]:
     """
     Returns a mapping from task names to their one-line descriptions.
+
+    :param driver_class: Class of driver object to instantiate.
     """
     return {
-        task: getattr(driver_class, task).__doc__.strip().split("\n")[0]
+        task: (getattr(driver_class, task).__doc__ or "").strip().split("\n")[0]
```
maddenp-noaa, line 26: Fixes a bug whereby an exception was raised if a task function/method did not provide a docstring.

src/uwtools/tests/api/test_drivers.py
```diff
@@ -0,0 +1,57 @@
+# pylint: disable=missing-function-docstring,protected-access
+
```
maddenp-noaa, line 2: Given that the API driver modules no longer contain any significant code of their own and simply call code from other modules, it seemed to me sensible to combine their unit tests into a single module. This breaks the one-to-one mapping between library modules and test modules, but saves so much duplicated code that I thought the benefit outweighed that departure from convention. The per-driver test modules are removed.

src/uwtools/api/esg_grid.py
```diff
@@ -2,55 +2,11 @@
 API access to the uwtools ``esg_grid`` driver.
```
NaureenBharwaniNOAA, line 2:
```suggestion
API access to the ``uwtools`` ``esg_grid`` driver.
```

src/uwtools/api/chgres_cube.py
```diff
@@ -2,60 +2,11 @@
 API access to the uwtools ``chgres_cube`` driver.
```
NaureenBharwaniNOAA, line 2:
```suggestion
API access to the ``uwtools`` ``chgres_cube`` driver.
```

src/uwtools/tests/api/test_drivers.py
```diff
@@ -0,0 +1,57 @@
...
+from datetime import datetime as dt
+from unittest.mock import patch
+
+import iotaa
+import pytest
+
+from uwtools.api import (
```
NaureenBharwaniNOAA, line 9: This definitely makes the duplication of code much less and I appreciate in the future it'll feel like a real addition by adding a driver vs. a copy-paste job.

src/uwtools/utils/api.py
```diff
@@ -25,10 +32,127 @@ def ensure_data_source(
     return str2path(data_source)
 
 
+def make_execute(driver_class: type[Driver], with_cycle: bool) -> Callable[..., bool]:
+    """
+    Returns a function that executes tasks for the given driver.
+
+    :param driver_class: The driver class whose tasks to execute.
+    :param with_cycle: Does the driver's constructur take the 'cycle' parameter?
```
elcarpenterNOAA, line 40:
```suggestion
    :param with_cycle: Does the driver's constructor take the 'cycle' parameter?
```

NaureenBharwaniNOAA:
```suggestion
    :param with_cycle: Does the driver's constructor take the 'cycle' parameter?
```

src/uwtools/api/global_equiv_resol.py
```diff
@@ -2,55 +2,11 @@
...
-    """
-    Returns a mapping from task names to their one-line descriptions.
-    """
-    return _support.tasks(_GlobalEquivResol)
+from uwtools.drivers.global_equiv_resol import GlobalEquivResol as _Driver
+from uwtools.drivers.support import graph
+from uwtools.utils.api import make_execute as _make_execute
+from uwtools.utils.api import make_tasks as _make_tasks
+
+execute = _make_execute(_Driver, with_cycle=True)
```
christinaholtNOAA, line 10:
```suggestion
execute = _make_execute(_Driver, with_cycle=False)
```

elcarpenterNOAA (approved):
Thank you. Nice work on this.

src/uwtools/utils/api.py
```diff
@@ -25,10 +32,127 @@ def ensure_data_source(
...
+            cycle=cycle,
+            config=config,
+            batch=batch,
+            dry_run=dry_run,
+            graph_file=graph_file,
+            stdin_ok=stdin_ok,
+        )
+
+    execute_cycle.__name__ = "execute"
+    assert _execute.__doc__ is not None
```
christinaholtNOAA, line 83: Is this one just to make sure we don't get odd failures in `re.sub()` calls if someone decided to remove the docstring on the function `_execute()` below?

christinaholtNOAA (approved):
LGTM! Just a tiny change request.

src/uwtools/utils/api.py
maddenp-noaa: It's actually for the typechecker. Without this assertion, the line after this one is flagged:
```
uwtools/utils/api.py:84: error: Argument 3 to "sub" has incompatible type "str | None"; expected "str"  [arg-type]
        execute_cycle.__doc__ = re.sub(r"\n *:param driver_class:.*\n", "\n", _execute.__doc__)
                                                                              ^~~~~~~~~~~~~~~~
```
I'm surprised, because it seems like static analysis could see that `_execute()` _has_ a docstring so that, although officially its type is technically `str` or `None`, it's a `str` in this case. My guess is that `__str__` may be attached to the class only at runtime: From a syntax perspective, the docstring is just a string literal given as the first statement in the class body. The fact that it's interpreted as a "docstring", and then attached to the `__str__` attribute, may not be known until runtime.

PR #476, opened 2024-04-30
Repin iotaa and remove file() dependency on driver executables

maddenp-noaa:
**Synopsis**

- Repin `iotaa` to `0.7.*` to avoid a regression in `0.8.x` until that is fixed.
- Remove `file()` dependency on driver executables.

**Type**

**Impact**

**Checklist**


src/uwtools/utils/tasks.py
```diff
@@ -4,11 +4,23 @@
...
 
+@external
+def executable(program: Union[Path, str]):
+    """
+    An executable program located on the current path.
+
+    :param program: Name of or path to the program.
+    """
+    yield "Executable program %s" % program
+    yield asset(program, lambda: bool(which(program)))
```
maddenp-noaa, line 21: I suspect this will be useful later, so I opted to keep it even though it is not used at the moment.

christinaholtNOAA (approved):
LGTM. Thanks for such a quick fix!

elcarpenterNOAA (approved):
LGTM.

PR #477, opened 2024-04-30
UW 555: Add Jedi Documentation

NaureenBharwaniNOAA:
**Synopsis**

Adds the documentation for the Jedi driver.

Rendered docs are available for review here: https://uwtools--477.org.readthedocs.build/en/477/index.html

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/yaml/components/jedi.rst
```diff
@@ -0,0 +1,44 @@
...
+^^^^^^^^^^^^^^
+
+See :ref:`this page <files_yaml>` for details.
+
+.. code-block:: text
+
+   jedi:
+     files_to_copy:
+        f1: /path/to/f1
+
```
NaureenBharwaniNOAA, line 35: Used the content from the `docs/shared/jedi.yaml`. Maybe a more real example would be useful here. I'm not sure what a real example would look like as of right now.

docs/sections/user_guide/cli/drivers/jedi.rst
```diff
@@ -0,0 +1,101 @@
+``jedi``
+========
+
+The ``uw`` mode for configuring and running the JEDI framework.
```
NaureenBharwaniNOAA, line 4: I haven't added any additional documentation resources for users to reference and started with this simple introduction.

docs/sections/user_guide/yaml/components/jedi.rst
maddenp-noaa: I'm curious to hear what @christinaholtNOAA thinks, but I'm pretty ignorant about what real use cases look like. It seems like it could be a big topic, but maybe also something that's well-documented in JEDI itself, such that we could just redirect people to their documentation once we have real use cases and know more.

elcarpenterNOAA (approved):
LGTM.

WeirAE (approved):
LGTM

christinaholtNOAA (approved):
LGTM! I am suggesting we still have a follow up ticket come from this, but one that we don't necessarily tackle until we see better use cases from other sources -- Land DA, MPAS/RRFS, global-workflow, etc.

PR #481, opened 2024-05-06
UW-463 UPP driver

maddenp-noaa:
**Synopsis**

A UPP driver.

Most of the changed pages are small doc fixes / updates / harmonizations.

**Type**

**Impact**

**Checklist**


src/uwtools/cli.py
```diff
@@ -964,21 +989,34 @@ def _parse_args(raw_args: List[str]) -> Tuple[Args, Checks]:
...
+        ]
+    }
+    drivers_with_cycle_and_leadtime = {
+        component: partial(
+            _add_subparser_for_driver, component, subparsers, with_cycle=True, with_leadtime=True
+        )
+        for component in [
+            STR.upp,
         ]
     }
```
maddenp-noaa, line 1018: Slightly longer than the tuple-based version, but possibly clearer in intent, too.

src/uwtools/drivers/driver.py
```diff
@@ -41,13 +42,21 @@ def __init__(
...
+            if not cycle:
+                raise UWError("When leadtime is specified, cycle is required")
+            self._validtime = cycle + timedelta(hours=leadtime)
         self._config.dereference(
-            context={**({"cycle": cycle} if cycle else {}), **self._config.data}
+            context={
+                **({"cycle": cycle} if cycle else {}),
+                **({"validtime": self._validtime} if leadtime else {}),
+                **self._config.data,
+            }
```
maddenp-noaa, line 59: Make both `cycle` and `validtime` `datetime` objects available for config realization.

Also, we previously had a no-args `deference()` call followed by one that also made `cycle` available, and I couldn't think of a reason why we needed the first one. I could be missing something that we'll have to fix later, but all of our tests continue to pass. If anyone knows a reason why we can't do a single realization, please let me know.

src/uwtools/drivers/driver.py
```diff
@@ -163,6 +172,7 @@ def _resources(self) -> Dict[str, Any]:
             "account": platform["account"],
             "rundir": self._rundir,
             "scheduler": platform["scheduler"],
+            "stdout": "%s.out" % self._runscript_path.name,  # config may override
```
maddenp-noaa, line 175: Provide a default output filename based on the runscript name, to be used in case the user does not specify anything. This will almost certainly be better than whatever name is chosen by the batch scheduler.

src/uwtools/drivers/driver.py
```diff
@@ -254,7 +264,7 @@ def _write_runscript(self, path: Path, envvars: Dict[str, str]) -> None:
             envvars=envvars,
             execution=[
                 "time %s" % self._runcmd,
-                "test $? -eq 0 && touch %s/done.%s" % (self._rundir, self._driver_name),
+                "test $? -eq 0 && touch %s.done" % self._runscript_path.name,
```
maddenp-noaa, line 267: We shouldn't need an absolute path here, as we want this file to be relative to the runscript, i.e. in the run directory, which should already be the case. This chang makes the runscript a little more concise, and also means one less thing to manually change if the run directory is copied to another path.

Also, we may now see a series of closely related files grouped together, e.g. for `upp` submitted to the batch system:
```
runscript.upp
runscript.upp.done
runscript.upp.out
runscript.upp.submit
```
The first is the runscript itself. The second is the empty sentinel file created if the component runs to completion and exits with success status. The third is the combined stdout/stderr streams from the runscript and everything that happens in it. And the last is the output produced by whatever command submits the runscript to the batch scheduler (should contain the batch job ID).

src/uwtools/drivers/upp.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns the name of this driver.
+        """
+        return STR.upp
+
+    @property
+    def _namelist_path(self) -> Path:
+        """
+        Path to the namelist file.
+        """
+        return self._rundir / "itag"
```
maddenp-noaa, line 129: Unfortunately, we're stuck with this for now.

src/uwtools/drivers/upp.py
```diff
@@ -0,0 +1,163 @@
...
+        Returns a common tag for graph-task log messages.
+
+        :param suffix: Log-string suffix.
+        """
+        return "%s f%03d %s %s" % (
+            self._cycle.strftime("%Y%m%d %HZ"),
+            self._leadtime,
+            self._driver_name,
+            suffix,
+        )
```
maddenp-noaa, line 163: The task name now includes a leadtime component.

src/uwtools/resources/jsonschema/upp.jsonschema
```diff
@@ -0,0 +1,188 @@
...
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
+          "properties": {
+            "base_file": {
+              "type": "string"
+            },
+            "update_values": {
```
maddenp-noaa, line 33: Much of the content in this block comes from [this UPP PR](https://github.com/NOAA-EMC/UPP/pull/940).

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -80,7 +80,7 @@ def driverobj(config_file, cycle):
     return chgres_cube.ChgresCube(config=config_file, cycle=cycle, batch=True)
 
 
-# Driver tests
+# Tests
```
maddenp-noaa, line 83: I harmonized a few of the driver test modules w.r.t. section names.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -57,34 +57,37 @@ def write(path, s):
 
 
 @fixture
-def driverobj(tmp_path):
-    cf = write(
```
maddenp-noaa, line 61: This doesn't need to be written to a file and read back in, it can be used as an in-memory object.

NaureenBharwaniNOAA:
Rendered documentation link: https://uwtools--481.org.readthedocs.build/en/481/

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -34,14 +34,12 @@ All tasks take the same arguments. For example:
...
-   usage: uw chgres_cube run --config-file PATH --cycle CYCLE [-h] [--version] [--batch] [--dry-run]
-                             [--graph-file PATH] [--quiet] [--verbose]
+   usage: uw chgres_cube run --cycle CYCLE [-h] [--version] [--config-file PATH] [--batch]
+                             [--dry-run] [--graph-file PATH] [--quiet] [--verbose]
 
    A run
 
    Required arguments:
-     --config-file PATH, -c PATH
-         Path to UW YAML config file
```
maddenp-noaa, line 44: I re-ran many of the CLI invocation examples and update the docs with their current output.

docs/index.rst
```diff
@@ -143,22 +143,28 @@ chgres_cube
...
+| **API**: ``import uwtools.api.upp``
 
+Driver for JEDI
+^^^^^^^^^^^^^^^
 
+| **CLI**: ``uw jedi -h``
+| **API**: ``import uwtools.api.jedi``
 
+JEDI
+""""
```
NaureenBharwaniNOAA, line 167: I feel like this is out of place or most likely was missed to be removed.

maddenp-noaa: Yes, thank, will fix.

docs/shared/upp.yaml
```diff
@@ -0,0 +1,41 @@
...
+    mpiargs:
+      - "--ntasks $SLURM_CPUS_ON_NODE"
+    mpicmd: srun
+  files_to_copy:
+    postxconfig-NT.txt: /path/to/postxconfig-NT.txt
+  files_to_link:
+    eta_micro_lookup.dat: /path/to/nam_micro_lookup.dat
+    params_grib2_tbl_new: /path/to/params_grib2_tbl_new
+  namelist:
+    update_values:
```
NaureenBharwaniNOAA, line 21: Does it make sense to add `base_file` here to show a complete example to the user?

maddenp-noaa: Good idea, thanks, will do.

src/uwtools/resources/jsonschema/upp.jsonschema
```diff
@@ -0,0 +1,188 @@
...
+              "additionalProperties": false,
+              "properties": {
+                "model_inputs": {
```
NaureenBharwaniNOAA, line 36: Are any of these properties `required` under `model_inputs`?

docs/sections/user_guide/yaml/components/upp.rst
```diff
@@ -0,0 +1,57 @@
...
+     - ``datestr``, ``filename``, ``filenameflat``, ``filenameflux``, ``grib``, ``ioform``, ``modelname``
+   * - ``nampgb``
+     - ``aqf_on``, ``d2d_chem``, ``d3d_on``, ``filenameaer``, ``gccpp_on``, ``gocart_on``, ``gtg_on``, ``hyb_sigp``, ``kpo``, ``kpv``, ``kth``, ``method_blsn``, ``nasa_on``, ``numx``, ``po``, ``popascal``, ``pv``, ``rdaod``, ``slrutah_on``, ``th``, ``vtimeunits``, ``write_ifi_debug_files``
+
+Read more on the UPP namelists, including variable meanings and appropriate values, `here <https://upp.readthedocs.io/en/develop/BuildingRunningTesting/InputsOutputs.html#itag>`_.
+
+run_dir:
+^^^^^^^^
+
+The path to the UPP run directory.
```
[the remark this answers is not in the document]
maddenp-noaa: IMO we should update the other drivers to use more general language. My guess is that this description was written for one driver and copied, even for drivers that have no namelist file. So, good call about consistency, but I think I'd prefer to be more general, similar to this. If anyone disagrees, please speak up.

Also, I'll add `run_dir:` to `shave`.

src/uwtools/api/sfc_climo_gen.py
```diff
@@ -7,6 +7,6 @@
 from uwtools.utils.api import make_execute as _make_execute
 from uwtools.utils.api import make_tasks as _make_tasks
 
-execute = _make_execute(_Driver, with_cycle=False)
```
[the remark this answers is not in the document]
maddenp-noaa: Nice catch, thanks, fixing!

src/uwtools/drivers/upp.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @property
+    def _runscript_path(self) -> Path:
+        """
+        Path to the runscript.
+        """
+        return self._rundir / f"runscript.{self._driver_name}"
+
+    @property
+    def _runcmd(self) -> str:
```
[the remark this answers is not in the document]
maddenp-noaa: It just depends on how the driver needs to be run. Historically, a lot of NWP components operate just by running the program (possibly preceded by `mpirun`, `mpiexec`, or `srun` if they are MPI parallel programs), with no arguments, because historically Fortran did not support command-line arguments. So, for a serial program, you'd just run
```
./component
```
or for a parallel MPI program, to run it with 8 parallel tasks
```
mpirun -n 8 component
```
It was assumed that any configuration required by the program would be supplied by namelists in a namelist file, and the program would open a file with a standard name that the user couldn't change (like the `itag` for `upp` as seen in this PR). Funny enough, MPI is newer than Fortran and so it _does_ support command like arguments, but those just tell MPI how to behave, and the program still gets its configuration from a namelist file.

C programs have always supported command like arguments, and modern Fortran programs do, too, so some components we're writing drivers for need those. So our default `_runcmd()` method in the `Driver` base class supports the classic Fortran-style no-argument invocation, and drivers that need extra command-line arguments need to override that method and provide their own implementation.

Yours is a good question, though: Maybe we should make the `Driver` base version of `_runcmd()` more flexible so that it's less likely that we'll need to override it. You might consider brining it up in standup and potentially writing a ticket for the work if the team likes the idea.

src/uwtools/resources/jsonschema/upp.jsonschema
```diff
@@ -0,0 +1,188 @@
...
+                        "GFS",
+                        "RAPR",
+                        "NMM"
+                      ],
+                      "type": "string"
+                    }
+                  },
+                  "type": "object"
+                },
+                "nampgb": {
```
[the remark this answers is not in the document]
maddenp-noaa: I don't think so, because the user might provide a mostly-correct `base_file` and only want to update a subset of the values. I think we can only describe what keys they _may_ provide and what their values look like; I don't think we can say that the _must_ provide anything. Please let me know if you see it otherwise.

src/uwtools/resources/jsonschema/upp.jsonschema
maddenp-noaa: Same answer as for `nampgb` :)

docs/sections/user_guide/yaml/components/chgres_cube.rst
```diff
@@ -5,6 +5,8 @@ chgres_cube
 
 Structured YAML to run :ufs-utils:`chgres_cube<chgres-cube>` is validated by JSON Schema and requires the ``chgres_cube:`` block, described below. If ``chgres_cube`` is to be run via a batch system, the ``platform:`` block, described :ref:`here <platform_yaml>`, is also required.
 
+.. include:: ../../../../shared/injected_cycle.rst
```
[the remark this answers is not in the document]
WeirAE: This is really useful and seems to work!

docs/sections/user_guide/yaml/components/upp.rst
```diff
@@ -0,0 +1,57 @@
...
+.. list-table::
+   :widths: 10 95
+   :header-rows: 1
+
+   * - Namelist
+     - Variables
+   * - ``model_inputs``
+     - ``datestr``, ``filename``, ``filenameflat``, ``filenameflux``, ``grib``, ``ioform``, ``modelname``
+   * - ``nampgb``
+     - ``aqf_on``, ``d2d_chem``, ``d3d_on``, ``filenameaer``, ``gccpp_on``, ``gocart_on``, ``gtg_on``, ``hyb_sigp``, ``kpo``, ``kpv``, ``kth``, ``method_blsn``, ``nasa_on``, ``numx``, ``po``, ``popascal``, ``pv``, ``rdaod``, ``slrutah_on``, ``th``, ``vtimeunits``, ``write_ifi_debug_files``
```
WeirAE, line 50: I think this table might look better shorter with the third row, second column variables on two lines

src/uwtools/drivers/upp.py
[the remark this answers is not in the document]
WeirAE: Thats a good point, and the fewer overrides the better

docs/sections/user_guide/yaml/components/upp.rst
maddenp-noaa: They are on two lines for me. It's meant to be flexible and rewrap based on the size of your browser window. Could you try and see if it adapts to your window size if you change it? And maybe post a screenshot if it's not doing that?

src/uwtools/tests/test_schemas.py
```diff
@@ -859,3 +864,127 @@ def test_schema_ungrib_run_dir(ungrib_prop):
...
+        assert "not of type 'integer'" in errors({"update_values": {"nampgb": {key: True}}})
+    # nampgb: Some arrays of numbers are supported:
+    nitems = 70
+    for key in ["po", "pv", "th"]:
+        assert not errors({"update_values": {"nampgb": {key: [3.14] * nitems}}})
+        assert "too long" in errors({"update_values": {"nampgb": {key: [3.14] * (nitems + 1)}}})
+        assert "not of type 'number'" in errors(
+            {"update_values": {"nampgb": {key: [True] * nitems}}}
+        )
+    # nampgb: Only one vtimeunits value is supported:
```
WeirAE, line 982: Very thorough!

docs/sections/user_guide/yaml/components/upp.rst
WeirAE: Yep, it does wrap on window size, that should probably be fine

WeirAE (approved):
Whew, this covered a ton! I think I checked everything, I'll double-check in the morning if its still pending but I think it looks good!

elcarpenterNOAA (approved):
LGTM.

NaureenBharwaniNOAA (approved):

src/uwtools/cli.py
```diff
@@ -642,6 +643,15 @@ def _add_arg_keys(group: Group) -> None:
     )
 
 
+def _add_arg_leadtime(group: Group) -> None:
+    group.add_argument(
+        _switch(STR.leadtime),
+        help="Leadtime in hours",
+        required=True,
+        type=int,
+    )
```
christinaholtNOAA, line 652: There will probably be a use case for sub-hourly output from our users (either immediately, or the not-to-distant future). How much trouble would it be to tackle that change here? I was originally thinking about it as a `timedelta` type object.

src/uwtools/drivers/driver.py
christinaholtNOAA: I was just noticing this in my example yesterday!

christinaholtNOAA (commented):
This is looking great! With so many changes, it's gotten me thinking about tiny little upgrades that might be helpful. Some are inline below. 

One that I didn't put inline...as I was looking through the thorough tests for namelist settings that are tested against `update_values`, I'm realizing that we don't actually schema-check the final namelist that we dump to the run directory (unless I'm missing it). A user could have total trash in the `base_file` and we wouldn't help them correct that at all. I am also now curious if we want to put a requirement on the `base_file` existence before we try to update it. All of these suggestions are definitely contingent on the user providing a `base_file` entry.

docs/sections/user_guide/cli/drivers/upp.rst
```diff
@@ -0,0 +1,106 @@
...
+
+   $ uw upp run --help
+   usage: uw upp run --cycle CYCLE --leadtime LEADTIME [-h] [--version] [--config-file PATH]
+                     [--batch] [--dry-run] [--graph-file PATH] [--quiet] [--verbose]
+
+   A run
+
+   Required arguments:
+     --cycle CYCLE
+         The cycle in ISO8601 format
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: I would definitely second this suggestion, when first making a driver with a cycle I had to look up the actual format and failed once or twice before I found an example somewhere in our code.

maddenp-noaa:
@christinaholtNOAA 
> we don't actually schema-check the final namelist that we dump to the run directory
We might consider the ability to provide an arbitrary namelist as the relief valve for a user who has modified the namelist requirements (and code) in an unanticipated way and wants to do something our schema doesn't permit. Previously, the entire `namelist:` block was optional, so a user could simply omit it and manually place a namelist file in the run directory. We decided to make `namelist:` required and use `base_file:` as the mechanism to supply a pre-made namelist file, so if we added a schema check of the namelist content, we'd revoke the ability of users to do arbitrary experiments. We could consider adding an optional `validate:` key under `namelist:`, defaulting to true, to suppress the check. That might preserve the relieve valve while defaulting to rigorous checking. I'll write a ticket for it and we can evaluate it during a grooming session.
>  I am also now curious if we want to put a requirement on the `base_file` existence before we try to update it.
Excellent idea. I'll add a ticket.

docs/sections/user_guide/cli/drivers/upp.rst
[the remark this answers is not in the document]
maddenp-noaa: Good idea, will update. I'm considering the following text:
```
The cycle in ISO8601 format (e.g. 2024-05-08T18)
```
My reasoning:
1. The difference between upper- and lower-case letters in the proposed text above leaks internal Python details that I don't think belong in the CLI.
2. The underscore in the proposed text appears to work, but as far as I can tell is not ISO8601-compliant, as times are to be preceded by `T`. This may be a convenience added by Python.
3. If/when we need to support sub-hourly cycles, we may need to think about other assumptions in the codebase, so we may not want to suggest that we're correctly handling that level of precision today. Specifying a datetime precise to the hour is correct, ISO8601-compliant, and concise, so I've left off the `:MM:SS` part, as it would generally just be `:00:00` anyway.
4. I think we should continue to mention ISO8601 so that users understand that they have more power here should they choose to look it up. Also, it's a service to the NWP community to raise awareness of this international standard and encourage best practice (i.e. using it). IMO, anyone working with time-based data should be aware of ISO8601.

docs/shared/upp.yaml
```diff
@@ -0,0 +1,42 @@
...
+  files_to_link:
+    eta_micro_lookup.dat: /path/to/nam_micro_lookup.dat
+    params_grib2_tbl_new: /path/to/params_grib2_tbl_new
+  namelist:
+    base_file: /path/to/base.nml
+    update_values:
+      model_inputs:
+        datestr: "{{ validtime.strftime('%Y-%m-%d_%H:%M:%S') }}"
+        filename: /path/to/dynf{{ user.leadtime }}.nc
+        filenameflux: /path/to/phyf{{ user.leadtime }}.nc
```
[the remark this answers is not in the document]
maddenp-noaa: Actually, `leadtime` is computed in the `user:` block. The value injected by the driver is `validtime`, which is a Python `datetime` object, like `cycle`. My reasoning:

1. Injecting as few variables as possible is a benefit, as it keeps the implementation simpler.
2. Either `leadtime` or `validtime` could be derived from the other, given `cycle`. But I'm not sure `timedelta`, which I think would be needed for an `int` -> `datetime` conversion, will be available to Jinja2 in configs. Operations on two `datetime` objects are certainly available.
3. Of `leadtime` and `validtime`, the latter is more general: More values can be immediately derived from it.
4. The consistency of having two Python `datetime` objects, instead of one `datetime` and one `int`, may be helpful as a memory aid.
5. On the command line, conversely, `leadtime` is more ergonomic: Saying `--cycle 2024-05-08T18 --leadtime 167` is IMO much simpler than trying to figure out what the correct `--validtime` argument would be. So I settled on `leadtime` as a user input, and `validtime` as a Jinja2 variable.

docs/shared/upp.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - source /etc/profile.d/slurm.sh
+    executable: /path/to/upp.x
+    mpiargs:
+      - "--ntasks $SLURM_CPUS_ON_NODE"
+    mpicmd: srun
+  files_to_copy:
+    postxconfig-NT.txt: /path/to/postxconfig-NT.txt
+  files_to_link:
+    eta_micro_lookup.dat: /path/to/nam_micro_lookup.dat
+    params_grib2_tbl_new: /path/to/params_grib2_tbl_new
```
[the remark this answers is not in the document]
maddenp-noaa: Hardlinks are a great idea, but I think they need to be opt-in, at least as a replacement for symlinks, because a user expecting symlinks will probably be expecting the provenance information implied e.g. by the RHS of the `->` info info shown by `ls -l` to be preserved. It might be appropriate to silently hardlink instead of copying when possible, though. I don't immediately see a downside to that.

src/uwtools/cli.py
maddenp-noaa: I think a `--leadtime HH[:MM[:SS]]` option on the command line, immediately parsed to a `timedelta`, and a native `timedelta` in the API and deeper, is probably a good idea and straightforward to implement. Let me see if it won't overly complicate this PR.

docs/sections/user_guide/cli/drivers/upp.rst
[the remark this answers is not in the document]
christinaholtNOAA: Yep. I think that makes sense. I definitely just botched my suggestion, and this one looks good to me!

docs/shared/upp.yaml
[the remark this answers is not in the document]
christinaholtNOAA: There are a couple of options to include here, but valid time is a derived piece of information from the command line input, and shouldn't be stored in the static config.yaml file. Here are a few thoughts: 

* We could support both `leadtime` and `validtime` as Jinja2 variables. If `leadtime` is a timedelta object already, I don't think Jinja has a problem with using it. I don't think Jinja would let you make a timedelta object without some additional context that includes the `timedelta` object.
* We could show this example using `{{ validtime - cycle }}` inline without the redirection to the `user:` section. That value won't get rendered until run time anyway, so it just means we will need to provide the user section to this 


I guess now that I'm walking through those points, it really just boils down to questioning what value the redirection adds.

src/uwtools/cli.py
maddenp-noaa: Implemented `--leadtime HH[:MM[:SS]]` support.

docs/sections/user_guide/cli/drivers/upp.rst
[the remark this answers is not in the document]
maddenp-noaa: Added ISO8601 example to CLI help.

src/uwtools/drivers/driver.py
maddenp-noaa: `validtime` is now replaced with a `leadtime` `timedelta` object.

maddenp-noaa:
I made some updates to support `--leadtime HH[:MM[:SS]]` instead of an integer number of hours. We should still consider whether we've made any assumptions in the overall codebase about leadtime resolution, but at least the current PR should move us in the right direction by supporting subhourly intervals for UPP.

christinaholtNOAA (approved):
Thanks for the changes!

PR #482, opened 2024-05-10
File mode updates

maddenp-noaa:
**Synopsis**

- Add support for custom Jinja2 filter `env` to insert values of environment variables.
- Fix dry-run mode for the file tool and update tests.
- Improve an error message.

**Type**

**Impact**

**Checklist**


src/uwtools/config/jinja2.py
```diff
@@ -310,7 +310,7 @@ def path_join(path_components: List[str]) -> str:
             raise UndefinedError()
         return os.path.join(*path_components)
 
-    filters = dict(path_join=path_join)
+    filters = dict(env=lambda var: os.environ[var], path_join=path_join)
```
maddenp-noaa, line 313: This adds an `env` filter to access specific environment variables in configs. For example, in `file copy` or `file link` mode, we could now have e.g.
```
files:
  f000.00: "{{ basedir }}/{{ 'YYYYMMDD' | env }}/00/gfs.t00z.pgrb2.0p25.f000"
  f012.00: "{{ basedir }}/{{ 'YYYYMMDD' | env }}/12/gfs.t12z.pgrb2.0p25.f000"
basedir: /path/to/data
```
in `config.yaml`. The hardcoded `basedir` value keeps that path DRY, and the reference to the `YYYYMMDD` environment variable would support an invocation like
```
YYYYMMDD=20240509 uw file link --config-file config.yaml --target-dir target files
```
While working on the video demo for the `file` tool, this seemed like something you'd obviously want to do when using that, to parameterize a file copy/link list by e.g. cycle. I'm reluctant to litter our use of Jinja2 with custom filters and functions (we currently only have one other filer: `path_join` filter, to create filesystem paths from components) but this one seems maybe indispensable. But if anyone has an idea how else we might do this -- or thinks that we shouldn't -- please let me know.

BTW, it would also be possible to add custom _functions_ to Jinja2, instead of filters. For example, instead of `{{ 'YYYYMMDD' | env }}`, we could have `{{ env('YYYYMMDD') }}`. I can't immediately find a reason to prefer one over the other, except that the filter version is more Jinja2-ish.

src/uwtools/file.py
```diff
@@ -36,10 +36,11 @@ def __init__(
         :param dry_run: Do not copy files
         :raises: UWConfigError if config fails validation.
         """
+        if dry_run:
+            dryrun()
```
maddenp-noaa, line 40: The dry-run option wasn't actually being used in the `file` mode. Now it is.

In the process of fixing this, though, I realized that API users may have a problem here: If a driver or other tool using `iotaa` is called with `dry_run=True`, `iotaa`' dry-run mode will be enabled and will not be subsequently disabled, even if another driver/tool is called with `dry_run=False`. I think an easy fix may be to just call `dryrun(enabled=dry_run)`, which will _always_ toggle `iotaa`'s dry-run mode to either `True` or `False`. I've added a reminder for myself to look into this, but thought I'd mention it.

src/uwtools/file.py
```diff
@@ -55,7 +56,7 @@ def _file_map(self) -> dict:
         for key in self._keys:
             nav.append(key)
             if key not in cfg:
-                raise UWConfigError("Config navigation %s failed" % " -> ".join(nav))
+                raise UWConfigError("Failed following YAML key(s): %s" % " -> ".join(nav))
```
maddenp-noaa, line 59: I ran into this error when testing, where my file dst/src block was under a top-level `files` key. The error read
```
Config navigation files failed
```
which I initially misinterpreted. The updated error would read
```
Failed following YAML key(s): files
```
which I hope is clearer. It definitely helped me, as I'd added `files` to my CLI invocation to access the block.

src/uwtools/tests/test_file.py
```diff
@@ -26,25 +27,7 @@ def assets(tmp_path):
...
-    assert stager._validate() is True
-
-
-@pytest.mark.parametrize("source", ("dict", "file"))
-def test_FileStager_bad_key(assets, source):
-    dstdir, cfgdict, cfgfile = assets
-    config = cfgdict if source == "dict" else cfgfile
-    with raises(UWConfigError) as e:
-        file.FileStager(target_dir=dstdir, config=config, keys=["a", "x"])
-    assert str(e.value) == "Config navigation a -> x failed"
```
maddenp-noaa, line 43: These tests moved down in the module.

src/uwtools/tests/test_file.py
```diff
@@ -55,8 +38,19 @@ def test_FileCopier_config_file(assets, source):
...
 
+def test_FileCopier_config_file_dry_run(assets):
+    dstdir, cfgdict, _ = assets
+    assert not (dstdir / "foo").exists()
+    assert not (dstdir / "subdir" / "bar").exists()
+    stager = file.FileCopier(target_dir=dstdir, config=cfgdict, keys=["a", "b"], dry_run=True)
+    stager.go()
+    assert not (dstdir / "foo").exists()
+    assert not (dstdir / "subdir" / "bar").exists()
+    iotaa.dryrun(False)
```
maddenp-noaa, line 49: A new test for dry-run mode.

src/uwtools/config/jinja2.py
christinaholtNOAA: I like the addition of the `env` filter. I think many of our users will appreciate it. 

I'm curious if we might also entertain adding an optional `--cycle` flag to this tool so that folks who want to pull pieces directly from driver config files would have a similar experience with this tool. It might be nice to have an option that is configurable in the template.

christinaholtNOAA (approved):
LGTM.

src/uwtools/config/jinja2.py
maddenp-noaa: That seems like a good idea, but probably a bit of work and so then probably worthy of a ticket and scheduling. We might as well add `--leadtime` then, too, to support all driver config YAML. I'll add a Jira ticket.

NaureenBharwaniNOAA (approved):

benjamin-cash:
This might be a place to address the issue I just ran into. Because the file move uses mkdir, it will fail if you give it a nested directory as --target-dir.

elcarpenterNOAA (approved):
Nice updates.

PR #483, opened 2024-05-12
UW-576: make_solo_mosaic driver

elcarpenterNOAA:
**Synopsis**

Driver for make_solo_mosaic. 

Successful run here: `/lfs4/BMC/wrfruc/elcarpenter/uwtools/workflow-tools/rundir`

[Documentation](https://uwtools--483.org.readthedocs.build/en/483/). 

**Type**

**Impact**

**Checklist**


docs/shared/make_solo_mosaic.yaml
```diff
@@ -0,0 +1,17 @@
+make_solo_mosaic:
+  config:
+    dir: /path/to/input
+    num_tiles: 1
+    mosaic_name: ocean_mosaic
+    periodx: 360
+    periody: 360
+    tile_file: C403_grid.tile7.halo6.nc
```
elcarpenterNOAA, line 8: These four flags are optional and the other two, `dir` and `num_tiles`, are required. 
1. Should the optional flags be included for in the yaml? 
2. If yes, should all 6 be sorted alphabetically, regardless of importance? 
3. Do the optional flags need to be accounted for in the schema and in the corresponding`yaml/components` file?

docs/sections/user_guide/yaml/components/make_solo_mosaic.rst
```diff
@@ -0,0 +1,41 @@
...
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/make_solo_mosaic.yaml
+
+UW YAML for the ``make_solo_mosaic:`` Block
+-------------------------------------------
+
+config:
+^^^^^^^
+
+Describes the required parameters to run a ``make_solo_mosaic`` configuration.
```
maddenp-noaa, line 21: I see in the prototype YAML config below that there are several parameters below `config:`. Should we list and describe them here, or at least list them and give a link to external docs describing what each means?

docs/shared/make_solo_mosaic.yaml
maddenp-noaa: I vote yes on all three. The schema can make the required parameters required, and can specify the value type for each/all of them.

docs/shared/make_solo_mosaic.yaml
```diff
@@ -0,0 +1,17 @@
...
+  execution:
+    batchargs:
+      cores: 1
+      walltime: "00:01:00"
+    executable: /path/to/make_solo_mosaic.exe
+  run_dir: /path/to/rundir
+platform:
+  account: wrfruc
```
maddenp-noaa, line 16:
```suggestion
  account: me
```
This is what we've got in the other prototype YAML files, to avoid getting too specific with the account names that we actually use.

src/uwtools/drivers/make_solo_mosaic.py
```diff
@@ -0,0 +1,79 @@
...
+    def _runcmd(self):
+        """
+        Returns the full command-line component invocation.
+        """
+        executable = self._driver_config["execution"]["executable"]
+        config = self._driver_config["config"]
+        flags = []
+        for k, v in config.items():
+            flags.append("--%s %s" % (k, v))
+        return f"{executable} " + " ".join(flags)
```
maddenp-noaa, line 71:
```suggestion
        flags = " ".join(f"--{k} {v}" for k, v in self._driver_config["config"].items())
        return f"{executable} {flags}"
```
A bit simpler with a listcomp + f-string.

src/uwtools/resources/jsonschema/make-solo-mosaic.jsonschema
```diff
@@ -0,0 +1,46 @@
...
+              "required": [
+                "dir"
+              ]
+            },
+            {
+              "required": [
+                "num_tiles"
+              ]
+            }
+          ],
```
maddenp-noaa, line 19: I think that says that at least one of these (but not necessarily both) is required. Is that right, or are they both required? If they're both required, we could replace this with a `"required": ["dir", "num_tiles"]` as a sibling to `"properties"`.

As mentioned above in an answer to you question, we could (and IMO should) add the other `config:` items here, specifying their types, but not adding them to `"required"`.

src/uwtools/tests/drivers/test_make_solo_mosaic.py
```diff
@@ -0,0 +1,119 @@
...
+        driverobj = make_solo_mosaic.MakeSoloMosaic(config=config_file, batch=True, dry_run=True)
+    assert driverobj._dry_run is True
+    dryrun.assert_called_once_with()
+
+
+def test_MakeSoloMosaic_provisioned_run_directory(driverobj):
+    with patch.multiple(
+        driverobj,
+        runscript=D,
+    ) as mocks:
```
maddenp-noaa, line 72:
```suggestion
    with patch.multiple(driverobj, runscript=D) as mocks:
```
nitpick: If you have a comma after the last item in a list, `black` will respect that and place list items on separate lines and not combine them, no matter how short the line. But this one is short enough to be on one line.

src/uwtools/tests/drivers/test_make_solo_mosaic.py
```diff
@@ -0,0 +1,119 @@
...
+        driverobj.provisioned_run_directory()
+    for m in mocks:
+        mocks[m].assert_called_once_with()
```
maddenp-noaa, line 75:
```suggestion
def test_MakeSoloMosaic_provisioned_run_directory(driverobj):
    with patch.object(driverobj, "runscript") as runscript:
        driverobj.provisioned_run_directory()
        runscript.assert_called_once_with()
```
Also, `patch.object()` is preferable, due to its comparative brevity, is preferable to `patch.multiple()` when only one item is being patched/mocked.

src/uwtools/tests/test_schemas.py
```diff
@@ -594,6 +601,32 @@ def test_schema_make_hgrid_run_dir(make_hgrid_prop):
...
+    errors = schema_validator("make-solo-mosaic", "properties", "make_solo_mosaic")
+    # Basic correctness:
+    assert not errors(config)
+    # All top-level keys are required:
+    for key in ("config", "execution", "run_dir"):
+        assert f"'{key}' is a required property" in errors(with_del(config, key))
+    # Additional top-level keys are not allowed:
+    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+
+
```
maddenp-noaa, line 622: I think we probably want a `test_schema_make_solo_mosaic_config()` here to exercise the `config:` block -- verifying that validation fails a config lacking the required items, that it fails items with incorrectly typed values, etc. (maybe not etc., maybe that's it). Probably this is related to a question asked earlier about what should _be_ in the schema.

christinaholtNOAA (approved):
LGTM.

docs/sections/user_guide/yaml/components/make_solo_mosaic.rst
```diff
@@ -0,0 +1,41 @@
...
+num_tiles:
+""""""""""
+
+Number of tiles in the mosaic.
+
+execution:
+^^^^^^^^^^
+
+See :ref:`here <execution_yaml>` for details.
+
```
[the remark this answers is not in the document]
elcarpenterNOAA: `dir` and `num_tiles` are both supposed to be under config and are required.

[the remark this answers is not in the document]
maddenp-noaa: Sorry, my mistake! I missed the hierarchy here.

src/uwtools/tests/test_schemas.py
```diff
@@ -594,6 +601,52 @@ def test_schema_make_hgrid_run_dir(make_hgrid_prop):
...
+def test_schema_make_solo_mosaic_config():
+    # Get errors function from schema_validator
+    errors = schema_validator(
+        "make-solo-mosaic", "properties", "make_solo_mosaic", "properties", "config"
+    )
```
maddenp-noaa, line 627: Can we receive the `make_solo_mosaic_prop` fixture here and then
```
errors = make_solo_mosaic_prop("config")
```
here. IIRC that's the pattern in the tests for the other driver configs, except for `test_schema_make_solo_mosaic()` and its analogs for other drivers, where we need to start a level higher in the schema.

maddenp-noaa (approved):
I left one final comment, but this looks great.

src/uwtools/tests/test_schemas.py
elcarpenterNOAA: Updated. Thank you for all of your feedback on this one.

PR #486, opened 2024-05-13
omnibus-2024-05-13

maddenp-noaa:
**Synopsis**

It looks like a lot, but it's not: Just three simple patterns, repeated:

- In UW YAML docs, differentiate lowest level of detail by marking keys bold (instead styling as a heading) and indenting. See e.g. the items under `boundary_conditions:` and `streams:` on [this page](https://uwtools--486.org.readthedocs.build/en/486/sections/user_guide/yaml/components/mpas.html).
- Factor out `_taskname()` logic in `Driver` subclasses: Move the most common (non-cycle, non-leadtime) treatment to the base class as the default behavior, then define -- and call from the appropriate subclasses -- new `_taskname_with_cycle()` and `_taskname_with_cycle_and_leadtime()` methods.
- Partially address an issue with the `iotaa` global dry-run variable by _always_ calling the `iotaa` `dryrun()` method with either `True` or `False` from drivers and from the `file` tool. This ensures that all drivers, plus the `file` tool, operate with the correct dry-run setting. There are still use cases, though, where the `iotaa` dry-run variable could be left set to `True` and would need to be intentionally toggled back to `False`: For example, some of the unit tests did this, breaking tests that ran later; but this PR updates those tests to run with dry-run disabled, which avoids the problem. Future work should probably make this more robust by ensuring that dry-run is toggled back to `False` at the end of any operation that sets it to `True` (possibly with a context manager).

**Type**

- [x] Documentation

**Impact**

**Checklist**


elcarpenterNOAA (commented):
If you want to keep the style consistent across all pages, `dir` through `tile_file` all belong under `config` in the `make_solo_mosaic` driver.

maddenp-noaa:
@elcarpenterNOAA Yes, thank for reminding me -- I'll update that new doc to match.

NaureenBharwaniNOAA (approved):
LGTM!

christinaholtNOAA (approved):
LGTM.

PR #490, opened 2024-05-15
UW-559 Add argument for directing drivers to an arbitrary block of an experiment YAML.

NaureenBharwaniNOAA:
**Synopsis**

In a config we currently support a high-level key with the drivers name. 
i.e. 
```
mpas:
    ....
    ....
```

- We are adding an argument to support the driver being nested further within the config by traversing down a set of keys.
- We also want to support a driver running multiple times and may have a different naming scheme than one chosen.
i.e.
```
foo:
    mpas:
        ....
        ....
bar:
    mpas:
        ....
        ....
``` 

**Type**

**Impact**

**Checklist**


src/uwtools/drivers/make_solo_mosaic.py
```diff
@@ -16,13 +16,20 @@ class MakeSoloMosaic(Driver):
...
+    ):
         """
         The driver.
 
         :param config: Path to config file (read stdin if missing or None).
         :param dry_run: Run in dry-run mode?
         :param batch: Run component via the batch system?
+        :param key_path: Does this driver require a sub-section of YAML to be output?
         """
         super().__init__(config=config, dry_run=dry_run, batch=batch)
```
NaureenBharwaniNOAA, line 34: Add key_path argument here

src/uwtools/tests/test_cli.py
```diff
@@ -562,23 +562,25 @@ def test__dispatch_to_driver():
     args: dict = {
         "action": "foo",
         "batch": True,
-        "config_file": "config.yaml",
+        "config_file": {"foo": {"bar": "config.yaml"}},
```
NaureenBharwaniNOAA, line 565: Trying to figure out how to come up with a test case for this currently.

maddenp-noaa:
A somewhat pedantic point maybe, but is this really a breaking change? Wouldn't everything that already worked continue to work? This seems like only additional functionality that users could just ignore if they didn't need it.

maddenp-noaa:
I don't see any doc updates, and the _I have reviewed the documentation and have made any updates necessitated by this change._ box is not ticked. I think we'd at least need to update the CLI `--help` examples to reflect the new option*, and probably we'd want a simple use example in the CLI docs.

* Feel free to ignore updating the `--help` output: I'll have a PR up soon that automates the generation of that content so that it will be easier for all of us going forward. It would be a waste of your time to manually update it right now. But I think a use example in the docs would still be helpful. Or do we have a different ticket for that work?

src/uwtools/cli.py
```diff
@@ -874,6 +884,7 @@ def _add_subparser_for_driver_task(
     _add_arg_batch(optional)
     _add_arg_dry_run(optional)
     _add_arg_graph_file(optional)
+    _add_arg_key_path(optional)
```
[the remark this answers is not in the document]
maddenp-noaa: This is where simple built-in integration tests would come in handy: If you'd added an integration test that executed, say, `uw upp --config-file config.yaml --key-path a.b [other options]`, it would fail because, without this line, `--key-path` would not be recognized.

But then I have to ask: How did you successfully test this new feature from the CLI without this line? I don't think you could have. What works best for me is doing the code development and manually testing the feature -- which for a CLI-exposed feature like this one would necessitate a command-line test -- until it appears to behave as I expect, and _then_ write unit tests, and _then_ write integration tests (someday soon!), and _then_ write documentation.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -33,8 +34,11 @@ def __init__(
         :param config: Path to config file (read stdin if missing or None).
         :param dry_run: Run in dry-run mode?
         :param batch: Run component via the batch system?
+        :param key_path: Path of keys to subsection of config file.
```
maddenp-noaa, line 37: I wonder if this might benefit from a little more specificity: It's not just _any_ subsection of the config we're looking for (and it's not necessarily a file -- it may be data streamed on `stdin` -- so we should probably not use that word here), it's the driver-configuration block. Maybe something like
```
```
? Here and in the other driver modules?

src/uwtools/drivers/driver.py
```diff
@@ -29,10 +29,11 @@ class Driver(ABC):
 
     def __init__(
         self,
-        config: Optional[Path] = None,
+        config: Optional[Union[dict, Path]],
```
maddenp-noaa, line 32: I think we made this change to support the unit tests. Am I recalling correctly? I think it's a good change, and maybe some day we'll have reason to make the same change in the subclasses, but not today.

NaureenBharwaniNOAA:
> I don't see any doc updates, and the _I have reviewed the documentation and have made any updates necessitated by this change._ box is not ticked. I think we'd at least need to update the CLI `--help` examples to reflect the new option*, and probably we'd want a simple use example in the CLI docs.
> 
> * Feel free to ignore updating the `--help` output: I'll have a PR up soon that automates the generation of that content so that it will be easier for all of us going forward. It would be a waste of your time to manually update it right now. But I think a use example in the docs would still be helpful. Or do we have a different ticket for that work?
> 
> **EDIT** Sorry, I realize now that this is for driver support, and it would be annoying to repeat an example of using `--key-path` for every driver. Maybe a generic use case could be constructed in a file in `docs/shared` and then that snipped included in each driver CLI page. It might show something like changing from
> 
> ```
> driver:
>     ... driver config ...
> ```
> 
> to
> 
> ```
> a:
>   b:
>     driver:
>       ... driver config ...
> ```
> 
> and how `--key-path a.b` could be used to select the correct driver.

I'll add this in the documentation. Originally when I made the draft PR I didn't add any yet as I was still figuring out the bool value to a List of strings. I'll update this soon.

src/uwtools/drivers/driver.py
NaureenBharwaniNOAA: Yes you are! 

It was this error message:
```
=> Running typechecker
+ mypy --install-types --non-interactive .
uwtools/tests/drivers/test_driver.py:123: error: Argument "config" to
"ConcreteDriver" has incompatible type "dict[str, dict[str, Any]]";
expected "Path | None"  [arg-type]
            config={"foo": {"bar": config}},
                   ^~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 127 source files)
```

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -93,6 +93,11 @@ Its contents are described in depth in section :ref:`chgres_cube_yaml`. Each of
 
      $ uw chgres_cube run --config-file config.yaml --cycle 2023-12-15T18 --batch --dry-run
 
+* Specifying the ``--key-path`` flag follows a path of keys leading through the config to extract the driver's configuration block.
+
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/arg_key_path.yaml
+
```
NaureenBharwaniNOAA, line 100: I added the `key_path` arg documentation here for a single driver. Once I get some feedback on how this looks, I can add it to all drivers. I also can factor all the information out to it's own section (not just the shared yaml) since this will be the same across all drivers.

docs/sections/user_guide/cli/drivers/chgres_cube.rst
NaureenBharwaniNOAA, line 1: @maddenp-noaa I didn't add the `--key-path` argument to the help output from `uw chgres_cube run --help` command here based on this comment:

> Feel free to ignore updating the --help output: I'll have a PR up soon that automates the generation of that content so that it will be easier for all of us going forward. It would be a waste of your time to manually update it right now. But I think a use example in the docs would still be helpful. Or do we have a different ticket for that work?

I believe that is #495 that you're currently working on. I see here that you factored it out into this [file](https://github.com/ufs-community/uwtools/pull/495/files#diff-0cdc92eda7b1121e63f56f7bedd12951e9b69114cf8274f607046755aa7046e3R1). I think the updating would depend on which merged first.

maddenp-noaa: I think it'll be easiest is you don't update the `--help` output, as it will create merge conflicts if you do. If I merge before you do, after you merge my changes onto your branch you'll just need to run `make docs` from your clone root and any changes related to your work in this PR should be reflected in the docs. Let's talk if my PR merges before yours.

docs/sections/user_guide/cli/drivers/chgres_cube.rst
maddenp-noaa: I think it would keep things DRYer if the explanatory text as well as the YAML example were include-able. For example, something like `key_path_nested.yaml`
```
foo:
  bar:
    driver:
      # driver config block
```
and `key_path_simple.yaml`
```
driver:
  # driver config block
```
and `key_path.rst`
```
The ``--key-path`` option can be used to navigate from the top of the config to the driver's configuration block. For example, specifying ``--key-path foo.bar`` with config

.. literalinclude:: shared/key_path_nested.yaml
.. highlight:: yaml

is equivalent to using config

.. literalinclude:: shared/key_path_simple.yaml
.. highlight:: yaml

without specifying ``--key-path``.
```
Then, I think, we could just add the single line
```
.. literalinclude:: ../../../../shared/key_path.rst
```
to each driver doc to get the explanatory text and nicely formatted YAML snippets. I'm not sure we have other examples of one include including another, so hopefully that's more-or-less right.

Does that proposed explanatory text make sense?

docs/shared/key_path.rst
```diff
@@ -0,0 +1,11 @@
+The ``--key-path`` option can be used to navigate from the top of the config to the driver's configuration block. For example, specifying ``--key-path foo.bar`` with config
+
+  .. highlight:: yaml
+  .. literalinclude:: ../../../../shared/key_path_nested.yaml
```
NaureenBharwaniNOAA, line 4: Failing the CI/CD run with an error: 
```
Warning, treated as error:
/Users/naureenbharwani/Desktop/UFS/uwtools/docs/shared/key_path.rst:6:Include file '/Users/naureenbharwani/Desktop/shared/key_path_nested.yaml' not found or reading it failed
```

The documentation is being generated correctly [here](https://uwtools--490.org.readthedocs.build/en/490/sections/user_guide/cli/drivers/chgres_cube.html).

docs/shared/key_path.rst
```diff
@@ -0,0 +1,11 @@
+The ``--key-path`` option can be used to navigate from the top of the config to the driver's configuration block. For example, specifying ``--key-path foo.bar`` with config
```
maddenp-noaa, line 1:
```suggestion
  The ``--key-path`` option can be used to navigate from the top of the config to the driver's configuration block. For example, specifying ``--key-path foo.bar`` with config
```

docs/shared/key_path.rst
```diff
@@ -0,0 +1,11 @@
...
+
+  .. highlight:: yaml
+  .. literalinclude:: ../../../../shared/key_path_nested.yaml
+
+is equivalent to using config
```
maddenp-noaa, line 6:
```suggestion
  is equivalent to using config
```

docs/shared/key_path.rst
```diff
@@ -0,0 +1,11 @@
...
+
+  .. highlight:: yaml
+  .. literalinclude:: ../../../../shared/key_path_simple.yaml
+
+without specifying ``--key-path``.
```
maddenp-noaa, line 11:
```suggestion
  without specifying ``--key-path``.
```

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -93,6 +93,9 @@ Its contents are described in depth in section :ref:`chgres_cube_yaml`. Each of
 
      $ uw chgres_cube run --config-file config.yaml --cycle 2023-12-15T18 --batch --dry-run
 
+
+.. include:: ../../../../shared/key_path.rst
```
maddenp-noaa, line 97: This looks good to me in the rendered docs. If no one else disagrees, I think we could add this line to the other driver `.rst` files, since they all inherit the `--key-path` functionality.

src/uwtools/cli.py
```diff
@@ -638,6 +638,16 @@ def _add_arg_key_eq_val_pairs(group: Group) -> None:
     )
 
 
+def _add_arg_key_path(group: Group) -> None:
+    group.add_argument(
+        _switch(STR.keypath),
+        help="Dot-separated path of keys leading through the config to the driver's configuration block",
```
NaureenBharwaniNOAA, line 644: I wanted to point out that I changed the help message for the argument as well.

docs/shared/key_path.rst
maddenp-noaa: I checked out your branch and had a look at this and think I know the solution. Try specifying
```
.. literalinclude:: /shared/key_path_nested.yaml
```
and
```
.. literalinclude:: /shared/key_path_simple.yaml
```
Apparently, and this was news to me, Sphinx interprets absolute paths as relative to "the top source directory" which, in our case, for docs, is `docs/`. So, for me, the above works.

Also, I think I misled you about the indentation: I think the lines in this file should be all the way to the left, but the line
```
.. include:: ../../../../shared/key_path.rst
```
in `chgres_cube.rst` should be indented to match its context.

maddenp-noaa (approved):
It all looks great, nice work.

src/uwtools/tests/test_cli.py
```diff
@@ -546,23 +546,25 @@ def test__dispatch_to_driver():
...
```
christinaholtNOAA, line 549: This is an interesting use case I had not considered before. Is the logic implemented to turn this into the contents of the config.yaml file? I see that it doesn't quite matter for the test, but it might give off the wrong idea as an example, if the logic does not exist. 

I don't think this should hold up this PR, to be clear.

christinaholtNOAA (approved):
LGTM. Just one clarification question, but I'm happy with it as-is.

src/uwtools/tests/test_cli.py
NaureenBharwaniNOAA: Great point, keeping the original text as is passes the tests as well. I originally added this in when debugging how to get 100% test coverage. I don't think it needs to be here and happy to remove if agree

maddenp-noaa: It's slightly complicated: To support the tests in `uwtools.tests.drivers.test_driver` via the `ConcreteDriver(Driver)` class defined there, we updated the signature of the `Driver` base class constructor to accept a `dict`, to avoid the need to write to disk. I suspect this is where @NaureenBharwaniNOAA got the idea to use a `dict` in the current test. This is an obviously valid change, because `Driver` immediately passes this value to the `YAMLConfig` constructor, which is built to receive a `dict` (among other value types).

But all the `Driver` subclasses currently require `config` to be `Optional[Path]` and don't support a `dict` value. We could easily change that later if we found a need, but I don't see it now.

That said, in the current test, the driver object is a `Mock` anyway, so it doesn't matter what the value is. As @christinaholtNOAA suggests, using a `dict` here might be confusing. I'd suggest changing it to a `str` like `"/path/to/config"`. I could change it in a near-future omnibus PR if the change doesn't make it into this PR.

NaureenBharwaniNOAA: Updating with this change now. I applied the change to the `test__dispatch_to_driver()` test case specifically. I believe that was the action to take.

PR #491, opened 2024-05-15
uw-579-resolve-walltime-inconsistencies

maddenp-noaa:
**Synopsis**

Make `execution.batchargs.walltime` required when `execution.batchargs` is present in a driver config. See [UW-579](https://jira.epic.oarcloud.noaa.gov/browse/UW-579) for details. This is not a breaking change because the code that looks up the value of `walltime` does so with `[]` syntax (not `.get()` syntax), so that that it was implicitly required already. Existing use cases that work should continue to work.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/resources/jsonschema/execution-serial.jsonschema
+++ b/src/uwtools/resources/jsonschema/execution-serial.jsonschema
@@ -2,6 +2,15 @@
   "additionalProperties": false,
   "properties": {
     "batchargs": {
+      "additionalProperties": true,
+      "properties": {
+        "walltime": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "walltime"
+      ],
       "type": "object"
     },
     "envcmds": {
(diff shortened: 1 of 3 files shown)


elcarpenterNOAA (approved):
Looks good to me.

WeirAE (approved):
LGTM

src/uwtools/resources/jsonschema/execution-serial.jsonschema
```diff
@@ -2,6 +2,15 @@
   "additionalProperties": false,
   "properties": {
     "batchargs": {
+      "additionalProperties": true,
```
maddenp-noaa, line 5: Although `true` is the default here, I added this line to be explicit about the fact that we allow any key-value pairs here, which are used to create additional batch arguments. For example, for the Slurm scheduler, a UW YAML `batchargs` value like `{walltime: "00:05:00", "foo": "bar"}` would produce directives
```
#SBATCH --foo=bar
#SBATCH --time=00:05:00
```
There are many directives available for each scheduler, and we've so far left room for users to specify them arbitrarily, apart from a few common, often-used ones -- of which `walltime` (which gets translated to the correct actual command-line argument depending on the scheduler in use) is the only _required_ argument.

PR #493, opened 2024-05-20
UW-580 config realize update

maddenp-noaa:
**Synopsis**

Replace the supplemental-files feature of `config realize` with a single update-config / update-format pair.

For example, given `input.yaml`
```
foo: "{{ x }}"
```
and ``update.yaml``
```
x: bar
```
we can now
```
% uw config realize --input-file input.yaml --update-file update.yaml --output-format yaml
foo: bar
x: bar
% cat input.yaml | uw config realize --input-format yaml --update-file update.yaml --output-format yaml
foo: bar
x: bar
% cat update.yaml | uw config realize --input-file input.yaml --update-format yaml --output-format yaml
foo: bar
x: bar
% echo "x: qux" | uw config realize --input-file input.yaml --update-format yaml --output-format yaml
foo: qux
x: qux
```
etc. This addresses a user request to be able to provide update values on `stdin`.

The `--update-format` option addresses another user request to be able to specify the format of update values.

Finally, and addressing a third user request, `--verbose` output now provides additional messages about when streams are in use. For example, from the `--verbose` output of the last `uw` command shown above:
```
[2024-05-20T19:56:50]    DEBUG Reading update from stdin
[2024-05-20T19:56:50]    DEBUG Writing output to stdout
```

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/config/tools.py
```diff
@@ -79,33 +78,24 @@ def config_check_depths_update(config_obj: Union[Config, dict], target_format: s
...
+    input_obj = _realize_config_update(input_obj, update_config, update_format)
     input_obj.dereference()
-    output_data = input_obj.data
-    if output_block is not None:
-        for key in output_block:
-            output_data = output_data.get(key, {})
-    config_check_depths_realize(output_data, output_format)
+    output_data, output_format = _realize_config_output_setup(
+        input_obj, output_file, output_format, output_block
+    )
```
maddenp-noaa, line 98: The old code between the function signature and here grew in complexity, so I factored it out to three separate functions -- `_realize_config_input_setup()`, `_realize_config_output_setup()`, and `_realize_config_update()` -- each with their own, new tests.

src/uwtools/config/tools.py
```diff
@@ -136,7 +126,7 @@ def _ensure_format(
     :raises: UWError if the format cannot be determined.
     """
     if isinstance(config, Config):
-        return FORMAT.yaml
+        return config.get_format()
```
maddenp-noaa, line 129: For the old supplemental files, we enforced that if `Config` objects were provided via the API, that they were `YAMLConfig` objects. We can loosen that requirement now and query the format from the `Config` object. Depth compatibility requirements are still enforced.

src/uwtools/config/tools.py
```diff
@@ -170,38 +160,82 @@ def _print_config_section(config: dict, key_path: List[str]) -> None:
...
+    """
+    Set up config-realize input.
+
+    :param input_config: Input config source (None => read stdin).
+    :param input_format: Format of the input config.
+    :return: The input Config object and its format name.
+    """
+    input_format = _ensure_format("input", input_format, input_config)
+    if not input_config:
+        log.debug("Reading input from stdin")
```
maddenp-noaa, line 175: This is an example of new debug logging showing users, when they specify `--verbose` (or connect `uwtools` to a verbose logger via the API), what is being read from `stdin`.

src/uwtools/config/tools.py
```diff
@@ -248,48 +282,24 @@ def _validate_depth(
     target_class = format_to_config(target_format)
     config = config_obj.data if isinstance(config_obj, Config) else config_obj
     if bad_depth(target_class.get_depth_threshold(), depth(config)):
-        raise log_and_error(
+        raise UWConfigError(
```
maddenp-noaa, line 285: This error was being printed twice via the CLI since the CLI catches and prints/logs UW exceptions. API users can catch the exception and log if they want to.

src/uwtools/config/tools.py
```diff
@@ -248,48 +282,24 @@ def _validate_depth(
...
             "Cannot %s depth-%s config to type-'%s' config" % (action, depth(config), target_format)
         )
 
 
-def _validate_format_output(input_fmt: str, output_fmt: str) -> None:
+def _validate_format(other_fmt_desc: str, other_fmt: str, input_fmt: str) -> None:
```
maddenp-noaa, line 290: This is now more generic as it handles both output formats and update formats -- so "other" instead of "output".

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -284,229 +278,224 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
        date: 20240105
        empty: null
        greeting: Good Night
-       message: 'Good Night Moon Good Night Moon '
+       message: Good Night Moon Good Night Moon
```
NaureenBharwaniNOAA, line 281: How did we get rid of all these quotes around the messages? I wasn't sure where this change was coming from, or if this was from an older PR and just needed to be updated.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -284,229 +278,224 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
...
        repeat: 2
 
-* By default, variables/expressions that cannot be rendered are passed through unchanged in the output. For example, given config file ``config.yaml`` with contents
-
-  .. code-block:: yaml
-
-     roses: "{{ color1 }}"
-     violets: "{{ color2 }}"
-     color1: red
+* With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
```
NaureenBharwaniNOAA, line 285:
```suggestion
* With the ``--dry-run`` flag specified, nothing is written to ``stdout`` (or content is saved to a file if ``--output-file`` is specified), but a report of what would have been written is logged to ``stderr``:
```

The original made it sound like nothing was written to a file if ``--output-file`` was specified. If that is the case maybe we can make this line clearer.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -284,229 +278,224 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
...
+       recipient: Moon
+       repeat: 2
 
-* If an input file is read alone from ``stdin``, ``uw`` will not know how to parse its contents:
+* Similarly, if an input file is read from ``stdin``, ``uw`` will not automatically know how to parse its contents:
 
   .. code-block:: text
 
-     $ cat config.yaml | uw config realize --output-file realized.yaml values1.yaml
+     $ cat config.yaml | uw config realize --update-file update.yaml --output-format yaml
```
NaureenBharwaniNOAA, line 322:
```suggestion
     $ cat config.txt | uw config realize --update-file update.yaml --output-format yaml
```

Based on the line (The format can be explicitly specified (`config.txt` is a copy of `config.yaml`) below I think you want this update. If not, maybe removing that line would be useful since it seems to be misplaced with `config.txt` as an example.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -284,229 +278,224 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
...
      Specify --input-format when --input-file is not specified
 
-* To read the config from ``stdin`` and realize to ``stdout``:
+  The format can be explicitly specified  (``config.txt`` is a copy of ``config.yaml``):
 
   .. code-block:: text
 
-     $ cat config.yaml | uw config realize --input-format yaml --output-format yaml values1.yaml
+     $ cat config.yaml | uw config realize --update-file update.yaml --output-format yaml --input-format yaml
```
NaureenBharwaniNOAA, line 329:
```suggestion
     $ cat config.txt | uw config realize --update-file update.yaml --output-format yaml --input-format yaml
```

Same with this one, if that's not what was intended feel free to disregard.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -284,229 +278,224 @@ and an additional supplemental YAML file ``values2.yaml`` with the following con
...
 
-  In this case, the format can be explicitly specified  (``config.txt`` is a copy of ``config.yaml``):
+      $ uw config realize --input-file config.yaml --output-format sh
+      [2024-05-20T19:17:02]    ERROR Cannot realize depth-2 config to type-'sh' config
+
+* It is possible to provide update values, rather than the input config, on ``stdin``. Usage rules are as follows:
+
+  * Only if either ``--update-file`` or ``--update-config`` are specified will ``uw`` attempt to read and apply update values to the input config.
+  * If ``--update-file`` is provided with an unrecognized (or no) extension, or if the update values are provided on ``stdin``, ``--update-format`` must be used to specify the correct format.
+  * At least one of the input config and the update config must be provided via file: They cannot be streamed from ``stdin`` simultaneously.
```
NaureenBharwaniNOAA, line 363:
```suggestion
  * At least one of the input config or the update config must be provided via file: They cannot be streamed from ``stdin`` simultaneously.
```

src/uwtools/api/config.py
```diff
@@ -147,13 +148,14 @@ def realize(
...
-    supplemental_configs: Optional[List[Union[dict, _Config, Path, str]]] = None,
+    update_config: Optional[Union[dict, _Config, Path, str]] = None,
+    update_format: Optional[str] = None,
     values_needed: bool = False,
     dry_run: bool = False,
     stdin_ok: bool = False,
 ) -> dict:
     """
-    Realize a config to a ``dict``, based on an input config and optional supplemental configs.
+    Realize a config to a ``dict``, based on a base input config and optional update config.
```
NaureenBharwaniNOAA, line 158:
```suggestion
    Realize a config to a ``dict``, based on a base input config and an optional update config.
```

src/uwtools/api/config.py
```diff
@@ -228,14 +231,15 @@ def validate(
 
 :param input_config: Input config file (``None`` or unspecified => read ``stdin``)
 :param input_format: Format of the input config (optional if file's extension is recognized)
-:param output_block: Path through keys to the desired output block
+:param update_config: Input config file (``None`` or unspecified => read ``stdin``)
```
NaureenBharwaniNOAA, line 234:
```suggestion
:param update_config: Update config file (``None`` or unspecified => read ``stdin``)
```

src/uwtools/cli.py
```diff
@@ -145,17 +145,19 @@ def _add_subparser_config_realize(subparsers: Subparsers) -> ActionChecks:
     optional = _basic_setup(parser)
     _add_arg_input_file(optional)
     _add_arg_input_format(optional, choices=FORMATS)
+    _add_arg_update_file(optional)
```
NaureenBharwaniNOAA, line 148: In other sections this is called `update_config` should we keep this consistent across files?

src/uwtools/cli.py
```diff
@@ -744,6 +737,27 @@ def _add_arg_total(group: Group) -> None:
...
+        required=required,
+        type=Path,
+    )
+
+
+def _add_arg_update_format(group: Group, choices: List[str], required: bool = False) -> None:
+    group.add_argument(
+        _switch(STR.updatefmt),
+        choices=choices,
+        help="Input format",
```
NaureenBharwaniNOAA, line 755:
```suggestion
        help="Update format",
```

Should this be update format since we are passing in the `update file` and not in the input file here?

NaureenBharwaniNOAA (commented):
Still going through some of the test cases, but here are some initial thoughts.

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: That was done [here](https://github.com/ufs-community/uwtools/pull/493/files#diff-a4378ed914a7ad3423e757c932d624dff449acbaff2574aaed9ec48e57c29c93R207).

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: I agree, will clarify.

maddenp-noaa: Oh, sorry, I misread the suggest and I actually disagree. I think the original PR text is accurate and grammatically correct.

maddenp-noaa: @NaureenBharwaniNOAA I mean, you're right about the meaning of the original text, but I believe this text just moved in the file and did not change, and I think the original text was correct: It says, IMO slightly less clumsily, that when `--dry-run` is specified, nothing is written to `stdout`; and, also, that if `--dry-run` is specified when `--output-file` is specified, nothing will be written to a file. Does that make sense?

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: I think this line is correct, but the "(`config.txt` is a copy of `config.yaml`) is in the wrong place and causing confusion. The purpose of this line is to show that, even though the file being streamed in on `stdin` is YAML, `uw` has no way to know that and so needs the `--input-format yaml` option (shown below) to tell it. I'll remove the bit about `config.txt`, which I think was a copy/paste error. Thanks!

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: I think this one becomes correct after removing the misleading reference to `config.txt` in the preceding commentary, which I'm about to do.

NaureenBharwaniNOAA: Yup definitely agree, I wasn't sure which way you wanted to go on the comments. This works great! Thanks.

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: @NaureenBharwaniNOAA What's your thinking on this and/or (no pun intended) do you have a reference for this? I thought about this when I wrote it and landed on "and" to avoid suggesting that this is a mutually exclusive choice. It's ok for either one to be true, but also for both of them to be true. Normal usage of "or" for English speakers tends to assume mutual exclusion, for example, if I say "We have two pieces of candy left. Do you want the caramel or the peppermint?" I don't expect you to say "Yes" and eat them both, though _mathematically_ this would be a correct interpretation of "or". The accepted answer in [this StackExchange topic](https://patents.stackexchange.com/questions/20738/should-i-use-and-or-or-for-at-least-one-of-structure-when-only-two-items-i) reflects my thinking, but I'm having trouble finding anything authoritative about this. Google Docs doesn't flag either construction is grammatically incorrect. I'd be happy to take a tie-breaker vote from another reviewer...

src/uwtools/cli.py
maddenp-noaa: I don't think so: This is the CLI, so users can only provide _files_ -- they cannot provide Python `dict` objects or `Config` objects on the command line. This is why it's also e.g. `_add_arg_input_file()` just above. Once we're at the API and beyond, there's a chance that users may be providing in-memory `dict` or `Config` objects, so the terminology changes from "file" to "config" to reflect the ambiguity.

src/uwtools/cli.py
maddenp-noaa: Yes, thanks!

docs/sections/user_guide/cli/tools/config.rst
NaureenBharwaniNOAA: My thinking was with either in front of the sentence to me it implied only one or the other. But if it's the case that 1. both are accepted or 2. either one can stand alone, this changes my opinion of the sentence.

elcarpenterNOAA: I think the English "or" is relatively ambiguous outside of further context. To counter the candy example, if a waiter were to ask at the end of a meal "Would you like coffee or dessert?" I know it's not exclusive -- I could (and will) get both. Neither the "and" nor the "or" are carrying much weight in this case: it's the "At least one of the" at the beginning of the sentence that conveys the meaning. I think it is clear enough with "and", but could see how someone could interpret it as "at least one of _each_". That's not to say they are _correct_ to interpret it that way, but, particularly for non-natives, the grammatically correct form is not always the clearest.

maddenp-noaa: If the two options were _categories_ instead of _elements_, I think the "and" could be problematic. For example TIL that a US Circuit Court [ruled](https://www.bipc.com/ptab-establishes-plain-meaning-of-at-least-one-of-a-and-b) that "at least one of A and B", where A and B are categories, requires at least two elements -- at least one each from each of the categories; whereas "at least one of A or B" could be just a single element from either category. Lawyers!

There's probably no right answer to this given that humans have to read and interpret the words. I'll probably leave it as-is unless someone else takes a strong position.

src/uwtools/tests/config/test_tools.py
```diff
@@ -670,51 +599,200 @@ def test__print_config_section_yaml_not_dict():
...
+    assert input_obj.data == {"section": {"foo": "bar"}}
+    assert input_format == FORMAT.ini
 
 
-def test__realize_config_update_file(realize_config_testobj, tmp_path):
-    values = {1: {2: {3: 99}}}
+def test__realize_config_input_setup_ini_stdin(caplog):
+    data = """
+    [section]
+    foo = bar
```
christinaholtNOAA, line 626: One interesting thing I noticed here was that the integer-supporting formats have integers in the tests. I'm curious if it's worth showing here that the integer is processed as a string from ini format...if that's true, of course.

christinaholtNOAA (approved):
There's a lot here, but I think it looks good in general.

src/uwtools/config/tools.py
```diff
@@ -170,38 +160,82 @@ def _print_config_section(config: dict, key_path: List[str]) -> None:
...
-            config_obj.update_values(supplemental_obj)
-            log.debug("After update, config has depth %s", config_obj.depth)
-    else:
-        log.debug("Input config has depth %s", config_obj.depth)
-    return config_obj
+    if update_config or update_format:
+        update_format = _ensure_format("update", update_format, update_config)
+        if not update_config:
+            log.debug("Reading update from stdin")
+        _validate_format("update", update_format, update_format)
```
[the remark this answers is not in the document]
maddenp-noaa: Yes, thanks! Fixed.

src/uwtools/tests/config/test_tools.py
maddenp-noaa: Good idea. I just extended this test to include an `int` value.

src/uwtools/cli.py
```diff
@@ -928,6 +942,13 @@ def _check_template_render_vals_args(args: Args) -> Args:
     return args
 
 
+def _check_update(args: Args) -> Args:
+    if args.get(STR.updatefile) is not None:
+        if args.get(STR.updatefmt) is None:
+            args[STR.updatefmt] = get_file_format(args[STR.updatefile])
+    return args
```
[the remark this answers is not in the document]
maddenp-noaa: It's not the case that at least one of `--input-file` and `--update-file` be provided: If neither are provided, then `uw` will read the input from `stdin` and will not perform an update (it will just process the input file with internal information or potentially environment variables if the new-ish `env` filter is used). If `--update-file` is provided, `uw` will perform an update via the specified file; if `--update-format` is provided and `--update-format` is not, `uw` will read the update info from `stdin` **unless** `--input-file` is also not provided, which is then an `error`. This is documented [here](https://github.com/ufs-community/uwtools/pull/493/files#diff-a4378ed914a7ad3423e757c932d624dff449acbaff2574aaed9ec48e57c29c93R359-R363).

docs/sections/user_guide/cli/tools/config.rst
gspetro-NOAA: I was going to ask what Circuit Court the UW Team would fall under, but actually, the Federal Circuit would be a logical choice if we were interested in patenting `uwtools`... :D On a more serious note, why not reword the sentence to make the meaning clear instead of hoping that readers agree with the Federal Circuit? ;) For example, "The input config, the update config, or both must be provided via file; they cannot be streamed from ``stdin`` simultaneously." (Note that that colon should be a semicolon followed by lowercase or else a period followed by new sentence.)

maddenp-noaa: @gspetro-NOAA Excellent alternative! I will use it. As much as I'd like to establish a precedent and not have to avoid this construction forever, clarity wins the day.

I'm not sure about the punctuation and capitalization, though. Doesn't it depend on the style one follows? I was initiated into AP Style*, where the text following a colon is capitalized if it's a complete sentence (✔️), and a colon is used to clarify or complete the preceding statement -- "to introduce an explanatory statement" (✔️ IMO because the text after the colon explains _why_ the text before it is a rule). But I think maybe you have an MLA background; good stuff if so, but old habits etc.

* I'm an AP apostate because I believe in the Oxford Comma.

gspetro-NOAA: Ha! I am a fellow apostate and have established the Oxford comma as one of the program's standard exceptions to AP Style (https://confluence.epic.oarcloud.noaa.gov/x/C4C2AQ). That said, I believe that AP Style's rule is to use the Oxford comma only in technical/scientific documents, so I just generalized that rule, since (a) the Oxford comma is just better (aka clearer) ;) and (b) much of what we do is technical anyway. 
As to the colon, the program doesn't have a specific rule (although I could make one... ;) ). We are loosely based on AP Style, but that is difficult to adhere to strictly without actually being given a copy of the AP Style Guide. It looks like technically, what you're doing may be allowed in AP Style, but without the style guide, the details are a little fuzzy. A colon can be used when followed by an illustration or elaboration--usually (but not necessarily?) in the form of a list. I typically see a colon used only when a list of examples follow. Technically, examples are a type of explanation/elaboration, so this seems to be a bit of a square/rectangle problem. A colon can be used with an explanatory statement that's a list, but can it be used with a broader array of explanatory statements? Possibly. But how broad an array? AP Style does say to use colons sparingly, and it seems like a period would suffice here. The idea already appears as a single bullet point, which itself suggests that the ideas are connected (otherwise they would be separate bullets). Taken to the extreme, whole paragraphs could sometimes pop in a bunch of colons since usually, ideas build on each other and continue to explain previous statements. In short, this seems like a case where you _can_ justify using a colon, but I would still air on the side of a period.

PR #495, opened 2024-05-21
UW-491 Automate injection of CLI commands/inputs/outputs into RST docs

maddenp-noaa:
**Synopsis**

Don't Panic. There are a **lot** of new files in this PR, but they are almost (?) exclusively content that was factored out of `.rst` files.

Automate injection of CLI commands, inputs, and outputs into RST docs in the CLI section.

This is implemented as a collection of `Makefile`s that recursively handle appropriate directories, executing code stored in `.cmd` files to generate outputs stored in `.out` files. Data inputs to the CLI examples are likewise factored out to separate files, and RST directives to include the contents of the `.out` and input files replace the original content. All of the above command and content files are added to git. Where appropriate (currently only adjacent to the docs for the `file` tool, which needs to create temporary on-disk files and directories), `.gitignore` rules have been put in place to mask ephemeral files from exposure to git.

An `examples` target has been added to `docs/Makefile` to perform updates to the new content files. This allows ReadTheDocs CI to continue to run only `make docs` to render the docs to HTML for publishing, without running `make examples`. However, the `make docs` target in the top-level `Makefile` has been updated to run _both_ `make examples` and `make docs`, and it what should be run by developers in a dev shell to update docs.

Note that, contrary to `make` convention, the `.out` files are _always_ updated by `make examples`, regardless of whether their corresponding `.cmd` files have been updated. (Unlike the more common case where, say, a `.o` file will only be regenerated when its corresponding `.c` file has a newer timestamp.) This is because it is not only changes to `.cmd` files that should trigger generation of new `.out` files, but updates to `uw` itself. So, `make docs` from the clone root (or `make examples` from `docs/`) will always regenerate `.out` files, and updated timestamps in log messages are to be expected. These are always safe to commit, and can be ignored in PR diffs.

As you'll see when reviewing the preview docs, the command and outputs are now split into separate code blocks in the rendered HTML. I think this is nice, as it makes it easier for users to copy and paste commands into the terminal for testing. You might notice that it starts to look something like a Jupyter notebook, which might be a hint that we're heading in the right direction with our plans to develop demo notebooks. I have added highlighting to the `uw` commands in the docs; in a couple of cases, I think this nicely draws attention to the pertinent `uw` command and away from some supporting commands that also have to be run for correct behavior.

Most of the changes are in the [CLI section](https://uwtools--495.org.readthedocs.build/en/495/sections/user_guide/cli/index.html) and deserve careful consideration, but you can see from the _Files Changed_ list what other docs got updates.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/utils/tasks.py
```diff
@@ -70,4 +70,7 @@ def symlink(target: Path, linkname: Path):
     yield asset(linkname, linkname.exists)
     yield existing(target)
     linkname.parent.mkdir(parents=True, exist_ok=True)
-    os.symlink(src=target, dst=linkname)
+    os.symlink(
+        src=target if target.is_absolute() else target.relative_to(linkname.parent, walk_up=True),
+        dst=linkname,
+    )
```
maddenp-noaa, line 76: This update supports target files with relative paths. For example (and this is taken from the dynamic docs for the `file link` tool/action), given `src/` directory
```
src
├── bar
└── foo
```
and `config.yaml`
```
foo: src/foo
subdir/bar: src/bar
```
command `uw file link --target-dir dst --config-file config.yaml` would previously have given us
```
dst
├── foo -> src/foo
└── subdir
    └── bar -> src/bar
```
which is wrong. With the change, we get
```
dst
├── foo -> ../src/foo
└── subdir
    └── bar -> ../../src/bar
```
which is correct.

docs/environment.yml
```diff
@@ -4,3 +4,4 @@ channels:
 dependencies:
   - sphinx_rtd_theme=1.3.0
   - sphinxcontrib-bibtex=2.6.1
+  - tree
```
maddenp-noaa, line 7: The dynamic content generators use `tree`, which may not be available on every developer's system. This ensures that it's made available when running `source docs/install-deps`, which was already required in the Contributor Guide for doc updates.

src/uwtools/cli.py
```diff
@@ -350,7 +350,7 @@ def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
 
     :param subparsers: Parent parser's subparsers, to add this subparser to.
     """
-    parser = _add_subparser(subparsers, STR.rocoto, "Realize and validate Rocoto XML Documents")
+    parser = _add_subparser(subparsers, STR.rocoto, "Realize and validate Rocoto XML documents")
```
maddenp-noaa, line 353: For consistency with other mode descriptions.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -504,137 +256,59 @@ and YAML file ``update.yaml`` with the following contents:
...
-     [2024-01-03T17:31:19]    ERROR      'type': 'object'}
-     [2024-01-03T17:31:19]    ERROR
-     [2024-01-03T17:31:19]    ERROR On instance['values']:
-     [2024-01-03T17:31:19]    ERROR     {'greeting': 'Hello'}
+  .. literalinclude:: config/values-bad.yaml
+     :language: yaml
+  .. literalinclude:: config/validate-fail.cmd
+     :language: text
+     :emphasize-lines: 1
+  .. literalinclude:: config/validate-fail.out
```
christinaholtNOAA, line 303: Looking at the rendered output, I'm not sure this one failed as we describe here in the text. Maybe I am misreading it?

src/uwtools/utils/tasks.py
christinaholtNOAA: Why not use the `pathlib` methods for both of these? Is there something I'm missing about using `target.relative_to(linkname.parent)`?

christinaholtNOAA (approved):
This PR looks like an exercise in tedium! Thank you for taking it on an improving our ability to accurately document the code!

I did the best I could here. I had to switch over to reading the rendered docs when I got to the generic tools because there are so many different files and references there. I could have missed something in any text changes you made in those sections since I wasn't reviewing diffs.

It looks really good, though! This will no doubt save us time and heartache in the long run.

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -3,63 +3,19 @@
...
-         Only log info, making no changes
-     --graph-file PATH
-         Path to Graphviz DOT output [experimental]
-     --quiet, -q
-         Print no logging messages
-     --verbose, -v
-         Print all logging messages
+.. literalinclude:: chgres_cube/run-help.cmd
+   :language: text
+   :emphasize-lines: 1
```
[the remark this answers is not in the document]
christinaholtNOAA: Just a quick update...this annoyed me at first with the basic one-liners, but when I got to the `uw config` section, I totally understand why we want it.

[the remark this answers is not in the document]
maddenp-noaa: 👍 to the multi-line commands benefiting from highlighting. And, yes, this is the line. The yellow, specifically, is probably a choice made by the specific theme used by ReadTheDocs and could possibly be controlled with some explicit CSS, but I didn't care enough (I might have preferred just bold, but I don't hate the highlighter) to interfere with.

docs/sections/user_guide/cli/tools/config.rst
maddenp-noaa: Nice catch, thank you! The command file actually used `values.yaml` instead of `values-bad.yaml`. I'm pushing a fix.

src/uwtools/utils/tasks.py
maddenp-noaa: The `pathlib` solution doesn't work, unfortunately. From [the docs](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) and in re: the function signature
```
PurePath.relative_to(other, walk_up=False)
```
> When `walk_up` is false (the default), the `path` must start with other.
But `walk_up` was added in 3.12, and the `False` behavior was the previous behavior, so
```
>>> Path("/foo/bar/baz").relative_to("/foo")
PosixPath('bar/baz')
```
is ok but
```
>>> Path("/foo/bar/baz").relative_to("/foo/qux")
  File "/usr/lib/python3.10/pathlib.py", line 818, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
ValueError: '/foo/bar/baz' is not in the subpath of '/foo/qux' OR one path is relative and the other is absolute.
```
is not. Compare to
```
>>> os.path.relpath("/foo/qux", "/foo/bar/baz")
'../../qux'
```
The release notes for Python 3.12 say
> Add `walk_up` optional parameter to [`pathlib.PurePath.relative_to()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) to allow the insertion of .. entries in the result; this behavior is more consistent with [`os.path.relpath()`](https://docs.python.org/3/library/os.path.html#os.path.relpath).

maddenp-noaa: So, we can switch to the `pathlib` solution with `walk_up` (which does work, I tested it) when we only support Python 3.12+, but that'll be years from now.

NaureenBharwaniNOAA (approved):
LGTM!

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -3,147 +3,82 @@
...
-
-* If there are differences between the config files, they will be shown below the dashed line. For example, with ``recipient: World`` removed from ``values1.nml``:
+  .. literalinclude:: config/compare-match.cmd
+     :language: text
+     :emphasize-lines: 1
+  .. literalinclude:: config/compare-match.out
+     :language: text
 
-  .. code-block:: text
+* If there are differences between the config files, they will be shown below the dashed line. For example, ``c.nml`` is missing the line ``recipient: World``:
```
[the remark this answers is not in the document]
maddenp-noaa: Probably a good idea to make it completely clear, will do.

PR #496, opened 2024-05-24
UW-586: Build driver for WaveWatchIII

WeirAE:
**Synopsis**

Build driver for WaveWatchIII. As per discussion, scope is limited to "library" version for the coupled model. This also requires splitting the base `Driver()` methods and should not expose this driver to the CLI or API.

AC:
new driver introduced in CLI and API
JSON Schema validation capability
Documentation for CLI, API and landing page addition

**Type**

- [ ] Documentation

**Impact**

- [ ] This is a non-breaking change (existing functionality continues to work as expected)

**Checklist**


src/uwtools/drivers/standalonedriver.py
```diff
@@ -0,0 +1,186 @@
...
+
+from iotaa import asset, task, tasks
+
+from uwtools.drivers.driver import Driver
+from uwtools.exceptions import UWConfigError
+from uwtools.scheduler import JobScheduler
+from uwtools.utils.processing import execute
+
+
+class StandaloneDriver(Driver):
```
maddenp-noaa, line 21: I don't immediately see a reason why this can't live with `Driver` in `driver.py`, as both are closely-related base classes. I don't necessarily see why this one needs its own module. I suppose one could say the same for _all_ the driver modules, though I see a benefit to keeping imports tailored somewhat for each driver (which, then again, is also true here, as you've split the imports nicely). I don't know. Maybe it's just the slightly unwieldy name of this module... I think it would be fine to see
```
from uwtools.drivers.driver import Driver
```
or
```
from uwtools.drivers.drivers import StandaloneDriver
```
Just a thought. I see pros and cons.

src/uwtools/tests/drivers/test_ww3.py
```diff
@@ -0,0 +1,97 @@
...
+
+def test_WaveWatchIII_namelist_file(driverobj):
+    with open(driverobj._driver_config["namelist"]["base_file"], "w", encoding="utf-8") as f:
+        print(
+            "&MISC CICE0 = 0.25, CICEN = 0.75, FLAGTR = 4 / "
+            "&FLX3 CDMAX = 3.5E-3 , CTYPE = 0 / END OF NAMELISTS",
+            file=f,
+        )
+    dst = driverobj._rundir / "namelists.nml"
+    assert not dst.is_file()
```
WeirAE, line 62: This should fix the coverage area as the other drivers use this test, but it is currently failing here

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,97 @@
...
+        yield asset(path, path.is_file)
+        yield None
+        try:
+            namelist = self._driver_config["namelist"]
+        except KeyError as e:
+            raise UWConfigError(
+                "Provide either a 'namelist' YAML block or the %s file" % path
+            ) from e
+        additional_files = namelist.get("additional_files", {})
+        for file in additional_files:
```
WeirAE, line 57: The logic here seems to have become convoluted. I'm torn between making this a 'softlink all specified namelists to the rundir', or fixing this so that it reads all the namelists and generates a merged one correctly?

maddenp-noaa: It would help me, at least, to distinguish between "namelists" and "namelist files": A namelist file can contain multiple namelists, and a namelist does't have to be in a file. And there may be different component- or app-imposed requirements around required names for namelists and files.

What are WW3's requirements w.r.t. namelists and namelist files? I mean, is it expecting to open one file, multiple files? Are the filenames hardcoded/predefined, or do we have control over filenames at run time? Are we certain there's no possibility of conflict between namelist names? Because, if not, concatenating namelists into a single file may not be safe.

I'm also looking at the JSON Schema for this component and trying to understand the use case. Could you describe that, or point to a design doc if I missed/forgot that? It seems that there are potentially multiple namelist files in play, and I don't see any support for `update_values` -- do we need that?

WeirAE: Unfortunately, its not terribly friendly regarding the file components. There are [11 namelists](https://github.com/NOAA-EMC/WW3/tree/develop/model/nml) that are used in the various applications and configurations. There are [different](https://github.com/NOAA-EMC/WW3/blob/14e6bedfcefeff8f0896acea5dade4100a3fbacd/model/bin/README.md) default parameters for some operational configurations. Global Workflow walks through the namelists in a [script](https://github.com/NOAA-EMC/global-workflow/blob/50c2b8951b29a3c883a778becbf8582f9519eb48/ush/parsing_namelists_WW3.sh#L4) so I thought it would be important for us to know and track the namelists used

WeirAE: I left `update_values` out as I was not familiar with any configurations where the user would override the values externally, and it makes it more complicated as well

maddenp-noaa: Those are 11 files files, but more than 11 namelists -- some of the files contain multiple namelists.

I see that some of the namelists (e.g. [this one](https://github.com/NOAA-EMC/WW3/blob/develop/model/nml/ww3_multi.nml#L145-L179)) contain both array-element _and_ object member syntax, and AFAIK there's no way to directly express that in YAML, though we could with some effort define a representation in YAML that could then map on to namelist syntax with custom logic. We haven't done this yet, though, as we've only dealt with simple namelist values.

So, this looks complicated. Would it be reasonable to require users to provide ready-to-go namelist files and have a task called something like `namelist_files()` that behaved similarly to, say, `files_copied()` from the `fv3` driver? If those 11 files with those specific names are required, we could enforce that via the schema.

WeirAE: Fair! WWIII also uses the file-namelist and namelist-namelist interchangeably, which is...ok, really more reasons to strangle away from Fortran where possible. That might go back to my earlier thought of just listing the components needed and marking the necessary files. In Global, it looks like
```# Build exes for prep jobs and post jobs:
prep_exes="ww3_grid ww3_prep ww3_prnc ww3_grid"
post_exes="ww3_outp ww3_outf ww3_outp ww3_gint ww3_ounf ww3_ounp ww3_grib"
```
The `YAML` might then be:
```ww3: 
  config: 
    switch: 
    prep_exes: 
    post_exes: 
    run_dir: 
```
And just use `files_copied()` here as well

maddenp-noaa: What would be the benefit of the `config:` level there? Isn't this all "config"? Or is there something _besides_ `config:` at that level that would need to be kept separate?

WeirAE: Ah, force of habit? I admit an attachment to that subblock, but it is sort of unnecessary here

christinaholtNOAA: If we look at the couple of options that are necessary in "library" mode taking the ufs-weather-model regression test suite as an example, I'm seeing only one option for running WW3 in coupled mode, and that's the ww3_shel.nml.

After diving deeper into the fortran code that's used in the ufs-weather-model, there are two options for providing this configuration. An old non-standard configuration file that's managed with a template: https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/parm/ww3_shel.inp.IN

The same code also promises to use a standard nml file, if it's present with the name `ww3_shel.nml`.

I'd say we support the namelist option and try to find out more about why the ufs-weather-model doesn't use it.

The code that does the work of parsing these two files is here: https://github.com/NOAA-EMC/WW3/blob/d9b3172f4197c65d471662c6952a668152d71230/model/src/wav_shel_inp.F90

We should not worry about any of the 11 other namelist files and focus only on this one.

christinaholtNOAA: @maddenp-noaa If we stick to the one namelist file, I feel like we're going to have some of the same complications that you mentioned surrounding array element and object member syntax. Can we take a short-cut for now to go the flat template route until we can come back to this more complicated problem in the future?

WeirAE: Thanks @christinaholtNOAA , its so easy to get into rabbit holes with this one

maddenp-noaa: @christinaholtNOAA I think templating is the only way to handle these data types today, though we should probably think about a generic way to handle these in the future -- a UW YAML syntax for these kinds of LHS names -- as I don't think we can avoid this forever.

(BTW I think I was wrong in calling the ones we see here "object member syntax". There _is_ object-oriented Fortran, but I think this is just an example of a [user-defined type](https://www.cs.emory.edu/~cheung/Courses/561/Syllabus/6-Fortran/struct.html), which is much more common, in my experience, in NWP Fortran than is actual OO constructs.)

The `f90nml` representation might provide a guide, e.g.
```
>>> f90nml.reads("&nl o(2)%m(4:5)=88,99 /").todict()
OrderedDict({'nl': OrderedDict({'o': [OrderedDict({'m': [88, 99], '_start_index': {'m': [4]}})], '_start_index': {'o': [2]}})})
```
The namelist syntax says that there's an array `o` of some user-defined type that has a member called `m`, which is another array; and we're setting the 4th and 5th elements of `m` on the 2nd element of `o` to the integers `88` and `99`. And that's how `f90nml` keeps track of all that internally. So we could in theory design a UW YAML representation of that, then detect and turn into `f90nml` `Namelist` data, then let `f90nml` serialize back to namelist-formatted data. We might flag such values with a custom YAML tag so that they'd get the proper treatment.

WeirAE: Would it be reasonable to cut this down to the UFS namelist for now, then spin off the syntax as another ticket? I think theres enough time in the sprint to knock that out while its on our minds too.

maddenp-noaa: @WeirAE I was just musing there about how we could eventually support those Fortran namelist constructs in YAML, not suggesting that we implement it now.

christinaholtNOAA: Totally agree with @maddenp-noaa. This is a nice to have that we can include in our backlog, but let's just go the templating route for this work.

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,32 @@
+{
+  "properties": {
+    "ww3": {
+      "additionalProperties": false,
+      "properties": {
+        "namelist": {
+          "additionalProperties": false,
+          "properties": {
+            "additional_files": {
```
christinaholtNOAA, line 9: We don't have a need for this with the change to the namelist task, right? In general, I think the namelist section of this driver should match the namelist section of the other drivers with `base_file` and `update_values` supported in the same way.

christinaholtNOAA (changes requested):
I took a second look at the code surrounding the new driver and provided a few comments. I didn't look at any of the tests in detail at this point.

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,97 @@
...
+            path=path,
+        )
+
+    @tasks
+    def provisioned_run_directory(self):
+        """
+        Run directory provisioned with all required content.
+        """
+        path = self._restart_path
+        path.parent.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
WeirAE: Thanks, I started moving that over in my changes

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: OK, I'll make that change as part of my update - is there a case where a user would use `update_values`, or more for symmetry?

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(config=config, dry_run=dry_run, batch=batch, key_path=key_path)
+
+    # Workflow tasks
+
+    @task
+    def namelist_file(self):
+        """
+        The namelist file.
+        """
+        path = self._rundir / "namelists.nml"
```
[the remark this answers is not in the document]
WeirAE: Should this also accept `ww3_shel.inp.IN`?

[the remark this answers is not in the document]
WeirAE: Nevermind, I got it!

christinaholtNOAA:
@WeirAE I think this looks nice as far as pulling out a base class/subclass for the driver and stand-alone driver. I'm concerned a bit about the wording and getting the relationships right on how we name these classes. I think this kind of activity is closely related to discussions we've had regarding clean code, so I'm going to trigger some semantics questions in offline conversations.

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: Honestly, nothing wrong with keeping it as an option, so I've added it in the latest update.

elcarpenterNOAA:
It seems to be missing the API and CLI documentation. Is that coming later, or in a separate PR?

WeirAE:
@elcarpenterNOAA This one has been limited to just the coupled library mode, which should be fully disconnected from the API and CLI. Though it might be worth still putting something in place to cover it?

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,24 @@
+.. _ww3_yaml:
+
+ww3
+===
+
+Structured YAML when including WaveWatchIII as part of a compiled coupled executable is validated by JSON Schema and requires the ``ww3:`` block, described below.
```
maddenp-noaa, line 6:
```suggestion
Structured YAML to configure WaveWatchIII as part of a compiled coupled executable is validated by JSON Schema and requires the ``ww3:`` block, described below.
```

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,24 @@
...
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/ww3.yaml
+
+UW YAML for the ``ww3:`` Block
+------------------------------
+
+namelists:
+^^^^^^^^^^
+
+The path to the namelist(s) used by the WaveWatchIII configuration. Namelists will be symlinked to the run directory.
```
maddenp-noaa, line 19: Did we decide to render templates into the run directory instead of linking files?

docs/shared/ww3.yaml
```diff
@@ -0,0 +1,12 @@
+ww3:
+  namelist:
+    base_file: "{{ user.ww3_app }}/ww3/ww3_shel.inp.IN"
```
maddenp-noaa, line 3: Is the reference to (and exitence) of the `user.ww3_app` useful here given that it's only used once? IMO this is useful when it DRYs out the config, but is maybe distracting when there's no duplication anyway.

src/uwtools/drivers/driver.py
```diff
@@ -154,6 +123,133 @@ def _driver_name(self) -> str:
...
+        """
+        Perform all necessary schema validation.
+        """
+        for schema_name in (self._driver_name.replace("_", "-"), "platform"):
+            validate_internal(schema_name=schema_name, config=self._config)
+
+
+class StandaloneDriver(Driver):
+    """
+    An abstract class for component drivers.
```
maddenp-noaa, line 187: Whatever class names we settle on, probably this class and its parent should have different docstrings to make it clearer why they are separate classes.

src/uwtools/drivers/driver.py
```diff
@@ -154,6 +123,133 @@ def _driver_name(self) -> str:
         Returns the name of this driver.
         """
 
+    @property
+    def _resources(self) -> Dict[str, Any]:
+        """
+        Returns configuration data for the runscript.
```
maddenp-noaa, line 129: Should we have anything related to a runscript in the base class? It seems to me that that would belong in the subclass that deals with standalone execution. Would platform information be a concern for an assets-only driver? Maybe you're aware of good reasons I'm not thinking of...

I do get that `run_dir()` needs to be in the base class, so that we have a directory to place assets in.

src/uwtools/tests/drivers/test_standalonedriver.py
```diff
@@ -0,0 +1,246 @@
+# pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
+"""
+Tests for uwtools.drivers.standalonedriver module.
```
maddenp-noaa, line 3:
```suggestion
Tests for uwtools.drivers.driver module.
```

maddenp-noaa: If both base and sub driver classes continue to live in the same `uwtools.drivers.driver` module, the test modules should probably be merged together, too, to try to keep the one-to-one mapping we (mostly) have between test modules and modules-under-test. That might call for test classes like `ConcreteDriver` and `ConcreteAssets`, if those are the names we use, so that they can co-exist in the test module.

src/uwtools/drivers/driver.py
WeirAE: Hm, maybe this is actually a bad doc string? The resources function just keeps the account and rundir info from the platform block, I think this is probably ok for the base if I fix the name

src/uwtools/drivers/driver.py
WeirAE: True! This may have gotten lost in revisions.
Also, offline discussion prefers `Driver(Assets)` and `ComponentDriver(ComponentAssets)` for naming so I'll be implementing it

docs/shared/ww3.yaml
WeirAE: Hmm, thats fair - I liked the distinct installation path but it is unnecessary here

docs/sections/user_guide/yaml/components/ww3.rst
WeirAE: Ah, I didn't change this after implementing it to look more like the `fv3` driver on suggestion, which does currently render the templates rather than linking

src/uwtools/tests/drivers/test_standalonedriver.py
WeirAE: Also fair - I had observed that but stubbornly kept them separated. They should have their own test classes and go together.

src/uwtools/drivers/driver.py
WeirAE: I've updated the text, but I can shift it still if that's whats wanted

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,82 @@
...
+        yield self._taskname(fn)
+        path = self._rundir / fn
+        yield asset(path, path.is_file)
+        yield None
+        self._create_user_updated_config(
+            config_class=NMLConfig,
+            config_values=self._driver_config["namelist"],
+            path=path,
+            schema=self._namelist_schema(),
+        )
```
maddenp-noaa, line 54: In line with other comments, and I think you're still working on it, but: This should probably be rendering a template instead of realizing a (namelist) config.

src/uwtools/drivers/driver.py
```diff
@@ -24,9 +24,9 @@
 from uwtools.utils.processing import execute
 
 
-class Driver(ABC):
+class Assets(ABC):
     """
-    An abstract class for component drivers.
+    An abstract class containing assets for all component drivers.
```
maddenp-noaa, line 29:
```suggestion
    An abstract class to provision assets for component drivers.
```

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,44 @@
...
+              "required": [
+                "base_file"
+              ]
+            },
+            {
+              "required": [
+                "update_values"
+              ]
+            }
+          ],
```
maddenp-noaa, line 19: In re: template render vs config realize, probably these are both required: We need the `base_file` that contains the template, and we need the `update_values` to render the template. I do like keeping these names, though, so that we won't have to change terminology later if we implement a real namelist config realize. So, I think this `anyOf` can be replaced with a
```
"required": ["base_file", "update_values"],
```

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,44 @@
...
+          "properties": {
+            "base_file": {
+              "type": "string"
+            },
+            "update_values": {
+              "$ref": "urn:uwtools:namelist"
+            },
+            "validate": {
+              "type": "boolean"
+            }
```
maddenp-noaa, line 29: We won't need `validate` until we switch to namelist realize instead of template render. It's not that we _couldn't_ validate a namelist file obtained by rendering a template, but there's no code support for that at the moment: The final-validation code path is only via `_create_user_udpated_config()`, which relies on `Config` subclasses that won't be in play here.

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Here is a prototype UW YAML ``ww3:`` block, explained in detail below:
+
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/ww3.yaml
+
+UW YAML for the ``ww3:`` Block
+------------------------------
+
+namelists:
```
NaureenBharwaniNOAA, line 16:
```suggestion
namelist:
```

In the shared ww3.yaml the block is called `namelist:`

src/uwtools/drivers/driver.py
```diff
@@ -196,6 +165,133 @@ def _resources(self) -> Dict[str, Any]:
...
+    """
+    An abstract class for standalone component drivers.
+    """
+
+    def __init__(
+        self,
+        config: Optional[Union[dict, Path]],
+        dry_run: bool = False,
+        batch: bool = False,
+        cycle: Optional[datetime] = None,
```
NaureenBharwaniNOAA, line 232: I was wondering if changes from @maddenp-noaa #499 were added into the PR yet?

I noticed the ordering of the input parameters here and how he updated them in his pervious PR to have `cycle` and `leadtime` next to each other. Or possibly this just should be updated with the new `Driver` class. If so, I'll keep my eye out for where it appears in other places.
```
        cycle: Optional[datetime] = None,
        key_path: Optional[List[str]] = None,
        leadtime: Optional[timedelta] = None,
```

NaureenBharwaniNOAA (commented):
Looking great so far!

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: I think it still bothers me requiring an `update_values` where the expected behavior is to provide a single base file and not change it. I'm not sure theres a better way at this stage though

docs/sections/user_guide/yaml/components/ww3.rst
WeirAE: I knew I was gonna miss one when I changed that, thanks!

src/uwtools/drivers/driver.py
WeirAE: This would be an error from resolving conflicts, thanks for keeping an eye out for this!

src/uwtools/resources/jsonschema/ww3.jsonschema
maddenp-noaa: I thought we were going to support Jinja2-templated files and render those. @christinaholtNOAA what do you recall about this?

docs/shared/ww3.yaml
```diff
@@ -0,0 +1,11 @@
+ww3:
+  namelist:
+    base_file: /path/to/ww3/ww3_shel.inp.IN
+    update_values:
+      input_nml:
+        input_forcing_winds: "C"
+    validate: true
```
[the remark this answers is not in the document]
WeirAE: I think I really just wanted to validate the existing files, we'll get that in later with the other features!

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: Discussed offline - we can probably make `update_values` optional with the expectation of a user providing a complete base file that can be overridden

src/uwtools/drivers/ww3.py
WeirAE: Discussed offline - this should use `render` much like `mpas_base` does for `stream` files. Still some issues in tests, though, working on it

src/uwtools/drivers/driver.py
maddenp-noaa: Am I right in thinking that we still need to deal with `platform` here because it's always checked by the `_validate()` method when a driver object is instantiated? I'm inclined to leave things as they are now, but feel a little bit uneasy since I haven't convinced myself that an assets-only driver needs to know platform details. If I could think of an example, I'd be satisfied that it should remain here. I think an alternative might be to remove `platform` from this `Assets` method (it's already included in `Driver`'s `_resources()` method), remove it from the `Assets` `_validate()` method, and have `Driver` implement its own `_validate()`, which would be identical to the one currently in `Assets` -- i.e. it would validate both the driver block and the `platform` block. If anyone can think of a single example where an `Assets` driver needs platform information, I'd drop it; but as thing stand, it feels weird to me to see platform information here. Anyone else have thoughts on this?

WeirAE: Mm, that isn't unfair - thinking of this more as one block of a series of some future `experiment` key, the experiment is probably where this should reside

src/uwtools/drivers/driver.py
```diff
@@ -196,6 +165,133 @@ def _resources(self) -> Dict[str, Any]:
...
+        :param leadtime: The leadtime.
+        """
+        super().__init__(
+            config=config,
+            dry_run=dry_run,
+            batch=batch,
+            cycle=cycle,
+            key_path=key_path,
+            leadtime=leadtime,
+        )
```
maddenp-noaa, line 253: If I'm not mistaken, this method is identical to the one in the superclass except for the docstring, which IMO is not useful enough to justify the duplication. Can it be removed and just inherit the identical (I think) `Assets` constructor?

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,82 @@
...
+        super().__init__(config=config, dry_run=dry_run, batch=batch, key_path=key_path)
+
+    # Workflow tasks
+
+    @task
+    def namelist_file(self):
+        """
+        The namelist file.
+        """
+        fn = "ww3_shel.nml"
```
maddenp-noaa, line 45: I know that this is the name of the namelist file that we decided to initially support, but is this the filename that WW3 expects to actually see in the run directory? Is there not maybe a canonical filename that the application will read, with the user expected to copy their file (perhaps starting from one of the example files provided with the application) into the run directory with that canonical name?

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,82 @@
+"""
+A library driver for ww3.
```
maddenp-noaa, line 2:
```suggestion
An assets driver for ww3.
```
To agree with our chosen nomenclature...

maddenp-noaa:
This is looking good, and close. I think it would be helpful, for consistency with other drivers, to add some basic coverage for `ww3.jsonschema` to `test_schemas.py`.

src/uwtools/drivers/ww3.py
WeirAE: Strictly, its expecting `ww3_shel.nml` or `ww3_shel.inp` but that's what WW3 is expecting to see

src/uwtools/drivers/driver.py
maddenp-noaa: Exactly. If this assets driver is called upon by a coupled-component driver, I think that latter driver (which would subclass `Driver`) would be where platform-specific considerations would arise.

Unless our currently one-and-only assets driver (WW3) has some use for platform info -- which would be the proof of existence I'm looking for -- I'd err on the side of making `Asset` ignorant of `platform`. We can reintroduce it later if we need to.

src/uwtools/drivers/ww3.py
maddenp-noaa: All good, then.

src/uwtools/drivers/driver.py
WeirAE: Huh, turns out it can! I can remove that from my habits now!

maddenp-noaa (approved):

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,32 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "namelist": {
+          "additionalProperties": false,
+          "properties": {
+            "base_file": {
+              "type": "string"
+            },
+            "update_values": {
+              "$ref": "urn:uwtools:namelist"
```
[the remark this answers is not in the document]
WeirAE: It was my intentional choice, although I think we could probably include the ones in [ww3_shel](https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/parm/ww3_shel.inp.IN) here. There are far too many parameters across the other namelists but they're not our problem right now if we're restricting ourselves to this one.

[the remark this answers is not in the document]
maddenp-noaa: Maybe we should leave it vague until later, if/when we're pretty sure requirements have solidified.

src/uwtools/drivers/ww3.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    A driver for ww3.
+    """
+
+    def __init__(
+        self,
+        config: Optional[Path] = None,
+        dry_run: bool = False,
+        batch: bool = False,
+        key_path: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
christinaholtNOAA: But what about the case where date/time information needs to be updated in the configuration files for this component? Or if input data is needed that has date time information in the drivers?

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,32 @@
...
+            }
+          },
+          "required": [
+            "base_file"
```
christinaholtNOAA, line 17: For other similar components, we've had the requirement that either `base_file` or `update_values` is required. We should probably stick with that decision here.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -94,40 +129,114 @@ def driverobj(config):
...
-# Tests
+# Asset Tests
 
 
-def test_Driver(driverobj):
-    assert Path(driverobj._driver_config["base_file"]).name == "base.yaml"
-    assert driverobj._batch is True
+def test_Assets(assetobj):
+    assert Path(assetobj._driver_config["base_file"]).name == "base.yaml"
+    assert assetobj._batch is True
```
christinaholtNOAA, line 137: Do we not want to continue checking that the Driver class does all these things (here and in the next couple of tests)?

src/uwtools/tests/drivers/test_ww3.py
```diff
@@ -0,0 +1,85 @@
...
+
+# Fixtures
+
+
+@fixture
+def config(tmp_path):
+    return {
+        "ww3": {
+            "namelist": {
+                "base_file": str(tmp_path / "ww3_shel.inp.IN"),
```
christinaholtNOAA, line 21: This is not a namelist file and should not be used in conjunction with the `update_values` block. In so much as the unit tests serve as examples, we probably do not want to mix the approach here.

src/uwtools/tests/test_schemas.py
```diff
@@ -1282,3 +1287,50 @@ def test_schema_upp_run_dir(upp_prop):
...
     assert "88 is not of type 'string'" in errors(88)
+
+
+# ungrib
+
+
+def test_schema_ww3():
+    config = {
+        "namelist": {
+            "base_file": "/tmp/ww3_shel.inp.IN",
```
christinaholtNOAA, line 1298: Same is true here.

christinaholtNOAA (changes requested):
There are still a few things here that I'd like to see corrected.

src/uwtools/drivers/ww3.py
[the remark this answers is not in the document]
WeirAE: Is there a time when that would happen for WW3? I don't think it would hurt though, I'll get that in

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: Thats fair, and that was @maddenp-noaa 's original thought too

src/uwtools/tests/drivers/test_ww3.py
WeirAE: Fair, though its also a valid input - should I put some logic that checks for this condition and bypasses the namelist validation, or at least accepts another input? Or perhaps this should be hard-coded and error on other inputs?

src/uwtools/tests/drivers/test_driver.py
WeirAE: The Driver version should be down at line 235-ish, though it is a little harder to follow with the two parts combined now. I am concernes that I may have missed or lost some parts in the transfer, though. Thanks, I'll go through all of it again to fix anything missed

src/uwtools/tests/drivers/test_ww3.py
maddenp-noaa: @christinaholtNOAA I'm not sure I understand your concern on this one. What is the "mix[ed] approach"? The output _is_ a namelist file, so I would think that `namelist:`, at least, is appropriate. Is it `base_file` and `update_values` that you're concerned about? My thinking in suggesting this to @WeirAE was that:

1. If we are later able to update our Fortran namelist handling to deal with the array and user-defined-type syntax used in the WW3 namelists, the change could be _internal only_ and not affect the UW YAML language used here. That is, users would not have to change their configs -- they probably wouldn't even need to change their input namelist (template) files, as `f90nml` could just overwrite the Jinja2 expressions with real namelist values. If we use some other terminology here, then the UW YAML would have to change, too.
2. We also use `base_file` and `update_values` for INI, SH, and YAML configs -- they're not specific to NML configs.
3. And, anyway, aren't we updating a base file with some values? It seems like appropriate -- not to mention consistent -- terminology.

But I might have totally misunderstood your intent here. Sorry if so.

src/uwtools/resources/jsonschema/ww3.jsonschema
maddenp-noaa: OK, at least here are the files that I was looking at: https://github.com/NOAA-EMC/WW3/tree/develop/model/nml. Is `w3_shel.nml` the thing we are provisioning?

src/uwtools/drivers/ww3.py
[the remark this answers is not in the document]
maddenp-noaa: I didn't think about it until just now, but e.g. https://github.com/NOAA-EMC/WW3/blob/develop/model/nml/ww3_shel.nml contains what looks like cycle information, so users would probably benefit to having `cycle` available when provisioning this namelist -- if, indeed, this namelist is what we are provisioning (see my confused comments elsewhere in this PR).

src/uwtools/tests/drivers/test_ww3.py
WeirAE: Aha, there is a way to link comments! To step in and sort a little, this was @christinaholtNOAA 's clarification on file limits [here](https://github.com/ufs-community/uwtools/pull/496#discussion_r1617908543) where the [input script](https://github.com/NOAA-EMC/WW3/blob/d9b3172f4197c65d471662c6952a668152d71230/model/src/wav_shel_inp.F90) used by UFS checks for either [ww3_shel.inp](https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/parm/ww3_shel.inp.IN) or [ww3_shel.nml](https://github.com/ufs-community/ufs-weather-model/blob/develop/tests/parm/ww3_shel.nml.IN). The .inp file is parsed by the input script, which we're not overriding, so I think it might be ok to accept but bypass it, and change to just the nml for the tests? It looks like atparse to me so we _could_ even handle it properly

WeirAE:
Unresolved a few prior threads for ease of further discussion

src/uwtools/resources/jsonschema/ww3.jsonschema
christinaholtNOAA: I'm sorry I got confused by this. I'm curious if we want to do something different with this section to indicate it's not your standard treatment of a namelist? Perhaps that's what got me off-course?

src/uwtools/tests/drivers/test_driver.py
christinaholtNOAA: I don't see evidence of `test_Driver_cycle_leadtime_error` or `test_Driver_dry_run `, and below that, there are several others that change from the original `Driver` to `Asset`. Would we not want to test all that behavior on Asset objects and on driver objects?

An answer of "no" is fine, I would just like to feel more comfortable about this not being an oversight.

src/uwtools/tests/drivers/test_ww3.py
christinaholtNOAA: I think Brian is correct. WW3 accepts either an ascii `ww3_shel.inp` file as its source of configuration information **or** it accepts a legit fortran namelist named `ww3_shel.nml`. We should not be suggesting that users can put a `ww3_shel.inp.IN` template input file in here since that most likely corresponds to the ascii version of the file. Even though the driver will fill it in as a template just fine, when it's placed in the run directory as `ww3_shel.nml` (hardcoded in the driver), it won't be formatted as expected and the program will (hopefully) die.

src/uwtools/tests/drivers/test_driver.py
maddenp-noaa: I don't think we need to test, in a subclass, behavior that is inherited from and that is tested by a base class. That would duplicate the test logic for, I think, no additional coverage benefit. I'll check out this branch and see if I agree that that's actually what's happening here -- the refactor isn't as clear to me in these diffs as it'll probably be looking at the code directly.

src/uwtools/tests/drivers/test_ww3.py
maddenp-noaa: Yes! I understand now. It was just a concern about the example filename here, `ww3_shel.inp.IN`, being liable to mislead users because _that_ file is not a namelist file. I thought it was a deeper concern about how we're provisioning the actual namelist file. Please let me know if I'm still off the mark.

src/uwtools/resources/jsonschema/ww3.jsonschema
maddenp-noaa: Any thought about what that something different might be? I think part of my confused comment in another thread actually applies here: Assuming that we're eventually able to support actual programmatic edits of namelist files with array and user-defined-type syntax, we'd want to replace template rendering with a call to `_create_user_updated_config()` and an `NMLConfig` object, in which case we'll want `base_file:` and `update_values:` here. If we stick with those keys now, then a future update will be an implementation detail that will hopefully be invisible to users. But if we change terminology here, users will have to update their configs. If that's not compelling, then we could use e.g. `template_file:` and `template_values:` here. Will users benefit from knowing "it's not your standard treatment of a namelist"? I'm thinking that they're going to know this, because they will not know how to represent e.g.
```
TYPE%FIELD%LIST     = 'HS DIR SPR WND ICE CUR LEV'
```
or
```
HOMOG_INPUT(3)%VALUE1     = 5.
```
in UW YAML and will end up in the docs, where (presumably) we explain the fact that this is done via template rendering. Plus, I suppose their app will be providing the Jinja2-templated base files and almost-correct UW YAML prototypes, so it should be pretty clear.

BTW, `f90nml` has guidance on [representing user-defined types](https://f90nml.readthedocs.io/en/latest/notes.html#derived-types) that looks workable, except that we'll run afoul of our own depth checks, which may require a little thought. But it works transparently:
```
>>> nml
{'dtype_nml': {'a': {'b': {'c': 1}}}}
>>> print(f90nml.Namelist(nml))
&dtype_nml
    a%b%c = 1
/
```
Clearly that's easily representable in YAML. I'm not sure yet about the array indexing...

src/uwtools/tests/drivers/test_ww3.py
christinaholtNOAA: Nope. I think we're on the same page now.

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,28 @@
...
+UW YAML for the ``ww3:`` Block
+------------------------------
+
+namelist:
+^^^^^^^^^
+
+The WaveWatchIII namelist file can be provisioned either by:
+1. Providing a path to a complete, ready-to-use namelist file as the value of the `base_file:` key and omitting the `update_values:` key, or
+2. Providing a path to a namelist file containing Jinja2 expressions as the value of the `base_file:` key and a mapping from variable names to values used to render those Jinja2 expressions as the value of the `update_values:` key.
+
```
christinaholtNOAA, line 22: Given the ongoing discussion below about the names of the subkeys in this section, I think it might be helpful here to provide a warning note about the behavior in this section not conforming to the same behavior as the other namelist sections.

src/uwtools/tests/drivers/test_ww3.py
WeirAE: Offline discussion: a clear breaking change is preferable, so we should define it as 'template_file' and 'template_values'

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,32 @@
...
+------------------------------
+
+namelist:
+^^^^^^^^^
+
+.. important:: Fortran namelists for WaveWatchIII are used here as templates to be realized. Input namelists will not have their configurations validated, only the associated YAML. Other files are not currently supported.:
+
+  **template_file:**
+
+  The path to the base ``ww3_shel.nml`` file that comes from the WW3 build.
```
maddenp-noaa, line 23:
```suggestion
  The path to the input template file containing Jinja2 expressions, based on the ``ww3_shel.nml`` file from the WaveWatchIII build.
```
True?

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,32 @@
...
+
+  **template_values:**
+
+  The set of key-value pairs that will render the appropriate XML entries in the template input file.
```
maddenp-noaa, line 27:
```suggestion
  Key-value pairs necessary to render all Jinja2 expressions in the input template file named by ``template_file:``.
```
Is there XML in play here? I don't recall seeing any.

docs/shared/ww3.yaml
```diff
@@ -0,0 +1,10 @@
+ww3:
+  namelist:
+    template_file: /path/to/ww3/ww3_shel.nml.IN
+    template_values:
+      input_nml:
+        input_forcing_winds: "C"
```
maddenp-noaa, line 6:
```suggestion
    template_values:
      input_forcing_winds: "C"
```
From the point of view of `uwtools`, the input template file is flat text, just containing Jinja2 expressions that need to be rendered. So there's no use for structure under `template_values:` except for a key-value mapping.

Given how the `namelist_file()` task method in the `WaveWatchIII` driver is written, I assume that a config structured like this would fail if given as input in a CLI command like `uw ww3 namelist_file --config-file config.yaml` where this is `config.yaml`. It's probably worth creating a simple `ww3_shel.nml.IN` file and doing that test before merging this.

src/uwtools/tests/drivers/test_ww3.py
```diff
@@ -0,0 +1,91 @@
+# pylint: disable=missing-function-docstring,protected-access,redefined-outer-name
+"""
+WW3 driver tests.
```
maddenp-noaa, line 3:
```suggestion
WaveWatchIII driver tests.
```
or `ww3 driver tests` -- we're using `ww3` as an in-code identifier for this component, which makes sense because `wwiii` is kind of awful (is that a Nintendo product?), but we could be consistent in our human-readable name.

src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -0,0 +1,32 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "namelist": {
+          "additionalProperties": false,
+          "properties": {
+            "template_file": {
+              "type": "string"
+            },
+            "template_values": {
+              "$ref": "urn:uwtools:namelist"
```
maddenp-noaa, line 13:
```suggestion
              "additionalProperties": {
                "type": "string"
              },
              "type": "object"
```
I think... We cannot reference the `namelist.jsonschema` here, because it describes a 2-level data structure (namelists -> variables -> values), and for template rendering we want a 1-level structure (keys -> values).

maddenp-noaa (changes requested):

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,32 @@
...
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/ww3.yaml
+
+UW YAML for the ``ww3:`` Block
+------------------------------
+
+namelist:
+^^^^^^^^^
+
+.. important:: Fortran namelists for WaveWatchIII are used here as templates to be realized. Input namelists will not have their configurations validated, only the associated YAML. Other files are not currently supported.:
```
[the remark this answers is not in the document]
WeirAE: I wanted to leave something to note that we are specifically not supporting the inp.in file, even though it is accepted by WW3, but maybe its sufficiently implied here?

src/uwtools/resources/jsonschema/ww3.jsonschema
WeirAE: Mm, and even then we probably want a clean break at all points, now that I think about it

docs/shared/ww3.yaml
WeirAE: Hmmm - so this is interesting because the entry being edited here in `ww3_shel.nml` is 
```
&INPUT_NML
  INPUT%FORCING%WINDS         = 'H'
```

docs/sections/user_guide/yaml/components/ww3.rst
WeirAE: Currently this is pointing directly to the namelist `ww3_shel.nml`.

docs/sections/user_guide/yaml/components/ww3.rst
WeirAE: cut-and-paste-itis strikes again

docs/sections/user_guide/yaml/components/ww3.rst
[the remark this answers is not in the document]
maddenp-noaa: That could be good advice, if made explicit. In the `template_file` section below, we specifically mention the `ww3_shel.nml` file, so maybe that would be a good place for an additional sentence like
```
Note that the non-namelist ``ww3_shel.inp`` file may not be used as the basis for the input template file.
```
?

docs/shared/ww3.yaml
maddenp-noaa: What we'd expect to see in the input template file would be something like
```
&INPUT_NML
  INPUT%FORCING%WINDS = {{ input_forcing_winds }}
```
right?

docs/sections/user_guide/yaml/components/ww3.rst
maddenp-noaa: How is it currently pointing to anything? The user has to set `template_file:` in their UW YAML to specify the input template file they want us to use, right? We'd just be saying here that they should specify a file that is _based on_ `ww3_shel.nml` from the WaveWatch III build, but that they have (or their app maintainer has) updated to contain Jinja2 expressions.

docs/sections/user_guide/yaml/components/ww3.rst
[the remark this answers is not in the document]
WeirAE: OK, yeah - no need to be coy about it, just be decisive

docs/sections/user_guide/yaml/components/ww3.rst
WeirAE: It is currently hard-coded within `ww3.py` as such, but perhaps we should explicitly call it `ww3_shel.nml.IN` to clearly differentiate it?

docs/shared/ww3.yaml
WeirAE: Ha, yes - there are many places in this driver where there are similar topics with different issues, I think this is what got me here

docs/sections/user_guide/yaml/components/ww3.rst
maddenp-noaa: Oh, right -- the _output_ filename is hardcoded, but the input filename is arbitrary. Given that we use the filename `ww3_shel.nml.IN` in our `shared/ww3.yaml`, how about e.g.
```
The path to the input template file containing Jinja2 expressions (perhaps named ``ww3_shel.nml.IN``), based on the ``ww3_shel.nml`` file from the WaveWatchIII build.
```
? It's not IMO really any of our business what they name their input template file, but if this makes the meaning clearer, I'm all for it.

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,32 @@
...
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/ww3.yaml
+
+UW YAML for the ``ww3:`` Block
+------------------------------
+
+namelist:
+^^^^^^^^^
+
+.. important:: The WaveWatchIII namelist file is provisioned by rendering an input template file containing Jinja2 expressions. Unlike namelist files provisioned by ``uwtools`` for other components, the WaveWatchIII namelist file will not be validated.Note that the non-namelist ``ww3_shel.inp`` file may not be used as the basis for the input template file.:
```
maddenp-noaa, line 19:
```suggestion
.. important:: The WaveWatchIII namelist file is provisioned by rendering an input template file containing Jinja2 expressions. Unlike namelist files provisioned by ``uwtools`` for other components, the WaveWatchIII namelist file will not be validated.
```
I think the note about the bad `ww3_shel.inp` file makes better sense adjacent to the info about the good `ww3_shel.nml`, below.

docs/sections/user_guide/yaml/components/ww3.rst
```diff
@@ -0,0 +1,32 @@
...
+
+  **template_file:**
+
+  The path to the input template file containing Jinja2 expressions (perhaps named ``ww3_shel.nml.IN``), based on the ``ww3_shel.nml`` file from the WaveWatchIII build.
```
maddenp-noaa, line 23:
```suggestion
  The path to the input template file containing Jinja2 expressions (perhaps named ``ww3_shel.nml.IN``), based on the ``ww3_shel.nml`` file from the WaveWatchIII build. Note that the non-namelist ``ww3_shel.inp`` file may not be used as the basis for the input template file.
```

src/uwtools/tests/drivers/test_ww3.py
```diff
@@ -0,0 +1,89 @@
...
+    assert isinstance(driverobj, ww3.WaveWatchIII)
+
+
+def test_WaveWatchIII_namelist_file(driverobj):
+    src = driverobj._driver_config["namelist"]["template_file"]
+    with open(src, "w", encoding="utf-8") as f:
+        yaml.dump({}, f)
+    assert not (driverobj._rundir / "ww3_shel.nml").is_file()
+    driverobj.template_file()
+    assert (driverobj._rundir / "ww3_shel.nml").is_file()
```
maddenp-noaa, line 63:
```suggestion
    dst = driverobj._rundir / "ww3_shel.nml"
    assert not dst.is_file()
    driverobj.template_file()
    assert dst.is_file()
```

maddenp-noaa (approved):

src/uwtools/drivers/ww3.py
```diff
@@ -28,29 +30,33 @@ def __init__(
...
+        super().__init__(
+            config=config, dry_run=dry_run, batch=batch, cycle=cycle, key_path=key_path
+        )
+        self._cycle = cycle
 
     # Workflow tasks
 
     @task
-    def namelist_file(self):
+    def template_file(self):
```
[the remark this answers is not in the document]
WeirAE: Thats true - I was trying to differentiate it but it is better to be clear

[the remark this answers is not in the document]
WeirAE: Reverted and clarified further in docstring

christinaholtNOAA (approved):
LGTM.

PR #497, opened 2024-05-28
omnibus-2024-05-23

maddenp-noaa:
**Synopsis**

Two small updates:

- Replace `--output-block` in `uw config realize` mode with `--key-path` to agree with similar usage in the drivers, but keeping mode/action-specific help messages for each use case.
- Eliminate "empty values" section in values-needed reporting, as there's really no such thing as an empty value in YAML (an unspecified value is synonymous with `null`).

**Type**

- [x] Documentation

**Impact**

**Checklist**


elcarpenterNOAA (approved):
Looks good to me.

src/uwtools/cli.py
```diff
@@ -149,7 +149,7 @@ def _add_subparser_config_realize(subparsers: Subparsers) -> ActionChecks:
     _add_arg_update_format(optional, choices=FORMATS)
     _add_arg_output_file(optional)
     _add_arg_output_format(optional, choices=FORMATS)
-    _add_arg_output_block(optional)
+    _add_arg_key_path(optional, helpmsg="Dot-separated path of keys to the block to be output")
```
christinaholtNOAA, line 152: I like keeping this help message with the new consistent name for the key.

christinaholtNOAA (approved):
LGTM!

PR #498, opened 2024-05-29
UW-590 Add --cycle and --leadtime to file tool

maddenp-noaa:
**Synopsis**

Add `--cycle` and `--leadtime` CLI options (and equivalent API arguments) to `uw file copy` and `uw file link`. The user-provided values are made available to Jinja2 when dereferencing the input config.

**Type**

- [x] Documentation

**Impact**

**Checklist**


NaureenBharwaniNOAA (approved):
LGTM!

docs/sections/user_guide/cli/tools/file/src/20240529/12/006/baz
```diff
@@ -0,0 +1 @@
+baz
```
WeirAE, line 1: Was this generated as part of making the .out files?

maddenp-noaa: No, I intentionally added it as content to avoid having a zero-length file, in case one wants to do a manual inspection of the destination directory created by a file-copy operation: The existing `foo` and `bar` files also have their own names as content, making it possible to verify that each file really was copied source -> destination, which wouldn't be possible if the files were empty.

WeirAE: Ah, I think the date directory was what flagged for me - could/should we store them together without the layers? I think I'm just wary of extra details within docs

WeirAE (approved):
Looks clean and solid to me!

christinaholtNOAA (approved):
Thanks for this feature. :)

PR #501, opened 2024-06-01
UW-556 XML tooling for MPAS streams files

maddenp-noaa:
**Synopsis**

Replace Jinja2-rendering of MPAS streams files with native XML generation and configuration support via UW YAML configuration language. For example, given the UW YAML
```
mpas_init:
  ...
  streams:
    input:
      filename_template: conus.static.nc
      input_interval: initial_only
      mutable: false
      type: input
    output:
      filename_template: conus.init.nc
      files:
        - foo
      io_type: pnetcdf
      mutable: false
      output_interval: initial_only
      type: output
```

in `config.yaml`, `uw mpas_init streams_file --cycle 2024-05-31T12 --config config.yaml` generates `streams.init_atmosphere`
```
<streams>
  <immutable_stream name="input" type="input" filename_template="conus.static.nc" input_interval="initial_only"/>
  <immutable_stream name="output" type="output" filename_template="conus.init.nc" io_type="pnetcdf" output_interval="initial_only">
    <file name="foo"/>
  </immutable_stream>
</streams>
```

The full set of optional attributes and child elements described in the [MPAS docs](https://www2.mmm.ucar.edu/projects/mpas/mpas_atmosphere_users_guide_8.1.0.pdf) are supported, though only the `io_type` attribute and the `file` child element are shown above.

Most of the changed files are documentation, and most of those are a simple change from `../../../../shared/*`-style relative references to simpler `/shared/*` references, taking advantage of the fact that Sphinx views such paths as relative to the documentation root directory.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/contributor_guide/developer_setup.rst
```diff
@@ -6,7 +6,7 @@ Creating a ``bash`` Development Shell
 
 If an existing conda (:miniforge:`Miniforge<>`, :miniconda:`Miniconda<>`, :anaconda:`Anaconda<>`, etc.) installation is available and writable, step 1 may be skipped.
 
-.. include:: ../../shared/miniforge_instructions.rst
+.. include:: /shared/miniforge_instructions.rst
```
maddenp-noaa, line 9: Many of the doc changes below follow this pattern: Use the relative-to-doc-root absolute reference to shared files, which are simpler and easier to maintain.

docs/sections/user_guide/cli/tools/config/realize-help.out
```diff
@@ -29,7 +29,7 @@ Optional arguments:
   --key-path KEY[.KEY...]
       Dot-separated path of keys to the block to be output
   --values-needed
-      Print report of values needed to render template
+      Print report of values needed to realize config
```
maddenp-noaa, line 32: I updated `uwtools.cli` so that `config realize` and `template render` pass accurate help messages to `_add_arg_values_needed()` so that the help output fits the mode/action. This output was re-rendered by `make docs` based on that change.

docs/sections/user_guide/yaml/components/mpas.rst
```diff
@@ -75,20 +75,11 @@ namelist:
...
-^^^^^^^^
-
-  **path:**
-
-  The path to the base ``streams.atmosphere`` file that comes from the MPAS build.
-
-  **values:**
-
-  The set of key-value pairs that will render the appropriate XML entries in the streams input file.
+.. include:: /shared/mpas_streams.rst
```
maddenp-noaa, line 85: The description of the `streams:` block is identical for MPAS Atm and MPAS Init, so has been factored out to a shared RST file.

docs/shared/mpas.yaml
```diff
@@ -44,11 +44,18 @@ mpas:
...
+      filename_template: conus.init.nc
+      input_interval: initial_only
+      mutable: false
+    - name: output
+      type: output
+      filename_template: history.$Y-$M-$D_$h.$m.$s.nc
+      output_interval: "6:00:00"
+      mutable: true
+      files:
+        - stream_list.atmosphere.output
```
maddenp-noaa, line 58: See the update UW YAML docs for this block and/or the new `mpas_streams.jsonschema` for background on this.

docs/shared/mpas_streams.rst
```diff
@@ -0,0 +1,44 @@
...
+
+  ``name:``
+
+  A unique name used to refer to the stream. (type: string)
+
+  ``type:``
+
+  One of ``input``, ``output``, ``input;output``, or ``none``. (type: string)
+
+**Optional attributes:**
```
maddenp-noaa, line 24: The MPAS docs describe more optional attributes, similar to `packages` in
```
<stream name="foo", packages="bar;baz" />
```
(supported in this PR) and optional elements, similar to `file` in
```
<stream name="foo">
    <file name="bar"/>
</stream>
```
(supported in this PR via the `files:` item).

Since UW-556 calls for only implementing what's currently required, I implemented in this PR only what I saw used in the MPAS App examples I looked at. Please let me know if I missed something, or if you think we should implement more of the optional attributes/elements now rather than in follow-on work.

src/uwtools/drivers/mpas_base.py
```diff
@@ -114,12 +115,22 @@ def streams_file(self):
...
-        render(
-            input_file=Path(self._driver_config["streams"]["path"]),
-            output_file=path,
-            values_src=self._driver_config["streams"]["values"],
-        )
+        yield None
+        streams = Element("streams")
+        for item in self._driver_config["streams"]:
+            stream = SubElement(streams, "stream" if item["mutable"] else "immutable_stream")
+            for attr in ["name", "type", "filename_template"]:
```
maddenp-noaa, line 122: Although attribute order is not semantically meaningful in XML, `lxml` appears to output attributes in the order they are defined, so I've tried to order the generation of attributes in the `for` loops here to match the MPAS docs and examples.

src/uwtools/resources/jsonschema/mpas-init.jsonschema
```diff
@@ -70,19 +70,7 @@
...
-            "values": {
-              "type": "object"
-            }
-          },
-          "required": [
-            "path",
-            "values"
-          ],
-          "type": "object"
+          "$ref": "urn:uwtools:mpas-streams"
```
maddenp-noaa, line 73: `mpas-init.jsonschema` and `mpas.jsonschema` now refer to a common sub-schema for the `streams:` block.

elcarpenterNOAA (approved):
Looks good to me! Thank you for making this right.

docs/shared/mpas_streams.rst
```diff
@@ -0,0 +1,44 @@
...
+
+  The interval at which the stream will be read. (type: string)
+
+  ``files:``
+
+  Names of files, each of which lists variables, one per line, to associate with the stream. (type: sequence of strings)
+
+  ``input_interval:``
+
+  The interval at which the stream will be read. Required if ``type:`` includes ``input``.  (type: string)
```
christinaholtNOAA, line 36: I don't see the example use case I saw above with `io_type`...now I'm concerned about versioning....

christinaholtNOAA (changes requested):
I'm slightly concerned about all the versions we have floating around with MPAS right now, and how much we might want to tighten down requirements on the contents of the stream files.

I also think it might be important to be able to update the values in base streams files without a user having to pass an entire stream file contents in the config. In many cases, we're using what comes from the MPAS build system and applying changes to achieve the necessary output. Those changes are often limited to the names of files and frequency of I/O. I'm afraid we lose some of that with this approach as-is.

docs/shared/mpas.yaml
maddenp-noaa: > losing our ability to update the values contained in the lists with a call to `update_values()`

There is no `update_values` in the current implementation, so there's nothing to lose: Streams are defined in UW YAML (like Rocoto workflows are defined in UW YAML) and serialized as XML, so there's nothing to update.

That said, I see from another comment that you might be asking for the ability to read and modify an existing streams file, which would make the concern pertinent again. I don't think `list` or `dict` matters, because we'd have to use XPath queries (or an equivalent mechanism) to select and update elements matching appropriate criteria; but `dict` might still be more appropriate because, although XML is not a mapping-like format (it's a sequence-like format, e.g. there's no language rule against identical repeated elements like `<food><fruit name="banana"/><fruit name="banana"/></food>` as there's nothing wrong with a `list` containing `["banana", "banana"]`), MPAS does request -- and maybe require, but I haven't checked -- that the `name` attribute on a stream element be unique. And element order is not meaningful in streams files, as far as I know.

I don't see `io_type` in the streams files I got from jet but, as mentioned in another comment, we could add support for any or all of the elements and attributes that MPAS supports.

But we should probably decide first if we want to continue with this kind of update. XML has different characteristics than the other config formats we deal with and could have some unexpected behaviors. Generating it from validated config is straightforward, but updating it programmatically could yield some surprises, given that the use cases are fairly new to us. As of this writing, I think that generating it is the safer and more reliable approach, but I think both ways are possible.

NaureenBharwaniNOAA (approved):
Looks good to me!

docs/shared/mpas.yaml
maddenp-noaa: Updates have been made so that the `streams` block is now a mapping from strings (stream names) to mappings of key-value config pairs. Support for the full range of MPAS-supported stream configuration options has been added.

christinaholtNOAA (approved):
Thanks for the rework on this one! LGTM.

WeirAE (approved):
Another massive, excellent fix - thanks!

PR #503, opened 2024-06-05
[main] Factor out run script done file.

christinaholtNOAA:
**Synopsis**

The run script done file definition existed in two places and was out of sync. This fix DRYs the name of that file ensuring the task is responsible for writing the asset that it requires to be complete.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


maddenp-noaa (approved):

WeirAE (approved):
Quick fixes, thanks!

src/uwtools/resources/info.json
```diff
@@ -1,4 +1,4 @@
 {
-  "version": "2.3.0",
+  "version": "2.3.1",
```
maddenp-noaa, line 2: Certainly no problem with this, though the version number on `main` doesn't really matter until (and sometimes even when) we do a release: There are already other commits on `main` that are not on the `2.3.1` branch (or `2.3.0` for that matter), so the apparent version number on `main` should never be consider authoritative -- only the version number of release branches is really meaningful. It seems like a good idea to keep up with the highest released version number here, though.

christinaholtNOAA: I concur. This was explicit in the SoP doc, so I went with it.

PR #506, opened 2024-06-10
UW-585: Build driver for SCHISM

WeirAE:
**Synopsis**

Build driver for SCHISM. As per discussion, scope is limited to "library" version for the coupled model. 

AC:
new driver introduced in CLI and API
JSON Schema validation capability
Documentation for CLI, API and landing page addition

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/yaml/components/schism.rst
+++ b/docs/sections/user_guide/yaml/components/schism.rst
@@ -1,0 +1,32 @@
+.. _schism_yaml:
+
+schism
+======
+
+Structured YAML to configure SCHISM as part of a compiled coupled executable is validated by JSON Schema and requires the ``schism:`` block, described below.
+
+Here is a prototype UW YAML ``schism:`` block, explained in detail below:
+
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/schism.yaml
+
+UW YAML for the ``schism:`` Block
+---------------------------------
+
+namelist:
+^^^^^^^^^
+
+  .. important:: The SCHISM namelist file is provisioned by rendering an input template file containing Jinja2 expressions. Unlike namelist files provisioned by ``uwtools`` for other components, the SCHISM namelist file will not be validated.
+
+  **template_file:**
+
+  The path to the input template file containing Jinja2 expressions (perhaps named ``param.nml.IN``), based on the ``param.nml`` file from the SCHISM build.
+
+  **template_values:**
+
+  Key-value pairs necessary to render all Jinja2 expressions in the input template file named by ``template_file:``.
+
+run_dir:
+^^^^^^^^
+
+The path to the run directory.
--- a/docs/sections/user_guide/yaml/components/schism.rst
+++ b/docs/sections/user_guide/yaml/components/schism.rst
@@ -8,7 +8,7 @@ Structured YAML to configure SCHISM as part of a compiled coupled executable is 
 Here is a prototype UW YAML ``schism:`` block, explained in detail below:
 
 .. highlight:: yaml
-.. literalinclude:: ../../../../shared/schism.yaml
+.. literalinclude:: /shared/schism.yaml
 
 UW YAML for the ``schism:`` Block
 ---------------------------------
--- a/src/uwtools/drivers/schism.py
+++ b/src/uwtools/drivers/schism.py
@@ -67,7 +67,6 @@ class SCHISM(Assets):
         yield self._taskname("provisioned run directory")
         yield [self.namelist_file()]
 
-
     # Private helper methods
 
     @property
--- a/src/uwtools/drivers/schism.py
+++ b/src/uwtools/drivers/schism.py
@@ -65,7 +65,7 @@ class SCHISM(Assets):
         Run directory provisioned with all required content.
         """
         yield self._taskname("provisioned run directory")
-        yield [self.namelist_file()]
+        yield self.namelist_file()
 
     # Private helper methods
 
--- a/src/uwtools/drivers/schism.py
+++ b/src/uwtools/drivers/schism.py
@@ -16,7 +16,7 @@ from uwtools.utils.tasks import file
 
 class SCHISM(Assets):
     """
-    A library driver for SCHISM.
+    An assets driver for SCHISM.
     """
 
     def __init__(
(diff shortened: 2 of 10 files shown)


src/uwtools/resources/jsonschema/ww3.jsonschema
```diff
@@ -11,8 +11,14 @@
             },
             "template_values": {
               "additionalProperties": {
-                "type": "string"
```
WeirAE, line 14: Noticed this was hardcoded to string in prior revisions to remove the namelist subschema and fixed here

src/uwtools/tests/test_schemas.py
```diff
@@ -1399,7 +1449,7 @@ def test_schema_upp_run_dir(upp_prop):
     assert "88 is not of type 'string'" in errors(88)
 
 
-# ungrib
+# wavewatch III
```
WeirAE, line 1452: Totally missed this previously and fixed here

docs/sections/user_guide/yaml/components/schism.rst
```diff
@@ -0,0 +1,32 @@
...
+.. highlight:: yaml
+.. literalinclude:: ../../../../shared/schism.yaml
+
+UW YAML for the ``schism:`` Block
+---------------------------------
+
+namelist:
+^^^^^^^^^
+
+  .. important:: The SCHISM namelist file is provisioned by rendering an input template file containing Jinja2 expressions. Unlike namelist files provisioned by ``uwtools`` for other components, the SCHISM namelist file will not be validated.
```
maddenp-noaa, line 19: Looking at [the namelist file](https://github.com/schism-dev/pyschism/blob/main/pyschism/param/param.nml) I see why this has to be the case. It seems to me like we're going to need to deal with the array-element and user-defined-type-member syntax soon, and maybe even before our next release so that we can provide native namelist editing and schema checking for this an WW3.

src/uwtools/drivers/schism.py
```diff
@@ -0,0 +1,77 @@
...
+            overrides=self._driver_config["namelist"]["template_values"],
+        )
+
+    @tasks
+    def provisioned_run_directory(self):
+        """
+        Run directory provisioned with all required content.
+        """
+        yield self._taskname("provisioned run directory")
+        yield [self.namelist_file()]
```
maddenp-noaa, line 68:
```suggestion
        yield self.namelist_file()
```
No need for the list syntax if we're only `yield`ing one asset.

maddenp-noaa, line 68: What about the `.gr3` and `.in` files mentioned in [your doc](https://docs.google.com/document/d/1kFb-tF5GkMINZMp7kMsfXoh-l9CSDa0uJVoXPyONVi0/edit)? Are they needed in the run directory? Or would that be handled by the driver for the coupled complnent?

WeirAE: That was one of the things I was pondering - there are a variety of files that are required, but they have to be created by the user in preprocessing so historically we have not handled those and might not want to in a library driver, BUT we might want to here. So I started with a baseline to pose the question if we want to add that here

src/uwtools/drivers/schism.py
```diff
@@ -0,0 +1,77 @@
...
+        self._cycle = cycle
+
+    # Workflow tasks
+
+    @task
+    def namelist_file(self):
+        """
+        Render the namelist from the template file.
+        """
+        fn = "param.nml"
```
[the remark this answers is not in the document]
WeirAE: resolving here to discuss further below

docs/sections/user_guide/yaml/components/schism.rst
WeirAE: Yeah, I'm seeing a consistent theme in these components, its going to be needed sooner than previously expected

WeirAE:
Really not sure why `test__values_needed` in `test_jinja2.py` is failing now in the 3.9.19 version, nothing here should affect that

src/uwtools/tests/drivers/test_schism.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+# Tests
+
+
+def test_SCHISM(driverobj):
+    assert isinstance(driverobj, schism.SCHISM)
+
+
+def test_WaveWatchIII_namelist_file(driverobj):
```
christinaholtNOAA, line 57: WWIII?

christinaholtNOAA (approved):
LGTM. Just one minor copy/paste error (I think).

src/uwtools/drivers/schism.py
WeirAE: As long as @christinaholtNOAA is ok with letting these be for now, it should be ok, but it also shouldn't be bad to add if needed.

src/uwtools/tests/test_schemas.py
maddenp-noaa, line 1452:
```suggestion
# ww3
```
I think these headings are meant to track the `.jsonschema` filenames, though they are not currently all correct (e.g. underscores instead of hyphens). I put harmonizing those on my todo list, but `ww3` would be correct in any case.

src/uwtools/tests/drivers/test_schism.py
```diff
@@ -0,0 +1,82 @@
...
+        },
+    }
+
+
+@fixture
+def config_file(config, tmp_path):
+    path = tmp_path / "config.yaml"
+    with open(path, "w", encoding="utf-8") as f:
+        yaml.dump(config, f)
+    return path
```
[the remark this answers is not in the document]
WeirAE: Ah, good to include that here rather than have to fuss with PR sequencing, thanks!

src/uwtools/resources/jsonschema/schism.jsonschema
```diff
@@ -0,0 +1,41 @@
...
+              "type": "string"
+            },
+            "template_values": {
+              "additionalProperties": {
+                "type": [
+                  "array",
+                  "boolean",
+                  "number",
+                  "string"
+                ]
```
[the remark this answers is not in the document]
WeirAE: I was going for an open approach pending that support but `Object` felt like just a little bit of a stretch at the time - you might be right tho, thats a really interesting use case!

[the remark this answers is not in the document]
maddenp-noaa: I think it makes sense here (and in the updated `w3` schema below) to be inclusive. You might just simplify these to
```
"template_values": {"type": "object", "minProperties": 1}
```
to allow all the JSON value types.

maddenp-noaa (approved):
I left one comment about simplifying / generalizing the `template_values` item in the `schism` and `ww3` schemas, but otherwise this looks good to me!

PR #510, opened 2024-06-14
UW-608 filter_topo driver

maddenp-noaa:
**Synopsis**

A driver for the `filter_topo` utility. A sample run on `hera`: `/home/Paul.Madden/git/uwtools/run.filter_topo`, configured by `/home/Paul.Madden/git/uwtools/filter_topo.yaml`.

I also noticed some repeated patterns in the drivers and their unit tests and took the opportunity to do some refactoring:

1. Many drivers implemented an identical, simple `runscript()` method, so I factored that out to the `Driver` base class and let them inherit it. Several drivers override this default with a custom method.
2. Many of the driver unit tests repeatedly re-test methods from the `Driver` base class, which have their own unit tests. So I DRYed out those tests and replaced them with a test that asserts that certain methods on a derived class are, in fact, identical to methods in the base class -- which we know work correctly because they have their own tests.

I also tidied up the docs in a few places.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/yaml/components/filter_topo.rst
```diff
@@ -0,0 +1,42 @@
...
+--------------------------------------
+
+config:
+^^^^^^^
+
+Configuration parameters for the ``orog_gsl`` component.
+
+  **input_grid_file:**
+
+  Path to the tiled input grid file.
```
maddenp-noaa, line 25: I nested `input_grid_file:` under a `config:` parent for consistency with the `orog_gsl` config, which has no namelist and all its other config parameters under `config:`. Generally, I think I'd find it unnecessary to have a separate section with only a single child value, but the consistency with other related drivers might be helpful.

src/uwtools/config/support.py
```diff
@@ -56,6 +56,15 @@ def format_to_config(fmt: str) -> Type:
     return cfgclass
 
 
+def from_od(d: Union[OrderedDict, Dict]) -> dict:
+    """
+    Returns a (nested) dict with content equivalent to the given (nested) OrdereDict.
+
+    :param d: A (possibly nested) OrderedDict.
+    """
+    return {key: from_od(val) if isinstance(val, dict) else val for key, val in d.items()}
```
maddenp-noaa, line 65: I needed this for a unit test, so factored it out of the `_represent_ordereddict()` method, below.

src/uwtools/drivers/driver.py
```diff
@@ -220,6 +220,17 @@ def run(self):
...
+    @task
+    def runscript(self):
+        """
+        The runscript.
+        """
+        path = self._runscript_path
+        yield self._taskname(path.name)
+        yield asset(path, path.is_file)
+        yield None
+        self._write_runscript(path)
```
maddenp-noaa, line 232: This was repeated verbatim in many driver modules, so I factored it out here as a base-class default. Some drivers override it.

src/uwtools/drivers/upp.py
```diff
@@ -131,13 +120,6 @@ def _namelist_path(self) -> Path:
         """
         return self._rundir / "itag"
 
-    @property
-    def _runscript_path(self) -> Path:
-        """
-        Path to the runscript.
-        """
-        return self._rundir / f"runscript.{self._driver_name}"
```
maddenp-noaa, line 139: This is provided verbatim in the `Driver` base class.

src/uwtools/tests/config/test_support.py
```diff
@@ -68,11 +74,13 @@ def test_log_and_error(caplog):
 
 
 def test_represent_namelist():
+    support.add_yaml_representers()
```
maddenp-noaa, line 77: Here and just below: These tests were passing by accident because some other test just happened to call this function before these tests ran. Given the new-ish parallelism in our unit tests, these tests were now sometimes failing. Since they require `add_yaml_representers()` to have been called, and since there's no harm (I think) in calling them repeatedly (and hopefully concurrently!) I added explicit calls and could no longer trigger a test failure.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -76,14 +77,28 @@ def config(tmp_path):
...
+        "_runcmd",
+        "_runscript",
+        "_runscript_done_file",
+        "_runscript_path",
+        "_scheduler",
+        "_validate",
+        "_write_runscript",
+        "run",
+    ]:
+        assert getattr(ChgresCube, method) is getattr(Driver, method)
```
maddenp-noaa, line 101: This is a pattern for many tests that follow:

1. Simplify the import of the driver module.
2. Assert that certain methods are actually the methods of the `Driver` base class, which have their own tests.
3. Remove tests that are essentially duplicates of tests in `test_driver.py`.
4. Ensure that all the methods that are explicitly implemented by the driver class have direct tests.

src/uwtools/tests/drivers/test_schism.py
```diff
@@ -36,14 +37,19 @@ def cycle():
...
 
-def test_SCHISM(driverobj):
-    assert isinstance(driverobj, schism.SCHISM)
+def test_SCHISM():
+    for method in [
+        "_driver_config",
+        "_taskname",
+        "_validate",
+    ]:
+        assert getattr(SCHISM, method) is getattr(Assets, method)
```
maddenp-noaa, line 52: Note that the SCHISM and WW3 driver derive from the `Assets` base class and so have a shorter list of inherited methods.

src/uwtools/config/support.py
```diff
@@ -56,6 +56,15 @@ def format_to_config(fmt: str) -> Type:
     return cfgclass
 
 
+def from_od(d: Union[OrderedDict, Dict]) -> dict:
+    """
+    Returns a (nested) dict with content equivalent to the given (nested) OrdereDict.
```
elcarpenterNOAA, line 61:
```suggestion
    Returns a (nested) dict with content equivalent to the given (nested) OrderedDict.
```

WeirAE (approved):
Solid driver, great refactor - thanks!

docs/shared/filter_topo.yaml
```diff
@@ -0,0 +1,27 @@
...
+      cores: 1
+      walltime: "00:01:00"
+    envcmds:
+      - module use /path/to/modules
+      - module load module_name
+    executable: /path/to/filter_topo
+  namelist:
+    update_values:
+      filter_topo_nml:
+        grid_file: /path/to/C{{ filter_topo.namelist.filter_topo_nml.res }}_mosaic.halo6.nc
```
christinaholtNOAA, line 15:
```suggestion
        grid_file: /path/to/C{{ filter_topo.namelist["update_values"]["filter_topo_nml"]["res"] }}_mosaic.halo6.nc
```
I wish we could do dots all the way down! In my experience, only the first one works that way.

src/uwtools/drivers/global_equiv_resol.py
```diff
@@ -5,7 +5,7 @@
 from pathlib import Path
 from typing import List, Optional
 
-from iotaa import asset, external, task, tasks
+from iotaa import asset, external, tasks
```
christinaholtNOAA, line 8: It's pretty cool that we've gotten some drivers all the way down to not needing their own task definitions.

src/uwtools/tests/drivers/test_filter_topo.py
```diff
@@ -0,0 +1,102 @@
...
+        "runscript",
+    ]:
+        assert getattr(FilterTopo, method) is getattr(Driver, method)
+
+
+def test_FilterTopo_input_grid_file(driverobj):
+    path = Path(driverobj._driver_config["run_dir"]) / "C403_grid.tile7.halo4.nc"
+    assert not path.is_file()
+    driverobj.input_grid_file()
+    assert path.is_symlink()
```
christinaholtNOAA, line 84: I must have skipped over it in the driver....this file gets linked? It does not end up getting overwritten by the program, right?

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -70,14 +71,52 @@ def cycle():
...
+        "_runscript",
+        "_runscript_done_file",
+        "_runscript_path",
+        "_scheduler",
+        "_validate",
+        "_write_runscript",
+        "run",
+        "runscript",
+    ]:
+        assert getattr(JEDI, method) is getattr(Driver, method)
```
christinaholtNOAA, line 95: Now that I've flown past a few of these, it makes me realize that perhaps we will never update these lists when/if we add new methods to the drivers. Can there be a test that will remind us to do that? I don't even know what that would look like.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -169,34 +186,14 @@ def file(path: Path):
...
-    driverobj.configuration_file()
-    assert not cfgfile.is_file()
-    assert regex_logged(caplog, f"{base_file}: State: Not Ready (external asset)")
-
-
 def test_JEDI__config_fn(driverobj):
     assert driverobj._config_fn == "jedi.yaml"
 
 
+def test_JEFI__driver_name(driverobj):
```
christinaholtNOAA, line 193:
```suggestion
def test_JEDI__driver_name(driverobj):
```
I appreciate the irony in this one!

christinaholtNOAA (commented):
This looks great.

I have a future question inspired by this test clean up work. Can we have a test that fails if we haven't tested all the methods on a driver? I was thinking something like checking that the `set(attributes of Driver subclass - attributes of Driver class)` show up in the tests in the driver test file.

docs/sections/user_guide/yaml/components/filter_topo.rst
maddenp-noaa: Because the requirements aren't generic: There's a fixed set of required files, and by being explicit about those, we can give users precise schema-checked guidance on what's required. If we found that we also needed some ability to copy generic files, we could add that, but with these tiny, old, single-purpose components, that strikes me as unlikely.

docs/sections/user_guide/yaml/components/sfc_climo_gen.rst
```diff
@@ -21,7 +21,7 @@ See :ref:`this page <execution_yaml>` for details.
 namelist:
 ^^^^^^^^^
 
-Supports ``base_file:`` and ``update_values:`` blocks (see the :ref:`updating_values` for details). Namelist options are described :ufs-utils:`here<id57>`.
+Supports ``base_file:`` and ``update_values:`` blocks (see :ref:`updating_values` for details). Namelist options are described :ufs-utils:`here <id56>`.
```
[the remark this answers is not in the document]
maddenp-noaa: I just noticed, when clicking the link, that it was one section off in the target doc.

docs/shared/filter_topo.yaml
maddenp-noaa: I just re-executed the example run referenced in the PR _Description_ with this all-dots syntax, and it seems fine to me. Do you have a counterexample?

src/uwtools/tests/drivers/test_filter_topo.py
maddenp-noaa: The program modifies the topo (aka orography) file, not this one. From [the docs](https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#id49):

> The filtered orography is written to the “orog_filt” record of the input orography file ...

If you look at the example run, the symlink `C403_grid.tile7.halo6.nc` in the run directory is symlinked to a file that you own, that I cannot write to, so I'm confident that it's not the one being written to.

src/uwtools/tests/drivers/test_jedi.py
maddenp-noaa: An automated version of this manually curated list would involve _generating_ this list by seeing which methods in a derived class are synonymous with the same-named method in the base class, but that would make this test a tautology.

So, I think that we either need to manually define this list to state what we believe to be true, or possibly just remove this test (and the corresponding tests in other modules) on the grounds that it doesn't provide much value, but...

What this test provides is the following: If today e.g `JEDI` inherits a method from `Driver` -- which has its own unit test(s) -- and tomorrow we implement a custom override method in `JEDI`, this test will fail, alerting us to the fact that we need to write a new unit test for that new custom override method.

I'm not sure I answered your question, though.

docs/sections/user_guide/cli/drivers/make_hgrid.rst
```diff
@@ -43,12 +43,6 @@ Its contents are described in section :ref:`make_hgrid_yaml`.
 
   The driver creates a ``runscript.make_hgrid`` file in the directory specified by ``run_dir:`` in the config and submits it to the batch system. Running with ``--batch`` requires a correctly configured ``platform:`` block in ``config.yaml``, as well as appropriate settings in the ``execution:`` block under ``make_hgrid:``.
 
-  Looking at the run command in ``runscript.make_hgrid`` shows us the specified executable as well as the YAML keys converted to appropriate command line flags.
-
-  .. code-block:: text
-
-     time make_hgrid --grid_type gnomonic_ed --do_schmidt --great_circle_algorithm --grid_name C96_foo --halo 1 --nest_grids 1 --istart_nest 10 --iend_nest 87 --jstart_nest 19 --jend_nest 78 --nlon 192 --parent_tile 6 --refine_ratio 2 --stretch_factor 1.0001 --target_lon -97.5 --target_lat 38.5
```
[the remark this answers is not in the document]
maddenp-noaa: I added content (see above) to dynamically generate this example output.

docs/sections/user_guide/cli/drivers/make_solo_mosaic.rst
```diff
@@ -35,6 +35,15 @@ Its contents are described in section :ref:`make_solo_mosaic_yaml`.
...
   The driver creates a ``runscript.make_solo_mosaic`` file in the directory specified by ``run_dir:`` in the config and runs it, executing ``make_solo_mosaic``.
 
+..
+  An example runscript:
+
+  .. literalinclude:: make_solo_mosaic/runscript.cmd
+     :language: text
+     :emphasize-lines: 5
+  .. literalinclude:: make_solo_mosaic/runscript.out
+     :language: text
```
maddenp-noaa, line 45: This block is currently commented out because the prototype `docs/shared/make_solo_mosaic.yaml` appears not to be valid (or its schema is incorrect):
```
% uw make_solo_mosaic validate --config-file docs/shared/make_solo_mosaic.yaml 
[2024-06-15T15:25:22]     INFO Validating config against internal schema make-solo-mosaic
[2024-06-15T15:25:22]    ERROR 2 UW schema-validation errors found
[2024-06-15T15:25:22]    ERROR Error at make_solo_mosaic -> config:
[2024-06-15T15:25:22]    ERROR   Additional properties are not allowed ('tile_file' was unexpected)
[2024-06-15T15:25:22]    ERROR Error at make_solo_mosaic -> config -> mosaic_name:
[2024-06-15T15:25:22]    ERROR   'ocean_mosaic' is not of type 'integer'
[2024-06-15T15:25:22]    ERROR YAML validation errors
```
We could fix this later and uncomment this block (remove `..` line 38).

docs/shared/filter_topo.yaml
christinaholtNOAA: Hmmm. I will revisit this later. Perhaps I'd assumed that was an issue when I'd had failures in the past, but it was something else?

christinaholtNOAA (commented):
LGTM.

christinaholtNOAA (approved):

elcarpenterNOAA (approved):
Looks good for me. Thanks for making all the additional updates.

PR #512, opened 2024-06-18
Better support for realizing scalar YAML values

maddenp-noaa:
**Synopsis**

Add YAML configuration to omit `...` end-of-stream marker when dumping scalar values. For example:

Old behavior:
```
(DEV-uwtools) ~/git/uwtools ∴ echo "{foo: {bar: baz}}" | uw config realize --input-format yaml --output-format yaml --key-path foo.bar
baz
...
```

New behavior:
```
(DEV-uwtools) ~/git/uwtools ∴ echo "{foo: {bar: baz}}" | uw config realize --input-format yaml --output-format yaml --key-path foo.bar
baz
```

This supports easier lookups of single values from input configs.

Works for e.g. namelist inputs, too:
```
(DEV-uwtools) ~/git/uwtools ∴ echo "&nl res = 768 /" | uw config realize --input-format nml --output-format yaml --key-path nl.res
768
```

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -1,17 +1,13 @@
+from collections import OrderedDict
 from pathlib import Path
 from types import SimpleNamespace as ns
 from typing import Optional
 
 import yaml
+from f90nml import Namelist  # type: ignore
 
 from uwtools.config.formats.base import Config
-from uwtools.config.support import (
-    INCLUDE_TAG,
-    UWYAMLConvert,
-    UWYAMLRemove,
-    add_yaml_representers,
-    log_and_error,
-)
+from uwtools.config.support import INCLUDE_TAG, UWYAMLConvert, UWYAMLRemove, from_od, log_and_error
 from uwtools.exceptions import UWConfigError
 from uwtools.strings import FORMAT
 from uwtools.utils.file import readable, writable
@@ -52,7 +48,7 @@ class YAMLConfig(Config):
         """
         The string representation of a YAMLConfig object.
         """
-        add_yaml_representers()
+        self._add_yaml_representers()
         return yaml.dump(self.data, default_flow_style=False).strip()
 
     # Private methods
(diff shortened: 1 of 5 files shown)


src/uwtools/config/formats/yaml.py
```diff
@@ -145,3 +141,52 @@ def get_format() -> str:
         Returns the config's format name.
         """
         return FORMAT.yaml
+
+    # Private methods
```
maddenp-noaa, line 145: I relocated some YAML-related code from `uwtools.config.support` to the `YAMLConfig` class. I half expected a circular-dependency issue to arise in the refactor, and assumed that that's the reason the code lived where it did, but was pleasantly surprised to find no issues incorporating it into the class where it seems like it belongs.

src/uwtools/config/formats/yaml.py
```diff
@@ -145,3 +141,52 @@ def get_format() -> str:
...
+    Write YAML without ...
+
+    end-of-stream marker.
+    """
+    self.write_plain_base(*args, **kwargs)
+    self.open_ended = False
+
+
+setattr(yaml.emitter.Emitter, "write_plain_base", yaml.emitter.Emitter.write_plain)
+setattr(yaml.emitter.Emitter, "write_plain", _write_plain_open_ended)
```
maddenp-noaa, line 192: These two `setattr()` calls and the `_write_plain_open_ended()` function they refer to are responsible for removing the `...` end-of-stream marker from the YAML-scalar output.

src/uwtools/tests/config/test_tools.py
```diff
@@ -347,6 +347,16 @@ def test_realize_config_remove_yaml_to_yaml_subtree(tmp_path):
...
 
 
+def test_realize_config_scalar_value(capsys):
+    stdinproxy.cache_clear()
+    tools.realize_config(
+        input_config=YAMLConfig(config={"foo": {"bar": "baz"}}),
+        output_format="yaml",
+        key_path=["foo", "bar"],
+    )
+    assert capsys.readouterr().out.strip() == "baz"
```
maddenp-noaa, line 357: A test for the new behavior. Previously, the RHS of the `assert` statement would have been `baz\n...`.

christinaholtNOAA (approved):
LGTM!

src/uwtools/config/formats/yaml.py
WeirAE: aha, nice!

WeirAE (approved):
LGTM

PR #513, opened 2024-06-18
UW-626 Update developer_setup.rst documentation with IDE troubleshooting

lexbuko22:
**Synopsis**

Added language for VS Code Python interpreter troubleshooting to the developer setup docs.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/contributor_guide/developer_setup.rst
+++ b/docs/sections/contributor_guide/developer_setup.rst
@@ -31,7 +31,7 @@ Future ``make devshell`` invocations will be almost instantaneous, as the underl
 
 If your development shell misses any functionality you’re used to in your main (``bash``) shell, you can create a ``~/.condevrc`` file, which will be sourced by ``make devshell``, and add desired environment-setup commands to it.
 
-If using an IDE such as VS Code, ensure that the correct Python interpretor, Python 3.12.3 ('DEV-uwtools'), is being used for the environment. In VS Code, this can be changed by using the Command Pallet and searching "Python: Select Interpretor".
+If using an IDE such as VS Code, ensure that the correct Python interpreter, Python 3.12.3 ('DEV-uwtools'), is being used for the environment. In VS Code, this can be changed by using the Command Palette and searching "Python: Select Interpreter".
 
 Using a ``bash`` Development Shell
 ----------------------------------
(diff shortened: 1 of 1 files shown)


docs/sections/contributor_guide/developer_setup.rst
```diff
@@ -31,6 +31,8 @@ Future ``make devshell`` invocations will be almost instantaneous, as the underl
 
 If your development shell misses any functionality you’re used to in your main (``bash``) shell, you can create a ``~/.condevrc`` file, which will be sourced by ``make devshell``, and add desired environment-setup commands to it.
 
+If using an IDE such as VS Code, ensure that the correct Python interpretor, Python 3.12.3 ('DEV-uwtools'), is being used for the environment. In VS Code, this can be changed by using the Command Pallet and searching "Python: Select Interpretor".
```
WeirAE, line 34:
```suggestion
If using an IDE such as VS Code, ensure that the correct Python interpreter, Python 3.12.3 ('DEV-uwtools'), is being used for the environment. In VS Code, this can be changed by using the Command Palette and searching "Python: Select Interpreter".
```

I might also suggest moving the specific text `Python 3.12.3 ('DEV-uwtools')` to the end, I don't think it looks the same in, say, PyCharm.

WeirAE (changes requested):
Its a good idea to add a little UW-specific text about our most common IDE here, so thanks! Just a technicality here and we can probably add more details later as well

docs/sections/contributor_guide/developer_setup.rst
maddenp-noaa: I agree with @WeirAE 's comment, and would add that the specific Python version may change over time, and `uwtools` supports a range of Python versions: 3.9.x through 3.12.x, as of this writing, but we'll update that when the next Python version is released and so would not want to have to find and update references to it. So, it would be helpful to find a more generic way to hint the user about what they should select, e.g.

> ...ensure that the Python interpreter belonging to the conda environment providing ``uwtools`` is selected.

Hopefully something like that would indicate fairly well which of the available Python interpreters they should use without being specific in a brittle way.

elcarpenterNOAA (approved):
Nice addition.

WeirAE (approved):
Looks good!

PR #517, opened 2024-06-24
omnibus-2024-06-24

maddenp-noaa:
**Synopsis**

1. Since Python 3.9, which happens to be the earliest version `uwtools` supports, type annotations can use the built-in type names instead of those provided by the `typing` module, e.g. `list` can be used instead of `List`, eliminating the need for a `from typing import List` statement. This PR simplifies our code by taking advantage of this fact for `list`, `dict`, `tuple`, and `set` types.
2. Use `from pytest import mark` and simplify `@pytest.mark.parametrize` decorations given this more explicit import.
3. In a number of tests, convert in-test `for` loops to `@mark.parametrize` decorations. Parametrization is preferable because an `assert` in a `for` loop terminates the entire test and reports a single error, while a parametrized test tests every parametrized value and reports every failure. (I will open a separate PR in the near future to update `test_schemas.py` to break up many of the tests that contains multiple `assert` statements into separate tests, some parametrized, to obtain similar benefits.)

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -4,7 +4,7 @@ API access to ``uwtools`` configuration management tools.
 
 import os
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 from uwtools.config.formats.fieldtable import FieldTableConfig as _FieldTableConfig
 from uwtools.config.formats.ini import INIConfig as _INIConfig
@@ -119,7 +119,7 @@ def realize(
     update_format: Optional[str] = None,
     output_file: Optional[Union[Path, str]] = None,
     output_format: Optional[str] = None,
-    key_path: Optional[List[Union[str, int]]] = None,
+    key_path: Optional[list[Union[str, int]]] = None,
     values_needed: bool = False,
     total: bool = False,
     dry_run: bool = False,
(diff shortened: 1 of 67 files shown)


src/uwtools/api/config.py
```diff
@@ -119,7 +119,7 @@ def realize(
     update_format: Optional[str] = None,
     output_file: Optional[Union[Path, str]] = None,
     output_format: Optional[str] = None,
-    key_path: Optional[List[Union[str, int]]] = None,
+    key_path: Optional[list[Union[str, int]]] = None,
```
christinaholtNOAA, line 122: This is nice!

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -97,8 +98,10 @@ def test_ChgresCube():
         "_validate",
         "_write_runscript",
         "run",
-    ]:
-        assert getattr(ChgresCube, method) is getattr(Driver, method)
+    ],
+)
+def test_ChgresCube(method):
```
christinaholtNOAA, line 103: Is my assumption correct that this could help even more with the ability to run tasks more quickly in parallel? I mean maybe not this test explicitly since it looks like it might be fairly inexpensive, but in general.

christinaholtNOAA (approved):
LGTM!

src/uwtools/tests/drivers/test_chgres_cube.py
maddenp-noaa: I hadn't thought of it but, yes, I think so: This test would previously have had to be run by a single worker, and now each of its cases can be executed concurrently.

PR #521, opened 2024-07-02
Allow API users to obtain copies of driver configs

maddenp-noaa:
**Synopsis**

- Add new property methods, `.config` and `.config_full`, to the `Assets` class (inherited by all subclasses, i.e. `Driver` and all the individual component drivers) to obtain _copies of_ the driver config and the full input config. Copies are provided for informational purposes and cannot be used to modify the driver's actual configuration.
- Now that driver objects can be directly created and manipulated via the `uwtools` API, provide `__repr__` and `__str__` methods.
- DRY out driver constructors and `_taskname()` methods by creating intermediate classes for time-invariant, cycle-based, and cycle-and-leadtime-based drivers. This results in a net reduction in code, and a simplification of all the driver modules.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,10 +1,3 @@
 # See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-# Code owners:
-
 * @NaureenBharwaniNOAA @christinaholtNOAA @elcarpenterNOAA @fgabelmannjr @maddenp-noaa @weirae
-
-# Documentation owners:
-
-.readthedocs.yaml @christinaholtNOAA @jprestop @maddenp-noaa
-docs/* @christinaholtNOAA @jprestop @maddenp-noaa
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -10,6 +10,7 @@ sys.path.insert(0, os.path.abspath("../src"))
 with open("../recipe/meta.json", "r", encoding="utf-8") as f:
     _metadata = json.loads(f.read())
 
+autoclass_content = "both"
 autodoc_mock_imports = ["f90nml", "iotaa", "jsonschema", "lxml", "referencing"]
 copyright = str(dt.datetime.now().year)
 extensions = ["sphinx.ext.autodoc", "sphinx.ext.extlinks", "sphinx.ext.intersphinx"]
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -83,11 +83,3 @@ class ChgresCube(DriverWithCycle):
         Returns the name of this driver.
         """
         return STR.chgrescube
-
-    def _taskname(self, suffix: str) -> str:
-        """
-        Returns a common tag for graph-task log messages.
-
-        :param suffix: Log-string suffix.
-        """
-        return self._taskname_with_cycle(suffix)
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -7,12 +7,12 @@ from pathlib import Path
 from iotaa import asset, task, tasks
 
 from uwtools.config.formats.nml import NMLConfig
-from uwtools.drivers.driver import Driver
+from uwtools.drivers.driver import DriverWithCycle
 from uwtools.strings import STR
 from uwtools.utils.tasks import file
 
 
-class ChgresCube(Driver):
+class ChgresCube(DriverWithCycle):
     """
     A driver for chgres_cube.
     """
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -39,7 +39,6 @@ class ChgresCube(Driver):
         super().__init__(
             config=config, dry_run=dry_run, batch=batch, cycle=cycle, key_path=key_path
         )
-        self._cycle = cycle
 
     # Workflow tasks
 
@@ -114,4 +113,4 @@ class ChgresCube(Driver):
 
         :param suffix: Log-string suffix.
         """
-        return self._taskname_with_cycle(self._cycle, suffix)
+        return self._taskname_with_cycle(suffix)
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -7,12 +7,12 @@ from pathlib import Path
 from iotaa import asset, task, tasks
 
 from uwtools.config.formats.nml import NMLConfig
-from uwtools.drivers.driver import DriverWithCycle
+from uwtools.drivers.driver import DriverCycleBased
 from uwtools.strings import STR
 from uwtools.utils.tasks import file
 
 
-class ChgresCube(DriverWithCycle):
+class ChgresCube(DriverCycleBased):
     """
     A driver for chgres_cube.
     """
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -2,9 +2,7 @@
 A driver for chgres_cube.
 """
 
-from datetime import datetime
 from pathlib import Path
-from typing import Optional
 
 from iotaa import asset, task, tasks
 
@@ -18,27 +16,6 @@ class ChgresCube(Driver):
     """
     A driver for chgres_cube.
     """
-
-    def __init__(
-        self,
-        cycle: datetime,
-        config: Optional[Path] = None,
-        dry_run: bool = False,
-        batch: bool = False,
-        key_path: Optional[list[str]] = None,
-    ):
-        """
-        The driver.
-
-        :param cycle: The cycle.
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param batch: Run component via the batch system?
-        :param key_path: Keys leading through the config to the driver's configuration block.
-        """
-        super().__init__(
-            config=config, dry_run=dry_run, batch=batch, cycle=cycle, key_path=key_path
-        )
 
     # Workflow tasks
 
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -7,6 +7,7 @@ import os
 import re
 import stat
 from abc import ABC, abstractmethod
+from copy import deepcopy
 from datetime import datetime, timedelta
 from functools import partial
 from pathlib import Path
@@ -32,28 +33,26 @@ class Assets(ABC):
     def __init__(
         self,
         config: Optional[Union[dict, Path]],
-        dry_run: bool = False,
-        batch: bool = False,
         cycle: Optional[datetime] = None,
         leadtime: Optional[timedelta] = None,
+        batch: bool = False,
         key_path: Optional[list[str]] = None,
+        dry_run: bool = False,
     ) -> None:
         """
         A component driver.
 
         :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param batch: Run component via the batch system?
         :param cycle: The cycle.
         :param leadtime: The leadtime.
+        :param batch: Run component via the batch system?
         :param key_path: Keys leading through the config to the driver's configuration block.
+        :param dry_run: Run in dry-run mode?
         """
-        dryrun(enable=dry_run)
-        self._config = YAMLConfig(config=config)
-        self._batch = batch
         has_leadtime = leadtime is not None
         if has_leadtime and not cycle:
             raise UWError("When leadtime is specified, cycle is required")
+        self._config = YAMLConfig(config=config)
         self._config.dereference(
             context={
                 **({"cycle": cycle} if cycle else {}),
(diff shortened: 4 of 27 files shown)


src/uwtools/drivers/chgres_cube.py
```diff
@@ -2,45 +2,21 @@
...
 
 from uwtools.config.formats.nml import NMLConfig
-from uwtools.drivers.driver import Driver
+from uwtools.drivers.driver import DriverCycleBased
 from uwtools.strings import STR
 from uwtools.utils.tasks import file
 
 
-class ChgresCube(Driver):
+class ChgresCube(DriverCycleBased):
```
maddenp-noaa, line 15: This is the first example of a theme that repeats below (beyond the `driver.py` diff): Simplify the driver class by inheriting from a new, more-specific base class that defines some common behavior. Some imports are no longer needed, and the constructor and `_taskname()` method from the immediate parent class is correct for this class.

src/uwtools/drivers/driver.py
```diff
@@ -32,39 +33,61 @@ class Assets(ABC):
...
+    def __repr__(self) -> str:
+        cycle = self._cycle.strftime("%Y-%m-%dT%H:%M") if hasattr(self, "_cycle") else None
+        leadtime = None
+        if hasattr(self, "_leadtime"):
+            h, r = divmod(self._leadtime.total_seconds(), 3600)
+            m, s = divmod(r, 60)
+            leadtime = "%02d:%02d:%02d" % (h, m, s)
+        return " ".join(
+            filter(None, [str(self), cycle, leadtime, "in", self._driver_config["run_dir"]])
+        )
```
maddenp-noaa, line 72: `__repr__` defines how an object of this class, `obj` will be displayed either by `repr(obj)` or by just typing `obj` in the REPL or e.g. in a Jupyter notebook cell. As the unit tests show, the representations are:

- for time-invariant objects: `driver-name in /path/to/run/dir`
- for cycle-based objects: `driver-name yyyy-mm-ddThh:mm in /path/to/run/dir`
- for cycle-and-leadtime-based objects `driver-name yyyy-mm-ddThh:mm HH:MM:SS in /path/to/run/dir`

Certainly these are up for discussion, but the new
```
>>> UPP(config="docs/shared/upp.yaml", cycle=dt.datetime.now(), leadtime=dt.timedelta(hours=6))
upp 2024-07-02T22:24 06:00:00 in /path/to/run/dir
```

is probably more informative than the old

```
>>> UPP(config="docs/shared/upp.yaml", cycle=dt.datetime.now(), leadtime=dt.timedelta(hours=6))
<uwtools.drivers.upp.UPP object at 0xffff9bf45a60>
```

src/uwtools/drivers/driver.py
```diff
@@ -32,39 +33,61 @@ class Assets(ABC):
...
+
+    def __str__(self) -> str:
+        return self._driver_name
```
maddenp-noaa, line 75: e.g.
```
>>> x = UPP(config="docs/shared/upp.yaml", cycle=dt.datetime.now(), leadtime=dt.timedelta(hours=6))
>>> str(x)
'upp'
```

This string is also what the driver expects to find as a key pointing to the driver-config block in UW YAML (e.g. `upp:` for the UPP block), so could be used to look up the driver-specific block from the `dict` returned by the new `config_full` property (see below) without harcoding the driver name in external Python or bash code; but, then, it would be easier to just use the new `config` property (see below). But, for sake of completeness, both are possible.

src/uwtools/drivers/driver.py
```diff
@@ -32,39 +33,61 @@ class Assets(ABC):
...
+
+    @property
+    def config(self) -> dict:
+        """
+        Return a copy of the driver config.
+        """
+        return deepcopy(self._driver_config)
```
maddenp-noaa, line 82: This will return `<block>` in e.g. for the UPP driver given UW YAML
```
upp:
    <block>
```

It returns a copy so that the receiver cannot use the returned value to interfere with the state of the driver object.

src/uwtools/drivers/driver.py
```diff
@@ -32,39 +33,61 @@ class Assets(ABC):
...
+
+    @property
+    def config_full(self) -> dict:
+        """
+        Return a copy of the full input config.
+        """
+        full_config: dict = self._config.data
+        return deepcopy(full_config)
```
maddenp-noaa, line 90: This returns a copy of the input config, which might include driver (e.g. `upp:`), `platform:`, `user:` etc. blocks. However, if `key_path` is being used, the returned value will be a copy of the block at the end of the key path, and so may exclude some top-level blocks. But, as with the new `config` property above, the returned `dict` should be fully dereferenced.

src/uwtools/drivers/driver.py
```diff
@@ -174,43 +197,126 @@ def _taskname(self, suffix: str) -> str:
...
         """
-        return "%s %s" % (self._driver_name, suffix)
+        cycle = getattr(self, "_cycle", None)
+        leadtime = getattr(self, "_leadtime", None)
+        timestr = (
+            (cycle + leadtime).strftime("%Y%m%d %H:%M:%S")
+            if cycle and leadtime is not None
+            else cycle.strftime("%Y%m%d %HZ") if cycle else None
+        )
+        return " ".join(filter(None, [timestr, self._driver_name, suffix]))
```
maddenp-noaa, line 207: This version of `_taskname()` now flexibly deals with the time-invariant / cycle-based / cycle-and-leadtime-based drivers so that multiple implementations are not needed.

src/uwtools/drivers/driver.py
```diff
@@ -174,43 +197,126 @@ def _taskname(self, suffix: str) -> str:
...
         """
-        Returns a common tag for graph-task log messages.
+        Perform all necessary schema validation.
+        """
+        schema_name = self._driver_name.replace("_", "-")
+        validate_internal(schema_name=schema_name, config=self._config)
 
-        :param suffix: Log-string suffix.
+
+class AssetsCycleBased(Assets):
```
maddenp-noaa, line 217: Here and for a bit below, new base classes for cycle-based, cycle-and-leadtime-based, and time-invariant drivers, both assets-style and standalone-style. We need these classes for their explicit argument lists: While no interfaces would break if, for example, `ChgresCube` inherited directly from `Assets`, in e.g. the REPL a `help(ChgresCube)` would show the signature and docstring from the `Assets` class' constructor, which includes optional `cycle` and `leadtime` arguments, which aren't valid for `ChgresCube`. So, while a user could still instantiate a `ChgresCube` object by just not providing `cycle` and `leadtime` arguments, it would be confusing to see them there as options. So, these intermediate classes provide correct signatures with only the actually-supported arguments for each driver inheriting from them.

src/uwtools/drivers/driver.py
```diff
@@ -174,43 +197,126 @@ def _taskname(self, suffix: str) -> str:
...
     An abstract class for standalone component drivers.
     """
 
+    def __init__(
+        self,
+        config: Optional[Path] = None,
+        cycle: Optional[datetime] = None,
+        leadtime: Optional[timedelta] = None,
+        dry_run: bool = False,
+        batch: bool = False,
```
maddenp-noaa, line 302: `Driver` now provides its own constructor so that it can receive the `batch` argument, which `Assets` should not receive. It sets the `self._batch` instance variable from the formal argument so that `Assets` is relieved of that duty.

src/uwtools/drivers/ioda.py
```diff
@@ -13,6 +13,8 @@ class IODA(JEDIBase):
     A driver for the IODA component.
     """
 
+    # Workflow tasks
```
maddenp-noaa, line 16: For consistency with other driver modules.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -35,40 +42,32 @@ def provisioned_run_directory(self):
...
     def _validate(self) -> None:
         pass
 
-    @task
-    def atask(self):
-        yield "atask"
-        yield asset("atask", lambda: True)
 
+class ConcreteAssetsCycleBased(Common, driver.AssetsCycleBased):
+    pass
```
maddenp-noaa, line 50: There are now six "concrete" classes, each extending the new `Assets` and `Driver` intermediate classes and inheriting their methods from the `Common` class just above. Two of the classes aren't current used by tests, but are here for completeness and because we'll probably find a use for them before long.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -108,46 +107,68 @@ def config(tmp_path):
...
+        config=config, cycle=dt.datetime(2024, 7, 2, 12), leadtime=dt.timedelta(hours=6)
+    )
+    expected = "concrete 2024-07-02T12:00 06:00:00 in %s" % obj._driver_config["run_dir"]
+    assert repr(obj) == expected
+
+
+def test_Assets_repr_time_invariant(config):
+    obj = ConcreteAssetsTimeInvariant(config=config)
+    expected = "concrete in %s" % obj._driver_config["run_dir"]
+    assert repr(obj) == expected
```
maddenp-noaa, line 143: The above three tests exercise the new, flexible `__repr__()` method defined in `Assets` and inherited by its subclass hierarchy.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -108,46 +107,68 @@ def config(tmp_path):
...
+    assert str(assetsobj) == "concrete"
+
+
+def test_Assets_config(assetsobj):
+    old = assetsobj._driver_config
+    new = assetsobj.config
+    # The user-accessible object is equivalent to the internal driver config:
+    assert new == old
+    # But they are separate objects:
+    assert not new is old
```
maddenp-noaa, line 156: Test the new `config` property.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -108,46 +107,68 @@ def config(tmp_path):
...
+
+
+def test_Assets_config_full(assetsobj):
+    old = assetsobj._config
+    new = assetsobj.config_full
+    # The user-accessible object is equivalent to the internal driver config:
+    assert new == old
+    # But they are separate objects:
+    assert not new is old
```
maddenp-noaa, line 165: Test the new `config_full` property.

docs/conf.py
```diff
@@ -10,6 +10,7 @@
 with open("../recipe/meta.json", "r", encoding="utf-8") as f:
     _metadata = json.loads(f.read())
 
+autoclass_content = "both"
```
maddenp-noaa, line 13: This lets Sphinx add driver-class constructor arguments to the HTML docs.

src/uwtools/drivers/driver.py
```diff
@@ -31,40 +32,62 @@ class Assets(ABC):
...
         A component driver.
 
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param batch: Run component via the batch system?
         :param cycle: The cycle.
         :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
         :param key_path: Keys leading through the config to the driver's configuration block.
```
maddenp-noaa, line 48: I reordered the arguments lists for the `Asset` / `Driver` class hierarchy, and the calls between them via `super()`. Non-optional arguments are of course first, but the optional arguments now appear in the same order as in the `Assets` constructor, with `batch` tacked on to the end for `Driver` and its subclasses.

I also harmonized the type annotation for the `config` arguments across all these classes.

.github/CODEOWNERS
```diff
@@ -1,10 +1,3 @@
 # See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
 
-# Code owners:
-
 * @NaureenBharwaniNOAA @christinaholtNOAA @elcarpenterNOAA @fgabelmannjr @maddenp-noaa @weirae
-
-# Documentation owners:
-
-.readthedocs.yaml @christinaholtNOAA @jprestop @maddenp-noaa
-docs/* @christinaholtNOAA @jprestop @maddenp-noaa
```
maddenp-noaa, line 10: Updated so that core UW team can approve doc changes as well as code changes.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -108,46 +107,64 @@ def config(tmp_path):
...
+
+def test_Assets_str(assetsobj):
+    assert str(assetsobj) == "concrete"
+
+
+def test_Assets_config(assetsobj):
+    # The user-accessible object is equivalent to the internal driver config:
+    assert assetsobj.config == assetsobj._driver_config
+    # But they are separate objects:
+    assert not assetsobj.config is assetsobj._driver_config
```
maddenp-noaa, line 154: Test the new `config` property.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -108,46 +107,64 @@ def config(tmp_path):
...
+
+
+def test_Assets_config_full(assetsobj):
+    # The user-accessible object is equivalent to the internal driver config:
+    assert assetsobj.config_full == assetsobj._config
+    # But they are separate objects:
+    assert not assetsobj.config_full is assetsobj._config
```
maddenp-noaa, line 161: Test the new `config_full` property.

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):
Nice work on this!

christinaholtNOAA (commented):
I'm late to the party, but this looks great!

PR #525, opened 2024-07-09
Config realize bugfix

maddenp-noaa:
# NOT READY FOR REVIEW

**Synopsis**

Consider the following reasonable thing you might want to do via the `uwtools` API (in particular, this might be useful in a loop to apply multiple updates to a base config initially represented as a `dict`, after which a `YAMLConfig`, `NMLConfig`, etc. might be instantiated with this built-up `dict` to write the final config to disk in the desired format):
```
import uwtools.api.config as config
cfgdict = config.realize_to_dict(input_config={"a": "apple"}, update_config={"b": "banana"})
```
Currently, this is the result:
```
  File "/home/maddenp/git/uwtools/src/uwtools/api/config.py", line 162, in realize_to_dict
    return _realize(**{**locals(), "output_file": Path(os.devnull), "output_format": None})
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: realize_config() got an unexpected keyword argument 'stdin_ok'
```
This is due to a bug in how `realize_to_dict()` is implemented. While fixing the proximate bug, however, I discovered a few other updates needed to support operations on in-memory `dict` objects. These are annotated via inline PR comments. After the updates:
```
>>> import uwtools.api.config as config
>>> cfgdict = config.realize_to_dict(input_config={"a": "apple"}, update_config={"b": "banana"})
>>> cfgdict
{'a': 'apple', 'b': 'banana'}
```

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -50,7 +50,7 @@ def get_fieldtable_config(
     :param stdin_ok: OK to read from ``stdin``?
     :return: An initialized ``FieldTableConfig`` object
     """
-    return _FieldTableConfig(config=_ensure_data_source(config, stdin_ok))
+    return _FieldTableConfig(config=_ensure_data_source(_str2path(config), stdin_ok))
 
 
 def get_ini_config(
@@ -64,7 +64,7 @@ def get_ini_config(
     :param stdin_ok: OK to read from ``stdin``?
     :return: An initialized ``INIConfig`` object
     """
-    return _INIConfig(config=_ensure_data_source(config, stdin_ok))
+    return _INIConfig(config=_ensure_data_source(_str2path(config), stdin_ok))
 
 
 def get_nml_config(
@@ -79,7 +79,7 @@ def get_nml_config(
     :param stdin_ok: OK to read from ``stdin``?
     :return: An initialized ``NMLConfig`` object
     """
-    return _NMLConfig(config=_ensure_data_source(config, stdin_ok))
+    return _NMLConfig(config=_ensure_data_source(_str2path(config), stdin_ok))
 
 
 def get_sh_config(
@@ -94,7 +94,7 @@ def get_sh_config(
     :param stdin_ok: OK to read from ``stdin``?
     :return: An initialized ``SHConfig`` object
     """
-    return _SHConfig(config=_ensure_data_source(config, stdin_ok))
+    return _SHConfig(config=_ensure_data_source(_str2path(config), stdin_ok))
 
 
 def get_yaml_config(
@@ -109,7 +109,7 @@ def get_yaml_config(
     :param stdin_ok: OK to read from ``stdin``?
     :return: An initialized ``YAMLConfig`` object
     """
-    return _YAMLConfig(config=_ensure_data_source(config, stdin_ok))
+    return _YAMLConfig(config=_ensure_data_source(_str2path(config), stdin_ok))
 
 
 def realize(
@@ -129,9 +129,9 @@ def realize(
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
     return _realize(
-        input_config=_ensure_data_source(input_config, stdin_ok),
+        input_config=_ensure_data_source(_str2path(input_config), stdin_ok),
         input_format=input_format,
-        update_config=_ensure_data_source(update_config, stdin_ok),
+        update_config=_ensure_data_source(_str2path(update_config), stdin_ok),
         update_format=update_format,
         output_file=_str2path(output_file),
         output_format=output_format,
@@ -178,7 +178,7 @@ def validate(
     :return: ``True`` if the YAML file conforms to the schema, ``False`` otherwise
     """
     return _validate_yaml(
-        schema_file=_str2path(schema_file), config=_ensure_data_source(config, stdin_ok)
+        schema_file=_str2path(schema_file), config=_ensure_data_source(_str2path(config), stdin_ok)
     )
 
 
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -124,14 +124,14 @@ def realize(
     total: bool = False,
     dry_run: bool = False,
     stdin_ok: bool = False,
-) -> None:
+) -> dict:
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
     input_config = (
         _YAMLConfig(config=input_config) if isinstance(input_config, dict) else input_config
     )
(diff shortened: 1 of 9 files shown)


src/uwtools/api/config.py
```diff
@@ -124,17 +124,14 @@ def realize(
     total: bool = False,
     dry_run: bool = False,
     stdin_ok: bool = False,
-) -> None:
+) -> dict:
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
-    input_config = (
-        _YAMLConfig(config=input_config) if isinstance(input_config, dict) else input_config
```
maddenp-noaa, line 132: It's no longer necessary to wrap a `dict` input in a `YAMLConfig` object, as `dict` objects are now accepted and handled correctly in the library.

src/uwtools/api/config.py
```diff
@@ -159,7 +158,7 @@ def realize_to_dict(  # pylint: disable=unused-argument
 
     See ``realize()`` for details on arguments, etc.
     """
-    return _realize(**{**locals(), "output_file": Path(os.devnull), "output_format": None})
+    return realize(**{**locals(), "output_file": Path(os.devnull), "output_format": _FORMAT.yaml})
```
maddenp-noaa, line 161: Note that
1. `key_path` and `total` are now accepted by `realize_to_dict()` as well, and passed through to `realize()`.
2. The output format is set to YAML because _some_ format needs to be available later in the library code, and YAML is an appropriate, generic format to associate with arbitrary `dict` objects, as there are no depth etc. concerns.

src/uwtools/config/tools.py
```diff
@@ -90,7 +90,7 @@ def realize_config(
     """
     NB: This docstring is dynamically replaced: See realize_config.__doc__ definition below.
     """
-    input_obj, input_format = _realize_config_input_setup(input_config, input_format)
```
maddenp-noaa, line 93: The returned `input_format` value was never used, so is no longer returned by `_realize_config_input_setup()` (see below).

src/uwtools/config/tools.py
```diff
@@ -127,11 +127,12 @@ def _ensure_format(
     """
     if isinstance(config, Config):
         return config.get_format()
+    if isinstance(config, Path):
+        return fmt or get_file_format(config)
+    if isinstance(config, dict):
+        return fmt or FORMAT.yaml
```
maddenp-noaa, line 133: Here's another case where, if we are operating on a plain `dict`, we associate it with the YAML format for convenience.

src/uwtools/tests/config/test_tools.py
```diff
@@ -607,9 +615,8 @@ def test__print_config_section_yaml_not_dict():
 def test__realize_config_input_setup_ini_cfgobj():
     data = {"section": {"foo": "bar"}}
     cfgobj = INIConfig(config=data)
-    input_obj, input_format = tools._realize_config_input_setup(input_config=cfgobj)
+    input_obj = tools._realize_config_input_setup(input_config=cfgobj)
```
maddenp-noaa, line 618: `_realize_config_input_setup()` no longer returns an input format value, so this is removed here and in multiple similar cases below.

src/uwtools/tests/utils/test_api.py
```diff
@@ -41,13 +41,6 @@ def test_ensure_data_source_stdin_ok():
     assert api.ensure_data_source(data_source=None, stdin_ok=True) is None
 
 
-def test_ensure_data_source_str_to_path():
-    val = "/some/path"
-    result = api.ensure_data_source(data_source=val, stdin_ok=False)
-    assert isinstance(result, Path)
-    assert result == Path(val)
```
maddenp-noaa, line 48: `_ensure_data_source()` no longer performs the `str` -> `Path` type conversion, so this test was obsolete.

src/uwtools/utils/api.py
```diff
@@ -5,31 +5,30 @@
...
 
+T = TypeVar("T")
+
 # Public
 
 
-def ensure_data_source(
-    data_source: Optional[Union[dict, Config, Path, str]], stdin_ok: bool
-) -> Any:
+def ensure_data_source(data_source: T, stdin_ok: bool) -> T:
```
maddenp-noaa, line 20: I think this is new in the `uwtools` codebase. Python's `TypeVar` provides behavior similar to so-called [generics](https://en.wikipedia.org/wiki/Generic_programming) in other languages. This says "I return a value that is of the same type as the `data_source` argument's value". This is a stronger statement than would be made by e.g.
```
def foo(val: Union[bool, int]) -> Union[bool, int]:
```
where, say, it would be fine for `foo()` to return an `int` value when its `val` argument was an `int`.

src/uwtools/tests/config/test_tools.py
```diff
@@ -520,16 +520,24 @@ def test_realize_config_values_needed_yaml(caplog):
...
-def test__ensure_format_bad():
+def test__ensure_format_bad_no_path_no_format():
     with raises(UWError) as e:
         tools._ensure_format(desc="foo")
-    assert str(e.value) == "Either foo file format or name must be specified"
+    assert str(e.value) == "Either foo path or format name must be specified"
 
 
 def test__ensure_format_config_obj():
     assert tools._ensure_format(desc="foo", config=YAMLConfig(config={})) == FORMAT.yaml
```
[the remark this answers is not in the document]
maddenp-noaa: Nice idea, will do.

src/uwtools/tests/config/test_tools.py
```diff
@@ -520,16 +520,24 @@ def test_realize_config_values_needed_yaml(caplog):
...
+def test__ensure_format_dict_explicit():
+    assert tools._ensure_format(desc="foo", fmt=FORMAT.yaml, config={}) == FORMAT.yaml
+
+
+def test__ensure_format_dict_implicit():
+    assert tools._ensure_format(desc="foo", config={}) == FORMAT.yaml
+
+
 def test__ensure_format_deduced():
     assert tools._ensure_format(desc="foo", config=Path("/tmp/config.yaml")) == FORMAT.yaml
```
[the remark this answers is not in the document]
maddenp-noaa: Updating.

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

WeirAE (approved):

PR #533, opened 2024-07-17
Added compare usage examples, sample files, and tests to example notebook

lexbuko22:
**Synopsis**

Added example 3 to the example Jupyter notebook for the config.compare() tool. Added the necessary example files to the fixtures folder. Added tests to the test file to test the outputs of the compare file.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


notebooks/fixtures/config_test_file_a.nml
```diff
@@ -0,0 +1,4 @@
+&values
+  greeting = "Hello"
+  recipient = "World"
+/
```
maddenp-noaa, line 4:
```suggestion
/
```
Here and below, adding suggestions to add a final newline to files.

notebooks/fixtures/config_test_file_b.nml
```diff
@@ -0,0 +1,4 @@
+&values
+  greeting = "Hello"
+  recipient = "Different World"
+/
```
maddenp-noaa, line 4:
```suggestion
/
```

notebooks/fixtures/config_test_file_c.nml
```diff
@@ -0,0 +1,4 @@
+&values
+  greeting = "Hello"
+  recipient = "World"
+/
```
maddenp-noaa, line 4:
```suggestion
/
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
     "cat ./fixtures/rendered_config.yaml"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ea5eff92",
+   "metadata": {},
+   "source": [
+    "## Example 3: Comparing two config files\n",
+    "Let's explore using the at using the `config.compare()` method to compare two two config files"
```
maddenp-noaa, line 209:
```suggestion
    "Let's explore using the `config.compare()` method to compare two two config files."
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "71d27692",
+   "metadata": {},
+   "source": [
+    "We again need to start by importing the `uwtools.api.config` from the uwtools python package."
```
maddenp-noaa, line 217: This may or may not be good for instructional purposes, but IIRC this shouldn't be necessary in a Jupyter notebook given that we've imported this already in a previous cell. Some instructional notebooks I've worked with have a cell near the top of the notebook that does a bunch of imports that will be needed to support cells that follow; I'm not necessarily asking for that, but it's something we could consider if we found that executing (or repeating) `import` statements as we go distracted from the story. Thoughts?

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+   "source": [
+    "from uwtools.api import config"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b0103f44",
+   "metadata": {},
+   "source": [
+    "The uwtools.api.config.compare() method has the following syntax:\n",
```
maddenp-noaa, line 235:
```suggestion
    "The `config.compare()` method has the following syntax:\n",
```
I think that, since we're doing `from uwtools.api import config`, we can thereafter ignore the `uwtools.api` part and treat `config` as the main namespace and concept, which cuts down on the verbosity a bit.

maddenp-noaa: Since we have this information already rendered nicely in our HTML docs -- here: https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html#uwtools.api.config.compare -- would it work to refer the reader to that, both to avoid repeating ourselves (and having to keep the information in this notebook in-sync with changes that might occur later), and to familiarize the notebook reader with the official docs? We might say something like
> Please review the [`config.compare()` documentation](https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html#uwtools.api.config.compare) for full information on this function's arguments.

maddenp-noaa (commented):
A few initial comments as I'm still reviewing...

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+    "\n",
+    "The method returns False if config files had differences, otherwise True."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b121dc29",
+   "metadata": {},
+   "source": [
+    "For example, lets compare the two nml files that contain different text."
```
maddenp-noaa, line 259:
```suggestion
    "For example, let's compare two Fortran namelist files with differences:"
```
1. It might help readers here to connect the `uwtools` designator `nml` with "Fortran namelist", something they might not already be aware of.
2. It's a subtle distinction but, with structured configuration formats like Fortran namelist, YAML, etc., it's helpful to keep in mind that they are not treated just as _text_ -- the `compare` tool understands their structure and can compare information in a way that takes the structure into account -- and that also ignores insignificant differences like ordering or whitespace. For these reasons, I think the word "text" here might suggest something that isn't pertinent.

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+       "False"
+      ]
+     },
+     "execution_count": 42,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# path of nml file that contains 'greeting = \"Hello\" recipient = \"World\"'\n",
```
maddenp-noaa, line 280: Rather than comments here stating the nature of the differences, what would you think about using the `%%bash` magic to display the files being compared? This could be done in a cell just before this one. It can be more powerful to show than to tell.

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+    "file_a = \"./fixtures/config_test_file_a.nml\"\n",
+    "\n",
+    "# path of nml file that contains 'greeting = \"Hello\" recipient = \"Different World\"'\n",
+    "different_file_b = \"./fixtures/config_test_file_b.nml\"\n",
+    "\n",
+    "config.compare(file_a,different_file_b,None,None)"
```
maddenp-noaa, line 286:
```suggestion
    "config.compare(file_a, different_file_b)"
```
The third and fourth arguments are optional, so we don't need to explicitly pass the `None` values in. Those arguments are only required when the format of the input files cannot be deduced from the extensions on their filenames, but we have `.nml` extensions here and so are good.

Also nitpick: Code style. It might be helpful to use `black` in a terminal to format code, e.g.
```
$ black --code "config.compare(file_a,different_file_b)"
config.compare(file_a, different_file_b)
```
Kind of a trivial example, I guess, but this could help to keep the code in the notebooks consistent with what we use internally in `uwtools`, which is IMO a clean, standard style. Well, it's the `black` style, which has become pretty popular.

If there are larger blocks, you could paste them into a temporary file and e.g.
```
$ black tmpfile
reformatted tmpfile

All done! ✨ 🍰 ✨
1 file reformatted.
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c16c38d0",
+   "metadata": {},
+   "source": [
+    "The config() method returns False to denote the files are different.\n",
```
maddenp-noaa, line 294:
```suggestion
    "The `config()` method returns `False` to denote the files are different.\n",
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+    "\n",
+    "Now to compare two files that contain identical text:"
```
maddenp-noaa, line 296:
```suggestion
    "Now to compare two semantically equivalent files:"
```
I think it would be a useful demonstration here to use two files that are _textually_ different but _semantically_ identical. For example, Fortran namelists are agnostic about whitespace and capitalization (outside of quoted strings), so given `ex3.3.nml`
```
&foo n=88, s="string" /
```
and `ex3.4.nml`
```
&FOO
  S = "string"
  N = 88
/
```
we get (I'm using the CLI here for demonstration purposes, but the CLI calls the same API code that we're demonstrating here):
```
(DEV-uwtools) ~/git/uwtools $ uw config compare --file-1-path ex3.3.nml --file-2-path ex3.4.nml
[2024-07-17T20:30:15]     INFO - ex3.3.nml
[2024-07-17T20:30:15]     INFO + ex3.4.nml
[2024-07-17T20:30:15]     INFO ---------------------------------------------------------------------
```
i.e. no differences.

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+    "\n",
+    "config.compare(file_a,identical_file_c,None,None)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b84db243",
+   "metadata": {},
+   "source": [
+    "The config() method returns True to denote the files are identical."
```
maddenp-noaa, line 331:
```suggestion
    "The `config()` method returns `True` to denote the files are semantically equivalent."
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,143 @@
...
+    }
+   ],
+   "source": [
+    "# path of nml file that contains 'greeting = \"Hello\" recipient = \"World\"'\n",
+    "file_a = \"./fixtures/config_test_file_a.nml\"\n",
+    "\n",
+    "# path of nml file that contains 'greeting = \"Hello\" recipient = \"World\"'\n",
+    "identical_file_c = \"./fixtures/config_test_file_c.nml\"\n",
+    "\n",
+    "config.compare(file_a,identical_file_c,None,None)"
```
maddenp-noaa, line 323:
```suggestion
    "config.compare(file_a, identical_file_c)"
```

notebooks/example.ipynb
maddenp-noaa, line 294: Another thing we could consider in these notebooks is enabling the `uwtools` logging, such that we get additional information when we use the tools. Logging can be enabled like this:
```
from uwtools.api.logging import use_uwtools_logger
use_uwtools_logger()
```
Then, for example:
```
>>> from uwtools.api.logging import use_uwtools_logger
>>> use_uwtools_logger()
>>> config.compare("ex3.1.nml", "ex3.2.nml")
[2024-07-17T20:44:29]     INFO - ex3.1.nml
[2024-07-17T20:44:29]     INFO + ex3.2.nml
[2024-07-17T20:44:29]     INFO ---------------------------------------------------------------------
[2024-07-17T20:44:29]     INFO foo:               n:  - 88 + 99
False
```
That's in a Python REPL, but behavior in a notebook cell should be similar. The input files `ex3.1.nml` and `ex3.2.nml` look like
```
&foo n=88, s="string" /
```
and
```
&FOO
  S = "string"
  N = 99
/
```
respectively.

notebooks/example.ipynb
elcarpenterNOAA, line 209:
```suggestion
    "Let's explore using the `config.compare()` method to compare two config files"
```

notebooks/example.ipynb
```diff
@@ -13,7 +13,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 45,
```
elcarpenterNOAA, line 16: I think we should stick with having this be a fresh run each time it is committed.

notebooks/example.ipynb
lexbuko22: My thought was if someone was jumping around the notebook looking for the specific example for the tool they were using, it shows a true start to finish? I see your point of it cluttering up the notebook, I am fine with removing it.

maddenp-noaa: I don't feel strongly about it, and you make a good point. Your call.

notebooks/example.ipynb
lexbuko22: Instead of updating these each time or changing them if examples are reordered, I think it may be best to delete all of the outputs of the cells. I think Paul recommended this earlier as a best practice as well. Are you fine with me making that change instead?

notebooks/example.ipynb
lexbuko22: Suggestion resolved in above conversation

notebooks/fixtures/config_test_file_a.nml
lexbuko22: Updated file to show textually different but semantically identical example you provided. Same for below.

notebooks/example.ipynb
maddenp-noaa: I think it might be the case that, if we delete the output cells, the nbviewer preview in binder may not have anything to show. That might be worth a test. Otherwise, I think clearing the cells and doing a _Run all cells_ might keep the numbers constant across commits and minimize the diffs.

notebooks/fixtures/config_test_file_a.nml
```diff
@@ -0,0 +1 @@
+&foo n=88, s="string" /
```
maddenp-noaa, line 1:
```suggestion
&foo n=88, s="string" /
```

notebooks/example.ipynb
```diff
@@ -182,24 +139,163 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "9d658b72",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# path of nml file that contains 'greeting = \"Hello\" recipient = \"World\"'\n",
```
maddenp-noaa, line 277: When all we have is source code (i.e. when we don't have notebooks), we must use code comments for commentary because we have no other choice. But code comments are distracting, and a selling point for notebooks is their ability to separate commentary and code into separate cells for better readability / storytelling. So I feel like we should have few to no code comments in these notbooks: Whatever we want to say, we can (and I think probably should) say in commentary, rather than code, cells. I'm sure there are edge cases where code comments might still be useful, but I think it would be good to leverage notebooks for what they're good at and keep the commentary and code separate. I realize that previous sections might also need an update along these lines.

notebooks/example.ipynb
```diff
@@ -182,24 +139,163 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8279b5c6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%bash\n",
+    "\n",
```
maddenp-noaa, line 204:
```suggestion
```
For consistency with earlier uses of `%%bash`, and compactness.

notebooks/example.ipynb
```diff
@@ -182,24 +139,163 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e3232247",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%bash\n",
+    "\n",
```
maddenp-noaa, line 216:
```suggestion
```

notebooks/example.ipynb
```diff
@@ -182,24 +139,163 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7fb40514",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from uwtools.api import config\n",
+    "from uwtools.api.logging import use_uwtools_logger\n",
+    "use_uwtools_logger()"
```
maddenp-noaa, line 177: Thanks for adding this! I'm executing the notebook locally, and I really like the way the output looks with the logging enabled.

maddenp-noaa (changes requested):
I like how this is looking. Almost there, I think, but I had a few additional, possibly final, comments.

notebooks/example.ipynb
elcarpenterNOAA: Paul's suggestion had missed the 2 twos, and it looks like that still needs to be updated.

lexbuko22: Yes it is! Removed one 'two'.

notebooks/example.ipynb
lexbuko22: Correct, it does look like it removes the nbviewer preview. I will Run All Cells, that should standardize the numbers in order.

notebooks/example.ipynb
lexbuko22: Removed the comments, these are also not necessary with the bash cell magic added.

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+    "## Example 3: Comparing two config files\n",
+    "Let's explore using the `config.compare()` method to compare two config files."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "71d27692",
+   "metadata": {},
+   "source": [
+    "We again need to start by importing the `uwtools.api.config` from the uwtools python package."
```
maddenp-noaa, line 217:
```suggestion
    "We again need to start by importing the `uwtools.api.config` from the `uwtools` python package."
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+      "&FOO\n",
+      "  S = \"string\"\n",
+      "  N = 99\n",
+      "/"
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "\n",
```
maddenp-noaa, line 287:
```suggestion
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+      ]
+     },
+     "execution_count": 11,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "file_a = \"./fixtures/config_test_file_a.nml\"\n",
+    "\n",
```
maddenp-noaa, line 320:
```suggestion
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+    "different_file_b = \"./fixtures/config_test_file_b.nml\"\n",
+    "\n",
```
maddenp-noaa, line 322:
```suggestion
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+      ]
+     },
+     "execution_count": 14,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "file_a = \"./fixtures/config_test_file_a.nml\"\n",
+    "\n",
```
maddenp-noaa, line 405:
```suggestion
```

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+    "identical_file_c = \"./fixtures/config_test_file_c.nml\"\n",
+    "\n",
```
maddenp-noaa, line 407:
```suggestion
```

maddenp-noaa (approved):
I left a few more click-to-commit suggesting removing unnecessary newlines in code cells, but apart from that this looks good to me.

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+    "from uwtools.api.logging import use_uwtools_logger\n",
+    "use_uwtools_logger()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b0103f44",
+   "metadata": {},
+   "source": [
+    "Please review the [config.compare() documentation](https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html#uwtools.api.config.compare) for full information on this function's arguments."
```
christinaholtNOAA, line 237: Do we also want to show users how to gain access to the same information here in the notebook?

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+    "\n",
+    "config.compare(file_a, identical_file_c)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b84db243",
+   "metadata": {},
+   "source": [
+    "The `config()` method returns `True` to denote the files are semantically equivalent."
```
christinaholtNOAA, line 416: Here `config` is the object. Maybe this should read "The `compare()` method...".

The same applies for the `False` case above.

notebooks/example.ipynb
```diff
@@ -200,6 +199,228 @@
...
+   "source": [
+    "Please review the [config.compare() documentation](https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html#uwtools.api.config.compare) for full information on this function's arguments."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b121dc29",
+   "metadata": {},
+   "source": [
+    "For example, let's compare two Fortran namelist files with differences:"
```
christinaholtNOAA, line 245: I'm curious if it would be easier to follow this multi-part example if we somehow emphasize this line, and the one before the comparison of semantically identical files. Maybe bold or larger font? Perhaps just making a third level under the headings used for the examples.

christinaholtNOAA (commented):
This looks good. Just some nit-picky comments below about style.

notebooks/example.ipynb
lexbuko22: Bolded both lines

notebooks/example.ipynb
lexbuko22: Corrected

notebooks/fixtures/config_test_file_c.nml
maddenp-noaa: @lexbuko22 It looks like you are resolving these comments without actually applying the suggestions. I'm unresolving them. Please ping us in Slack if you don't seem to be able to apply the suggestions via the web interface.

maddenp-noaa: Or maybe I haven't entered the suggestions correctly. I think you need to click _Commit suggestion_, or _Add suggestion to batch_ (and then later commit the batch, though I haven't tried this myself) instead of _Resolve conversation_. Apologies if that's what you've already done -- please let me know if so and I'll see if there's something wrong with the way I'm making commit-able suggestions.

notebooks/example.ipynb
```diff
@@ -200,6 +199,223 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "&foo n=88, s=\"string\" /"
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "\n",
```
maddenp-noaa, line 264:
```suggestion
    "%%bash\n",
```

maddenp-noaa (changes requested):

notebooks/fixtures/config_test_file_c.nml
lexbuko22: This was my mistake, I did the add suggestion to batch for each of these, but it looks like it only captured the ending "/" and not the file updates. I will get this corrected and recommit.

maddenp-noaa: All files should end with a newline, and the GitHub diff shows that some do not:
image
The newline is invisible in the diff, but I'd hope that committing the proposed suggestion will add the newline. Please try and let us know if it doesn't! (And if it doesn't you can add those newlines manually and push.)

maddenp-noaa (approved):

notebooks/example.ipynb
lexbuko22: Do you mean showing an example of the help(config.compare) or are you saying it is redundant and I should remove it? Or something else?

PR #534, opened 2024-07-17
UW-631 Provide mechanism to schema-check external drivers

NaureenBharwaniNOAA:
**Synopsis**
User supplied schemas are now required to be validated. We can do this by allowing external drivers to provide a JSON Schema file to validate the unique part of their configs. 

Utilizing existing functionality from the `schema_file` argument from the CLI that already returns the path to a schema file to use for validation and within `uwtools.config.validator` module an existing method,`validate_yaml`, validates a YAML config against the JSON Schema given the schema file. With both of these in action we were able to add the `schema_file` parameter to each of the Classes in the `uwtools.drivers.driver` module, and then add an if/else that does two checks:

1. If the `schema_file` parameter was passed from the user, we will validate an external driver via the JSON Schema file path provided to us.
2. If the `schema_file` was **_NOT_** passed from the user, then we will continue with an internal validation of a uwtools standalone driver, which will be schema checked by a JSON Schema in the `uwtools.resources.jsonschema` module.

**Type**
- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -515,16 +515,6 @@ class DriverCycleLeadtimeBased(Driver):
         batch: bool = False,
         schema_file: Optional[Path] = None,
     ):
-        """
-        The driver.
-
-        :param cycle: The cycle.
-        :param leadtime: The leadtime.
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param key_path: Keys leading through the config to the driver's configuration block.
-        :param batch: Run component via the batch system?
-        """
         super().__init__(
             cycle=cycle,
             leadtime=leadtime,
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -489,15 +489,6 @@ class DriverCycleBased(Driver):
         batch: bool = False,
         schema_file: Optional[Path] = None,
     ):
-        """
-        The driver.
-
-        :param cycle: The cycle.
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param key_path: Keys leading through the config to the driver's configuration block.
-        :param batch: Run component via the batch system?
-        """
         super().__init__(
             cycle=cycle,
             config=config,
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -272,13 +272,6 @@ class AssetsTimeInvariant(Assets):
         key_path: Optional[list[str]] = None,
         schema_file: Optional[Path] = None,
     ):
-        """
-        The driver.
-
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param key_path: Keys leading through the config to the driver's configuration block.
-        """
         super().__init__(
             config=config,
             dry_run=dry_run,
@@ -541,14 +534,6 @@ class DriverTimeInvariant(Driver):
         batch: bool = False,
         schema_file: Optional[Path] = None,
     ):
-        """
-        The driver.
-
-        :param config: Path to config file (read stdin if missing or None).
-        :param dry_run: Run in dry-run mode?
-        :param key_path: Keys leading through the config to the driver's configuration block.
-        :param batch: Run component via the batch system?
-        """
         super().__init__(
             config=config,
             dry_run=dry_run,
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -213,7 +213,7 @@ class Assets(ABC):
         """
         schema_name = self._driver_name.replace("_", "-")
         if schema_file:
-            validate_external(...)
+            validate_external(schema_file=schema_file, config=self._config)
         else:
             validate_internal(schema_name=schema_name, config=self._config)
 
@@ -460,7 +460,7 @@ class Driver(Assets):
         Perform all necessary schema validation.
         """
         if schema_file:
-            validate_external(...)
+            validate_external(schema_file=schema_file, config=self._config)
         else:
             for schema_name in (self._driver_name.replace("_", "-"), "platform"):
                 validate_internal(schema_name=schema_name, config=self._config)
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -47,6 +47,7 @@ class Assets(ABC):
         :param config: Path to config file (read stdin if missing or None).
         :param dry_run: Run in dry-run mode?
         :param key_path: Keys leading through the config to the driver's configuration block.
+        :param schema_file: Path to schema file to use for validation.
         """
         self._config = YAMLConfig(config=config)
         self._config.dereference(
@@ -229,6 +230,7 @@ class AssetsCycleBased(Assets):
         config: Optional[Union[dict, Path]] = None,
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
+        schema_file: Optional[Path] = None,
     ):
         """
         The driver.
@@ -237,8 +239,15 @@ class AssetsCycleBased(Assets):
         :param config: Path to config file (read stdin if missing or None).
         :param dry_run: Run in dry-run mode?
         :param key_path: Keys leading through the config to the driver's configuration block.
+        :param schema_file: Path to schema file to use for validation.
         """
-        super().__init__(cycle=cycle, config=config, dry_run=dry_run, key_path=key_path)
+        super().__init__(
+            cycle=cycle,
+            config=config,
+            dry_run=dry_run,
+            key_path=key_path,
+            schema_file=schema_file,
+        )
         self._cycle = cycle
 
 
@@ -254,6 +263,7 @@ class AssetsCycleAndLeadtimeBased(Assets):
         config: Optional[Union[dict, Path]] = None,
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
+        schema_file: Optional[Path] = None,
     ):
         """
         The driver.
@@ -263,9 +273,15 @@ class AssetsCycleAndLeadtimeBased(Assets):
         :param config: Path to config file (read stdin if missing or None).
         :param dry_run: Run in dry-run mode?
         :param key_path: Keys leading through the config to the driver's configuration block.
+        :param schema_file: Path to schema file to use for validation.
         """
         super().__init__(
-            cycle=cycle, leadtime=leadtime, config=config, dry_run=dry_run, key_path=key_path
+            cycle=cycle,
+            leadtime=leadtime,
+            config=config,
+            dry_run=dry_run,
+            key_path=key_path,
+            schema_file=schema_file,
         )
         self._cycle = cycle
         self._leadtime = leadtime
@@ -281,6 +297,7 @@ class AssetsTimeInvariant(Assets):
         config: Optional[Union[dict, Path]] = None,
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
+        schema_file: Optional[Path] = None,
     ):
         """
         The driver.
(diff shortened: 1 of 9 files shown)


src/uwtools/tests/drivers/test_driver.py
```diff
@@ -235,7 +236,7 @@ def test_Assets__rundir(assetsobj):
     assert assetsobj._rundir == Path(assetsobj._driver_config["rundir"])
 
 
-def test_Assets__validate(assetsobj):
+def test_Assets__validate_internal(assetsobj):
```
NaureenBharwaniNOAA, line 239: Test name updated from `_validate` to `validate_external`. Although this test case didn't change maybe updating the test name to remove this extra underscore would be appropriate. 
 test_Assets__validate_internal to  test_Assets_validate_internal

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -245,6 +246,17 @@ def test_Assets__validate(assetsobj):
         }
 
 
+def test_Assets__validate_external(config):
```
NaureenBharwaniNOAA, line 249: Same idea here with `validate_external`. Maybe updating the test name to remove this extra underscore would be appropriate. 
test_Assets__validate_external to  test_Assets_validate_external

NaureenBharwaniNOAA:
The ticket description for [UW-631](https://jira.epic.oarcloud.noaa.gov/browse/UW-631), included "tests and documentation are updated". Tests have been updated accordingly, and after taking a look at the documentation for the `--schema-file` flag for the [uwtools.api.config](https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html) I only saw one place where it would be appropriate to update changing

```
Parameters: schema_file – The JSON Schema file to use for validation
```

to 

```
Parameters: schema_file – The Path to JSON Schema file to use for validation
```

This looks to be more consistent with the descriptions we have of the parameter in other places. However, I wasn't able to find the spot to update this. If anyone has any ideas that'd be great. Besides that I don't see any additional changes needed in the documentation. I left this "unchecked" in the checklist for the PR for now.

src/uwtools/drivers/driver.py
```diff
@@ -410,12 +487,15 @@ def _scheduler(self) -> JobScheduler:
...
         """
         Perform all necessary schema validation.
         """
-        for schema_name in (self._driver_name.replace("_", "-"), "platform"):
-            validate_internal(schema_name=schema_name, config=self._config)
+        if schema_file:
+            validate_external(schema_file=schema_file, config=self._config)
+        else:
+            for schema_name in (self._driver_name.replace("_", "-"), "platform"):
+                validate_internal(schema_name=schema_name, config=self._config)
```
maddenp-noaa, line 498:
```suggestion
        if schema_file:
            validate_external(schema_file=schema_file, config=self._config)
        else:
            validate_internal(schema_name=self._driver_name.replace("_", "-"), config=self._config)
        validate_internal(schema_name="platform", config=self._config)
```
I think we always want to validate the `platform:` block against its `uwtools`-internal schema.

src/uwtools/tests/drivers/test_driver.py
maddenp-noaa: I think this new name is accurate: This test exists to test the `_validate()` method (see line 242). It's only incidental that it's interacting with the `validate_internal()` method, which we're mocking here. It's the behavior of `_validate()` -- namely, that it calls `validate_internal()` with the expected/correct arguments -- that this test is interested in.

I think I see where the name confusion might arise, though. This test could be described as
> Test `_validate()` with an internal schema.

which is not to be confused with a test of the `validate_internal()` function imported into this module. Would a more verbose name like `test_Assets__validate_with_internal_schema()` (and, similarly, below, `test_Assets__validate_with_external_schema()` help?

src/uwtools/tests/drivers/test_driver.py
maddenp-noaa: I think the name here is correct: We are (indirectly, via the `self._validate(schema_file)` in the `Assets` constructor, calling the `_validate()` method, and asserting that it then calls `validate_external()` with correct arguments. Since `validate_external()` (and `validate_internal()` in the previous test) has _its own_ unit tests, we know that it will work correctly if called correctly, so all we have to do here is assert that it is called correctly when it should be.

src/uwtools/drivers/driver.py
```diff
@@ -39,7 +39,18 @@ def __init__(
...
+        """
+        A component driver.
+
+        :param cycle: The cycle.
+        :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        :param schema_file: Path to schema file to use for validation.
+        """
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: Sounds good, I'll update this and remove the static docstrings.

src/uwtools/drivers/driver.py
```diff
@@ -515,6 +632,7 @@ def _add_docstring(class_: type, omit: Optional[list[str]] = None) -> None:
     :param dry_run: Run in dry-run mode?
     :param key_path: Keys leading through the config to the driver's configuration block.
     :param batch: Run component via the batch system?
+    :param schema_file: Path to schema file to use to validate an external driver.
```
NaureenBharwaniNOAA, line 635: @maddenp-noaa I added the `schema_file` parameter here to create the docstring dynamically. I also updated the test case in `test_driver` for the `_add_docstring` method. I didn't see the docstrings get added to each of the classes like I expected it would appear once I run the tests again and pushed to the PR. Is there something else I'm missing to get this to dynamically generate correctly?

maddenp-noaa: If this problem persists after the static docstrings in the classes above are removed, we can try to diagnose, but I think that should be done first.

NaureenBharwaniNOAA: I believe I removed all the places where I added the static doc strings. This is the only place I added it for the add_docstring method itself

src/uwtools/drivers/driver.py
```diff
@@ -39,7 +39,17 @@ def __init__(
...
     ) -> None:
+        """
+        A component driver.
+
+        :param cycle: The cycle.
+        :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        """
```
maddenp-noaa, line 52:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -218,8 +231,23 @@ def __init__(
...
     ):
-        super().__init__(cycle=cycle, config=config, dry_run=dry_run, key_path=key_path)
+        """
+        The driver.
+
+        :param cycle: The cycle.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        """
```
maddenp-noaa, line 243:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -235,9 +263,24 @@ def __init__(
...
     ):
+        """
+        The driver.
+
+        :param cycle: The cycle.
+        :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        """
```
maddenp-noaa, line 276:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -253,8 +296,21 @@ def __init__(
...
+        schema_file: Optional[Path] = None,
     ):
-        super().__init__(config=config, dry_run=dry_run, key_path=key_path)
+        """
+        The driver.
+
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        """
```
maddenp-noaa, line 307:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -270,9 +326,25 @@ def __init__(
...
+        The driver.
+
+        :param cycle: The cycle.
+        :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        :param batch: Run component via the batch system?
+        """
```
maddenp-noaa, line 340:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -448,9 +523,24 @@ def __init__(
...
     ):
+        """
+        The driver.
+
+        :param cycle: The cycle.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        :param batch: Run component via the batch system?
+        """
```
maddenp-noaa, line 536:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -468,14 +558,26 @@ def __init__(
...
+        The driver.
+
+        :param cycle: The cycle.
+        :param leadtime: The leadtime.
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        :param batch: Run component via the batch system?
+        """
```
maddenp-noaa, line 572:
```suggestion
```

src/uwtools/drivers/driver.py
```diff
@@ -492,8 +594,23 @@ def __init__(
...
     ):
-        super().__init__(config=config, dry_run=dry_run, key_path=key_path, batch=batch)
+        """
+        The driver.
+
+        :param config: Path to config file (read stdin if missing or None).
+        :param dry_run: Run in dry-run mode?
+        :param key_path: Keys leading through the config to the driver's configuration block.
+        :param batch: Run component via the batch system?
+        """
```
maddenp-noaa, line 606:
```suggestion
```

maddenp-noaa (changes requested):

src/uwtools/drivers/driver.py
maddenp-noaa: I just added some suggestion/comments showing what I think needs to be removed. Sorry if it wasn't clear previously, and I hope these help.

src/uwtools/drivers/driver.py
NaureenBharwaniNOAA: I see what you mean by removing **_all_** of these static docstrings. I actually only added the new additional `schema_file` parameter to the docstring and that is what I removed. I didn't add the block you suggested to remove, they were already existing for me when I did a `git pull` and created a new working branch.

maddenp-noaa: Yes, I figured something like that had happened. Almost there, only a couple more to go...

maddenp-noaa (approved):
Looks good! I think that this will be worth merging with one more approval, as @elcarpenter will need this to complete her UW-630 work.

elcarpenterNOAA (approved):
LGTM!

WeirAE (approved):
LGTM!

NaureenBharwaniNOAA:
> Looks good! I think that this will be worth merging with one more approval, as @elcarpenter will need this to complete her UW-630 work.

Interesting! I know originally the thought was UW-631 needed to be merged after UW-630. Which part did both of you discover would be great for UW-630 to use to complete the development?

maddenp-noaa:
@NaureenBharwaniNOAA The UW-631 code can now load, given user specs, an external task class from an external module. When that class is then instantiated, though, it will attempt to validate its config, which is also external -- but there's no code to load and use an external schema. So @elcarpenterNOAA has made progress by placing the external schema file in `src/uwtools/resources/jsonschema`, but really needs to test using a truly external file -- which your work provides for.

PR #537, opened 2024-07-22
UW-630: Allow uwtools to execute a standalone driver module

elcarpenterNOAA:
**Synopsis**
Standalone driver modules at arbitrary paths can now be executed. 

- A `uw execute` mode has been added to the Tools section (NB: this is the first mode that does not have a single action to go along with it). 
- `cycle` and `leadtime` are made optional because we cannot introspect the class before the command line arguments have been parsed. 
- `make_tasks()` has removed and replaced with calls to `partial()`
- A documentation ticket has been added to the backlog. 

Ticket questions:

- Does --help work? No, no it doesn't. 
- What happens if an external driver is named the same as one of ours? We execute the user-provided driver, not ours. 

**Type**

**Impact**

**Checklist**


src/uwtools/api/chgres_cube.py
```diff
@@ -2,12 +2,13 @@
...
-from uwtools.drivers.support import graph
+from uwtools.drivers.support import graph, tasks
 from uwtools.utils.api import make_execute as _make_execute
-from uwtools.utils.api import make_tasks as _make_tasks
 
 _driver = ChgresCube
 execute = _make_execute(_driver, with_cycle=True)
-tasks = _make_tasks(_driver)
+tasks = partial(tasks, _driver)
 __all__ = [_driver.__name__, "execute", "graph", "tasks"]
```
elcarpenterNOAA, line 14: This step is repeated in all internal drivers. `_make_tasks()` is made obsolete if we import `partial()` from `functools`.

src/uwtools/drivers/driver.py
```diff
@@ -307,6 +300,13 @@ def __init__(
 
     # Workflow tasks
 
+    @tasks
+    @abstractmethod
+    def provisioned_rundir(self):
+        """
+        Run directory provisioned with all required content.
+        """
+
```
elcarpenterNOAA, line 309: Relocated from the `Assets` base class to the `Driver` base class so that we could use the simplest driver in our tests.

src/uwtools/tests/utils/test_api.py
```diff
@@ -90,13 +90,6 @@ def test_make_execute_leadtime_no_cycle_error(execute_kwargs):
     assert "When leadtime is specified, cycle is required" in str(e)
 
 
-def test_make_tasks():
-    func = api.make_tasks(driver_class=TestDriver)
-    assert func.__name__ == "tasks"
-    taskmap = func()
-    assert list(taskmap.keys()) == ["atask", "run", "runscript", "validate"]
-
-
```
elcarpenterNOAA, line 99: `make_tasks()` was removed, so test is obsolete.

src/uwtools/utils/api.py
```diff
@@ -132,22 +132,6 @@ def execute_cycle_leadtime(  # pylint: disable=unused-argument
     return execute
 
 
-def make_tasks(driver_class: DriverT) -> Callable[..., dict[str, str]]:
-    """
-    Returns a function that maps task names to descriptions for the given driver.
```
elcarpenterNOAA, line 137: Made obsolete by `partial()`.

src/uwtools/api/driver.py
```diff
@@ -4,6 +4,103 @@
...
+from uwtools.logging import log
+from uwtools.utils.api import ensure_data_source
+
+
+def execute(  # pylint: disable=unused-argument
+    module: str,
+    classname: str,
+    task: str,
+    schema_file: str,
+    module_dir: Optional[Union[Path, str]] = None,
```
elcarpenterNOAA, line 24: New execute function for `uw invoke`. `cycle` and `leadtime` need to be optional arguments, since we have no way to presume the requirements of the user-made driver. We also needed to add `module`, `classname`, `module_dir` arguments that are absent in other `execute()` functions.

src/uwtools/api/driver.py
```diff
@@ -24,5 +121,34 @@ def _add_classes():
         __all__.append(classname)
 
 
-__all__: list[str] = []
+def _get_driver_class(
+    classname: str, module: str, module_dir: Optional[Union[Path, str]] = None
+) -> Optional[Type]:
```
elcarpenterNOAA, line 126: Here we find the module (either via the path provided or in the CWD), import it, and return the driver class.

src/uwtools/cli.py
```diff
@@ -351,6 +354,60 @@ def _dispatch_file_link(args: Args) -> bool:
...
+    _add_arg_leadtime(optional)
+    _add_arg_batch(optional)
+    _add_arg_dry_run(optional)
+    _add_arg_graph_file(optional)
+    _add_arg_key_path(
+        optional,
+        helpmsg="Dot-separated path of keys leading through the config "
+        "to the driver's configuration block",
+    )
+    return {STR.invoke: _add_args_verbosity(optional)}
```
christinaholtNOAA: Users won't be able to add additional args, huh?

src/uwtools/drivers/driver.py
maddenp-noaa: I'll add that it came up, in the context of the `CDEPS` driver, that some assets-only drivers may not have a single definition of "provisioned run directory", or even one that could be easily constructed with `if` statements. A user might just want to make multiple task calls on an assets-only driver to get what they need a la carte. But I think standalone drivers _can_ define a provisioned run directory: At the very least, it will contain a runscript. So this seems to me like a sound change.

src/uwtools/cli.py
```diff
@@ -679,6 +745,24 @@ def _add_arg_leadtime(group: Group, required: bool = False) -> None:
     )
 
 
+def _add_arg_module(group: Group) -> None:
+    group.add_argument(
+        _switch(STR.module),
+        help="Name of driver module",
```
christinaholtNOAA, line 751: How would the team feel about adding a bit extra help here to read "Name of driver module (the filename that holds the class, but without the py suffix)" or something like it?

I feel like especially in NWP the word module is overloaded and I've answered this question a lot of times when discussing with developers.

Interestingly, if I provide the full path here (relative and including the py suffix), it still works. Is that by design or accidental? Do we need the `--module-dir` arg?

```
...uwtools/src/uwtools/tests $ uw invoke --classname TestDriver --module ./fixtures/testdriver.py --task eighty_eight --schema-file ./fixtures/testdriver.jsonschema --cycle 2023-05-02 -c ./fixtures/simple.yaml --leadtime 20
[2024-07-22T08:26:56]     INFO 0 UW schema-validation errors found
[2024-07-22T08:26:56]     INFO 88: Initial state: Ready
```

src/uwtools/tests/fixtures/testdriver.yaml
```diff
@@ -0,0 +1,4 @@
+testdriver:
+  execution:
+    executable: "{{ testdriver.rundir }}/testdriver"
+  rundir: /path/to/rundir/
```
maddenp-noaa, line 4:
```suggestion
testdriver: null
```
Given that `TestDriver` doesn't actually use anything from its config, and that its schema supports any JSON object, maybe this can be simplified to avoid giving the impression that anything more complex is happening. I tried this suggestion on your branch and it seemed to be fine (all tests passed).

src/uwtools/tests/api/test_driver.py
```diff
@@ -1,25 +1,182 @@
...
+# Fixtures
+
+
+@fixture
+def args():
+    return ns(
+        classname="TestDriver",
+        config=fixture_path("testdriver.yaml"),
+        module="testdriver",
+        module_dir=fixture_path("testdriver.py").parent,
```
christinaholtNOAA, line 26: Out of curiosity, why have you gone with two separate flags for the CLI (module name and module dir) instead of just having the user provide the module path and treat it like you do here?

src/uwtools/tests/test_cli.py
```diff
@@ -161,6 +161,42 @@ def test__add_subparser_template_translate(subparsers):
...
+            key_path=["foo", "bar"],
+            dry_run=False,
+            config="/path/to/config",
+            module_dir="path/to/dir",
+            graph_file=None,
+            cycle=cycle,
+            leadtime=None,
+            batch=True,
+            stdin_ok=True,
+        )
```
maddenp-noaa, line 197:
```suggestion
    args: dict = {
        "module": "testdriver",
        "classname": "TestDriver",
        "schema_file": "/path/to/testdriver.jsonschema",
        "module_dir": "/path/to/dir",
        "batch": True,
        "config_file": "/path/to/config",
        "cycle": cycle,
        "leadtime": None,
        "dry_run": False,
        "graph_file": None,
        "key_path": ["foo", "bar"],
        "task": "eighty_eight",
        "stdin_ok": True,
    }
    with patch.object(cli.uwtools.api.driver, "execute") as execute:
        cli._dispatch_invoke(args=args)
        execute.assert_called_once_with(
            classname="TestDriver",
            module="testdriver",
            task="eighty_eight",
            schema_file="/path/to/testdriver.jsonschema",
            key_path=["foo", "bar"],
            dry_run=False,
            config="/path/to/config",
            module_dir="/path/to/dir",
            graph_file=None,
            cycle=cycle,
            leadtime=None,
            batch=True,
            stdin_ok=True,
        )
```
Added leading `/` to paths for consistency.

src/uwtools/api/driver.py
```diff
@@ -24,5 +121,34 @@ def _add_classes():
...
+    """
+    Returns the driver class.
+
+    :param classname: Class of driver object to instantiate.
+    :param module: Name of driver module.
+    :param module_dir: Path to directory that contains module.
+    """
+    module_dir = Path(module_dir) if module_dir else Path.cwd()
+    file_path = (module_dir / module).with_suffix(".py")
```
christinaholtNOAA, line 135: I like this python magic. :)

src/uwtools/cli.py
```diff
@@ -49,6 +50,7 @@ def main() -> None:
     try:
         setup_logging(quiet=True)
         args, checks = _parse_args(sys.argv[1:])
+        args[STR.action] = args[STR.action] if STR.action in args else args[STR.mode]
```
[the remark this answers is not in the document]
elcarpenterNOAA: I'm afraid not...

src/uwtools/api/driver.py
elcarpenterNOAA: I hadn't considered that inconsistency. In that case, I think we should use `uw execute` for this mode.

christinaholtNOAA (commented):
This looks good. It gave me plenty of good pointers when I inevitably butchered my first attempts at running the TestDriver manually. It "hung" once before I realized I didn't provide a config file. We might consider adding a troubleshooting section to our UG.

src/uwtools/cli.py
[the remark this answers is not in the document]
christinaholtNOAA: Is it bc there's no `get` method on a Namespace object?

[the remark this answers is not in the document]
maddenp-noaa: I want to circle back to this, but maybe we're missing something in `cli.py` because `uw -h` does't show the `invoke` mode. I think we had this issue with another mode once and agreed that integration / regression tests will help catch this. Unless I'm doing something wrong, I think we need to mention `STR.invoke` in another place to get it wired into the top-level CLI.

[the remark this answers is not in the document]
maddenp-noaa: I updated the suggestion above. The default value shouldn't be `STR.mode`, but `args[STR.mode]`, I think.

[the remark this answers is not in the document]
elcarpenterNOAA: Paul, I changed the name from `invoke` to `execute` and I see `execute` listed when I use `uw -h`.

src/uwtools/cli.py
maddenp-noaa: I think external drivers should conform to UW conventions, which define a set of supported arguments. Beyond that,  I think configuration should go in the YAML config. (I have ongoing misgivings about even `cycle`, `leadtime`, and `batch` due to the complications they've introduced. I know we want to support using the same base config for multiple cycles/leadtimes, but -- and this may not have always been the case -- that can be done, simply if a bit more verbosely, by pre-processing a driver config with `config realize`. It can even be done with pipes, so no intermediate files are necessary. That said, the flags _are_ handy, and if the CLI doesn't get further complicated, maybe we've dealt well enough with the complexity.)

christinaholtNOAA: Just making sure I was correct in this assumption. Thanks for expanding on the excellent reasons not to go beyond those flags we officially support.

src/uwtools/cli.py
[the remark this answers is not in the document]
maddenp-noaa: 🤦 of course. So fast though! Nice.

src/uwtools/cli.py
maddenp-noaa: Interesting. It might be a nice simplification to eliminate `--module-dir` and have just `--module` (and its API equivalent) and have the value be a module name, with or without the `.py` extension. As a user, I'd hope for my module load be loaded

1. If I give `/path/to/foo.py`, `../other/dir/foo.py`, `./foo.py`, etc. -- any explicit path should work.
2. If I give `foo.py` and `foo.py` exists in my current directory.
3. If I give `foo` and the path containing `foo.py` exists on `PYTHONPATH`.

I think our current code implements 1 and 2, except that it would not need to append `.py` because the user would supply it. For 3, I think we might just need a fallback to try `importlib.import_module()`, which consults `sys.path` / `PYTHONPATH` automatically.

@elcarpenterNOAA What do you think about this? If it also seems like good and simpler semantics to you, and if you want to talk about the implementation, please let me know.

maddenp-noaa: If we implemented that, the help message might read
> .py filename or path, or name of module on sys.path

I think most users will use the `.py` form, but IMO the other option, which honors the default `sys.path` as well as `PYTHONPATH`, should exist for some power-user use cases.

src/uwtools/api/driver.py
```diff
@@ -4,6 +4,103 @@
...
+    classname: str,
+    module: str,
+    module_dir: Optional[Union[Path, str]] = None,
+) -> dict[str, str]:
+    """
+    Returns a mapping from task names to their one-line descriptions.
+
+    :param classname: Class of driver object to instantiate.
+    :param module: Name of driver module.
+    :param module_path: Path to module file.
```
NaureenBharwaniNOAA, line 92:
```suggestion
    :param module_dir: Directory to module file.
```

Feel free to change the docstring if you don't think this accurately represents what's happening here. I believe the param name was updated, but the docstring was missed.

src/uwtools/api/driver.py
```diff
@@ -4,6 +4,103 @@
...
+    """
+    if not (class_ := _get_driver_class(classname, module, module_dir)):
+        msg = "Could not get tasks from class %s in module %s"
```
NaureenBharwaniNOAA, line 95: How does formatting work? I thought for filling in a string using `%s` syntax, there needed to be variables at the end of the string. 

I thought it would look more like this `msg = "Could not get tasks from class %s in module %s" % (var1, var2)`

src/uwtools/api/driver.py
```diff
@@ -4,6 +4,103 @@
...
+
+def tasks(
+    classname: str,
+    module: str,
+    module_dir: Optional[Union[Path, str]] = None,
+) -> dict[str, str]:
+    """
+    Returns a mapping from task names to their one-line descriptions.
+
+    :param classname: Class of driver object to instantiate.
```
NaureenBharwaniNOAA, line 90:
```suggestion
    :param classname: Name of driver class to instantiate.
```

Consistency with other docstrings in this module.

src/uwtools/api/driver.py
```diff
@@ -4,6 +4,103 @@
...
+    :param module: Name of driver module.
```
NaureenBharwaniNOAA, line 91:
```suggestion
    :param module: Name of driver module to load.
```

src/uwtools/cli.py
```diff
@@ -248,6 +251,60 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+    _add_arg_config_file(optional)
+    _add_arg_cycle(optional)
+    _add_arg_leadtime(optional)
+    _add_arg_batch(optional)
+    _add_arg_dry_run(optional)
+    _add_arg_graph_file(optional)
+    _add_arg_key_path(
+        optional,
+        helpmsg="Dot-separated path of keys leading through the config "
+        "to the driver's configuration block",
```
NaureenBharwaniNOAA, line 280: I realize this is also similar to the line [here](https://github.com/ufs-community/uwtools/blob/main/src/uwtools/cli.py#L902) in the `Support` methods.  

Should we update so it looks more closely to the `Mode Config` methods [here](https://github.com/ufs-community/uwtools/blob/main/src/uwtools/cli.py#L156)?

`_add_arg_key_path(optional, helpmsg="Dot-separated path of keys to the block to be output")`

maddenp-noaa (approved):
Looks great. If we've missed anything, we'll have opportunities for redress in the new sprint, as we'll be starting to try to extend this and put together a realistic case with a coupled driver.

NaureenBharwaniNOAA (approved):
Great work on this! LGTM!

PR #541, opened 2024-07-23
Validation fix

maddenp-noaa:
**Synopsis**

Update driver config schemas to require the driver's own block. Fix some prototype YAML files so that they validate against their corresponding drivers' schemas.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/docs/shared/ioda.yaml
+++ b/docs/shared/ioda.yaml
@@ -12,7 +12,6 @@ ioda:
       - module load some-module
       - module load ioda-module
     executable: /path/to/a/ioda/exe
-    mpicmd: time
   files_to_copy:
     d/f2: /path/to/f2
     f1: /path/to/f1
--- a/docs/shared/mpas.yaml
+++ b/docs/shared/mpas.yaml
@@ -1,8 +1,4 @@
 mpas:
-  boundary_conditions:
-    interval_hours: 6
-    offset: 0
-    path: /path/to/boundary/conditions/directory
   execution:
     batchargs:
       cores: 32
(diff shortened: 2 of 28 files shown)


docs/shared/ioda.yaml
```diff
@@ -12,7 +12,6 @@ ioda:
       - module load some-module
       - module load ioda-module
     executable: /path/to/a/ioda/exe
-    mpicmd: time
```
maddenp-noaa, line 15: Not allowed.

docs/shared/mpas.yaml
```diff
@@ -1,8 +1,4 @@
 mpas:
-  boundary_conditions:
-    interval_hours: 6
-    offset: 0
-    path: /path/to/boundary/conditions/directory
```
maddenp-noaa, line 5: Schema expects `lateral_boundary_conditions`. Updated.

docs/shared/mpas.yaml
```diff
@@ -58,7 +58,7 @@ mpas:
       mutable: true
       output_interval: "6:00:00"
       precision: single
-      reference_time: 2024-06-06 00:00:00
+      reference_time: "2024-06-06 00:00:00"
```
maddenp-noaa, line 61: Needs to be quoted, as YAML will otherwise parse it, strangely, as an integer.

src/uwtools/resources/jsonschema/make-solo-mosaic.jsonschema
```diff
@@ -10,7 +10,7 @@
               "type": "string"
             },
             "mosaic_name": {
-              "type": "integer"
+              "type": "string"
```
maddenp-noaa, line 13: Per the UW YAML docs, it seems like this should be a string.

src/uwtools/resources/jsonschema/make-solo-mosaic.jsonschema
```diff
@@ -21,7 +21,7 @@
             "periody": {
               "type": "integer"
             },
-            "tile_name": {
+            "tile_file": {
```
maddenp-noaa, line 24: The UW YAML docs (and the prototype YAML file) call this `file_file`.

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

WeirAE (approved):

docs/shared/ioda.yaml
christinaholtNOAA: This is done for every command, right?

christinaholtNOAA (approved):
LGTM!

docs/shared/ioda.yaml
maddenp-noaa: Only this one needed to change, as `ioda` uses the `execution-serial` schema. Other drivers use the full `execution` schema that _does_ allow `mpicmd`.

PR #546, opened 2024-07-26
Fixes for chgres_cube driver

maddenp-noaa:
**Synopsis**

Fixes what appear to be issues with the `chgres_cube` driver. See inline comments.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,12 +11,8 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 
 [*.py]
-# Use 80 for Docstrings and comments.
 indent_size = 4
 max_line_length = 100
-
-[docs/**.txt]
-max_line_length = 80
 
 [Makefile]
 indent_style = tab
--- a/.github/scripts/install-conda.sh
+++ b/.github/scripts/install-conda.sh
@@ -1,6 +1,6 @@
 set -eux
 source $(dirname ${BASH_SOURCE[0]})/common.sh
-url=https://github.com/conda-forge/miniforge/releases/download/23.1.0-4/Miniforge3-23.1.0-4-Linux-x86_64.sh
+url=https://github.com/conda-forge/miniforge/releases/download/24.3.0-0/Miniforge3-Linux-x86_64.sh
 installer=/tmp/$(basename $url)
 wget --no-verbose -O $installer $url
 bash $installer -bfp $CI_CONDA_DIR
--- a/docs/sections/user_guide/cli/drivers/chgres_cube.rst
+++ b/docs/sections/user_guide/cli/drivers/chgres_cube.rst
@@ -1,6 +1,8 @@
 ``chgres_cube``
 ===============
 
+.. include:: /shared/idempotent.rst
+
 The ``uw`` mode for configuring and running the :ufs-utils:`chgres_cube<chgres-cube>` component.
 
 .. literalinclude:: chgres_cube/help.cmd
--- a/docs/shared/idempotent.rst
+++ b/docs/shared/idempotent.rst
@@ -1,0 +1,3 @@
+.. note::
+
+   The ``uwtools`` drivers are *idempotent*, meaning that actions they successfully complete during one invocation are not repeated in subsequent invocations. For example, an asset like a configuration file will not be recreated when the driver is run again, even if its UW YAML configuration changes. To force recreation, remove the asset(s) in question -- up to and including the entire provisioned run directory -- then re-run the driver, which will recreate any missing assets based on the current configuration.
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -28,19 +28,31 @@ class ChgresCube(DriverCycleBased):
         yield self._taskname(f"namelist file {fn}")
         path = self._rundir / fn
         yield asset(path, path.is_file)
-        config_files = self._driver_config["namelist"]["update_values"]["config"]
-        input_paths = [
-            Path(config_files[k])
-            for k in ("mosaic_file_target_grid", "varmap_file", "vcoord_file_target_grid")
-        ] + [
-            Path(config_files["data_dir_input_grid"]) / config_files[k]
-            for k in ("atm_files_input_grid", "grib2_file_input_grid", "sfc_files_input_grid")
-            if k in config_files
-        ]
-        base_file = self._driver_config["namelist"].get("base_file")
-        yield [file(input_path) for input_path in input_paths] + (
-            [file(Path(base_file))] if base_file else []
-        )
+        input_files = []
+        namelist = self._driver_config["namelist"]
+        if base_file := namelist.get("base_file"):
+            input_files.append(base_file)
+        if update_values := namelist.get("update_values"):
+            config_files = update_values["config"]
+            for k in ["mosaic_file_target_grid", "varmap_file", "vcoord_file_target_grid"]:
+                input_files.append(config_files[k])
+            for k in [
+                "atm_core_files_input_grid",
+                "atm_files_input_grid",
+                "atm_tracer_files_input_grid",
+                "grib2_file_input_grid",
+                "nst_files_input_grid",
+                "orog_files_input_grid",
+                "orog_files_target_grid",
+                "sfc_files_input_grid",
+            ]:
+                if k in config_files:
+                    v = config_files[k]
+                    if isinstance(v, str):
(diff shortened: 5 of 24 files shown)


.editorconfig
```diff
@@ -11,12 +11,8 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 
 [*.py]
-# Use 80 for Docstrings and comments.
 indent_size = 4
 max_line_length = 100
 
-[docs/**.txt]
-max_line_length = 80
-
```
maddenp-noaa, line 20: Unrelated update: We don't have `.txt` files under `docs/` that we want formatting applied to, and we use 100 columns for comments and docstrings in Python code.

docs/sections/user_guide/cli/drivers/chgres_cube.rst
```diff
@@ -1,6 +1,8 @@
 ``chgres_cube``
 ===============
 
+.. include:: /shared/idempotent.rst
```
maddenp-noaa, line 4: Tangentially related: See below for the text of this note but, until we potentially implement something like the discussed-in-Slack `--reset` switch to allow drivers to recreate their run directory, users may benefit from being informed about the idempotent nature of the drivers and the fact that they'll need to remove items they want the driver to recreate.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -28,19 +28,31 @@ def namelist_file(self):
...
+                "orog_files_target_grid",
+                "sfc_files_input_grid",
+            ]:
+                if k in config_files:
+                    v = config_files[k]
+                    if isinstance(v, str):
+                        input_files.append(v)
+                    else:
+                        input_files += v
+        yield [file(Path(input_file)) for input_file in input_files]
```
maddenp-noaa, line 55: For files specified in the `chgres_cube` namelist, provide separate treatment for 1. Required single files (look up values with `[]` -- we expect the value to be available because the schema now requires it), and 2. Optional files that may be either single files or lists of files (look up with `.get()`).

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
```diff
@@ -37,21 +37,30 @@
                   },
                   "properties": {
                     "atm_core_files_input_grid": {
+                      "items": {
+                        "type": "string"
```
maddenp-noaa, line 41: Specify that `array` items must be of type `string`.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
```diff
@@ -197,6 +221,11 @@
                       "type": "string"
                     }
                   },
+                  "required": [
+                    "mosaic_file_target_grid",
+                    "varmap_file",
+                    "vcoord_file_target_grid"
```
maddenp-noaa, line 227: These were already looked up with `[]` in the driver code, so I've assumed that they are required and am enforcing that via this schema.

docs/shared/idempotent.rst
```diff
@@ -0,0 +1,3 @@
+.. note::
+
+   The ``uwtools`` drivers are *idempotent*, meaning that actions they successfully complete during one invocation are not repeated in subsequent invocations. For example, an asset like a configuration file will not be recreated when the driver is run again, even if its UW YAML configuration changes. To force recreation, remove the asset(s) in question -- up to and including the entire provisioned run directory -- then re-run the driver, which will recreate any missing assets based on the current configuration.
```
WeirAE, line 3: Do we want to perhaps put this up front in the base intro rather than updating all of the drivers with the same text?

maddenp-noaa: I thought about that, but don't think we can assume that a user who is looking for documentation for a specific driver will have bothered to read the more general docs: I suspect they'll drill down right to the thing they're interested in, and may not read the high-level background.

WeirAE: Fair - fortunately we've at least made the text referenceable so we can adjust later. Erring on the side of over-explaining is for the best here too.

WeirAE (approved):
Its a shame we couldn't keep the operator usage, but fortunately a quick fix!

src/uwtools/tests/test_schemas.py
```diff
@@ -387,55 +447,106 @@ def test_schema_cdeps_streams(cdeps_config, section):
...
-    assert not errors(config)
+    assert not errors(chgres_cube_config)
     # Some top-level keys are required:
     for key in ("execution", "namelist", "rundir"):
-        assert f"'{key}' is a required property" in errors(with_del(config, key))
+        assert f"'{key}' is a required property" in errors(with_del(chgres_cube_config, key))
     # Additional top-level keys are not allowed:
-    assert "Additional properties are not allowed" in errors({**config, "foo": "bar"})
+    assert "Additional properties are not allowed" in errors({**chgres_cube_config, "foo": "bar"})
+    # "rundir" must be a string:
```
NaureenBharwaniNOAA, line 459: This string doesn't make the test case below. The test case below is checking to see if "rundir" is present and not a string. Should we add an additional test case? Something along the lines of 
`assert not "'rundir' must be a string" in errors(chgres_cube_config, {"rundir": 88})`

```suggestion
    # "rundir" is a required property:
```

NaureenBharwaniNOAA (approved):
This looks great! Thanks for the new additions.

src/uwtools/tests/test_schemas.py
maddenp-noaa: Nice catch, thanks! I'm fixing this manually and adding the type test.

src/uwtools/tests/test_schemas.py
```diff
@@ -387,55 +447,106 @@ def test_schema_cdeps_streams(cdeps_config, section):
...
+    assert "88 is not of type 'string'" in errors(with_set(namelist, 88, "base_file"))
     # Just update_values is ok:
-    assert not errors(update_values)
+    assert not errors(with_del(namelist, "base_file"))
     # config is required with update_values:
-    assert "'config' is a required property" in errors({"update_values": {}})
+    assert "'config' is a required property" in errors(
+        with_del(namelist, "update_values", "config")
+    )
     # A combination of base_file and update_values is ok:
```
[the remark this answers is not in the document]
maddenp-noaa: The same assertion is already present in test function `test_schema_chgres_cube()` under `# Basic correctness`, so I'll remove this pointless duplicate.

.github/scripts/install-conda.sh
```diff
@@ -1,6 +1,6 @@
 set -eux
 source $(dirname ${BASH_SOURCE[0]})/common.sh
-url=https://github.com/conda-forge/miniforge/releases/download/23.1.0-4/Miniforge3-23.1.0-4-Linux-x86_64.sh
+url=https://github.com/conda-forge/miniforge/releases/download/24.3.0-0/Miniforge3-Linux-x86_64.sh
```
maddenp-noaa, line 3: Update Miniforge version used in CI.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
christinaholtNOAA: According to the [UFS_UTILS docs](https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#global-chgres-cube-namelist-options), `varmap_file` is only required for the grib2 input case. The other two in this required list are present in all three input format cases (grib2, nemsio, and netcdf). Perhaps we should remove that literal requirement here and in the code?

I did not follow up with a code deep-dive to confirm the docs are correct.

src/uwtools/tests/test_schemas.py
```diff
@@ -105,6 +105,66 @@ def cdeps_config():
...
+                    "vgfrc_from_climo": True,
+                    "vgtyp_from_climo": True,
+                    "wam_cold_start": True,
+                    "wam_parm_file": "v",
+                }
+            },
+            "validate": True,
+        },
+        "rundir": "/tmp",
+    }
```
christinaholtNOAA, line 165: Ugh.

src/uwtools/resources/jsonschema/chgres-cube.jsonschema
maddenp-noaa: I made `varmap_file` optional. Thanks for the fact check.

christinaholtNOAA (approved):
Thanks for that change!

PR #551, opened 2024-07-31
UW-632 Schema Injection

maddenp-noaa:
**Synopsis**

Provide a mechanism to support existing drivers requiring `rundir:` (`Assets` and all its subclasses) and `execution:` (`Driver` and all its subclasses) blocks in their UW YAML configs to inherit those from a block belonging to a coupled executable, by instantiating such constituent drivers with a new `coupler=` argument specifying the name of the coupled-executable block. See below for an example.

Note that the diffs in most of the changed files fall into two trivial categories:

1. Replacement of magic strings related to drivers with references to the `STR` dataclass.
2. Addition of calls to a new function to add constructor arguments to the docstrings for component drivers.

**Type**

- [x] Documentation

**Impact**

**Checklist**

Consider this UW YAML config, `coastal.yaml`, for a hypothetical coupled driver:
```
coastal:
  execution:
    executable: /path/to/coastal.exe
  rundir: /path/to/run/dir
cdeps:
  atm_in:
    update_values:
      datm_nml:
        datamode: GFS
        export_all: true
        factorfn_data: 'null'
        factorfn_mesh: 'null'
        flds_co2: false
        flds_presaero: false
        flds_wiso: false
        iradsw: 1
        model_maskfile: /path/to/mesh1.nc
        model_meshfile: /path/to/mesh1.nc
        nx_global: 100
        ny_global: 100
        restfilm: 'null'
  atm_streams:
    streams:
      stream01:
        dtlimit: 1.5
        mapalgo: redist
        readmode: single
        stream_data_files:
          - /path/to/data_file1.nc
        stream_data_variables:
          - temperature Sa_tbot
        stream_lev_dimname: 'null'
        stream_mesh_file: /path/to/mesh1.nc
        stream_offset: 0
        stream_vectors:
          - u
          - v
        taxmode: limit
        tinterpalgo: linear
        yearAlign: 2008
        yearFirst: 2008
        yearLast: 2008
    template_file: stream.jinja2
schism:
  namelist:
    template_file: /path/to/schism/param.nml.IN
    template_values:
      dt: 100
ww3:
  namelist:
    template_file: /path/to/ww3/ww3_shel.nml.IN
    template_values:
      input_forcing_winds: C
```
Ultimately, `coastal.exe` will be executed, but it will be linked to CDEPS, SCHISM, and WaveWatchIII code, so its driver will need to call on those constituent components' drivers to provision assets (namelist files, stream files) to the run directory. The `schism:` block does not contain the required `rundir:` entry, though, so will not pass validation when a hypothetical `Coastal` driver tries to instantiate a `SCHISM` driver object:
```
>>> schism_driver = SCHISM(cycle=dt.datetime(2024, 7, 31, 12), config="coastal.yaml")
1 UW schema-validation error found
Error at schism:
  'rundir' is a required property
...
uwtools.exceptions.UWConfigError: YAML validation errors
```
Since the `coastal:` block defines `rundir:`, the new `coupler=` argument can be provided to specify that run-time facts required by the `SCHISM` driver be obtained from that block:
```
>>> schism_driver = SCHISM(cycle=dt.datetime(2024, 7, 31, 12), config="coastal.yaml", coupler="coastal")
>>> schism_driver._rundir
PosixPath('/path/to/run/dir')
```
With an instantiated `SCHISM` driver in hand, the `Coastal` driver could call e.g. `schism_driver.namelist_file()` to execute the task to provision the SCHISM namelist file.


src/uwtools/config/formats/base.py
```diff
@@ -22,7 +23,7 @@ class Config(ABC, UserDict):
     several configuration-file formats.
     """
 
-    def __init__(self, config: Optional[Union[dict, Path]] = None) -> None:
+    def __init__(self, config: Optional[Union[dict, str, Path]] = None) -> None:
```
maddenp-noaa, line 26: Permit `str` or `Path` values for `config`.

src/uwtools/config/validator.py
```diff
@@ -47,37 +47,37 @@ def validate(schema: dict, config: dict) -> bool:
...
+    :param config: The config to validate.
     :raises: UWConfigError if config fails validation.
     """
 
     log.info("Validating config against internal schema %s", schema_name)
     schema_file = get_schema_file(schema_name)
     log.debug("Using schema file: %s", schema_file)
-    if not validate_external(config=config, schema_file=schema_file):
-        raise UWConfigError("YAML validation errors")
+    validate_external(config=config, schema_file=schema_file)
```
maddenp-noaa, line 63: `validate_external()` now itself raises a `UWConfigError` if validation fails.

src/uwtools/drivers/cdeps.py
```diff
@@ -127,3 +128,6 @@ def _model_stream_file(self, group: str, path: Path, template_file: str) -> None
             output_file=path,
             values_src=self._driver_config[group],
         )
+
+
+set_driver_docstring(CDEPS)
```
maddenp-noaa, line 133: This call attaches parameter descriptions from the parent class to the driver class' docstring. The pattern is repeated in each driver module.

Before:
image
After:
image

src/uwtools/drivers/chgres_cube.py
```diff
@@ -29,10 +30,10 @@ def namelist_file(self):
         path = self._rundir / fn
         yield asset(path, path.is_file)
         input_files = []
-        namelist = self._driver_config["namelist"]
-        if base_file := namelist.get("base_file"):
+        namelist = self._driver_config[STR.namelist]
```
maddenp-noaa, line 33: Here and in many cases that follow in the `uwtools.drivers` package, replace magic strings with references to `STR` object members. This change has been made where the strings refer to `uwtools` concepts; where strings refer to driver-specific concepts, they've been left as-is.

src/uwtools/drivers/driver.py
```diff
@@ -36,21 +37,25 @@ def __init__(
...
+                **({STR.cycle: cycle} if cycle else {}),
+                **({STR.leadtime: leadtime} if leadtime is not None else {}),
+                **self._config_full.data,
             }
         )
+        self._config = self._config_full
         for key in key_path or []:
             self._config = self._config[key]
+        if coupler:
+            self._config[self._driver_name][STR.rundir] = self._config_full[coupler][STR.rundir]
```
maddenp-noaa, line 58: If a `coupler` name was provided, set the driver config's `rundir:` item from the named block.

src/uwtools/drivers/driver.py
```diff
@@ -36,21 +37,25 @@ def __init__(
         self,
         cycle: Optional[datetime] = None,
         leadtime: Optional[timedelta] = None,
-        config: Optional[Union[dict, Path]] = None,
+        config: Optional[Union[dict, str, YAMLConfig, Path]] = None,
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
         schema_file: Optional[Path] = None,
+        coupler: Optional[str] = None,
```
maddenp-noaa, line 44: Accept an optional `coupler` argument naming a block in the UW YAML config from which the `rundir:` value can be taken.

src/uwtools/drivers/driver.py
```diff
@@ -282,11 +295,12 @@ def __init__(
         self,
         cycle: Optional[datetime] = None,
         leadtime: Optional[timedelta] = None,
-        config: Optional[Union[dict, Path]] = None,
+        config: Optional[Union[dict, str, YAMLConfig, Path]] = None,
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
         batch: bool = False,
         schema_file: Optional[Path] = None,
+        coupler: Optional[str] = None,
```
maddenp-noaa, line 303: Accept an optional `coupler` argument naming a block in the UW YAML config from which the `rundir:` and `execution:` values can be taken.

src/uwtools/drivers/driver.py
```diff
@@ -295,8 +309,13 @@ def __init__(
             dry_run=dry_run,
             key_path=key_path,
             schema_file=schema_file,
+            coupler=coupler,
         )
         self._batch = batch
+        if coupler:
+            self._config[self._driver_name][STR.execution] = self._config_full[coupler][
+                STR.execution
+            ]
```
maddenp-noaa, line 318: If a `coupler` name was provided, set the driver config's `execution:` item from the named block. A parent class' constructor will take care of `rundir:`.

src/uwtools/drivers/support.py
```diff
@@ -14,6 +17,22 @@ def graph() -> str:
...
 
 
+def set_driver_docstring(driver_class: Type) -> None:
+    """
+    Appends inherited parameter descriptions to the driver's own docstring.
+
+    :param driver_class: The class whose docstring to update.
+    """
+    head_old = driver_class.__doc__
+    body_old = driver_class.__mro__[1].__doc__
```
maddenp-noaa, line 27: `__mro__` is the "method resolution order" list: An ordered list of classes to consult to look for a method name. The class at index 0 is the current class, and index 1 is the parent class. The parent class, in this case, has the docstring with the parameter descriptions needed here.

docs/conf.py
```diff
@@ -12,6 +12,7 @@
 
 autoclass_content = "both"
 autodoc_mock_imports = ["f90nml", "iotaa", "jsonschema", "lxml", "referencing"]
+autodoc_typehints = "description"
```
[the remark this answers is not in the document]
christinaholtNOAA: I really appreciate this! It's nice that it does it for you with the right switch.

src/uwtools/config/validator.py
```diff
@@ -47,37 +47,37 @@ def validate(schema: dict, config: dict) -> bool:
...
 
 
 def validate_external(
     schema_file: Path, config: Union[dict, YAMLConfig, Optional[Path]] = None
```
christinaholtNOAA, line 67: Any reason not to have this function signature match `validate_internal`?

src/uwtools/drivers/driver.py
WeirAE: If I'm following correctly, `coupler` as used here just specifies the top level block name - first, is there a reason we have to specify it rather than just automatically pulling it from the YAML? Second, this is the name of the app, not anything involving component coupling, I think it would be misleading to use this variable name

src/uwtools/drivers/support.py
christinaholtNOAA: Would it be clearer to name this thing that many folks may never have heard of?

```
parent_class = driver_class.__mro__[1]
body_old = parent_class.__doc__
```

src/uwtools/drivers/driver.py
maddenp-noaa: > first, is there a reason we have to specify it rather than just automatically pulling it from the YAML?
I don't understand the question yet. What exactly would you pull from the YAML? Consider these [lines of code](https://github.com/ufs-community/uwtools/pull/551/files#diff-a26e1cd8178e1d78c4c251014c587627f1256afd357b0bf356f0f77fc9688eb2R57-R58): These are in the constructor for the `Assets` class, which will eventually be called when, for example, a coupled driver tries to instantiate a `SCHISM` driver object. From the perspective of `SCHISM`, how would it know which part of the YAML config to look at if it weren't told via this `coupler` argument? It knows its _own_ name, because it can call `self._driver_name` to get that, and so can locate its _own_ config block. But how would it know about the names of any other config blocks? This is a genuine question: I'm not assuming that the current code is the best answer, but I don't immediately see a simplification. Maybe you do!

> this is the name of the app, not anything involving component coupling, I think it would be misleading to use this variable name
It's involving component coupling because that's the only use case we have at the moment. And can we imagine another use case where, say, an `Assets`-type driver needs to be instantiated and asked to provision resources but has no idea what its run directory is? Or where a `Driver`-type driver doesn't know its own executable name? Those situations both arise in coupling; do they arise elsewhere? I'm definitely open to other names, though -- please share some ideas.

src/uwtools/config/validator.py
maddenp-noaa: Yes: `validate_internal` only knows the schema _name_; the path to the file is determined programmatically, based on `uwtools` internal conventions, via `importlib`, because the built-in `.jsonschema` files are package resources. So, `validate_internal` operates on a name, and `validate_external` operates on a path (file).

src/uwtools/drivers/jedi.py
```diff
@@ -8,6 +8,7 @@
 from iotaa import asset, refs, run, task, tasks
 
 from uwtools.drivers.jedi_base import JEDIBase
```
NaureenBharwaniNOAA, line 10: Question on this, I saw the other file was also modified for `src/uwtools/drivers/jedi_base.py`. Here we're importing the JEDIBase class and then setting the `set_driver_docstring(JEDI)` at the end of this file, but we don't do the same action for `jedi_base.py`. My thought process was the jedi driver here inherited from the jedi base class. 

Would it make sense to move the `set_driver_docstring(JEDI)` to the `JEDIBase` Class instead something like `set_driver_docstring(JEDIBase)`and have the `JEDI` Class inherit from it? If you have an explanation based on the choice that'd be great!

src/uwtools/rocoto.py
```diff
@@ -348,9 +348,7 @@ def _config_validate(self, config: Union[dict, YAMLConfig, Optional[Path]]) -> N
         :raises: UWConfigError if config fails validation.
         """
         schema_file = resource_path("jsonschema/rocoto.jsonschema")
-        ok = validate_yaml(schema_file=schema_file, config=config)
-        if not ok:
-            raise UWConfigError("YAML validation errors")
+        validate_yaml(schema_file=schema_file, config=config)
```
NaureenBharwaniNOAA, line 351: This is a great simplification! I was excited to see how this method would change with the update and it looks much cleaner now that `validate_external` which is imported as `validate_yaml` here raises an error as it's return value instead a bool now.

src/uwtools/drivers/driver.py
WeirAE: Module-wise, any time we use an assets-type driver, such as SCHISM or CDEPS, as part of an app, we'll need that information. This may be an issue in wording - coupled/coupling should only be used to define the relationship between two components in a given app. We have the convenience where our first apps only have a very specific coupling mode that is part of a common use, such as FV3-CMAQ for SRW-AQM or data-ATM in our current coastal app, but the coupling is not generally singularly defined by an app

WeirAE: For the simplification, ah, I've been pondering a best way to do that - should it perhaps know from the CLI instantiation? `uw coastal run` should know to look for a `coastal` block and store or pass that information....hm. I'll have to try some ideas

src/uwtools/drivers/jedi.py
maddenp-noaa: Good question. The `IODA` class has a docstring that provides the header line
```
A driver for the IODA component.
```
and the JEDI driver provides header line
```
A driver for the JEDI component.
```
The new `set_driver_docstring()` function, called at the bottom of each (non-base -- so not `jedi_base.py` or `mpas_base.py`) driver module assembles the driver's docstring by combining 1. The header line from the drive, like those shown above, with the parameter descriptions from the parent class. So I think it's necessary to do this on the final driver class, and not on the base class, as otherwise the appropriate header line will not be available.

src/uwtools/drivers/driver.py
maddenp-noaa: How/why NWP components are coupled is not our concern in `uwtools` -- flux isn't a concept we're concerned about in setting up files and directories on disk :) Here, currently, we're trying to solve the more prosaic problem of being able to validate a driver schema when it's missing information it's nominally supposed to have, because that information is provided elsewhere, and because we don't want to force our users to repeat the same information in multiple blocks just to pass a schema check.

> `uw coastal run` should know to look for a `coastal` block and store or pass that information
It does! That's exactly the point. A future `Coastal` driver would know its name -- `coastal` -- and could therefore pass the `coupler="coastal"` argument to the `CDEPS`, `SCHISM`, `WaveWatchIII` constructors, which could then get their `rundir` value from the `coastal` block. It would be the `Coastal` driver informing them to do so. A `Coastal` driver might have a `provisioned_rundir()` task method something like
```
@tasks
def provisioned_rundir(self):
    yield "Provisioned run directory"
    yield [
        cdeps_driver.atm_nml(),
        cdeps_driver.atm_stram(),
        schism_driver.namelist(),
        ww3_driver.namelist(),
    ]
```
Somewhere -- maybe in that task method, maybe elsewhere -- the `cdeps_driver` object needs to be instantiated:
```
cdeps_driver = CDEPS(cycle=<cycle>, config=<config>)
```
But `<config>` would not have `rundir:` defined in the `cdeps:` block -- because it is defined in the `coastal:` block. But `Coastal` can instantiate the CDEPS driver like
```
cdeps_driver = CDEPS(cycle=<cycle>, config=<config>, coupler=self._driver_name)
```
and the CDEPS constructor will now be able to validate its `cdeps:` block in the config because the `rundir:` value from `coastal:` will be added to it before validation occurs.

WeirAE: > How/why NWP components are coupled is not our concern
Then we probably shouldn't use the word `coupler`! Perhaps `parent` or `app`?

>  A future Coastal driver would know its name 
I think my original imagining of this YAML block would push the components down one level, like task->subtask, so the `rundir` would still be carried through. My second thought was more in instantiating the top block then keeping that information in memory, but that was a bad habit from the old design. If I come up with something cleaner than having to pass it like your CDEPS instantiation example later it would be nice, but I think this is ok for now

maddenp-noaa: > Perhaps `parent` or `app`?

Let's see what kind of feedback we get from other reviewers. Maybe we can settle on a good term.

maddenp-noaa: We should probably not read too much into the fact that I happened to call the block `coastal` and talked about a `Coastal` driver, and the fact that those terms relate to a UFS _app_. They could have been called `foo` and `Foo`. So, I think "app" is not the right term here; "parent" isn't bad.

The term "parent` and your comment "push the components down one level" give me an idea, though. If we structured the YAML like
```
cc: # i.e. coupled component
    execution:
        <block>
    rundir: <value>
    cdeps:
        <block>
    schism:
        <block>
    ww3:
        <block>
```

*and* if we mandated that, instead of this new `coupler=` argument, coupled drivers provide the existing `key_path=` argument when instantiating a subcomponent driver, we could implement logic like (pseudocode):

```
if I do not have a rundir:
    for each level above me in key_path:
        if that level has a rundir:
            set my rundir to its rundir
            break
    if i still do not have a rundir:
        panic
```

And similarly for the `execution` block for `Driver` drivers.

I think it could work, but I have some concerns:

1. It seems more implicit and explicit: It triggers behavior that users may not know they are requesting. It adds additional hidden meaning to `key_path`.
2. This logic would always be active (while in the current PR code it's controlled by an orthogonal-to-everything-else specific argument), and so would execute whenever `rundir` is missing and `key_path` is present so, as it searches up through parent blocks, it could potentially find a `rundir` that doesn't mean what we think it means.
3. The current PR code would allow you to have a e.g. `fv3` block that _does_ define `rundir`, such that it could be used independently by `uw fv3`. Its rundir would be overridden by the one from the block specified by `coupler=` for the coupled use case. We'd lose that possibiity.
4. The schema for `cc` would have to either a) permit any key directly under `cc`, or explicitly permit any driver names (`cdeps`, `fc3`, `schism`, etc.) that might be permitted there. We wouldn't be able to schema check these blocks immediately via their own `.jsonschema` schemas, because they wouldn't contain e.g. `rundir`, so validation fo the `cc` block would be provisional and could fail at run time when the subcomponents are instantiated and their constructors check their schemas. I guess that's kind of the case with the current PR code, too, though.

It's an interesting idea, but it seems riskier to me than the more explicit method we have here.

WeirAE: I wasn't sure if we were going to have this structure of multiple components called by another thing outside of an app... although I guess if we were running the weather model itself, that sort of ruins that thought. I think we still have a terminology problem, but aside from that I do like that structure more now that its spelled out, I do have to admit it would be riskier than the explicit method like you said though

WeirAE, line 44:
```suggestion
        parent: Optional[str] = None,
```

WeirAE (dismissed):

src/uwtools/drivers/driver.py
christinaholtNOAA: I am not a fan of the nested structure proposed as an alternative. I agree it's riskier and doesn't seem as readable as the original idea that everything is at the same level.

Let's consider the use case of in-line post for this. Does _coupler_ make sense there from a `uwtools` perspective (not a flux swap)? I don't love _parent_ and definitely do not support _app_. Both are overloaded terms in general, and the ufs-weather-model executable should definitely not be confused with an app at this point. 

All of these relationships are with respect to the component run by the ufs-weather-model unless I'm missing a significant use case. It's often referred to as an umbrella repo. Does something in that realm make sense? It also controls all the components -- does _controller_ make sense?

src/uwtools/config/validator.py
christinaholtNOAA: I apparently wrote a comment here and lost it another of the many tabs. :(

I was too general in my original comment. I was really just suggesting that we might follow the same pattern with the type hint for `config`:

```
config: Optional[Union[dict, YAMLConfig, Path]]
```

src/uwtools/drivers/jedi.py
christinaholtNOAA: Great question! I appreciated the answer, too. :)

src/uwtools/drivers/driver.py
WeirAE: Hmmm - ok, I can go with _controller_

maddenp-noaa: I'm happy enough with _controller_. Let's try that and see how it goes.

src/uwtools/config/validator.py
maddenp-noaa: Oh, I see now: Yes, the `Optional` could move outside for consistency. Doing that now. Thanks.

christinaholtNOAA (approved):
LGTM! Thanks for the changes.

WeirAE (approved):
Honestly, everything was so solid that naming was all I had for feedback! LGTM

NaureenBharwaniNOAA (approved):
This looks great to me! I see the link that was broken and caused the tests to fail works now. I assume this had to do with the discussion yesterday when the epic link was broken.

PR #552, opened 2024-07-31
Create required directories for CDEPS, SCHISM, WaveWatchIII

maddenp-noaa:
**Synopsis**

These drivers have tasks that call a Jinja2 `render()` function to create run-time assets based on input template files, but were not creating the necessary parent directories. I initially added `mkdir` statements in the drivers themselves, but eventually realized that these calls rely on the `writable()` context manager in `uwtools.utils.file`, so I added the `mkdir` there. This could help future use cases succeed, too. We probably now have some unnecessary explicit `mkdir` calls in our drivers or elsewhere, so I'll be on the lookout for those and try to remove them.

I also updated the unit tests to exercise the bug by defining the run directory for the affected drivers as `tmp_path / "run"` instead of just `tmp_path`: Since `pytest` ensures that `tmp_dir` exists, we didn't experience the bug; but the bug surfaces when an additional subdirectory is added to the run-directory path.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/config/test_jinja2.py
+++ b/src/uwtools/tests/config/test_jinja2.py
@@ -186,7 +186,7 @@ def test_register_filters_path_join(key):
 
 
 def test_render(values_file, template_file, tmp_path):
-    outfile = str(tmp_path / "out.txt")
+    outfile = tmp_path / "out.txt"
     expected = "roses are red, violets are blue"
     result = render_helper(input_file=template_file, values_file=values_file, output_file=outfile)
     assert result == expected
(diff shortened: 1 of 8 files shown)


src/uwtools/tests/config/test_jinja2.py
```diff
@@ -186,7 +186,7 @@ def test_register_filters_path_join(key):
 
 
 def test_render(values_file, template_file, tmp_path):
-    outfile = str(tmp_path / "out.txt")
+    outfile = tmp_path / "out.txt"
```
maddenp-noaa, line 189: Unnecessary, and downstream code expects a `Path` anyway. Ditto for the diff just below.

elcarpenterNOAA (approved):

PR #553, opened 2024-07-31
Accept str values for config, target_dir in file tool

maddenp-noaa:
**Synopsis**

1. In addition to the `copy()` and `link()` convenience functions we provide via `uwtools.api.file`, we now also provide direct access to the `Copier` and `Linker` objects, so that external drivers can instantiate them and use their tasks directly, e.g. `yield`ing them as dependency tasks in their own `iotaa` tasks. We already ensured that `str` values provided by users in `copy()` or `link()` calls are converted to the `Path` objects expected by `Copier` and `Linker`. But it will be convenient for users to be able to provide `str` values when they directly call the constructors of those classes. So, the necessary conversions are added.
2. We previously renamed `FileCopier` -> `Copier` (and similarly `FileLinker` -> `Linker`) to avoid the awkward repetition in e.g. `uwtools.file.FileLinker`. `FileStager` -> `Stager` is added here.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -23,7 +23,7 @@ lint() {
   msg Running linter
   (
     set -eux
-    pylint -j 4 .
+    pylint -j 4 setup.py uwtools
   )
   msg OK
 }
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -23,7 +23,7 @@ lint() {
   msg Running linter
   (
     set -eux
-    pylint -j 4 setup.py uwtools
+    pylint -j 4 .
   )
   msg OK
 }
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -23,7 +23,7 @@ lint() {
   msg Running linter
   (
     set -eux
-    pylint -j 4 .
+    pylint -j 1 .
   )
   msg OK
 }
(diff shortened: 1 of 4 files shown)


src/uwtools/api/file.py
```diff
@@ -36,7 +35,7 @@ def copy(
     """
     copier = Copier(
         target_dir=Path(target_dir) if target_dir else None,
-        config=_ensure_data_source(_str2path(config), stdin_ok),
+        config=_ensure_data_source(config, stdin_ok),
```
maddenp-noaa, line 38: The `str2path()` conversion is now provided in `Copier`, so we don't need to do it here. Ditto in the `Linker` constructor call below.

src/uwtools/file.py
```diff
@@ -42,8 +43,8 @@ def __init__(
         :raises: UWConfigError if config fails validation.
         """
         dryrun(enable=dry_run)
-        self._target_dir = target_dir
-        self._config = YAMLConfig(config=config)
+        self._target_dir = str2path(target_dir)
+        self._config = YAMLConfig(config=str2path(config))
```
maddenp-noaa, line 47: Do the `str2path()` conversion here now.

recipe/run_test.sh
```diff
@@ -23,7 +23,7 @@ lint() {
   msg Running linter
   (
     set -eux
-    pylint -j 4 .
+    pylint -j 1 .
```
maddenp-noaa, line 26: Trying to avoid an issue with tests failing in CI when parallelism is used.

elcarpenterNOAA (approved):

christinaholtNOAA (approved):
LGTM.

NaureenBharwaniNOAA (approved):

WeirAE (approved):

PR #554, opened 2024-08-01
Allow native with cores and nodes.

christinaholtNOAA:
**Synopsis**

The Rocoto native flag may be used to specify resources if no nodes/cores flags are available, but it may also be used alongside one of those for any variety of reasons. When the bug was found, the reservation flag was the culprit for finding the bug.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/resources/jsonschema/rocoto.jsonschema
+++ b/src/uwtools/resources/jsonschema/rocoto.jsonschema
@@ -290,6 +290,23 @@
     },
     "task": {
       "additionalProperties": false,
+      "anyOf": [
+        {
+          "required": [
+            "cores"
+          ]
+        },
+        {
+          "required": [
+            "native"
+          ]
+        },
+        {
+          "required": [
+            "nodes"
+          ]
+        }
+      ],
       "dependentSchemas": {
         "exclusive": {
           "not": {
@@ -328,23 +345,6 @@
           ]
         }
       },
-      "oneOf": [
-        {
-          "required": [
-            "cores"
-          ]
-        },
-        {
-          "required": [
-            "native"
-          ]
-        },
-        {
-          "required": [
-            "nodes"
-          ]
-        }
-      ],
       "properties": {
         "account": {
           "type": "string"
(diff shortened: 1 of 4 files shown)


src/uwtools/resources/jsonschema/rocoto.jsonschema
```diff
@@ -290,6 +290,23 @@
...
+          "required": [
+            "native"
+          ]
+        },
+        {
+          "required": [
+            "nodes"
+          ]
+        }
+      ],
```
christinaholtNOAA, line 309: I assume the formatter did this move when it was alphabetized?

src/uwtools/tests/test_schemas.py
```diff
@@ -1664,6 +1664,18 @@ def test_schema_rocoto_workflow_cycledef():
...
+def test_schena_rocoto_task_resources():
+    errors = schema_validator("rocoto", "$defs", "task", "properties")
+    # Basic resource options
+    assert not errors([{"cores": 1}])
+    assert not errors([{"native": "abc"}])
+    assert not errors([{"native": {"cyclestr": {"value": "def"}}}])
+    assert not errors([{"nodes": "1:ppn=12"}])
+    # Combined valid resources
+    assert not errors([{"cores": 1, "native": "abc"}])
+    assert not errors([{"nodes": "1:ppn=12", "native": "abc"}])
```
christinaholtNOAA, line 1676: I really wanted there to be an extra "don't allow cores and nodes" rule but do not have time to work out the logic with this bug fix. Maybe in a future PR?

NaureenBharwaniNOAA, line 1676:
```suggestion
    assert not errors([{"native": "abc", "nodes": "1:ppn=12"}])
```

I reordered these so they were in alphabetical ordering.

maddenp-noaa: Do you mean that we'd like to enforce that `cores` and `nodes` are mutually exclusive? I think that can be done in the schema and tested here. If you open a Jira ticket for it, it'd be a nice fill-in task.

maddenp-noaa (approved):

WeirAE (approved):

src/uwtools/tests/test_schemas.py
christinaholtNOAA: I think that is how we should go, yes. @WeirAE Does your 👍 mean that you're creating a Jira ticket?

PR #557, opened 2024-08-02
omnibus-2024-08-02

maddenp-noaa:
**Synopsis**

1. Simplify access to drivers' internal config. Previously, we had `.config`, `.config_full`, and `._driver_config` properties. `.config` provided the block including the driver's own block and the `platform` block (i.e. `{"driver-name": ..., "platform": ...}`); `.config_full` provided the full input config, even content not relevant to the driver; `._driver_config` provided the block under `driver-name`. This PR simplifies this: The private instance variable `._config_full` now provides the `{"driver-name": ..., "platform": ...}` block, and `.config_full` provides a public read-only copy of that; and the private instance variable `._config` now provides the block under `driver-name`, and `.config` provides a public read-only copy of that. So, `.config` replaces the old `._driver_config`, and `.config_full` replaces the old `.config`. The driver no longer retains a copy of the complete input config, discarding the irrelevant parts of that input. Many references to these properties are updated to make use of this simpler arrangement.

2. Update `uwtools.api.driver` to explicitly import the `Assets` and `Driver` class families, instead of dynamically adding these to the module. I found in starting external coupled-driver development that running `pylint` on that driver code flagged e.g. `from uwtools.api.driver import DriverCycleBased`, reporting that `uwtools.api.driver` has no `DriverCycleBased` member. From a static analysis perspective, this is true: The class is added to the module only when it is imported and its top-level code executed, so `pylint` cannot know this. Without this change, all external driver developers would have to suppress complaints from static analyzers like `pylint`.

3. Refactor the key-path walking and error-checking code to a public function, and call it from the `Assets` constructor so that key-path walking there benefits from error checking and reporting.

4. Add public `.cycle` and `.leadtime` properties on the appropriate driver classes. I see no reason to hide these, and where external drivers need to access these values, using e.g. `self.cycle` is IMO less surprising than `self._cycle`.

5. Remove some `.mkdir()` calls where they are no longer necessary when they rely on downstream code that already makes necessary parent directories.

6. I recently realized that `Path()` accepts multiple arguments, similar to the legacy `os.path.join`, so I've simplified some calls like `Path(a) / b` to `Path(a, b)`. It's only one character shorter, but IMO looks simpler and more familiar.

7. Fix some API docs.

See some inline comments, too.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/api/driver.py
```diff
@@ -91,25 +100,6 @@ def tasks(module: str, classname: str) -> dict[str, str]:
...
-    "DriverCycleLeadtimeBased",
-    "DriverTimeInvariant",
-]
-
-
-def _add_classes():
-    m = import_module("uwtools.drivers.driver")
-    for classname in _CLASSNAMES:
-        setattr(sys.modules[__name__], classname, getattr(m, classname))
-        __all__.append(classname)
```
maddenp-noaa, line 110: The dynamic addition of the driver classes to the module is no longer needed since they're being explicitly imported now.

src/uwtools/api/driver.py
```diff
@@ -161,5 +151,19 @@ def _get_driver_module_implicit(module: str) -> Optional[ModuleType]:
...
+        AssetsTimeInvariant,
+        Driver,
+        DriverCycleBased,
+        DriverCycleLeadtimeBased,
+        DriverTimeInvariant,
+        execute,
+        graph,
+        tasks,
+    )
+]
```
maddenp-noaa, line 169: The docs currently on `main` are not showing `execute()` or `tasks()`. This update should ensure that all the current module contents are included.

src/uwtools/config/tools.py
```diff
@@ -110,6 +110,29 @@ def realize_config(
...
+        keys.append(key)
+        pathstr = " -> ".join(keys)
+        try:
+            subconfig = config[key]
+        except KeyError as e:
+            raise log_and_error(f"Bad config path: {pathstr}") from e
+        if not isinstance(subconfig, dict):
+            raise log_and_error(f"Value at {pathstr} must be a dictionary")
+        config = subconfig
+    return config, pathstr
```
maddenp-noaa, line 133: This function was factored out of `_print_config_section()`, below. It has its own, new unit tests, and is also called from the `Assets` class constructor to better deal with its key-path-walking needs.

src/uwtools/drivers/cdeps.py
```diff
@@ -54,7 +53,7 @@ def atm_stream(self):
         yield self._taskname(f"stream file {fn}")
         path = self._rundir / fn
         yield asset(path, path.is_file)
-        template_file = self._driver_config["atm_streams"]["template_file"]
+        template_file = self.config["atm_streams"]["template_file"]
```
maddenp-noaa, line 56: A LOT of lines like this follow, so hopefully you can visually pattern-match on and ignore them after this. Old references any of `._driver_config`, `._confg`, `._config_full`, `.config`, or `.config_full` are updated to use the new versions (except `._driver_config`, which is removed) as described in the PR Description.

src/uwtools/drivers/driver.py
```diff
@@ -43,19 +45,21 @@ def __init__(
...
                 **({STR.cycle: cycle} if cycle else {}),
                 **({STR.leadtime: leadtime} if leadtime is not None else {}),
-                **self._config_full.data,
+                **config.data,
             }
         )
-        self._config = self._config_full
-        for key in key_path or []:
-            self._config = self._config[key]
+        self._config_full, _ = walk_key_path(config.data, key_path or [])
```
maddenp-noaa, line 56: After the input config is dereferences, walk the key path and point `self._config_full` to the block containing the driver-name and `platform` sub-blocks.

src/uwtools/drivers/driver.py
```diff
@@ -43,19 +45,21 @@ def __init__(
...
+        try:
+            self._config = self._config_full[self._driver_name]
+        except KeyError as e:
+            raise UWConfigError("Required '%s' block missing in config" % self._driver_name) from e
```
maddenp-noaa, line 60: Then point `self._config` to just the driver-name sub-block.

src/uwtools/drivers/driver.py
```diff
@@ -66,27 +70,24 @@ def __repr__(self) -> str:
...
         return self._driver_name
 
     @property
     def config(self) -> dict:
         """
-        Return a copy of the driver config.
+        A copy of the driver config.
         """
-        return deepcopy(self._driver_config)
+        return deepcopy(self._config)
```
maddenp-noaa, line 83: This property method (and `config_full()`, just below) are simpler and more consistent w.r.t. naming now.

src/uwtools/drivers/driver.py
```diff
@@ -123,24 +124,11 @@ def _create_user_updated_config(
...
-    def _driver_config(self) -> dict[str, Any]:
-        """
-        Returns the config block specific to this driver.
-        """
-        name = self._driver_name
-        try:
-            driver_config: dict[str, Any] = self._config[name]
-            return driver_config
-        except KeyError as e:
-            raise UWConfigError("Required '%s' block missing in config" % name) from e
```
maddenp-noaa, line 142: No longer needed: `self.config` provides this now.

src/uwtools/drivers/driver.py
```diff
@@ -263,6 +260,20 @@ def __init__(
...
+        The cycle.
+        """
+        return self._cycle
+
+    @property
+    def leadtime(self):
+        """
+        The leadtime.
+        """
+        return self._leadtime
```
maddenp-noaa, line 275: Here are new read-only properties for the `cycle` and `leadtime` attributes. One or both of them are added to driver classes where appropriate.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -141,56 +141,68 @@ def driverobj(config):
...
-    with patch.object(ConcreteAssetsTimeInvariant, "_validate", driver.Assets._validate):
-        with raises(UWConfigError):
-            ConcreteAssetsTimeInvariant(config=config, schema_file=controller_schema)
-        assert ConcreteAssetsTimeInvariant(
-            config=config, schema_file=controller_schema, controller="controller"
-        )
+def test_Assets_fail_bad_config(config):
+    with raises(UWConfigError) as e:
+        ConcreteAssetsTimeInvariant(config=config["concrete"])
+    assert str(e.value) == "Required 'concrete' block missing in config"
```
maddenp-noaa, line 150: There are some new tests in this module (e.g. for the new `cycle` and `leadtime` properties), but most of the diffs are reorg. There's no real behavior change.

src/uwtools/py.typed
maddenp-noaa, line 1: This file signals to `mypy` and other typecheckers that the `uwtools` package supports typing. This is useful for use cases like external drivers where, if the driver is also providing type annotations, the `uwtools` API-function return types can be checked for conformity.

src/uwtools/tests/config/test_tools.py
```diff
@@ -520,6 +520,23 @@ def test_realize_config_values_needed_yaml(caplog):
...
+
+def test_walk_key_path_fail_bad_leaf_value():
+    with raises(UWError) as e:
+        tools.walk_key_path({"a": {"b": {"c": "cherry"}}}, ["a", "b", "c"])
+    assert str(e.value) == "Value at a -> b -> c must be a dictionary"
+
+
+def test_walk_key_path_pass():
+    expected = ({"c": "cherry"}, "a -> b")
+    assert tools.walk_key_path({"a": {"b": {"c": "cherry"}}}, ["a", "b"]) == expected
```
maddenp-noaa, line 537: New tests for the factored-out `walk_key_path()`.

src/uwtools/tests/drivers/test_cdeps.py
```diff
@@ -87,7 +87,7 @@ def test_CDEPS_streams(driverobj, group):
     template_file = driverobj._rundir.parent / "template.jinja2"
     with open(template_file, "w", encoding="utf-8") as f:
         print(dedent(template).strip(), file=f)
-    driverobj._driver_config[f"{group}_streams"]["template_file"] = template_file
+    driverobj._config[f"{group}_streams"]["template_file"] = template_file
```
maddenp-noaa, line 90: Note that unit tests that need to modify a driver's config after it is instantiated should do so via the read-write `._config` / `._config_full` attributes, not the read-only `.config` / `.config_full` properties.

christinaholtNOAA (approved):
LGTM. Just some minor suggestions.

src/uwtools/api/driver.py
maddenp-noaa: I agree, but I don't see a way around it: The old solution, using a single list of strings and dynamically importing and then building up `__all__` from that list had the unfortunate downside of obscuring the fact that these classes are members of the module. If anyone can think of a DRYer solution that doesn't reintroduce the visibility problem, I'm definitely interested.

src/uwtools/drivers/driver.py
maddenp-noaa: I'm adding some code comments about these, and updating the docstrings for the corresponding property methods to be more explicit about what they're returning.

src/uwtools/api/driver.py
christinaholtNOAA: It's fine. Just generally voicing my dissent at Python/pylint over this.

src/uwtools/drivers/cdeps.py
maddenp-noaa: `.config` and `.config_full` were introduced on `main` 3 weeks ago and are not in a release yet, so no one should be relying on them. Are you aware of any use cases?

src/uwtools/api/driver.py
maddenp-noaa: If the issue was only internal to `uwtools`, I'd stick with the dynamic-add behavior and `# pylint: disable` the complaint. But I think the duplication and maintenance burden thereof are worth it for a better user experience. I think we will -- as any good tool / library developer does -- find ourselves dealing with some painful complexity internally to create a (hopefully!) magical user experience.

WeirAE (approved):
Finally got through everything, LGTM!

NaureenBharwaniNOAA (approved):
LGTM!

src/uwtools/drivers/driver.py
```diff
@@ -463,32 +476,36 @@ def _scheduler(self) -> JobScheduler:
     def _validate(self, schema_file: Optional[Path] = None) -> None:
         """
         Perform all necessary schema validation.
+
+        :param schema_file: The JSON Schema file to use for validation.
```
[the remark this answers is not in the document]
maddenp-noaa: I think one point I had the thought that user-facing code, docs, etc. shouldn't necessarily mention the specific validation technology, but maybe that's unavoidable, and it's certainly visible now. I'll keep this in mind and see if any improvements come to mind.

PR #558, opened 2024-08-05
Public driver rundir attr, redefine config_full

maddenp-noaa:
**Synopsis**

1. Make the `rundir` property on driver objects public. It's useful to access this attribute in the context of task methods and, while that use case doesn't conflict with a private-attribute designation, it seems awkward to me, and I can't see a reason to hide this piece of information. Almost all the diffs are updates to conform to this change.
2. Redefine the `config_full` attribute to be the full, dereferenced input config, which allows a e.g. coupled driver to pass it to a constituent driver's constructor. The `platform` config item, previously but no longer accessible via the `config_full` attribute, is now dealt with internally via a private attribute and does not need to be publicly exposed.

**Type**

**Impact**

Non-breaking because we have not yet released a `uwtools` exposing the `config_full` attribute.

**Checklist**

### DIFF
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -23,7 +23,7 @@ lint() {
   msg Running linter
   (
     set -eux
-    pylint -j 1 .
+    pylint -j 4 .
   )
   msg OK
 }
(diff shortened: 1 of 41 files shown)


elcarpenterNOAA (approved):
LGTM.

WeirAE (approved):
LGTM once the pylint bug gets squashed!

PR #573, opened 2024-08-15
Bug Fix for the template render tool

Byrnetp:
**Synopsis**

`template.render` and `template.render_to_str` currently return `None` when the `values_needed` parameter is set to `True`, which leads to an error being raised. This fix has the functions return the unrendered template instead. A unit test is added that fails for the old behavior and passes for the new behavior.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -170,7 +170,7 @@ def render(
     :param searchpath: Paths to search for extra templates.
     :param values_needed: Just report variables needed to render the template?
     :param dry_run: Run in dry-run mode?
-    :return: The rendered template, or None.
+    :return: The unrendered template if values_needed is True, the rendered template, or None.
     """
     _report(locals())
     values = _supplement_values(
@@ -180,11 +180,11 @@ def render(
     undeclared_variables = template.undeclared_variables
 
     # If a report of variables required to render the template was requested, make that report and
-    # then return.
+    # then return the unrendered template.
 
     if values_needed:
         _values_needed(undeclared_variables)
-        return None
+        return str(template)
 
     # Render the template. If there are missing values, report them and return an error to the
     # caller.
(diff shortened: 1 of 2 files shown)


maddenp-cu (approved):
Looks good, though I see that the CI formatting check has failed. Probably `make format && make test` in a development shell and then committing any changes made by that will overcome the failure.

maddenp-noaa:
Do we need to issue bugfix releases for any previous versions? I suppose we'll need at least a `2.4.1`, but if the underlying issue predates the `2.4.0` release, we should probably backport the fix. To test, you could create one-off environments with the latest versions from each `major.minor.x` series (`2.3.3`, `2.2.0`, `2.1.1`, `2.0.2`) with e.g. `conda create -y -n uwtools-2.3.3 -c ufs-community uwtools=2.3.3` and do a manual test to see if the bug is present at that version. If so, we should consider doing a bugfix release for each series. (It may be that the manual test fails in a different way because the code was different enough that we shouldn't expect the test to work, in which case we can probably ignore that series.) Others on the team can help with the actual releases, but a quick survey would be helpful if you have time to do one.

Byrnetp:
After testing with each an environment for each of the versions you mentioned, I found the same bug exists in versions `2.3.3` and `2.2.0`. Versions `2.1.1` and `2.0.2` returned `True` and log the needed values , but they also required the `values` parameter to be set or they returned a `TypeError`, where the later versions didn't require `values_src` to be specified. 

I took a look at the code in each of the version branches in this repository too, and it looks like the bug should exist in `2.1.1` as well: it looks like `None` is returned when `values_needed` is `True`. I'm not sure why in my environment testing `True` was returned instead.

christinaholtNOAA (approved):
LGTM.

maddenp-noaa:
@Byrnetp Nice work, thank you. I'll prep a `2.4.1` bugfix release and also see about backporting this fix to the earlier ones.

PR #575, opened 2024-08-16
Allow any UPP namelist values to be set.

christinaholtNOAA:
**Synopsis**

UPP Driver users need to more flexibility as they are adding new namelist settings. 

We may want to consider making this change to all drivers.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/resources/jsonschema/upp.jsonschema
+++ b/src/uwtools/resources/jsonschema/upp.jsonschema
@@ -31,7 +31,6 @@
               "type": "string"
             },
             "update_values": {
-              "additionalProperties": false,
               "properties": {
                 "model_inputs": {
                   "additionalProperties": false,


christinaholtNOAA:
I haven't added tests here just yet...I'd like to get feedback on whether we should open up all namelist settings in a similar way first.

maddenp-noaa:
If we have to do this for any specific driver, then we have to. The problem is that this weakens the namelist schema check significantly: Any misspelled key will now be accepted, and we won't be able to advise the user that they've made a mistake. What we previously settled on was: Use strict namelist schemas for components that are unlikely to be under active development, and leave actively-developed component namelist schemas more open. Is that no longer viable?

Out of curiosity, what was the immediate need for this change? Are there changes to UPP that aren't reflected in our schema?

christinaholtNOAA:
This change will significantly weaken the power of uwtools to catch problems with preparing a namelist. We chose in group to instead add the requested field for the GSL developer since it's likely to go into UPP at some point, and to move toward making changes that will allow a user to specify their own schema for uwtools drivers in the future.

PR #580, opened 2024-08-18
Move execute() to its own API namespace, disambiguate

maddenp-noaa:
**Synopsis**

1. Move `execute()`, `graph()`, and `tasks()` from `uwtools.api.driver` to a new `uwtools.api.execute`, to correspond better to the CLI mode name, making it easier for users to find the API docs.
2. Rename `execute()` in `uwtools.utils.processing` to `shellcmd()` to help disambiguate the term "execute".

This modifies the API and so is a breaking change.

**Type**

- [x] Documentation

**Impact**

**Checklist**


elcarpenterNOAA (approved):
Thank you!

src/uwtools/drivers/driver.py
```diff
@@ -25,7 +25,7 @@
 from uwtools.scheduler import JobScheduler
 from uwtools.strings import STR
 from uwtools.utils.file import writable
-from uwtools.utils.processing import execute
+from uwtools.utils.processing import shellcmd
```
christinaholtNOAA, line 28: I feel a little like I'm channeling Mike Kay for this one, but we moved from a verb to a noun for a function. I noticed a log message below moved from "Executing..." to "Running..." Could we consider `runshellcmd` for this one?

christinaholtNOAA (approved):
I'm so sorry....


If you hate it, leave it. This one is internal like we've already discussed, so not as big a deal.

src/uwtools/drivers/driver.py
maddenp-noaa: Good point. `runshellcmd` strikes me as somewhat clunky -- how about `run_shell_cmd`?

WeirAE (approved):
A necessary step!

src/uwtools/drivers/driver.py
maddenp-noaa: I went ahead with `run_shell_cmd`. If anyone hates it we can easily change it later -- we can now do an unambiguous search-and-replace for this function name since it doesn't conflict with `execute`.

PR #582, opened 2024-08-19
Config classes to API

maddenp-noaa:
**Synopsis**

Provide direct access to `Config` subclasses (`YAMLConfig` et al.) via `uwtools.api.config`.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/Makefile
```diff
@@ -2,7 +2,7 @@ BUILDDIR     = build
 LINKCHECKDIR = $(BUILDDIR)/linkcheck
 SOURCEDIR    = .
 SPHINXBUILD  = sphinx-build
-SPHINXOPTS   = -a -n -W
+SPHINXOPTS   = -a -n -W --keep-going
```
maddenp-noaa, line 5: This lets Sphinx collect and report all the errors/warnings it finds instead of halting after finding/reporting the first one.

docs/sections/user_guide/api/config.rst
```diff
@@ -2,4 +2,5 @@
 ======================
 
 .. automodule:: uwtools.api.config
+   :inherited-members: UserDict
```
maddenp-noaa, line 5: This tells Sphinx to produce documentation for methods inherited from parent classes, but to stop at and not report on methods inherited from `UserDict`. So, for example, the docs for `YAMLConfig` show usage for `derefernce()`, inherited from `Config` but do not delve deeper into the inheritance hierarchy as e.g. docs for `UserDict` are available [here](https://docs.python.org/3/library/collections.html#collections.UserDict).

src/uwtools/api/config.py
```diff
@@ -6,12 +6,12 @@
...
-from uwtools.config.formats.ini import INIConfig as _INIConfig
-from uwtools.config.formats.nml import NMLConfig as _NMLConfig
-from uwtools.config.formats.sh import SHConfig as _SHConfig
-from uwtools.config.formats.yaml import Config as _Config
-from uwtools.config.formats.yaml import YAMLConfig as _YAMLConfig
+from uwtools.config.formats.base import Config as _Config
+from uwtools.config.formats.fieldtable import FieldTableConfig
+from uwtools.config.formats.ini import INIConfig
+from uwtools.config.formats.nml import NMLConfig
+from uwtools.config.formats.sh import SHConfig
```
maddenp-noaa, line 13: Mark the imported `*Config` classes public.

src/uwtools/api/config.py
```diff
@@ -246,3 +246,20 @@ def validate(
...
+    "compare",
+    "get_fieldtable_config",
+    "get_ini_config",
+    "get_nml_config",
+    "get_sh_config",
+    "get_yaml_config",
+    "realize",
+    "realize_to_dict",
+    "validate",
+]
```
maddenp-noaa, line 265: This is necessary to instruct Sphinx to produce docs for imported classes as well as things defined in this module.

src/uwtools/config/formats/base.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
maddenp-noaa, line 1: This was causing a problem for Sphinx, as it led to type annotations being hidden from it at module-load time. The workaround, below, is to use the `UserDict` type instead of `Config`, which is correct in this case (the way the so-typed object is being used is appropriate for all `UserDict` instances, and instances of subclasses like `Config`).

src/uwtools/config/formats/base.py
```diff
@@ -25,8 +23,6 @@ class Config(ABC, UserDict):
 
     def __init__(self, config: Optional[Union[dict, str, Path]] = None) -> None:
         """
-        Construct a Config object.
```
maddenp-noaa, line 28: In a number of places here and below, docstring content that looked fine just in code looked odd, incomplete, etc. when rendered into RST docs. Be sure to review the rendered documentation in Read the Docs to see how all of this looks.

src/uwtools/config/formats/base.py
```diff
@@ -36,10 +32,10 @@ def __init__(self, config: Optional[Union[dict, str, Path]] = None) -> None:
         else:
             self._config_file = str2path(config) if config else None
             self.data = self._load(self._config_file)
-        if self.get_depth_threshold() and self.depth != self.get_depth_threshold():
+        if self._get_depth_threshold() and self._depth != self._get_depth_threshold():
```
maddenp-noaa, line 35: This is slightly awkward, as elsewhere where either this or another of the methods now marked private is accessed a `pylint: disable=protected-access` suppressor is now needed, but I did not like seeing these irrelevant-to-end-users methods appearing in the RST docs. I think the ends (the final RST documentation) justifies the means, but it's too bad that we must (I think) make some internal compromises for the best end-user experience.

src/uwtools/config/formats/base.py
```diff
@@ -50,6 +46,43 @@ def __repr__(self) -> str:
 
     # Private methods
 
+    def _characterize_values(self, values: dict, parent: str) -> tuple[list, list]:
```
maddenp-noaa, line 49: Here and below, several formerly-public methods now marked private as they're likely to mystify end users if they appear in docs.

src/uwtools/config/formats/fieldtable.py
```diff
@@ -52,37 +65,25 @@ def dump_dict(cls, cfg: dict, path: Optional[Path] = None) -> None:
         """
         Dumps a provided config dictionary in Field Table format.
 
-        FMS field and tracer managers must be registered in an ASCII table called 'field_table'.
+        FMS field and tracer managers must be registered in an ASCII table called ``field_table``.
```
maddenp-noaa, line 68: We may find that we need yet more RST syntax in docstrings that are now being rendered to RtD. I got the obvious ones, but be on the lookout when perusing the RtD docs for more cases where formatting could look nicer.

src/uwtools/config/formats/sh.py
```diff
@@ -12,17 +12,15 @@
 
 class SHConfig(Config):
     """
-    Concrete class to handle bash config files.
+    Work with key=value shell configs.
```
maddenp-noaa, line 15: We probably should mention `bash` specifically, as this is also useful for `sh`, `ksh`, `zsh`, and other Bourne-family shells.

src/uwtools/config/formats/yaml.py
```diff
@@ -48,11 +48,20 @@
...
     # Private methods
 
+    @classmethod
+    def _add_yaml_representers(cls) -> None:
+        """
+        Add representers to the YAML dumper for custom types.
+        """
+        yaml.add_representer(UWYAMLConvert, UWYAMLConvert.represent)
+        yaml.add_representer(Namelist, cls._represent_namelist)
+        yaml.add_representer(OrderedDict, cls._represent_ordereddict)
```
maddenp-noaa, line 63: Repositioned.

src/uwtools/config/formats/sh.py
WeirAE: Hmm, we did call it `SHConfig`, maybe we should be more explicit about the valid shells while we're at it?

src/uwtools/config/formats/base.py
WeirAE: Thats an interesting point - now that you mention it, I'm surprised we actually left these public at first, but a necessary sacrifice

WeirAE (approved):
Some fussy Sphinx rules and swapping of public/private methods, but worth the trouble - LGTM

src/uwtools/config/formats/sh.py
maddenp-noaa: All that matters is that the shell understands `key=value` variable assignment. There are a LOT of shells that this applies to -- too many to usefully try to enumerate here, I think. It's even potentially usable with a number of non-shell languages, e.g.
```
$ cat vars.py
a="apple"
pi=3.14
$ python -c "import vars; print(vars.a, vars.pi)"
apple 3.14
```

WeirAE: Oh, certainly. I mean "...bash and other..." or "...any Bourne shell..." to emphasize that the 'sh' naming is more inclusive, but its not really necessary and probably over-complicating

christinaholtNOAA (approved):
I was having a bit of trouble tracking the reorg vs changes in the tests, but I _think_ it looks good.

PR #585, opened 2024-08-21
Docstring cleanup

maddenp-noaa:
**Synopsis**

While working on something else, I became distracted by divergences from conventions in our docstrings. I've tried to harmonize those in this PR, though I'm sure I've missed some stragglers that, if we agree on the conventions described below, we can chip away at in future work as we find them.

The conventions I've tried to adhere to -- which are the case in most of our docstrings already:

- For action-oriented functions/methods (most of them), use active language: "Do a thing", not "Does a thing", "Return a value", not "Returns a value", etc.
- For language objects that _are_ rather than _do_ -- classes, `@property` methods, driver workflow tasks, etc. -- simply state what the thing is: "The valid names", not "Return the valid names".
- For functions/methods that exist mainly to produce and return a value, it's ok to omit the `:return:` docstring item if the docstring's first line starts with "Return ...". In some places, having both is useful, but see the previous bullet for cases when neither is necessary.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/yaml.py
+++ b/src/uwtools/config/formats/yaml.py
@@ -88,7 +88,7 @@ class YAMLConfig(Config):
 
     def _load(self, config_file: Optional[Path]) -> dict:
         """
-        Reads and parses a YAML file.
+        Read and parse a YAML file.
 
         See docs for Config._load().
 
@@ -124,6 +124,7 @@ class YAMLConfig(Config):
 
         :param dumper: The YAML dumper.
         :param data: The f90nml Namelist to serialize.
+        :return: A YAML mapping.
         """
         namelist_dict = data.todict()
         return dumper.represent_mapping("tag:yaml.org,2002:map", namelist_dict)
@@ -137,6 +138,7 @@ class YAMLConfig(Config):
 
         :param dumper: The YAML dumper.
         :param data: The OrderedDict to serialize.
+        :return: A YAML mapping.
         """
 
         return dumper.represent_mapping("tag:yaml.org,2002:map", from_od(data))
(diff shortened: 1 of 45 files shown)


elcarpenterNOAA (approved):

src/uwtools/scheduler.py
```diff
@@ -181,20 +181,23 @@ def _managed_directives(self) -> dict[str, Any]:
...
-        Returns the scheduler's resource-request prefix.
+        The scheduler's resource-request prefix.
         """
         return "#BSUB"
 
     @property
     def _processed_props(self) -> dict[str, Any]:
+        """
+        Pre-processed runscript directives,
+        """
```
christinaholtNOAA, line 192: I'm curious how this one passed the linter...

christinaholtNOAA (approved):
This looks great! 

I like the idea of reflecting the nature of the method/function/property in the way we describe it. It's pretty consistent with naming methods/variables etc based on their part of speech (noun vs verb).

NaureenBharwaniNOAA (approved):

src/uwtools/config/formats/yaml.py
```diff
@@ -187,9 +189,7 @@ def dump_dict(cls, cfg: dict, path: Optional[Path] = None) -> None:
 
 def _write_plain_open_ended(self, *args, **kwargs) -> None:
     """
-    Write YAML without ...
-
-    end-of-stream marker.
+    Write YAML without the "..." end-of-stream marker.
```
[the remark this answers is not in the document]
maddenp-noaa: I think `docformatter` was interpreting the unquoted `...` as meaningful in some way.

WeirAE (approved):

src/uwtools/scheduler.py
maddenp-noaa: https://github.com/pylint-dev/pylint/issues/3177#issuecomment-539981952 -- which was news to me, but testing it out it seems to be true.

PR #587, opened 2024-08-21
GH 586 bad stdin_ok prompt

### LINKED ISSUE #586: `uwtools.api.config.realize` incorrectly prompts me to use `stdin_ok=True` flag when writing out from a dictionary or config object
**Expected behavior**
`uwtools.api.config.realize` should accept a dictionary (or a config object) for writing out a file. 

**Current behavior**
While this does appear to work, you get a spurious error that prompts you to include the `stdin_ok=True` flag, which doesn't make sense in the Python API context.

**Machines affected**
All. This was with uwtools release version 2.4.1.

**To Reproduce**
Here is a minimal reproducible example:

```
#!/usr/bin/env python3
  
"""
Example program demonstrating issue
"""

import uwtools.api.config as uwconfig

dict_to_write = {}
dict_to_write["test_field"]="test value"
uwconfig.realize(input_config=dict_to_write,output_file="test_output.yaml")
```

Run the above script with uwtools in your environment to reproduce:

```
$ python uw_issue_example.py 
  File "/scratch1/BMC/hmtb/HAFS/add_vx_workflow/HAFS/ush/uw_issue_example.py", line 11, in <module>
    uwconfig.realize(input_config=dict_to_write,output_file="test_output.yaml")
  File "/scratch1/BMC/hmtb/HAFS/add_vx_workflow/HAFS/ush/conda/envs/hafs_vx/lib/python3.12/site-packages/uwtools/api/config.py", line 131, in realize
    update_config=_ensure_data_source(_str2path(update_config), stdin_ok),
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch1/BMC/hmtb/HAFS/add_vx_workflow/HAFS/ush/conda/envs/hafs_vx/lib/python3.12/site-packages/uwtools/utils/api.py", line 31, in ensure_data_source
    raise UWError("Set stdin_ok=True to permit read from stdin")
uwtools.exceptions.UWError: Set stdin_ok=True to permit read from stdin
```

Include the `stdin_ok=True` flag and you will get the expected output. But this flag should not be required from the API.

maddenp-noaa:
**Synopsis**

Fixes #586

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -125,14 +125,12 @@ def realize(
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
-    update_config = _str2path(update_config)
-    assert not isinstance(update_config, str)
     if update_config is None and update_format is not None:  # i.e. updates will be read from stdin
         update_config = _ensure_data_source(update_config, stdin_ok)
     return _realize(
         input_config=_ensure_data_source(_str2path(input_config), stdin_ok),
         input_format=input_format,
-        update_config=update_config,
+        update_config=_str2path(update_config),
         update_format=update_format,
         output_file=_str2path(output_file),
         output_format=output_format,
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -125,6 +125,8 @@ def realize(
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
+    if update_config is None and update_format is None:
+        update_config = {}
     return _realize(
         input_config=_ensure_data_source(_str2path(input_config), stdin_ok),
         input_format=input_format,
--- a/src/uwtools/api/config.py
+++ b/src/uwtools/api/config.py
@@ -125,12 +125,14 @@ def realize(
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
-    if update_config is None and update_format is None:
-        update_config = {}
+    update_config = _str2path(update_config)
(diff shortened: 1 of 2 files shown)


src/uwtools/api/config.py
```diff
@@ -125,10 +125,12 @@ def realize(
     """
     NB: This docstring is dynamically replaced: See realize.__doc__ definition below.
     """
+    if update_config is None and update_format is not None:  # i.e. updates will be read from stdin
+        update_config = _ensure_data_source(update_config, stdin_ok)
```
maddenp-noaa, line 129: Previously, with the call to `_ensure_data_source()` inline on line 131, the `update_config` value was *always* being checked so that, if it was `None` and `stdin_ok` was `False` (the default), an exception would be raised. But it's completely valid for no update values to be provided because no update is desired, in which case the exception would be incorrect.

In order for `update_config` to be read from `stdin`, `update_format` must be provided, as `uwtools` would not otherwise know how to parse the stream. So, the presence of `update_format` implies that, if `update_config` isn't provided, the user wants to read it from `stdin`, and those criteria can be used to trigger the `stdin_ok` check for `update_config`.

Things are simpler for `input_config` because "no input" is not a valid configuration like "no update" is.

christinaholtNOAA (approved):
LGTM!

WeirAE (approved):
Ah, great quick fix!

elcarpenterNOAA (approved):

PR #588, opened 2024-08-22
Add --show-schema support to drivers

maddenp-noaa:
**Synopsis**

Add `--show-schema` flag to UW drivers to dump to `stdout` a dereferenced form of a built-in driver's built-in schema. Here, "dereferenced" means that `$ref` references pointing to `urn:uwtools:*` URIs have been replaced by the schema content they refer to, so that the schema shown is standalone and complete.

Don't Panic: Several small, repeated changes to code and docs constitute the bulk of the PR.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/cli.py
```diff
@@ -52,9 +52,9 @@ def main() -> None:
         setup_logging(quiet=True)
         args, checks = _parse_args(sys.argv[1:])
         args[STR.action] = args.get(STR.action, args[STR.mode])
-        for check in checks[args[STR.mode]][args[STR.action]]:
+        for check in checks[args[STR.mode]].get(args[STR.action], []):
             check(args)
-        setup_logging(quiet=args[STR.quiet], verbose=args[STR.verbose])
+        setup_logging(quiet=args.get(STR.quiet, False), verbose=args.get(STR.verbose, False))
```
maddenp-noaa, line 57: For drivers, the argument value called `action` here would be the driver task name (`provisioned_rundir`, `run`, `validate`, etc.). (For other modes it'd be e.g. `realize` in `config realize`, `render` in `template render`, etc.) But when `--show-schema` is specified, a task (action) wouldn't be expected: We'd want to be able to say `uw upp --show-schema` without specifying any particular task. So, in the upper diff, the case where no `action` is defined in handled.

Also, for drivers, the `--quiet` and `--verbose` flags are added at the task level (e.g. `uw upp run`), not at the mode level (e.g. `uw upp`), so we will also not have that information when `--show-schema` is specified. And since it does't really make sense in this context anyway, we can default to `False` for both.

src/uwtools/cli.py
```diff
@@ -977,8 +985,9 @@ def _add_subparser_for_driver(
     :param with_leadtime: Does this driver require a leadtime?
     """
     parser = _add_subparser(subparsers, name, "Execute %s tasks" % name)
-    _basic_setup(parser)
-    subparsers = _add_subparsers(parser, STR.action, STR.task.upper())
+    optional = _basic_setup(parser)
+    _add_arg_show_schema(optional)
+    subparsers = _add_subparsers(parser, STR.action, STR.task.upper(), required=False)
```
maddenp-noaa, line 990: The task name now has to be optional under `uw <driver-name>`, because it will not be provided together with `--show-schema`. It's now technically optional (`required=False`), but a check is added later that it exists when `--show-schema` is _not_ specified.

src/uwtools/cli.py
```diff
@@ -1106,7 +1115,13 @@ def _dispatch_to_driver(name: str, args: Args) -> bool:
     :param name: Name of the driver to dispatch to.
     :param args: Parsed command-line args.
     """
-    execute: Callable[..., bool] = import_module("uwtools.api.%s" % name).execute
+    module = import_module("uwtools.api.%s" % name)
+    if args.get(STR.showschema):
+        print(json.dumps(module.schema(), sort_keys=True, indent=2))
+        return True
```
maddenp-noaa, line 1121: I `--show-schema` was specified, print the schema and return successfully.

src/uwtools/cli.py
```diff
@@ -1106,7 +1115,13 @@ def _dispatch_to_driver(name: str, args: Args) -> bool:
...
+    if not args.get(STR.action):
+        _abort("No %s specified" % STR.task.upper())
```
maddenp-noaa, line 1123: Here's the check that a task has been specified when `--show-schema` is not.

src/uwtools/config/formats/yaml.py
```diff
@@ -100,9 +100,10 @@ def _load(self, config_file: Optional[Path]) -> dict:
                 config = yaml.load(f.read(), Loader=loader)
                 if isinstance(config, dict):
                     return config
+                t = type(config).__name__
                 raise UWConfigError(
-                    "Parsed a %s value from %s, expected a dict"
-                    % (type(config).__name__, config_file or "stdin")
+                    "Parsed a%s %s value from %s, expected a dict"
+                    % ("n" if t[0] in "aeiou" else "", t, config_file or "stdin")
```
maddenp-noaa, line 106: I didn't like seeing `Parsed a int ...` in an error message.

src/uwtools/config/validator.py
```diff
@@ -19,6 +20,25 @@
 # Public functions
 
 
+def bundle(schema: dict) -> dict:
```
maddenp-noaa, line 23: This function recursively dereferences the schema, resulting in a complete, standalone schema. The JSON Schema maintainers refer to this process as [bundling](https://json-schema.org/blog/posts/bundling-json-schema-compound-documents).

src/uwtools/config/validator.py
```diff
@@ -95,13 +116,10 @@ def _prep_config(config: Union[dict, YAMLConfig, Optional[Path]]) -> YAMLConfig:
     return cfgobj
 
 
-def _validation_errors(config: Union[dict, list], schema: dict) -> list[ValidationError]:
+@cache
+def _registry() -> Registry:
```
maddenp-noaa, line 120: Factored out since it's needed at two call sites now.

src/uwtools/drivers/driver.py
```diff
@@ -154,9 +168,9 @@ def _create_user_updated_config(
 
     # Public helper methods
 
-    @property
+    @classmethod
```
maddenp-noaa, line 171: Sadly, use of `@classmethod` and `@property` together [has been deprecated](https://docs.python.org/3.11/library/functions.html#classmethod) in Python.

src/uwtools/drivers/driver.py
```diff
@@ -191,6 +205,13 @@ def _namelist_schema(
                 schema = schema[schema_key]
         return schema
 
+    @classmethod
+    def _schema_name(cls) -> str:
```
maddenp-noaa, line 209: Also needed at multiple call sites, so factored out.

src/uwtools/tests/drivers/test_cdeps.py
```diff
@@ -112,7 +112,7 @@ def test_CDEPS_streams(driverobj, group):
 
 
 def test_CDEPS_driver_name(driverobj):
-    assert driverobj.driver_name == "cdeps"
+    assert driverobj.driver_name() == CDEPS.driver_name() == "cdeps"
```
maddenp-noaa, line 115: The driver name is now available from both instance and class contexts.

src/uwtools/api/cdeps.py
```diff
@@ -2,13 +2,27 @@
...
+    Return the driver's schema.
+    """
+    return _driver.schema()
+
+
+def tasks() -> dict[str, str]:
+    """
+    Return a mapping from task names to their one-line descriptions.
+    """
+    return _tasks(_driver)
```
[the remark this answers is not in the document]
maddenp-noaa: To be clear, this problem exists on `main` -- it's not a result of the work on this branch. I just noticed it when I was adding `schema()` and thought I should fix it.

NaureenBharwaniNOAA (approved):
This looks great! Bundling definitely is a great technique for maintainability and efficiency.

src/uwtools/drivers/driver.py
```diff
@@ -463,7 +482,7 @@ def _runscript_path(self) -> Path:
         """
         The path to the runscript.
         """
-        return self.rundir / f"runscript.{self.driver_name}"
+        return self.rundir / f"runscript.{self.driver_name()}"
```
christinaholtNOAA, line 485: This is a frustrating breaking change. I see that the combined use of `@classmethod` and `@property` is being removed in Python 3.13.

christinaholtNOAA (approved):
This is going to be so helpful for understanding how to formulate YAMLs for drivers!

src/uwtools/drivers/driver.py
maddenp-noaa: Frustrating breaking change by us in `uwtools`, or by them in Python? Both maybe. I'll need to make a tiny update to the Coastal driver due to this, but I don't think we have much exposure on this outside our own drivers yet.

christinaholtNOAA: Both. But more so in Python.

maddenp-noaa:
@christinaholtNOAA 
> This is going to be so helpful for understanding how to formulate YAMLs for drivers!
I hadn't even thought of that use case -- great insight! Our prototype YAML files show _some_ of the config options, but the schema tells much more of the story. We've discussed the tradeoff between strict/complete schemas for e.g. namelist content and more permissive ones, and I think this use case lends even more weight to the script/complete argument, in that a schema with more information is a better guide to figuring out what one's options are.

PR #592, opened 2024-08-25
Add --schema-file option to internal drivers

maddenp-noaa:
**Synopsis**

- Add `--schema-file` option to all built-in drivers.
- Update and DRY out driver docs.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/conf.py
```diff
@@ -14,6 +14,7 @@
 autodoc_mock_imports = ["f90nml", "iotaa", "jsonschema", "lxml", "referencing"]
 autodoc_typehints = "description"
 copyright = str(dt.datetime.now().year)
+exclude_patterns = ["**/shared/*.rst"]
```
maddenp-noaa, line 17: This prevents Sphinx from trying to process fragment `.rst` files in `shared/` directories.

docs/sections/user_guide/cli/Makefile
```diff
@@ -1,4 +1,4 @@
-SUBDIRS = $(shell find . -maxdepth 1 -mindepth 1 -type d)
+SUBDIRS = $(shell find . -maxdepth 1 -mindepth 1 -type d | sort)
```
maddenp-noaa, line 1: Doesn't matter much, but a deterministic processing order can be helpful when debugging.

docs/sections/user_guide/cli/drivers/Makefile
maddenp-noaa, line 1: This file was previously a symlink to the identical file one directory up, but now it needs to be slightly different and was replaced. See below.

docs/sections/user_guide/cli/drivers/cdeps/help.rst
```diff
@@ -0,0 +1 @@
+../shared/help.rst
```
maddenp-noaa, line 1: In each of the `cli/drivers/*` directories, symlink to `help.rst` and, below, `schema-options.rst`. These unfortunately cannot be referred to with relative links like `../shared/idempotent.rst` (see the `index.rst` diff just below) because they themselves contain links to file to include (`help.cmd`, `help.out`, etc.), and Sphinx looks for those files in the directory containing the including file, rather than in the directory containing `index.rst`. Since `drivers/shared/` does not contain `help.cmd` et al., the doc build fails. By symlinking `help.rst` into e.g. the `cdeps/` directory, `help.rst` can find the CDEPS-specific `help.cmd`, `help.out`, etc. files and include them properly. The upshot is that the `drivers/*` directories all look more similar to one another, with duplicate content factored out and fewer differences in content that could lead to copy/paste errors.

When creating a new driver subdirectory, we'll have to keep in mind that several of its files will be symlinks.

docs/sections/user_guide/cli/drivers/cdeps/index.rst
```diff
@@ -1,23 +1,11 @@
 ``cdeps``
 =========
 
-.. include:: /shared/idempotent.rst
+.. include:: ../shared/idempotent.rst
```
maddenp-noaa, line 4: We can use a relative path here because `idempotent.rst` doesn't transitively `include` anything that it can't locate.

docs/sections/user_guide/cli/drivers/cdeps/index.rst
```diff
@@ -1,23 +1,11 @@
...
-.. literalinclude:: cdeps/help.out
-   :language: text
-
-All tasks take the same arguments. For example:
-
-.. literalinclude:: cdeps/run-help.cmd
-   :language: text
-   :emphasize-lines: 1
-.. literalinclude:: cdeps/run-help.out
-   :language: text
```
maddenp-noaa, line 20: Here and just below are two repetitive blocks that are now factored out of each driver `index.rst` file and replaced with an `include` directive. The diffs in this file and its directory repeat many times below for each of the other drivers.

docs/sections/user_guide/cli/drivers/index.rst
```diff
@@ -6,20 +6,20 @@ Drivers
...
+   jedi/index
+   make_hgrid/index
+   make_solo_mosaic/index
+   mpas/index
+   mpas_init/index
+   orog_gsl/index
+   sfc_climo_gen/index
+   shave/index
+   ungrib/index
+   upp/index
```
maddenp-noaa, line 25: Instead of e.g. a `cdeps.rst` file and a `cdeps/` directory (repeat for each driver), there's now just a `cdeps/` directory, which contains an `index.rst` file. This reduces clutter and unnecessary differences between each driver's set of files.

docs/sections/user_guide/cli/drivers/shared/help.rst
```diff
@@ -0,0 +1,13 @@
...
+.. literalinclude:: help.out
+   :language: text
+
+All tasks take the same arguments. For example:
+
+.. literalinclude:: run-help.cmd
+   :language: text
+   :emphasize-lines: 1
+.. literalinclude:: run-help.out
+   :language: text
```
maddenp-noaa, line 13: Here's the factored-out help info that was (nearly) identical in each driver's `.rst` file. The reorganization of the driver-doc files made the `index.rst` file that includes this file and the driver's `.cmd` and `.out` files siblings, so what used to be e.g. `cdeps/help.cmd` can now just be `help.cmd`, such that the directory part of the filename is unneeded. Now this file can be used by each driver.

docs/sections/user_guide/cli/drivers/shared/schema-options.rst
```diff
@@ -0,0 +1,9 @@
+* Specifying the ``--show-schema`` flag, with no other options, prints the driver's schema:
+
+  .. literalinclude:: show-schema.cmd
+     :language: text
+     :emphasize-lines: 1
+  .. literalinclude:: show-schema.out
+     :language: text
+
+* Use the ``--schema-file`` option to specify a custom :json-schema:`JSON Schema<>` file with which to validate the driver config. A custom schema could range in complexity from the simplest, most permissive schema, ``{}``, to one based on the built-in schema shown by ``--show-schema``.
```
maddenp-noaa, line 9: In addition to factoring out the `--show-schema` info from each driver, this shared fragment now adds info on the `--schema-file` option.

src/uwtools/api/execute.py
```diff
@@ -41,7 +41,7 @@ def execute(
     :param module: Path to driver module or name of module on sys.path.
     :param classname: Name of driver class to instantiate.
     :param task: Name of driver task to execute.
-    :param schema_file: Path to schema file.
+    :param schema_file: The JSON Schema file to use for validation.
```
maddenp-noaa, line 44: This parameter description was different, for no apparent reason, from the description for the same parameter used in several other modules.

WeirAE (approved):
Whew, a marathon of small but fantastic updates, LGTM

elcarpenterNOAA (approved):

christinaholtNOAA (changes requested):
I think what is here looks great. I feel like there may still be some work to be done in the driver code.

It looks like `get_schema_file` is still called there for the class method `schema` and in the function `_namelist_schema`. To my knowledge, that function only looks for the uwtools-provided schema file where we might want to instead use the user-provided one, if provided.

maddenp-noaa:
@christinaholtNOAA I think the current `get_schema_file()` behavior is correct: It always returns the built-in schema file. I suppose it could behave either way, but I don't see any sense in giving back to the user the schema they just gave us. They have it. Maybe we can be clearer in docs/docstrings, though. I'll check.

But `_namelist_schema` definitely needs to be updated. We discussed it [here](https://epic-slack-chan.slack.com/archives/C02S342LAHM/p1724093170393229?thread_ts=1724089172.465809&cid=C02S342LAHM) in the context of external drivers, but there's an even greater need to update it in the context of internal drivers using external schemas. I'll open a Jira ticket for it.

maddenp-noaa:
It just occurred to me: `--show-schema` and `--schema-file` are in different parts of the CLI: The former is mode-level (`uw <driver> --show-schema`), while the latter is action (task) level (e.g. `<driver> <task> --schema-file`). So `--schema-file` will always show the internal schema. But we should probably also have some `log.debug()` statements around the schema-selection logic so that `--verbose` would show what's going on. I'll see about adding those when I fix the `_namelist_schema()` logic.

maddenp-noaa:
I updated this PR to make `_namelist_schema()` on class `Assets` public, and updated it to use an external schema if one was provided. I also added some additional debug log messages related to schema selection and use, and added some detail to existing ones to try to make them even more informative. Some docs were also updated as a result.

docs/sections/user_guide/cli/tools/rocoto/realize-exec-file.out
```diff
@@ -1,5 +1,5 @@
-[2024-08-02T00:43:07]     INFO 0 UW schema-validation errors found
-[2024-08-02T00:43:07]     INFO 0 Rocoto validation errors found
+[2024-08-26T23:11:41]     INFO 0 UW schema-validation errors found in Rocoto config
+[2024-08-26T23:11:41]     INFO 0 Rocoto XML validation errors found
```
maddenp-noaa, line 2: Added `XML` to this message to make it extra clear that it's not the UW YAML config that was validated on the previous line.

docs/sections/user_guide/cli/tools/execute.rst
```diff
@@ -61,4 +61,4 @@ Config ``rand.yaml``
   .. literalinclude:: execute/alt-schema.out
      :language: text
 
-* Other arguments behave identically to the same-named arguments to built-in ``uwtools`` drivers (see :ref:`drivers`).
+* Other arguments behave identically to the same-named arguments to internal ``uwtools`` drivers (see :ref:`drivers`).
```
maddenp-noaa, line 64: We use "external" and "internal" to describe two kinds of driver, so in two places I changed `built-in` to `internal` for consistency with `uwtools` terminology.

docs/sections/user_guide/cli/tools/config/validate-pass-stdin.out
```diff
@@ -1 +1 @@
-[2024-05-23T19:39:15]     INFO 0 UW schema-validation errors found
+[2024-08-26T22:54:27]     INFO 0 UW schema-validation errors found in config
```
maddenp-noaa, line 1: This may not seem very valuable, but in the case of drivers that validate both their own config and the `platform:` block, we'll now have e.g.
```
[2024-08-27T00:05:16]     INFO Validating config against internal schema: sfc-climo-gen
[2024-08-27T00:05:16]     INFO 0 UW schema-validation errors found in sfc_climo_gen config
[2024-08-27T00:05:16]     INFO Validating config against internal schema: platform
[2024-08-27T00:05:16]     INFO 0 UW schema-validation errors found in platform config
```
It was probably clear from the order what each message meant, but this is even more explicit.

The specific case for this line involves `uw config validate`, so we have no idea what we're validating except that it's some kind of config.

src/uwtools/config/validator.py
```diff
@@ -20,85 +20,94 @@
...
     for k, v in schema.items():
+        newkeys = [*(keys or []), k]
+        key_path = ".".join(newkeys)
         if isinstance(v, dict):
-            if list(v.keys()) == [key] and v[key].startswith("urn:uwtools:"):
+            if list(v.keys()) == [ref] and v[ref].startswith("urn:uwtools:"):
                 # i.e. the current key's value is of the form: {"$ref": "urn:uwtools:.*"}
-                bundled[k] = bundle(_registry().get_or_retrieve(v[key]).value.contents)
+                uri = v[ref]
+                log.debug("Bundling referenced schema %s at key path: %s", uri, key_path)
```
maddenp-noaa, line 40: Added some debug log messages here to trace the bundling process, which may be helpful if/when something doesn't work as expected.

christinaholtNOAA (approved):
Thanks! Those changes LGTM. :)

PR #598, opened 2024-08-29
UW-666 Jupyter Notebook: File System Tool

Byrnetp:
**Synopsis**

- Introduces a notebook for the file system tool
- Adds unit tests for the notebook
- Adds Binder link on the user guide main page and a button on the `user_guide/api/fs` page of Read the Docs. 

**Type**

- [x] Documentation

**Impact**

**Checklist**


notebooks/Makefile
```diff
@@ -16,3 +16,4 @@ test-nb: lint unittest
 
 unittest:
 	pytest tests
+	rm -rf tmp
```
Byrnetp, line 19: The `fs` notebook creates lots of copies, symbolic links, and empty directories to files that don't need to be saved, so I place them in a `tmp` directory and clean them up after unit testing.

notebooks/fixtures/fs/file2.txt
```diff
@@ -0,0 +1 @@
+Fun Fact: A group of zebras is called a "zeal" or a "dazzle".
```
christinaholtNOAA, line 1: I appreciate this addition. ✨ 🦓  ✨

christinaholtNOAA (approved):
I think this one came together quite nicely. I don't want to hold the PR up, but I'm curious if it might benefit from a Table of Contents at the top that has clickable links to the section headers. If it's an easy add, it might add value.

WeirAE (approved):
Great work again!

Byrnetp:
Good idea @christinaholtNOAA, I think a table of contents definitely adds some value since this notebook got a bit long.

notebooks/Makefile
maddenp-noaa: FYI about `make` and `Makefile`s: In a multiline recipe like this, the first failing command terminates the recipe, so if `pytest tests` fails (exits with non-zero status, like if a test fails), `rm -rf tmp` won't run. If a subsequent test run depends on these files having been removed, that might be a problem -- in which case doing the `rm` first might be necessary. But if the tests just overwrite their previous output, this might not be needed at all, since it's already `.gitignore`d. There are similar temporary files created by `make docs` that we don't even try to delete, because `git` has been told to ignore them and they don't make any trouble. Your call, and you might have reasons I'm not thinking of for doing this `rm`, but it might not be needed.

maddenp-noaa (commented):
A couple of small ideas while I run through the notebook itself.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "fs.copy(\n",
+    "    config={\"missing-copy.nml\":\"fixtures/fs/missing-file.nml\"},\n",
+    "    target_dir=\"tmp/copy-target\"\n",
+    ")"
```
maddenp-noaa, line 215: This one is interesting: The output cell changes if you re-run the cell. I get three lines of output if I restart the Jupyter kernel and run the cell, but then only one line if I run the cell again. I think this might have to do with caching of task output and beyond the scope of this PR to address, but could you confirm that you see it too? I might know how to fix it if so, but in a different PR. I think it might confuse users to see different output each time they run a cell if they're not changing anything.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6f93ced5-0953-462b-9739-e74333c94e64",
+   "metadata": {},
+   "source": [
+    "The missing copy does not appear in the targeted directory."
```
maddenp-noaa, line 223:
```suggestion
    "The missing copy does not appear in the target directory."
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "77864d80-e6f4-48c2-a5d5-88fc512106a9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from uwtools.api import fs\n",
+    "from pathlib import Path"
```
maddenp-noaa, line 37:
```suggestion
     "from pathlib import Path\n",
     "from shutil import rmtree\n",
     "from uwtools.api import fs\n",
    "from uwtools.api.logging import use_uwtools_logger\n",
   "\n",
    "use_uwtools_logger()"
```
This change leads to more informative output, I think, but also points out a potential issue with different output from cells if they're run more than once, which I think is something Jupyter users do often -- I know I do. This is related to the fact that this tool is idempotent, and doesn't re-do work it's already done, so if e.g. a cell copies some files and logs some nice output about what it's done, re-running the cell will produce less output since the work is already done. `rmtree()` might be useful here -- I'll make a suggestion below and see what you think.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "fs.copy(\n",
+    "    config=\"fixtures/fs/copy-config.yaml\",\n",
+    "    target_dir=Path(\"tmp/copy-target\")\n",
+    ")"
+   ]
```
maddenp-noaa, line 140:
```suggestion
   "source": [
    "rmtree(\"tmp\", ignore_errors=True)\n",
    "fs.copy(\n",
    "    config=\"fixtures/fs/copy-config.yaml\",\n",
    "    target_dir=Path(\"tmp/copy-target\")\n",
    ")"
   ]
```
Well, I would expect the addition of this `rmtree()` call to make this cell behave the same way no matter how many times the user runs it, but that doesn't seem to be the case. I'm confused by similar behavior in a later cell that I commented on and wonder if this may be related to caching. We should probably figure out what's going on here.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b2527839-c217-428d-a686-c684a682c0e8",
+   "metadata": {},
+   "source": [
+    "### Using the `keys` parameter<!--copy-->\n",
+    "\n",
+    "If the destination keys are not found at the top level of the configuration file, like in the example below, the keys under which this block exists can be provided to the `keys` parameter. If `keys` is not specified with this config file, `copy()` raises a `UWConfigError`. "
```
maddenp-noaa, line 258:
```suggestion
    "Consider the following configuration, in which the destination/source mapping is not located at the top level of the configuration:"
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1340 @@
...
+    "%%bash\n",
+    "cat fixtures/fs/copy-keys-config.yaml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5311866f-a1f5-4243-81a8-2c52172e091a",
+   "metadata": {},
+   "source": [
+    "A list of string names is given to the `keys` parameter, each of which represents a key and are ordered from the top level down. The last key in the list directly contains the block of destination keys and their corresponding source values."
```
maddenp-noaa, line 290:
```suggestion
    "Without additional information, `copy()` would raise a `UWConfigError` given this configuration. However, the list of keys leading to the destination/source mapping can be provided with the `keys` parameter:"
```

maddenp-noaa (commented):
Standup is about to start, and maybe we can talk about the (maybe) caching issue I'm seeing. I'll send this set of comments, but am still working through the notebook.

notebooks/fs.ipynb
Byrnetp: I see the same change in output on my end when running this cell: three lines on the first run and then one line after.

maddenp-noaa:
@Byrnetp A new `iotaa` version (0.8.3) is available that I think solves the caching issue that was interfering with repeatable output in `uwtools` notebook cells. Could you try it out? If you're working in a `uwtools` development shell, you should be able to kill the `jupyter` process, run `conda install iotaa=0.8.3`, then restart `jupyter` and re-run the notebook. (Before restarting `jupyter`, you can `conda list | grep iotaa` to verify the old and/or new installed version.) If a cell starts with an `rmtree()` call to clear the old output, then I believe you should now be able to execute the cell repeatedly and get the same output each time. Please let me know if that's not true. I'll try it out, too.

Byrnetp:
After updating `iotaa` to version 0.8.3, the issue is resolved on my end. Thanks @maddenp-noaa!

NaureenBharwaniNOAA (approved):
LGTM!

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "source": [
+    "help(fs.Copier)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "061ac341-96cb-4af6-94ce-4f1e4d342b63",
+   "metadata": {},
+   "source": [
+    "A `Copier` class is instantiated using the same parameters as `copy()`, but copying is not performed until `Copier.go()` is called."
```
maddenp-noaa, line 495:
```suggestion
    "A `Copier` object is instantiated using the same parameters as `copy()`, but copying is not performed until `Copier.go()` is called."
```
In OO terms, an object (or instance) is instantiated from a class. "Instance" is probably the better term, but "instance is instantiated" sounds awkward.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "source": [
+    "help(fs.Linker)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3312a98b-9f5d-41bd-ad02-f69d291cc947",
+   "metadata": {},
+   "source": [
+    "A `Linker` class is instantiated using the same parameters as `link()`, but links are not created until `Linker.go()` is called."
```
maddenp-noaa, line 1036:
```suggestion
    "A `Linker` object is instantiated using the same parameters as `link()`, but links are not created until `Linker.go()` is called."
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1a1adba8-2daf-4fb6-b224-980b134f011c",
+   "metadata": {},
+   "source": [
+    "### Using the `Copier` class\n",
+    "\n",
+    "An alternative to using `copy()` is using the `Copier` class to stage files to copy and carry out the copy with `Copier.go()`."
```
maddenp-noaa, line 423:
```suggestion
    "An alternative to using `copy()` is to instantiate a `Copier` object , then call its `go()` method."
```
- To "stage", the the parlance of these modes, is to actually put the things -- copies, links, directories -- so I feel like it's `go()` that does the staging. I tried here to separate that term from the act of just instantiating the `Copier` object, which doesn't make any changes yet.
- This suggested change eliminated `carry out` from the sentence, but if it had remained it should have become `carrying out` to agree with the two instances `using`: `using... using... carrying...` rather than `using... using... carry`.

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "source": [
+    "help(fs.MakeDirs)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "cf3b50de-4a8b-4f51-96bb-a477b2c53430",
+   "metadata": {},
+   "source": [
+    "A `MakeDirs` class is instantiated using the same parameters as `makedirs()`, but directories are not created until `MakeDirs.go()` is called."
```
maddenp-noaa, line 1481:
```suggestion
    "A `MakeDirs` object is instantiated using the same parameters as `makedirs()`, but directories are not created until `MakeDirs.go()` is called."
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "29a9457a-e4f3-460a-b873-cf1bf236c9de",
+   "metadata": {},
+   "source": [
+    "### Using the `Linker` class\n",
+    "\n",
+    "An alternative to using `link()` is using the `Linker` class to stage files to link and carry out the action with `Linker.go()`."
```
maddenp-noaa, line 964:
```suggestion
    "An alternative to using `link()` is to instantiate a `Linker` object , then call its `go()` method."
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "742ce55e-fded-4961-931d-49bd75c09901",
+   "metadata": {},
+   "source": [
+    "### Using the `MakeDirs` class\n",
+    "\n",
+    "An alternative to using `makedirs()` is using the `MakeDirs` class to stage directories to create and carry out the action with `MakeDirs.go()`."
```
maddenp-noaa, line 1410:
```suggestion
    "An alternative to using `makedirs()` is to instantiate a `MakeDirs` object , then call its `go()` method."
```

notebooks/fs.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "source": [
+    "help(fs.link)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "710edac4-ba97-4599-a0f3-bc75ba2210e2",
+   "metadata": {},
+   "source": [
+    "Links to be created are specified by a mapping from keys destination-pathname keys to source-pathname values, either in a YAML file or a a Python ``dict``."
```
maddenp-noaa, line 631:
```suggestion
    "Links to be created are specified by a mapping from keys destination-pathname keys to source-pathname values, either in a YAML file or a Python ``dict``."
```

maddenp-noaa (approved):
I had a few final suggestions for your consideration, but this looks great.

elcarpenterNOAA (approved):
These are coming along really nicely. Good work!

PR #603, opened 2024-09-06
Fixes from integration with SRW's make_orog

christinaholtNOAA:
**Synopsis**

A set of changes to the drivers run as part of SRW's make_orog task. 
* Run commands seemed not to work as expected, so changed them to use config files.
* More information was needed in the UW YAML in a couple of instances.

Also ordered the tests in each of the drivers.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/cli/drivers/filter_topo/help.out
+++ b/docs/sections/user_guide/cli/drivers/filter_topo/help.out
@@ -12,6 +12,8 @@ Optional arguments:
 
 Positional arguments:
   TASK
+    filtered_output_file
+      The filtered output file staged from raw input
     input_grid_file
       The input grid file
     namelist_file
--- a/docs/sections/user_guide/cli/drivers/orog_gsl/help.out
+++ b/docs/sections/user_guide/cli/drivers/orog_gsl/help.out
@@ -12,6 +12,8 @@ Optional arguments:
 
 Positional arguments:
   TASK
+    input_config_file
+      The input config file
     input_grid_file
       The input grid file
     provisioned_rundir
--- a/docs/sections/user_guide/yaml/components/filter_topo.rst
+++ b/docs/sections/user_guide/yaml/components/filter_topo.rst
@@ -20,9 +20,17 @@ config:
 
 Configuration parameters for the ``orog_gsl`` component.
 
+  **filtered_orog:**
+
+  Name of the filtered output file.
+
   **input_grid_file:**
 
   Path to the tiled input grid file.
+
+  **input_raw_orog:**
+
+  Path to the raw orography file. The output of the ``orog`` driver.
 
 namelist:
 ^^^^^^^^^
--- a/docs/sections/user_guide/yaml/components/shave.rst
+++ b/docs/sections/user_guide/yaml/components/shave.rst
@@ -27,7 +27,7 @@ Describes the required parameters to run a ``shave`` configuration.
 
   Name of the grid file with extra points to be shaved.
 
-  **nh4:**
+  **nhalo:**
 
   The number of halo rows/columns.
 
@@ -38,6 +38,10 @@ Describes the required parameters to run a ``shave`` configuration.
   **ny:**
 
   The j/y dimensions of the compute domain (not including halo)
+
+  **output_grid_file:**
+
+  The path to the output file.
 
 rundir:
 ^^^^^^^
--- a/docs/shared/filter_topo.yaml
+++ b/docs/shared/filter_topo.yaml
@@ -1,6 +1,8 @@
 filter_topo:
   config:
+    filtered_orog: C403_filtered_orog.tile7.nc
     input_grid_file: /path/to/C403_grid.tile7.halo6.nc
+    input_raw_orog: /path/to/out.oro.nc
   execution:
     batchargs:
       cores: 1
--- a/docs/shared/orog.yaml
+++ b/docs/shared/orog.yaml
@@ -4,12 +4,12 @@ orog:
       cores: 1
       walltime: 00:05:00
     executable: /path/to/orog
-    files_to_link:
-      fort.15: /path/to/fix/thirty.second.antarctic.new.bin
-      landcover30.fixed: /path/to/fix/landcover30.fixed
-      fort.235: /path/to/fix/gmted2010.30sec.int
-    grid_file: /path/to/netcdf/grid/file
-    rundir: /path/to/run/dir
+  files_to_link:
+    fort.15: /path/to/fix/thirty.second.antarctic.new.bin
+    landcover30.fixed: /path/to/fix/landcover30.fixed
+    fort.235: /path/to/fix/gmted2010.30sec.int
+  grid_file: /path/to/netcdf/grid/file
+  rundir: /path/to/run/dir
 platform:
   account: me
   scheduler: slurm
--- a/src/uwtools/drivers/filter_topo.py
+++ b/src/uwtools/drivers/filter_topo.py
@@ -37,7 +37,7 @@ class FilterTopo(DriverTimeInvariant):
         The filtered output file staged from raw input.
         """
         src = Path(self.config["config"]["input_raw_orog"])
-        dst = Path(self.config["config"]["filtered_orog"])
+        dst = self.rundir / self.config["config"]["filtered_orog"]
         yield self.taskname(f"Raw orog input {str(dst)}")
         yield asset(dst, dst.is_file)
         yield filecopy(src=src, dst=dst)
--- a/src/uwtools/drivers/filter_topo.py
+++ b/src/uwtools/drivers/filter_topo.py
@@ -10,7 +10,7 @@ from uwtools.config.formats.nml import NMLConfig
 from uwtools.drivers.driver import DriverTimeInvariant
 from uwtools.drivers.support import set_driver_docstring
 from uwtools.strings import STR
-from uwtools.utils.tasks import symlink
+from uwtools.utils.tasks import filecopy, symlink
 
 
 class FilterTopo(DriverTimeInvariant):
@@ -27,9 +27,20 @@ class FilterTopo(DriverTimeInvariant):
         """
         src = Path(self.config["config"]["input_grid_file"])
         dst = Path(self.config[STR.rundir], src.name)
-        yield self.taskname("Input grid")
+        yield self.taskname(f"Input grid {str(src)}")
         yield asset(dst, dst.is_file)
         yield symlink(target=src, linkname=dst)
+
+    @task
+    def filtered_output_file(self):
+        """
+        The filtered output file staged from raw input.
+        """
+        src = Path(self.config["config"]["input_raw_orog"])
+        dst = Path(self.config["config"]["filtered_orog"])
+        yield self.taskname(f"Raw orog input {str(dst)}")
+        yield asset(dst, dst.is_file)
+        yield filecopy(src=src, dst=dst)
 
     @task
     def namelist_file(self):
@@ -56,6 +67,7 @@ class FilterTopo(DriverTimeInvariant):
         yield self.taskname("provisioned run directory")
         yield [
             self.input_grid_file(),
+            self.filtered_output_file(),
             self.namelist_file(),
             self.runscript(),
         ]
--- a/src/uwtools/drivers/orog.py
+++ b/src/uwtools/drivers/orog.py
@@ -37,7 +37,7 @@ class Orog(DriverTimeInvariant):
         The input grid file.
         """
         grid_file = Path(self.config["grid_file"])
-        yield self.taskname("Input grid file")
+        yield self.taskname(f"Input grid file {grid_file}")
         yield asset(grid_file, grid_file.is_file) if str(grid_file) != "none" else None
 
     @task
@@ -62,8 +62,9 @@ class Orog(DriverTimeInvariant):
                 "blat",
             ]
             inputs = " ".join([str(inputs[i]) for i in ordered_entries])
-        outgrid = self.config["grid_file"]
-        orogfile = self.config.get("orog_file")
+        outgrid = "'{}'".format(self.config["grid_file"])
+        if orogfile := self.config.get("orog_file"):
+            orogfile = "'{}'".format(orogfile)
         mask_only = ".true." if self.config.get("mask") else ".false."
         merge_file = self.config.get("merge", "none")  # string none is intentional
         content = [i for i in [inputs, outgrid, orogfile, mask_only, merge_file] if i is not None]
@@ -81,6 +82,23 @@ class Orog(DriverTimeInvariant):
             self.input_config_file(),
             self.runscript(),
         ]
+
+
+    @task
+    def runscript(self):
+        """
+        The runscript.
+        """
+        path = self._runscript_path
+        yield self.taskname(path.name)
+        yield asset(path, path.is_file)
+        yield None
+        envvars = {
+            "KMP_AFFINITY": "disabled",
+            "OMP_NUM_THREADS": self.config.get(STR.execution, {}).get(STR.threads, 1),
+            "OMP_STACKSIZE": "2048m",
+        }
+        self._write_runscript(path=path, envvars=envvars)
 
     # Public helper methods
 
--- a/src/uwtools/drivers/orog.py
+++ b/src/uwtools/drivers/orog.py
@@ -83,7 +83,6 @@ class Orog(DriverTimeInvariant):
             self.runscript(),
         ]
 
-
     @task
     def runscript(self):
         """
--- a/src/uwtools/drivers/orog_gsl.py
+++ b/src/uwtools/drivers/orog_gsl.py
@@ -9,6 +9,7 @@ from iotaa import asset, task, tasks
 from uwtools.drivers.driver import DriverTimeInvariant
 from uwtools.drivers.support import set_driver_docstring
 from uwtools.strings import STR
+from uwtools.utils.file import writable
 from uwtools.utils.tasks import symlink
 
 
@@ -29,7 +30,7 @@ class OrogGSL(DriverTimeInvariant):
         yield asset(path, path.is_file)
         yield None
         inputs = [str(self.config["config"][k]) for k in ("tile", "resolution", "halo")]
-        with open(path, "w", encoding="utf-8") as f:
+        with writable(path) as f:
             print("\n".join(inputs), file=f)
 
     @task
--- a/src/uwtools/drivers/orog_gsl.py
+++ b/src/uwtools/drivers/orog_gsl.py
@@ -9,7 +9,6 @@ from iotaa import asset, task, tasks
 from uwtools.drivers.driver import DriverTimeInvariant
 from uwtools.drivers.support import set_driver_docstring
 from uwtools.strings import STR
-from uwtools.utils.file import writable
 from uwtools.utils.tasks import symlink
 
 
--- a/src/uwtools/drivers/orog_gsl.py
+++ b/src/uwtools/drivers/orog_gsl.py
@@ -9,6 +9,7 @@ from iotaa import asset, task, tasks
 from uwtools.drivers.driver import DriverTimeInvariant
 from uwtools.drivers.support import set_driver_docstring
 from uwtools.strings import STR
+from uwtools.utils.file import writable
 from uwtools.utils.tasks import symlink
 
 
@@ -18,6 +19,19 @@ class OrogGSL(DriverTimeInvariant):
     """
 
     # Workflow tasks
+
+    @task
+    def input_config_file(self):
+        """
+        The input config file.
+        """
+        path = self._input_config_path
+        yield self.taskname(str(path))
+        yield asset(path, path.is_file)
+        yield None
+        inputs = [str(self.config["config"][k]) for k in ("tile", "resolution", "halo")]
+        with writable(path) as f:
+            print("\n".join(inputs), file=f)
 
     @task
     def input_grid_file(self):
@@ -40,6 +54,7 @@ class OrogGSL(DriverTimeInvariant):
         """
         yield self.taskname("provisioned run directory")
         yield [
+            self.input_config_file(),
             self.input_grid_file(),
             self.runscript(),
             self.topo_data_2p5m(),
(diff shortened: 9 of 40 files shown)


docs/sections/user_guide/cli/drivers/filter_topo/help.out
```diff
@@ -12,6 +12,8 @@ Optional arguments:
 
 Positional arguments:
   TASK
+    filtered_output_file
+      The filtered output file staged from raw input
```
christinaholtNOAA, line 16: This ended up being required in the rundir for filter_topo. It's a copy of the raw orog output since it modifies the file in place.

docs/sections/user_guide/cli/drivers/orog_gsl/help.out
```diff
@@ -12,6 +12,8 @@ Optional arguments:
 
 Positional arguments:
   TASK
+    input_config_file
+      The input config file
```
christinaholtNOAA, line 16: Here and in the `shave` driver, the run command was causing problems bc the program said it could read Unit 5 from the file in `/tmp`. I went with the `exec < config_file.in` type approach to the run command, so added the creation of the config file for each of those drivers.

docs/sections/user_guide/yaml/components/filter_topo.rst
```diff
@@ -20,10 +20,18 @@ config:
 
 Configuration parameters for the ``orog_gsl`` component.
 
+  **filtered_orog:**
+
+  Name of the filtered output file.
+
```
christinaholtNOAA, line 26: I added this as a configurable item, although its name is decided in the filter_topo program. We need access to the name downstream, so I thought it might as well be defined here.

docs/sections/user_guide/yaml/components/shave.rst
```diff
@@ -27,7 +27,7 @@ Describes the required parameters to run a ``shave`` configuration.
 
   Name of the grid file with extra points to be shaved.
 
-  **nh4:**
+  **nhalo:**
```
christinaholtNOAA, line 30: `nh4` is a SRW construct to indicate specifically that there are 4 halo cells. Changed this name to reflect that it is the number of halo cells.

docs/sections/user_guide/yaml/components/shave.rst
```diff
@@ -39,6 +39,10 @@ Describes the required parameters to run a ``shave`` configuration.
 
   The j/y dimensions of the compute domain (not including halo)
 
+  **output_grid_file:**
+
+  The path to the output file.
```
christinaholtNOAA, line 44: The `shave` program requires that the user name the output file and path, so adding that here.

docs/shared/orog.yaml
```diff
@@ -4,12 +4,12 @@ orog:
...
-      landcover30.fixed: /path/to/fix/landcover30.fixed
-      fort.235: /path/to/fix/gmted2010.30sec.int
-    grid_file: /path/to/netcdf/grid/file
-    rundir: /path/to/run/dir
+  files_to_link:
+    fort.15: /path/to/fix/thirty.second.antarctic.new.bin
+    landcover30.fixed: /path/to/fix/landcover30.fixed
+    fort.235: /path/to/fix/gmted2010.30sec.int
+  grid_file: /path/to/netcdf/grid/file
+  rundir: /path/to/run/dir
```
christinaholtNOAA, line 12: Just fixing bad indentation.

src/uwtools/drivers/filter_topo.py
```diff
@@ -27,10 +27,21 @@ def input_grid_file(self):
...
+    @task
+    def filtered_output_file(self):
+        """
+        The filtered output file staged from raw input.
+        """
+        src = Path(self.config["config"]["input_raw_orog"])
+        dst = self.rundir / self.config["config"]["filtered_orog"]
+        yield self.taskname(f"Raw orog input {str(dst)}")
+        yield asset(dst, dst.is_file)
+        yield filecopy(src=src, dst=dst)
```
christinaholtNOAA, line 43: Needs to be a copy bc it's updated in place by `filter_topo`.

src/uwtools/drivers/orog.py
```diff
@@ -37,7 +37,7 @@ def grid_file(self):
         The input grid file.
         """
         grid_file = Path(self.config["grid_file"])
-        yield self.taskname("Input grid file")
+        yield self.taskname(f"Input grid file {grid_file}")
```
christinaholtNOAA, line 40: I was finding it hard to know which file I was missing during the debugging process, so added the path here.

src/uwtools/drivers/orog.py
```diff
@@ -82,6 +83,22 @@ def provisioned_rundir(self):
...
+        path = self._runscript_path
+        yield self.taskname(path.name)
+        yield asset(path, path.is_file)
+        yield None
+        envvars = {
+            "KMP_AFFINITY": "disabled",
+            "OMP_NUM_THREADS": self.config.get(STR.execution, {}).get(STR.threads, 1),
+            "OMP_STACKSIZE": "2048m",
+        }
+        self._write_runscript(path=path, envvars=envvars)
```
christinaholtNOAA, line 100: This program is OMP-enabled, but not MPI. It needed a few extras in its runscript.

src/uwtools/drivers/orog_gsl.py
```diff
@@ -80,15 +95,20 @@ def driver_name(cls) -> str:
...
 
     @property
     def _runcmd(self):
         """
         The full command-line component invocation.
         """
-        inputs = [str(self.config["config"][k]) for k in ("tile", "resolution", "halo")]
         executable = self.config[STR.execution][STR.executable]
-        return "echo '%s' | %s" % ("\n".join(inputs), executable)
+        return "%s < %s" % (executable, self._input_config_path.name)
```
christinaholtNOAA, line 111: Here's where I changed the run command.

src/uwtools/drivers/shave.py
```diff
@@ -16,13 +19,34 @@ class Shave(DriverTimeInvariant):
...
+        path = self._input_config_path
+        yield self.taskname(str(path))
+        yield asset(path, path.is_file)
+        config = self.config["config"]
+        input_file = Path(config["input_grid_file"])
+        yield asset(input_file, input_file.is_file)
+        flags = [
+            config[key] for key in ["nx", "ny", "nhalo", "input_grid_file", "output_grid_file"]
+        ]
+        content = "{} {} {} '{}' '{}'".format(*flags)
```
christinaholtNOAA, line 36: How dumb! This thing really does need those last two file paths in quotes.

src/uwtools/drivers/orog.py
```diff
@@ -62,8 +62,9 @@ def input_config_file(self):
                 "blat",
             ]
             inputs = " ".join([str(inputs[i]) for i in ordered_entries])
-        outgrid = self.config["grid_file"]
-        orogfile = self.config.get("orog_file")
+        outgrid = "'{}'".format(self.config["grid_file"])
+        if orogfile := self.config.get("orog_file"):
+            orogfile = "'{}'".format(orogfile)
```
christinaholtNOAA, line 67: These quotes are required or the program does not run correctly. It does not complain, but does produce garbage output.

src/uwtools/resources/jsonschema/shave.jsonschema
```diff
@@ -9,8 +9,8 @@
             "input_grid_file": {
               "type": "string"
             },
-            "nh4": {
-              "minimum": 1,
+            "nhalo": {
+              "minimum": 0,
```
christinaholtNOAA, line 13: Some of the SRW use cases run with `nhalo=0`.

src/uwtools/tests/drivers/test_filter_topo.py
```diff
@@ -79,6 +83,17 @@ def test_FilterTopo(method):
...
+def test_FilterTopo_driver_name(driverobj):
+    assert driverobj.driver_name() == FilterTopo.driver_name() == "filter_topo"
+
+
+def test_FilterTopo_filtered_output_file(driverobj):
+    path = Path(driverobj.config["rundir"], "C403_filtered_orog.tile7.nc")
+    assert not path.is_file()
+    driverobj.filtered_output_file()
+    assert path.is_file()
+
```
christinaholtNOAA, line 95: This is a new one for this code.

src/uwtools/tests/drivers/test_orog.py
```diff
@@ -137,8 +141,8 @@ def test_Orog_input_config_file_old(driverobj):
     content = [l.strip("\n") for l in content]
     assert len(content) == 5
     assert len(content[0].split()) == 9
-    assert content[1] == driverobj.config["grid_file"]
-    assert content[2] == driverobj.config.get("orog_file")
+    assert content[1] == f"'{driverobj.config["grid_file"]}'"
+    assert content[2] == f"'{driverobj.config.get("orog_file")}'"
```
christinaholtNOAA, line 145: Here and in previous test, making sure those dang quotes exist.

src/uwtools/tests/drivers/test_orog.py
```diff
@@ -150,8 +154,13 @@ def test_Orog_provisioned_rundir(driverobj):
...
 
-def test_Orog_driver_name(driverobj):
-    assert driverobj.driver_name() == Orog.driver_name() == "orog"
+def test_Orog_runscript(driverobj):
+    with patch.object(driverobj, "_runscript") as runscript:
+        driverobj.runscript()
+        runscript.assert_called_once()
+        args = ("envcmds", "envvars", "execution", "scheduler")
+        types = [list, dict, list, Slurm]
+        assert [type(runscript.call_args.kwargs[x]) for x in args] == types
```
christinaholtNOAA, line 163: This one is new for the new addition.

src/uwtools/tests/drivers/test_orog_gsl.py
```diff
@@ -72,6 +72,20 @@ def test_OrogGSL(method):
...
+
+
+def test_OrogGSL_input_config_file(driverobj):
+    driverobj.input_config_file()
+    inputs = [str(driverobj.config["config"][k]) for k in ("tile", "resolution", "halo")]
+    with open(driverobj._input_config_path, "r", encoding="utf-8") as cfg_file:
+        content = cfg_file.readlines()
+    content = [l.strip("\n") for l in content]
+    assert len(content) == 3
+    assert content == inputs
```
christinaholtNOAA, line 86: Also a new one.

src/uwtools/tests/drivers/test_shave.py
```diff
@@ -70,25 +72,36 @@ def test_Shave(method):
...
+    nhalo = driverobj.config["config"]["nhalo"]
+    input_file_path = driverobj._config["config"]["input_grid_file"]
+    Path(input_file_path).touch()
+    output_file_path = driverobj._config["config"]["output_grid_file"]
+    driverobj.input_config_file()
+    with open(driverobj._input_config_path, "r", encoding="utf-8") as cfg_file:
+        content = cfg_file.readlines()
+    content = [l.strip("\n") for l in content]
+    assert len(content) == 1
+    assert content[0] == f"{nx} {ny} {nhalo} '{input_file_path}' '{output_file_path}'"
```
christinaholtNOAA, line 91: This one is new, too.

src/uwtools/tests/test_schemas.py
```diff
@@ -1889,16 +1892,19 @@ def test_schema_shave():
...
-        # nx, ny, and nh4 must be positive integers:
-        elif key in ["nx", "ny", "nh4"]:
+        # nx, ny, and nhalo must be integers:
+        elif key in (keyvals := {"nx": 1, "ny": 1, "nhalo": 0}):
+            minval = keyvals[key]
             assert "not of type 'integer'" in str(errors({key: "/path/"}))
-            assert "0 is less than the minimum of 1" in str(errors({key: 0}))
+            assert f"{minval - 1} is less than the minimum of {minval}" in str(
+                errors({key: minval - 1})
+            )
```
christinaholtNOAA, line 1907: Looking for feedback that this new approach is okay after setting nhalo's min val to 0. It somehow felt like I was being bad, but maybe not?

docs/sections/user_guide/yaml/components/filter_topo.rst
```diff
@@ -20,10 +20,18 @@ config:
...
   **input_grid_file:**
 
   Path to the tiled input grid file.
 
+  **input_raw_orog:**
+
+  Path to the raw orography file. The output of the ``orog`` driver.
```
maddenp-noaa, line 33: Is the `orog` driver the only place that can come from? The second sentence seems maybe a little over-specified given that filter topo was reading raw orog data long before `uwtools` existed. I'm not sure we have documentation yet that defines the inputs of components in terms of `uwtools` functionality and not just general requirements.

docs/sections/user_guide/yaml/components/filter_topo.rst
maddenp-noaa: Am I right in thinking that this is just created in the run directory, such that this really is just a filename and not a path?

Also, I think this came up in another context, but I suspect it's a bad idea to expose in configuration files items that are not configurable, if that's what you mean. If we _had_ to do this, then at least I suppose the schema should enforce exactly the one correct value. But I feel like this is component-specific knowledge that should be hidden away in its driver. Could it be a class attribute or static method on the driver class in question, so that other drivers could consult it if they need to know it?

docs/shared/filter_topo.yaml
```diff
@@ -1,6 +1,8 @@
 filter_topo:
   config:
+    filtered_orog: C403_filtered_orog.tile7.nc
```
maddenp-noaa, line 3: See comment above about this: If this is a magic value that cannot be changed, I think maybe it shouldn't appear in config.

src/uwtools/drivers/orog_gsl.py
```diff
@@ -19,6 +20,19 @@ class OrogGSL(DriverTimeInvariant):
...
+        """
+        The input config file.
+        """
+        path = self._input_config_path
+        yield self.taskname(str(path))
+        yield asset(path, path.is_file)
+        yield None
+        inputs = [str(self.config["config"][k]) for k in ("tile", "resolution", "halo")]
+        with writable(path) as f:
+            print("\n".join(inputs), file=f)
```
maddenp-noaa, line 34:
```suggestion
        with open(path, "w", encoding="utf-8") as f:
            print("\n".join(inputs), file=f)
```
I know `writable()` is convenient, but using it implies that `path` might be `None`, in which case it will write to `stdout`, which I doubt is something we'd ever want to happen in this context. To avoid any chance of bizarre bugs, I think a few more characters to make our yes-really-a-disk-file intent explicit are worth it.

src/uwtools/drivers/orog_gsl.py
```diff
@@ -80,15 +95,20 @@ def driver_name(cls) -> str:
         return STR.oroggsl
 
     # Private helper methods
+    @property
```
maddenp-noaa, line 98:
```suggestion
    # Private helper methods

    @property
```

src/uwtools/drivers/orog_gsl.py
maddenp-noaa: Ah, with the quoted items the `echo` version was probably in trouble. Makes sense.

src/uwtools/drivers/shave.py
```diff
@@ -16,13 +19,34 @@ class Shave(DriverTimeInvariant):
...
+        with writable(path) as f:
+            print(content, file=f)
```
maddenp-noaa, line 38:
```suggestion
        with open(path, "w", encoding="utf-8") as f:
            print(content, file=f)
```

src/uwtools/resources/jsonschema/execution-serial.jsonschema
```diff
@@ -12,6 +12,9 @@
     },
     "executable": {
       "type": "string"
+    },
+    "threads": {
+      "type": "integer"
```
maddenp-noaa, line 17: The semantics might be debatable, but my position is that OpenMP is not serial, and that a `threads` value doesn't belong in a serial config. Would it work to remove this and instead change `filter-topo.jsonschema` to `"$ref": "urn:uwtools:execution" instead of `"$ref": "urn:uwtools:execution-serial"`?

src/uwtools/tests/test_schemas.py
maddenp-noaa: Looks like a nice test to me.

src/uwtools/tests/test_schemas.py
```diff
@@ -1889,16 +1892,19 @@ def test_schema_shave():
...
-    for key in ("input_grid_file", "nx", "ny", "nh4"):
+    for key in ("input_grid_file", "nx", "ny", "nhalo"):
         # All config keys are required:
         assert f"'{key}' is a required property" in errors({})
         # A string value is ok for input_grid_file:
         if key == "input_grid_file":
             assert "not of type 'string'" in str(errors({key: 42}))
-        # nx, ny, and nh4 must be positive integers:
-        elif key in ["nx", "ny", "nh4"]:
+        # nx, ny, and nhalo must be integers:
```
maddenp-noaa, line 1901:
```suggestion
        # nx, ny, and nhalo must be integers >= their respective minimum values:
```

docs/sections/user_guide/yaml/components/filter_topo.rst
christinaholtNOAA: Yeah, I don't know what to do about this one. The exact name needs to be created based on formulaic knowledge that the experiment YAML has access to (tile number, CRES, etc.). We could duplicate that piece of information from the Fortran code inside the driver instead of putting it here. The Fortran code that does it looks like:

```
tile_file = trim(topo_file)//'.tile'//trim(text)//'.nc'
```

Where `text` is the tile number, and `topo_file` is the file path prefix specified in the namelist. When we expand to include support for a global domain, we'll need this to include all 6 files (6 tiles) that are copied in. 

I'm stuck on this one...I can def see the perks of both sides. Currently, the driver objects are not communicating their output file name(s), but it is certainly a nice to have feature. For all drivers, we currently have to set it in the config as _a priori_ knowledge. Maybe could leave it that way in this iteration and return to this as a moderate upgrade for all drivers?

docs/sections/user_guide/yaml/components/filter_topo.rst
christinaholtNOAA: Should I change this one to reference the output of the ufs_utils `orog` program? I do think that's the only thing that can create this file, independent of running with a UW driver.

src/uwtools/drivers/orog_gsl.py
christinaholtNOAA: I was really enjoying a tiny moment of lazy here! 😉

src/uwtools/resources/jsonschema/execution-serial.jsonschema
christinaholtNOAA: I see what you mean. It feels a little gross that there are so many MPI-related variables in the `execution` schema. This may be one of our only programs that is threaded, but not parallelized with MPI.

docs/sections/user_guide/yaml/components/filter_topo.rst
maddenp-noaa: Since it has to be constructed based on the config, it'd have to be provided via an instance method instead of a class/static method. But I'm still in favor of doing this -- constructing the filename as the Fortran code you show does -- in the driver, and would vote for not doing a release where users have to construct this ultimately derivative value. If we don't do it now, could we plan on fixing it before a v2.5.0 release?

docs/sections/user_guide/yaml/components/filter_topo.rst
maddenp-noaa: Yes, I think even just not referring directly to the UW driver would be good; referencing the `orog` _program_ would  be general enough IMO.

src/uwtools/resources/jsonschema/execution-serial.jsonschema
maddenp-noaa: Not necessarily in this PR, but we could rename `execution.jsonschema` -> `execution-parallel.jsonschema` and `execution-serial.jsonschema` -> `execution.jsonschema` if that feels better. Or have `-parallel` and `-serial` in both names to be more explicit.

docs/sections/user_guide/yaml/components/filter_topo.rst
maddenp-noaa: I guess it could still be a static/class method, with the components of the filename passed in as (maybe some optional) arguments. If an instance of the driver called it, it could pass in the values from its config, but non-driver-instance callers could pass in the filename components. I think that's still better than having users just provide the filename, as it makes it clear that the filename is not arbitrary, makes sure it's of the right form, and makes the connection between the filename and the overall configuration obvious.

src/uwtools/drivers/shave.py
christinaholtNOAA: @maddenp-noaa I applied this change and realized that `writable` does the helpful additional task of ensuring the directory where the file is meant to land exists.

We use it in the base class in the definition of `_write_runscript` and it's called as part of `dump` methods when we're creating supported config formats. I'm not sure whether I should just leave the `writable` call here (and above) because it is consistent, or whether I should have a `makedirs` step explicitly of this level of the drivers where it doesn't appear in the others.

maddenp-noaa: I think use of `writable()` in the `dump` methods makes sense because we might be dumping those to a file or to `stdout`. Its use in `_write_runscript()` is questionable, as that _must_ be a file (because we set the executable bit on the file immediate after).

But maybe I'm being too strict here: As long as there's no chance that the `path` argument to `writable()` is `None`, it should be safe to use. That's probably the case here, so please feel free to resolve this conversation.

maddenp-cu (approved):
Whatever you decide to do w.r.t. the hardcoded orog filename(s) will be fine. On principle, and just repeating what I already argued, I think that values that are not configurable should not appear in configuration, but should be derived from actual configurable values. That might not be practically attainable immediately.

src/uwtools/drivers/shave.py
christinaholtNOAA: How bad would it be to add a line with `assert path is not None`?

christinaholtNOAA: Also, I'm seeing now that we are internally defining a path that is not None with this class.

maddenp-noaa: Yeah, I can't see a possibility of a `None` value. I think it's fine.

elcarpenterNOAA (approved):

nbharwani-cu (approved):

WeirAE (approved):
Whew, I think I got through everything - great fixes!

Byrnetp (approved):

src/uwtools/resources/jsonschema/execution-serial.jsonschema
christinaholtNOAA: It might be nice to have `execution-parallel` and `execution-serial`.

PR #604, opened 2024-09-06
UW-653 controller key path

maddenp-noaa:
**Synopsis**

To support coupled driver applications where subcomponents defer to a "controller" component's config block for their `execution` and/or `rundir` values, we initially implemented a single key identifying to identify the controller component's block at the top of the config. In UW YAML in general, however, config blocks can be nested arbitrarily deeply, and a key path provided to navigate to them. This PR adds the ability to locate the controller component's block via an arbitrary key path. This is a breaking change because the `controller` argument now requires ` list[str]` value; but we only have a single coupled component at the moment, so it should be an easy update when this change is released in a future v2.5.0.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -49,7 +49,7 @@ class Assets(ABC):
         dry_run: bool = False,
         key_path: Optional[list[str]] = None,
         schema_file: Optional[Path] = None,
-        controller: Optional[str] = None,
+        controller: Optional[list[str]] = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


WeirAE (approved):
Thats a good catch, it should be fine to set now and update for the release!

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

Byrnetp (approved):

PR #607, opened 2024-09-17
UW-667 Jupyter notebook: Config Tool Part I

Byrnetp:
**Synopsis**

- Introduces a notebook for the config tool that covers the realize, realize_to_dict, and the get_format_config family of functions. (e.g. get_yaml_config)
- Adds unit tests for the notebook
- Adds Binder link on the user guide main page and a button on the user_guide/api/config page of Read the Docs.

**Type**

- [x] Documentation

**Impact**

**Checklist**


notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "\n",
+    "* [Getting Config Objects](#Getting-Config-Objects)\n",
+    "* [Config Depth Limitations](#Config-Depth-Limitations)\n",
+    "* [Realizing Configs](#Realizing-Configs)\n",
+    "    * [Updating configs](#Updating-configs)\n",
+    "    * [Using the `key_path` parameter](#Using-the-key_path-parameter)\n",
+    "    * [Using the `values_needed` parameter](#Using-the-values_needed-parameter)\n",
+    "    * [Using the `total` parameter](#Using-the-total-parameter)\n",
+    "* [Realizing Configs to a Dictionary](#Realizing-Configs-to-a-Dictionary)\n",
+    "<!--cell 0-->"
```
Byrnetp, line 24: Adding notebook cells that intentionally raise an exception made it necessary to target specific cell indices with testbook, so I added a comment to each markdown cell with its index.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "d023d283-7e1d-4e75-95b8-5033bea42a59",
+   "metadata": {},
+   "source": [
+    "# Config Tool\n",
+    "\n",
+    "The `uwtools` API's `config` module provides functions to create and manipulate configuration files, objects, and dictionaries.\n",
+    "\n",
+    "For more information, please see the <a href=\"https://uwtools.readthedocs.io/en/main/sections/user_guide/api/config.html\">uwtools.api.config</a> Read the Docs page.\n",
```
NaureenBharwaniNOAA, line 12: I think it might be a good idea to pin to the `stable` version over the `main` version of the docs. My thought was that `stable` will present a version of our most recent release, where `main` will pull directly from the latest on our `main` branch in Git. I'm not sure if there's a preference for this. I'd say if I was a user, I'd probably want something in a release rather than in development. Thoughts from the rest of the team?

https://uwtools.readthedocs.io/en/stable/sections/user_guide/api/config.html

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+      "File \u001b[0;32m~/NOAA/uwtools/src/uwtools/config/tools.py:107\u001b[0m, in \u001b[0;36mrealize_config\u001b[0;34m(input_config, input_format, update_config, update_format, output_file, output_format, key_path, values_needed, total, dry_run)\u001b[0m\n\u001b[1;32m    105\u001b[0m     \u001b[38;5;28;01mreturn\u001b[39;00m {}\n\u001b[1;32m    106\u001b[0m \u001b[38;5;28;01mif\u001b[39;00m total \u001b[38;5;129;01mand\u001b[39;00m unrendered(\u001b[38;5;28mstr\u001b[39m(input_obj)):\n\u001b[0;32m--> 107\u001b[0m     \u001b[38;5;28;01mraise\u001b[39;00m UWConfigRealizeError(\u001b[38;5;124m\"\u001b[39m\u001b[38;5;124mConfig could not be totally realized\u001b[39m\u001b[38;5;124m\"\u001b[39m)\n\u001b[1;32m    108\u001b[0m output_class \u001b[38;5;241m=\u001b[39m format_to_config(output_format)\n\u001b[1;32m    109\u001b[0m output_class\u001b[38;5;241m.\u001b[39mdump_dict(cfg\u001b[38;5;241m=\u001b[39moutput_data, path\u001b[38;5;241m=\u001b[39moutput_file)\n",
+      "\u001b[0;31mUWConfigRealizeError\u001b[0m: Config could not be totally realized"
```
NaureenBharwaniNOAA, line 832: I think it would be helpful to add a comment that the reason this error message comes up is because the `id` key-value pair was not rendered since the user didn't supply the value. The idea gets a little lost since the error message doesn't formally say that `id` was the key that could not be rendered and just says `Config could not be totally realized`.

The next example is helpful to see once `id` is added back in that it can now fully realize the config. Either adding a comment before or after would be great.

NaureenBharwaniNOAA (commented):
This notebook is looking great! 

Just a few small comments. I wasn't able to run the notebook since I got a `ModuleNotFoundError: No module named 'uwtools'`. I was wondering if others ran into the same problem or this was just an issue for me?

notebooks/config.ipynb
Byrnetp: I think this is probably a good idea in this case and in most cases but am curious to hear other's thoughts as well. There may be times where it would be nice to add a feature to the notebooks that exists on `main` but not in a release yet (the file -> fs change comes to mind) where we might need to initially link the `main` version of the docs and change it later.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dc745e95-d1ce-435c-a488-13b761979e36",
+   "metadata": {},
+   "source": [
+    "## Config Depth Limitations\n",
+    "\n",
+    "Some config formats have limitations on the depth of their nested configs. For example, Shell configs can only contain key-value pairs at the top level.\n",
+    "<!--cell 10-->"
```
christinaholtNOAA, line 179: Does it make sense to include a link in this sentence that describes sh config files? The same goes for INI below.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   "source": [
+    "help(config.realize)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "501da514-a654-4511-928f-b2ad7db102b2",
+   "metadata": {},
+   "source": [
+    "The `input_config` parameter takes a config from a string path, <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object, Python `dict`, or `Config` objects like the `YAMLConfig` object from the <a href=\"#Getting-Config-Objects\">Getting Config Objects</a> section. The format of the input config can be specified with `input_format`. Configs are written to `stdout` if `output_file` is set to `None` or to a file specified by `output_file` and `output_format`. Providing the formats of input and output config files is optional if their file extensions are ini, nml, sh, or yaml.\n",
```
christinaholtNOAA, line 386:
```suggestion
    "The `input_config` parameter takes a config from a string path, <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object, Python `dict`, or UW `Config` objects like the `YAMLConfig` object from the <a href=\"#Getting-Config-Objects\">Getting Config Objects</a> section. The format of the input config can be specified with `input_format`. Configs are written to `stdout` if `output_file` is set to `None` or to a file specified by `output_file` and `output_format`. Providing the formats of input and output config files is optional if their file extensions are `ini`, `nml`, `sh`, or `yaml`.\n",
```

Just a proposed addition of `UW`. Maybe it doesn't make sense given you're immediately explaining what kind of objects.

It might be nice to use the code highlight on the file formats at the end of the text, though.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "    output_file=Path('tmp/config1.yaml')\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "fd1823c4-6c3d-4d4a-a614-4d1238588bdd",
+   "metadata": {},
+   "source": [
+    "The function returns a `dict` version of the config, and the file is written in the YAML format as indicated by the file extension.\n",
```
christinaholtNOAA, line 419:
```suggestion
    "The `realize` method on a `Config` object returns a `dict` version of the config, and the file is written in the YAML format as indicated by the file extension.\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ef0136e4-f549-467f-9341-f77f84738bb0",
+   "metadata": {},
+   "source": [
+    "### Updating configs\n",
+    "\n",
+    "Configs can be updated by providing a second config with the `update_config` parameter. If the update config contains keys that match the base config, they will be overwritten. If the base config contains Jinja2 expressions, like the one below, they will be rendered using the values in the update config wherever possible.\n",
```
christinaholtNOAA, line 514:
```suggestion
    "Configs can be updated by providing a second config with the `update_config` parameter. If the update config contains keys that match the base config, the base config values for those keys will be overwritten. Once updated,  if the config contains Jinja2 expressions, like the one below, they will be rendered in the config wherever possible.\n",
```

I believe we should be able to provide an update config with Jinja2 expressions and they'd be rendered as well. I added some suggested changes here that reflect that.

In other words, once the updated values are applied to the base config, the new total config is dereferenced -- as many Jinja2 expressions as possible are rendered.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+      "  sender_id = \"{{ id }}\"\n",
+      "  message = \"{{ greeting }}, {{ recipient }}!\"\n",
+      "  sent = .FALSE.\n",
+      "/\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "cat fixtures/config/update-config.nml"
```
christinaholtNOAA, line 538: Does it make sense to name this one `base-config.nml`?

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+     "execution_count": 16,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "config.realize(\n",
+    "    input_config='fixtures/config/update-config.nml',\n",
+    "    update_config={\"memo\": {\"greeting\":\"Salutations\", \"recipient\":\"Mars\", \"sent\": True}},\n",
+    "    output_file='tmp/config-update.nml'\n",
```
christinaholtNOAA, line 575: And maybe this one is `updated-config.nml`?

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat fixtures/config/keys-config.yaml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "633778ca-391c-4ba4-b3b4-b3f599af5d41",
+   "metadata": {},
+   "source": [
+    "Using `key_path` allows only the portion of the config contained within the given list of keys to be written to a file or, in this case, to `stdout`. Note that the key-value pairs from the update config are used to render values and appear in the returned config dictionary, but they don't appear in the config written to `stdout`.\n",
```
christinaholtNOAA, line 651: This is an interesting feature I did not realize.

christinaholtNOAA (commented):
I like the story you're telling in this notebook. Just a few suggestions so far.

I made it through the notebook, but haven't looked at the tests yet. I'll try to come back to it later today.

notebooks/config.ipynb
Byrnetp: I added a link for INI configs (https://en.wikipedia.org/wiki/INI_file), but am having trouble finding something that seems suitable for Shell configs. Is there any source that you had in mind or any specific information that I could add there?

notebooks/config.ipynb
Byrnetp: Sounds good to me, I added UW before `Config` here and in other places I used this.

notebooks/config.ipynb
Byrnetp: I agree, this name and the one below were confusing. I updated them both.

notebooks/config.ipynb
christinaholtNOAA: Hmmmm. That's a great question. Maybe it's too basic to have a link. Just stating that bash variables follow the standard bash syntax `key=variable` would be enough?

elcarpenterNOAA (approved):
These are turning out really well. Nice work!

notebooks/config.ipynb
Byrnetp: I went ahead and made this change for this notebook and the template notebook too.

notebooks/config.ipynb
Byrnetp: I updated the sentence to include this. Let me know if there's anything else I could add there.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dc745e95-d1ce-435c-a488-13b761979e36",
+   "metadata": {},
+   "source": [
+    "## Config Depth Limitations\n",
+    "\n",
+    "Some config formats have limitations on the depth of their nested configs. Shell configs, for example, contain key-value pairs using the standard bash syntax `key=value` and can only contain key-value pairs at the top level.\n",
```
maddenp-noaa, line 178:
```suggestion
    "Some config formats have limitations on the depth of their nested configs. Shell configs, for example, may only contain top-level, bash-syntax `key=value` pairs.\n",
```
Hopefully still clear, and shorter -- avoids repetition of `key-value pairs` and then the more-or-less-same `key=value` term. I don't think we have to say _too_ much, as the meaning is made quite clear by the example config just below.

notebooks/tests/test_config.py
```diff
@@ -0,0 +1,105 @@
...
+        config_dict = yaml.safe_load(config_str)
+    with open("fixtures/config/base-config.nml", "r", encoding="utf-8") as f:
+        update_config_str = f.read().rstrip()
+    with open("fixtures/config/keys-config.yaml", "r", encoding="utf-8") as f:
+        keys_config_str = f.read().rstrip()
+    with testbook("config.ipynb", execute=non_err_cells) as tb:
+        # Ensure that cell output text matches expectations.
+        assert tb.cell_output_text(21) == str(config_dict)
+        assert tb.cell_output_text(23) == config_str
+        assert tb.cell_output_text(25) == str(config_dict)
```
christinaholtNOAA, line 55: If you loaded the config files with the UW API and named them `config` for the purposes of this , the object you get back can be modified to match these tests.

For the string comparison, you should be able to go with `assert tb.cell_output_text(21) == str(config)`, and for the dict comparison, convert a UW Config object to a `dict`, so the test looks like `assert tb.cell_output_text(25) == str(dict(cfg))`.

christinaholtNOAA (approved):
I think the tests look good in general. I am getting increasingly concerned about the numbering and being able to easily map the test to the cell I'm looking at in the static version of the notebook.

I've left a few suggestions below, but you can feel free to do with them what you like.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+      "\u001b[0;31mUWConfigError\u001b[0m: Cannot instantiate depth-2 INIConfig with depth-1 config"
+     ]
+    }
+   ],
+   "source": [
+    "config.get_ini_config(\n",
+    "    config={\"greeting\":\"Salutations\", \"recipient\":\"Mars\"}\n",
+    ")"
+   ]
```
maddenp-noaa, line 310: image

We could alternatively demonstrate errors like this, which looks a lot less messy and I think draws the user's attention to the important part. This is actually more like what an API user would do, too: They'd want to catch exceptions thrown in `uwtools` and handle them, rather than letting them crash their program. What do you think?

This also makes me think that we probably need to expose `uwtools` exceptions via the API, too, so that users can `import` them and specifically catch them. Here I'm catching a generic `Exception` object, but a linter would complain about this, and a user probably wouldn't really want to cast such a wide net, and catch other exceptions like `KeyError` or `DivideByZeroError`. I think it's still ok for a demonstration, but ideally we could `from uwtools.api.exceptions import UWConfigError` and then `except UWConfigError as e:`.

maddenp-noaa (commented):
I'm still reviewing, but this is great so far. Here are a few initial comments/ideas...

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   "source": [
+    "help(config.realize)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "501da514-a654-4511-928f-b2ad7db102b2",
+   "metadata": {},
+   "source": [
+    "The `input_config` parameter takes a config from a string path, <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object, Python `dict`, or UW `Config` objects like the `YAMLConfig` object from the <a href=\"#Getting-Config-Objects\">Getting Config Objects</a> section. The format of the input config can be specified with `input_format`. Configs are written to `stdout` if `output_file` is set to `None` or to a file specified by `output_file` and `output_format`. Providing the formats of input and output config files is optional if their file extensions are `ini`, `nml`, `sh`, or `yaml`.\n",
```
maddenp-noaa, line 386:
```suggestion
    "The `input_config` parameter takes a config from a string path, <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object, Python `dict`, or UW `Config` object like the `YAMLConfig` object from the <a href=\"#Getting-Config-Objects\">Getting Config Objects</a> section. The `input_format` argument must be provided for `dict` inputs or for files without recognized extensions. Configs are written to `stdout` if `output_file` is unspecified or explicitly set to `None`, or to the file specified by `output_file`. The `output_format` argument must be provided when writing to `stdout` or to a file without a recognized extension. Recognized extensions are: `.ini`, `.nml`, `.sh`, and `.yaml`.\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat tmp/config1.yaml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "902c4863-e772-49c6-88f2-31e1ab76b418",
+   "metadata": {},
+   "source": [
+    "Input and output formats are not required to match. This can be used to convert a config from one format to another.\n",
```
maddenp-noaa, line 448:
```suggestion
    "Input and output formats are not required to match. This can be used to convert some configs from one format to another.\n",
```
This ability depends on depth- and format-compatibility of the configs. For example:
```
(DEV-uwtools) ~/git/uwtools $ cat a.nml
&a n=88 /
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.nml --output-format nml
&a
    n = 88
/
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.nml --output-format yaml
a:
  n: 88
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.nml --output-format ini
[2024-09-23T16:37:46]    ERROR Accepted output formats for input format nml are nml or yaml
```
IIRC we actually only support output to 1. the same format as the input, or 2. YAML, e.g.
```
(DEV-uwtools) ~/git/uwtools $ cat a.ini 
[a]
n = 88
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.ini --output-format ini
[a]
n = 88
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.ini --output-format yaml
a:
  n: '88'
(DEV-uwtools) ~/git/uwtools $ uw config realize --input-file a.ini --output-format nml
[2024-09-23T16:40:17]    ERROR Accepted output formats for input format ini are ini or yaml
```
Oh, and maybe YAML -> anything is ok, provided the YAML depth conforms to the output format.

You might need to check the code to verify this if you think we should be explicit, or we could just hedge here given that the `uwtools` will complain (informatively) if an unsupported conversion is attempted.

In addition to the depth problem, there are issues with expressiveness between the formats. For example, INI and SH configs are untyped: Their values are always `str` in Python:
```
>>> from uwtools.api.config import get_ini_config
>>> c = get_ini_config("a.ini")
>>> type(c["a"]["n"])
<class 'str'>
```
And so it would be dangerous to use these values to populate, say, an NML config where `int` values are expected.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   "source": [
+    "help(config.get_yaml_config)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "606da5b3-4bff-4148-a9a5-908aa7dd5e8c",
+   "metadata": {},
+   "source": [
+    "`get_yaml_config()` can take input from a Python `dict` or a YAML file like the one below.\n",
```
maddenp-noaa, line 83:
```suggestion
    "The `stdin_ok` argument can be used to permit reads from `stdin`, but this is a rare use case beyond the scope of this notebook that will not be discussed here.\n",
    "\n",
    "`get_yaml_config()` can take input from a Python `dict` or a YAML file like the one below.\n",
```
I feel like we should mention this, if only to say that we're not going to mention it again.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ddbc989a-a9ed-4649-8e87-f74d1ff43b89",
+   "metadata": {},
+   "source": [
+    "### Using the `total` parameter\n",
+    "\n",
+    "The `total` parameter is used to specify if all Jinja2 expressions should be rendered before the final config is written. Consider the config below which contains multiple expressions.\n",
```
maddenp-noaa, line 780:
```suggestion
    "The `total` parameter is used to specify that all Jinja2 expressions must be rendered before the final config is written. Consider the config below which contains multiple expressions.\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat fixtures/config/base-config.nml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7c0d64be-c97c-4d07-b636-54b1c8bb5d0c",
+   "metadata": {},
+   "source": [
+    "As was shown in the <a href=\"#Updating-configs\">Updating configs</a> section, not all Jinja2 expressions are required to be rendered when `total` is set to its default value, `False`. However, a complete rendering of all expressions will be required if `total` is set to `True`. If this is the case and not enough values are provided to fully render the config, a `UWConfigRealizeError` will be raised. Notice that values are provided for `greeting` and `recipient`, but not for `id`.\n",
```
maddenp-noaa, line 812:
```suggestion
    "As was shown in the <a href=\"#Updating-configs\">Updating configs</a> section, by default not all Jinja2 expressions are required to be rendered. However, when `total` is set to `True` and not enough values are provided to fully realize the config, a `UWConfigRealizeError` is raised. Notice that values are provided for `greeting` and `recipient`, but not for `id`.\n",
```

maddenp-noaa (commented):
I've made it to _Realize Configs to a Dictionary_, will review more this afternoon.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "The `uwtools` API's `config` module provides functions to create and manipulate configuration files, objects, and dictionaries.\n",
+    "\n",
+    "For more information, please see the <a href=\"https://uwtools.readthedocs.io/en/stable/sections/user_guide/api/config.html\">uwtools.api.config</a> Read the Docs page.\n",
+    "\n",
+    "## Table of Contents\n",
+    "\n",
+    "* [Getting Config Objects](#Getting-Config-Objects)\n",
+    "* [Config Depth Limitations](#Config-Depth-Limitations)\n",
+    "* [Realizing Configs](#Realizing-Configs)\n",
+    "    * [Updating configs](#Updating-configs)\n",
```
maddenp-noaa, line 19:
```suggestion
    "    * [Updating Configs](#Updating-configs)\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat tmp/realize-config.sh"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ef0136e4-f549-467f-9341-f77f84738bb0",
+   "metadata": {},
+   "source": [
+    "### Updating configs\n",
```
maddenp-noaa, line 512:
```suggestion
    "### Updating Configs\n",
```
For consistency with other headings.

notebooks/config.ipynb
maddenp-noaa, line 812:
```suggestion
    "As was shown in the <a href=\"#Updating-configs\">Updating Configs</a> section, not all Jinja2 expressions are required to be rendered when `total` is set to its default value, `False`. However, a complete rendering of all expressions will be required if `total` is set to `True`. If this is the case and not enough values are provided to fully render the config, a `UWConfigRealizeError` will be raised. Notice that values are provided for `greeting` and `recipient`, but not for `id`.\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "\n",
+    "For more information, please see the <a href=\"https://uwtools.readthedocs.io/en/stable/sections/user_guide/api/config.html\">uwtools.api.config</a> Read the Docs page.\n",
+    "\n",
+    "## Table of Contents\n",
+    "\n",
+    "* [Getting Config Objects](#Getting-Config-Objects)\n",
+    "* [Config Depth Limitations](#Config-Depth-Limitations)\n",
+    "* [Realizing Configs](#Realizing-Configs)\n",
+    "    * [Updating configs](#Updating-configs)\n",
+    "    * [Using the `key_path` parameter](#Using-the-key_path-parameter)\n",
```
maddenp-noaa, line 20:
```suggestion
    "    * [Using the `key_path` Parameter](#Using-the-key_path-parameter)\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "    * [Using the `values_needed` parameter](#Using-the-values_needed-parameter)\n",
```
maddenp-noaa, line 21:
```suggestion
    "    * [Using the `values_needed` Parameter](#Using-the-values_needed-parameter)\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "    * [Using the `total` parameter](#Using-the-total-parameter)\n",
```
maddenp-noaa, line 22:
```suggestion
    "    * [Using the `total` Parameter](#Using-the-total-parameter)\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat tmp/updated-config.nml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a20024ac-0f33-4000-942d-29c99dc0502e",
+   "metadata": {},
+   "source": [
+    "### Using the `key_path` parameter\n",
```
maddenp-noaa, line 618:
```suggestion
    "### Using the `key_path` Parameter\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "    key_path=['keys', 'to', 'config']\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "903ea23f-4625-48f0-9efc-1c0a106d5bf6",
+   "metadata": {},
+   "source": [
+    "### Using the `values_needed` parameter\n",
```
maddenp-noaa, line 696:
```suggestion
    "### Using the `values_needed` Parameter\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "    values_needed=True\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ddbc989a-a9ed-4649-8e87-f74d1ff43b89",
+   "metadata": {},
+   "source": [
+    "### Using the `total` parameter\n",
```
maddenp-noaa, line 778:
```suggestion
    "### Using the `total` Parameter\n",
```

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    "%%bash\n",
+    "cat tmp/config-total.nml"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e121dab6-5d16-40b0-aa16-161c329a8e9a",
+   "metadata": {},
+   "source": [
+    "## Realizing Configs to a Dictionary\n",
```
maddenp-noaa, line 925: I wonder if this whole section could be minimized, just saying (paraphrasing) "This does exactly what `config.realize()` does, but only returns a `dict` value and writes no output.` In fact, the `help()` output just directs the user to see the `realize()` help, so isn't very enlightening. Maybe one small example, like the one shown below, would suffice -- especially if the suggestion I offered above, to use a `_ = config.realize(...)` assignment to suppress the `dict` output that `realize()` also returns, is accepted.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+      "{'id': '456', 'greeting': 'Hello', 'recipient': 'World'}\n"
+     ]
+    }
+   ],
+   "source": [
+    "config_dict = config.realize_to_dict(\n",
+    "    input_config={\"id\": \"456\"},\n",
+    "    update_config=\"fixtures/config/get-config.yaml\"\n",
+    ")\n",
+    "print(config_dict)"
```
maddenp-noaa, line 1012:
```suggestion
    "config.realize_to_dict(\n",
    "    input_config={\"id\": \"456\"},\n",
    "    update_config=\"fixtures/config/get-config.yaml\"\n",
    ")\n"
```
Since Jupyter displays the cells result anyway, this more concise version shows exactly the same output.

maddenp-noaa (commented):
I think that's it for a first round from me -- sorry it's taken so long. I'd be happy to do another quick pass this afternoon if the timing works out. Otherwise, nice work, and I'll leave it to others to approve, etc.

notebooks/config.ipynb
Byrnetp: I like this style of demonstrating errors a lot! Besides cutting down on the length of the notebook and drawing attention to the important part, I think it will also make unit testing easier too. I'll update the error demo cells to look like this.

notebooks/config.ipynb
```diff
@@ -0,0 +1,1046 @@
...
+    }
+   ],
+   "source": [
+    "config.realize(\n",
+    "    input_config=\"fixtures/config/keys-config.yaml\",\n",
+    "    update_config={\"greeting\": \"Good morning\", \"recipient\": \"Venus\"},\n",
+    "    output_file=None,\n",
+    "    output_format='yaml',\n",
+    "    key_path=['keys', 'to', 'config']\n",
+    ")"
```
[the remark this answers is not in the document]
Byrnetp: I agree that having the two outputs together there maybe isn't the best. What would you think of splitting that demo into two code cells? I can assign the dictionary returned by `realize()` to a variable that I can print in a follow-up code cell so that I don't clutter the `stdout` output. Or do you think the `dict` isn't worth showing there?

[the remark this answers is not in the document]
maddenp-noaa: I don't think the `dict` is worth showing in this section: I think the `config.realize_to_dict()` section shows that off well. This section is about producing _external_ output, either in a file or on `stdout`, and the fact that `config.realize()` happens to return a `dict` is incidental in this context. (It's not incidental in `uwtools`, which takes advantage of that fact to define `realize_to_dict()` as a specially crafted call to `realize()`.)

notebooks/config.ipynb
maddenp-noaa: Good questions. One option might be to pin to a _specific_ release (not even just the latest release) that we are sure the notebooks are compatible with. Then, when we release a new version, we could intentionally revisit the notebooks, run their tests but also read through them to make sure they still make sense: It's one thing to pass automated code tests, it's another to ensure that the cell commentary still actually describes the code cells.

If we pinned to `stable` (which, if I'm not mistaken, is essentially what we're doing with the `uwtools` package itself in `binder/environment.yml` by not specifying a specific version number) then, when `notebooks` is integrated into the `main` branch, our CI tests should at least ensure that the code correct via the unit tests. But we'd still have to have the discipline to go review the prose part of the notebooks.

I think I'd prefer to pin both docs and the package to a specific release, and then schedule occasional work to update those pins and read through the notebooks to make sure they're still correct.

It would be good to hear from others, and ultimately from PO @christinaholtNOAA.

notebooks/tests/test_config.py
```diff
@@ -0,0 +1,105 @@
...
+
+all_cells = [*range(0, 57)]
+err_cells = [13, 17, 45]
+non_err_cells = list(filter(lambda n: n not in err_cells, all_cells))
+
+
+def test_get_config():
+    with open("fixtures/config/get-config.yaml", "r", encoding="utf-8") as f:
+        config1_str = f.read().rstrip()
+        config1_dict = yaml.safe_load(config1_str)
```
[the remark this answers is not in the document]
maddenp-noaa: @christinaholtNOAA Could you elaborate on your concern here? I'm also not sure I'm parsing "UW-supported constructors in a fixture file" correctly -- could you give an example of what you mean.

At first glance, I like this test: It uses the lowest-possible-level mechanism -- `.read()` to get a string version of a file, `yaml.safe_load()` to turn that string into a `dict` -- to establish the "expected" values, then asserts that the more-complicated UW-based techniques demonstrated in the notebook produce equivalent values. Using higher-level UW techniques for the "expected" values introduces the possibility of tautology (testing the same thing on both sides of `==`), like `assert YAMLConfig("a.yaml") == YAMLConfig("a.yaml")` is a tautology, and `assert YAMLConfig("a.yaml") == get_yaml_config("a.yaml")` isn't much better as it's exercising nearly identical code. Of course, all of the above eventually calls some form of `yaml.load()`, but IMO constructing "expected" values in the most basic way possible is sound practice.

But I'm curious if I'm missing something, especially about "UW-supported constructors".

maddenp-noaa (approved):
@Byrnetp This is looking great -- very thorough, and the notebook tells a coherent story as the reader works through it. I might have eventually had other comments, but am running out of time before vacation, so will 👍 this nice work and 🤞 for a quick tie-off of loose ends and a merge.

notebooks/config.ipynb
Byrnetp: I added your suggestion and did my best to capture the limitations in a few sentences in this markdown cell.

notebooks/config.ipynb
Byrnetp: I agree, I updated the section to remove the help call and emphasize the similarity to `realize()` sooner.

notebooks/tests/test_config.py
```diff
@@ -0,0 +1,105 @@
...
+        assert (
+            "'sender_id': '{{ id }}'" in tb.cell_output_text(31)
+            and "'message': 'Salutations, Mars!'" in tb.cell_output_text(31)
+            and "'sent': True" in tb.cell_output_text(31)
+        )
+        assert (
+            "sender_id = '{{ id }}'" in tb.cell_output_text(33)
+            and "message = 'Salutations, Mars!'" in tb.cell_output_text(33)
+            and "sent = .true." in tb.cell_output_text(33)
+        )
```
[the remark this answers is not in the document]
Byrnetp: I like this cleaner approach. I changed a few tests to look like this.

notebooks/tests/test_config.py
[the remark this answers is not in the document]
Byrnetp: I've been hesitant to use the UW API in my testing for the reasons @maddenp-noaa stated, and I'm curious to know more about the concerns here too.

notebooks/config.ipynb
christinaholtNOAA: I agree with Paul's assessment. Let's pin to a specific release and revisit the notebooks occasionally. Let's make sure that the version is very clear to the user of the notebook.

NaureenBharwaniNOAA (approved):
Looks great!

notebooks/tests/test_config.py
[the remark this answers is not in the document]
christinaholtNOAA: Sorry, I often use _constructor_ and _tag_ interchangeably here and probably shouldn't -- _tag_ is the `!int` that shows up in the YAML file, while _constructor_ is the chunk of code that tells _pyyaml_ what to do with the value. If we wanted to use UW-supported tags in our examples, then we'd need to load the YAML file with `get_yaml_config`.

I'm not saying we want to or need to here, but it feels like it could lead to problems where we might want some consistency in our tests later on.

notebooks/config.ipynb
Byrnetp: I updated this notebook and the template notebook to mention a specific release and link to the same Read the Docs version of the page, but left the fs notebook alone since there isn't a numbered release that has the file -> fs change yet.

PR #609, opened 2024-10-07
Update jsonschema file name from execution to execution-parallel

NaureenBharwaniNOAA:
**Synopsis**

Updated all naming instances of `execution` to `execution-parallel`.
Ensured that all test cases pass.

**Type**

**Impact**

**Checklist**


src/uwtools/tests/test_schemas.py
```diff
@@ -725,7 +725,7 @@ def test_schema_execution_threads():
 def test_schema_execution_serial():
     config = {"executable": "fv3"}
     batchargs = {"batchargs": {"queue": "string", "walltime": "string"}}
-    errors = schema_validator("execution")
+    errors = schema_validator("execution-serial")
```
NaureenBharwaniNOAA, line 728: I saw this was originally `execution`, but feel that it should have been `execution-serial` since it is testing the `execution-serial` schema and not `execution-parallel`.

src/uwtools/tests/test_schemas.py
```diff
@@ -667,15 +667,15 @@ def test_schema_esg_grid_rundir(esg_grid_prop):
...
+# execution-parallel
 
 
 def test_schema_execution():
     config = {"executable": "fv3"}
     batchargs = {"batchargs": {"queue": "string", "walltime": "string"}}
     mpiargs = {"mpiargs": ["--flag1", "--flag2"]}
     threads = {"threads": 32}
-    errors = schema_validator("execution")
+    errors = schema_validator("execution-parallel")
```
NaureenBharwaniNOAA, line 678: The comment above has been updated to `execution-parallel`. I can update these test names to be `test_schema_execution_parallel` if we'd like. I felt like it might make the test names longer, but will ultimately be clearer.

christinaholtNOAA: I like the idea of changing the test names. Following the pattern used for the serial version, it makes sense to go with the `execution_parallel` order here.

However, I found it easier to read with this example from the next test.

```
test_schema_parallel_execution_executable
```
What do others think about the ordering of `execution` and `parallel` in the test names?

christinaholtNOAA (approved):
Looks good. I'll leave it to you on which way to go about the test naming. I do like the idea of adding parallel to the test names.

src/uwtools/tests/test_schemas.py
NaureenBharwaniNOAA: I like this idea to! Makes it easy to understand what the test case is testing

elcarpenterNOAA (approved):

PR #611, opened 2024-10-11
Jupyter notebook: Config Tool Part II

Byrnetp:
**Synopsis**

* Adds sections to the config tool notebook covering `compare()`, `validate()`, and the UW `Config` classes. 
* Adds unit tests for the config notebook
* Removes the example notebook along with its unit tests and fixtures

**Type**

- [ ] Documentation

**Impact**

**Checklist**


notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+    "echo ----------------------------------------------\n",
+    "cat fixtures/config/alt-config.nml "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c4361a7c-9c16-4bc0-b3e6-985d85f1450c",
+   "metadata": {},
+   "source": [
+    "`compare()` returns `True` if the configs contain identical key-value pairs, and `False` otherwise. If a logger has been initialized, a log is displayed which reports which files are being compared and the values that differ, if any. Files are passed to `config_1_path` and `config_2_path` as a string filename or <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object. Corresponding formats are passed to `config_1_format` and `config_2_format`, with these parameters being optional if the format is recognized.\n",
```
christinaholtNOAA, line 1060:
```suggestion
    "`compare()` returns `True` if the configs contain identical key-value pairs, and `False` otherwise. If a logger has been initialized, information is logged on which files are being compared and the values that differ, if any. Files are passed to `config_1_path` and `config_2_path` as a string filename or <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object. Corresponding optional formats may be passed using `config_1_format` and `config_2_format` and are only needed if the format suffix is not recognized.\n",
```

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+   "source": [
+    "help(config.validate)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1eeca5ad-e969-4670-bddd-8e20b38f3d6c",
+   "metadata": {},
+   "source": [
+    "Below is an example of a schema used to validate a config. It ensures that the required keys are present and the value types match expectations.\n",
```
christinaholtNOAA, line 1208: It would be great if we could link to JSON Schema docs in this text block.

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+      "    \"greeting\", \"recipient\"\n",
+      "  ],\n",
+      "  \"type\": \"object\"\n",
+      "}\n"
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "cat fixtures/config/validate.jsonschema"
```
christinaholtNOAA, line 1242: Does it make sense to add a reminder cell showing what's in the `get-config.yaml` that this schema is going to be validating since we've more recently seen a couple of other YAML contents?

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+    "    config='fixtures/config/get-config.yaml'\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8d151205-4a95-4b46-aa1d-30ec30d96e88",
+   "metadata": {},
+   "source": [
+    "If validation errors exist, the logger provides the number, location, and details of any errors found.\n",
```
christinaholtNOAA, line 1290:
```suggestion
    "The `config` argument also accepts a dictionary. In the next example, validation errors exist, and the logger reports the number of errors found along with their locations and details.\n",
```

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+    ")\n",
+    "print(fruits)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1a50a362-d063-4ae1-9b6a-d98d0bf84c4f",
+   "metadata": {},
+   "source": [
+    "### Comparing Config Objects\n",
```
christinaholtNOAA, line 1572: This is interesting. I'd expect that we'd want to explicitly mention that all the config objects have this capability and that this is an alternative to using the `compare()` tool above from the API, which only takes file paths as input.

It may also be worth mentioning that while a dictionary doesn't inherently have a file format (INI, sh, etc.), the internals of UW Tools will specify how deep these formats can be. Those format limitations are not applied to this function (as best I can tell...please correct me if I read that wrong) so all format objects will give you the same behavior on any two dictionaries.

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+    "    dict2={'fruit count':{'apples':'3', 'grapes':'8', 'kiwis':'1'}}\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3caa77b7-2eca-498b-841f-b37c91068974",
+   "metadata": {},
+   "source": [
+    "When there are differences between the two configs, `False` is returned. When a logger is initialized, the values that differ are displayed.\n",
```
christinaholtNOAA, line 1607: Here, we should mention that when a second dictionary is not provided, the comparison happens between `dict1` and the object itself -- `fruit` in this example.

notebooks/config.ipynb
```diff
@@ -964,12 +971,831 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "945d1e71-2704-49de-9a7c-dc8bf58c12fe",
+   "metadata": {},
+   "source": [
+    "### Updating Values\n",
+    "\n",
+    "The `update()` method adds new or updated key-value pairs to the stored config, and these are provided as a `dict` or `UserDict` via the `src` parameter. \n",
```
christinaholtNOAA, line 1771:
```suggestion
    "The `update_from()` method adds new or updated key-value pairs to the stored config, and these are provided as a dictionary via the `src` parameter. \n",
```

christinaholtNOAA (changes requested):
Great work! I've added some suggestions below.

notebooks/config.ipynb
Byrnetp: Agreed, I added the link to this block.

notebooks/config.ipynb
Byrnetp: Yeah I think it's worthwhile to show this file here. I added a code block before this one to do so.

notebooks/config.ipynb
Byrnetp: I like these suggestions and added the information here to the markdown cell. The method used here doesn't have the depth limitations corresponding to each format applied, as you said, but it does assume a section/key/value structure, and I indicate this in the updated markdown. 

I think this might create a problem for some formats, especially Shell configs, which would not be able to compare any other dictionary to one that a SHConfig object contains.

notebooks/config.ipynb
Byrnetp: Good point. I had mentioned this in a previous cell, but I think it's better to have this information here.

christinaholtNOAA: I'm sorry I missed that earlier on. Thanks for pulling it into this space for folks who don't retain info for very long, like me. :)

christinaholtNOAA (approved):
LGTM. Thanks for addressing those earlier comments.

notebooks/config.ipynb
```diff
@@ -964,12 +971,860 @@
...
+   ],
+   "source": [
+    "config.compare(\n",
+    "    config_1_path=Path('fixtures/config/base-config.nml'),\n",
+    "    config_2_path='fixtures/config/alt-config.nml',\n",
+    "    config_1_format='nml',\n",
+    "    config_2_format='nml'\n",
+    ")"
+   ]
+  },
```
NaureenBharwaniNOAA, line 1099: What do you think about adding an example of when two files are trying to be compared of differing types so users can see that error?

From the docs: "Comparisons are supported only for configs of the same format, e.g. YAML vs YAML, Fortran namelist vs Fortran namelist, etc. uw will flag invalid comparisons"

i.e. I used existing fixture files to mock up an example
```
config.compare(
    config_1_path=Path('fixtures/config/get-config.yaml'),
    config_2_path=Path('fixtures/config/base-config.nml')
)

[2024-10-16T14:54:58]    ERROR Formats do not match: yaml vs nml
```

notebooks/config.ipynb
```diff
@@ -964,12 +971,860 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "e14c0158-baa4-45d0-a86c-d9a245021229",
+   "metadata": {},
+   "source": [
+    "## Validating Configs\n",
+    "\n",
+    "The `config.validate()` function checks if a given config conforms to a specified JSON schema.\n",
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: @christinaholtNOAA should we stick to the [phrasing](https://docs.google.com/presentation/d/1_gsEC9QB_l8qYyWTeZUY_vZt9wKcd_R7Z5xo1vVuwnw/edit#slide=id.g300d7275eb6_0_243) action instead of function? There's a couple other places this comes up. The mode being `config` the action being `validate`.

notebooks/config.ipynb
Byrnetp: Yeah, I agree this is a great example to include. I added it at the end of the `config.compare()` section.

notebooks/config.ipynb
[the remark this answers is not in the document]
Byrnetp: I know from a [discussion in a previous PR](https://github.com/ufs-community/uwtools/pull/581#discussion_r1721076527) that the term "mode" wouldn't apply here since it's outside of the CLI context, but I'm not sure if the same goes for "action" or not. I'm happy to change to whichever term fits best.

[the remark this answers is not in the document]
NaureenBharwaniNOAA: Yeah that's a great point. "action" should fall into the same context here as "mode" for the CLI. I'm happy with it as is then. I'm actually curious what the best fitting term would be now. There isn't any specific language int he docs that reference it a certain way.

NaureenBharwaniNOAA (approved):
Great work on this! The notebook looks solid and having the config tool in one notebook.

elcarpenterNOAA (approved):

notebooks/config.ipynb
```diff
@@ -964,12 +971,900 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d9a330d1-35ff-4843-90fa-dd19e6eb98e9",
+   "metadata": {},
+   "source": [
+    "### Writing Configs in a Specified Format\n",
+    "\n",
+    "The each of the `config` tool's classes provide methods that write configs of their format. With the `fruits` object, which is an instance of `INIConfig`, INI configs are written. `dump()` is one of these methods, which writes the config stored in the object to a file specified by providing the `path` parameter with a <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object. If `path` is `None`, the config is written to `stdout`.\n",
```
elcarpenterNOAA, line 1753:
```suggestion
    "Each of the `config` tool's classes provide methods that write configs of their format. With the `fruits` object, which is an instance of `INIConfig`, INI configs are written. `dump()` is one of these methods, which writes the config stored in the object to a file specified by providing the `path` parameter with a <a href=\"https://docs.python.org/3/library/pathlib.html#pathlib.Path\">Path</a> object. If `path` is `None`, the config is written to `stdout`.\n",
```

PR #612, opened 2024-10-20
Fix for Rocoto streq dependencies.

christinaholtNOAA:
**Synopsis**

Fixes a bug associated with using the Rocoto `streq` dependency. When specifying the schema-approved YAML option for the dependency as in this very abbreviated example:

```
rocoto:
  task:
    dependency:
      streq:
        left: foo
        right: bar
```

The resulting XML is invalid against the Rocoto RelaxNG schema:

```
<streq/>
```
And should appear as:

```
<streq/>
  <left>foo</left>
  <right>bar</right>
<streq/>
```

To fix future instances of similar failures, I added validation checks to each config object used to test the rocoto mechanics.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/rocoto.py
+++ b/src/uwtools/rocoto.py
@@ -261,7 +261,9 @@ class _RocotoXML:
         :param config: Configuration data for the tag.
         :param tag: Name of new element to add.
         """
-        self._set_attrs(SubElement(e, tag), config)
+        e = SubElement(e, tag)
+        for k, v in config.items():
+            self._add_compound_time_string(e, v, k)
 
     def _add_task_dependency_taskdep(self, e: _Element, config: dict) -> None:
         """
--- a/src/uwtools/tests/test_rocoto.py
+++ b/src/uwtools/tests/test_rocoto.py
@@ -165,15 +165,15 @@ class Test__RocotoXML:
             },
         }
         errors = schema_validator("rocoto", "$defs")
-        taskname = "metatask_test"
-        assert not errors({taskname: config})
+        metataskname = "metatask_test"
+        assert not errors({metataskname: config})
         orig = instance._add_metatask
         with patch.multiple(instance, _add_metatask=D, _add_task=D) as mocks:
-            orig(e=root, config=config, name_attr=taskname)
+            orig(e=root, config=config, name_attr=metataskname)
         metatask = root[0]
         assert metatask.tag == "metatask"
         assert metatask.get("mode") == "parallel"
-        assert metatask.get("name") == taskname
+        assert metatask.get("name") == metataskname
         assert metatask.get("throttle") == "42"
         assert {e.get("name"): e.text for e in metatask.xpath("var")} == {"baz": "3", "qux": "4"}
         mocks["_add_metatask"].assert_called_once_with(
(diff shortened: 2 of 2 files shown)


src/uwtools/rocoto.py
```diff
@@ -261,7 +261,9 @@ def _add_task_dependency_strequality(self, e: _Element, config: dict, tag: str)
         :param config: Configuration data for the tag.
         :param tag: Name of new element to add.
         """
-        self._set_attrs(SubElement(e, tag), config)
+        e = SubElement(e, tag)
+        for k, v in config.items():
+            self._add_compound_time_string(e, v, k)
```
christinaholtNOAA, line 266: This is the fix for the `streq` piece. This dependency needs two subelements, which may contain compound time strings.

src/uwtools/tests/test_rocoto.py
```diff
@@ -119,10 +119,12 @@ def test__add_compound_time_string_basic(self, config, instance, root):
         assert child.text == str(config)
 
     def test__add_compound_time_string_cyclestr(self, instance, root):
-        config = {"cyclestr": {"attrs": {"baz": "42"}, "value": "qux"}}
+        config = {"cyclestr": {"attrs": {"offset": "00:05:00"}, "value": "qux"}}
```
christinaholtNOAA, line 122: Many of the changes you see to the config dicts in this file are to get them to pass the validation step and/or to be more realistic if others peek in here for examples.

src/uwtools/tests/test_rocoto.py
```diff
@@ -12,7 +12,7 @@
 from uwtools import rocoto
 from uwtools.config.formats.yaml import YAMLConfig
 from uwtools.exceptions import UWConfigError, UWError
-from uwtools.tests.support import fixture_path
+from uwtools.tests.support import fixture_path, schema_validator
```
maddenp-noaa, line 15: I appreciate how DRY it is to do the schema checks in this module since we already have config dicts defined here. On the other hand, we have a separate set of Rocoto schema checks in `test_schemas.py` and I think there's some value in having a comprehensive set of schema tests in one place, as it's clear where to find them, to add new ones, etc. I'm torn and am not sure I have a strong opinion. Did you land on placement of the schema checks here after considering both options?

maddenp-cu (approved):

nbharwani-cu (approved):

src/uwtools/tests/test_rocoto.py
christinaholtNOAA: I think the comprehensive set of schema checks makes sense in `test_schemas.py`, but I don't think that necessarily needs to be the only place schema checking happens. Here, it ended up being important to ensure that we're validating the small configurations that we're using to check the code, and that validation makes it a much stronger test IMO. 

Rocoto dependencies are the most complicated (I think) schemas and code to check, given that there are so many options and combinations. I think it makes sense to use targeted small subsets of dictionaries to test with like we're doing int `test_rocoto.py`, so I don't see us DRYing things out by just making one big YAML for schema tests and code tests.

maddenp-noaa: > by just making one big YAML for schema tests and code tests
Just for the record, I don't think that (one big YAML) is what's done in `test_schemas.py`: There are lots of tiny config fragments that do granular testing of sub-schemas, similar to the dicts in this module. But I have no objects to going this route if you considered both and chose this one intentionally.

Byrnetp (approved):

PR #616, opened 2024-10-22
Add default output() method to Driver base class

### LINKED ISSUE #615: Add output method for base driver class.
# AC

- [x] Add default instance `@property` method `output()` to `Driver` base class.
- [x] Define `uwtools.exceptions.UWNotImplemented` and raise it from the default `output()`.
- [x] Test default `output()` via `ConcreteDriver*` test subclass(es) in `test_driver.py`. 
- [x] Add unit test to each existing driver test module (`test_chgres_cube.py` et al.) where the tested driver subclasses `Driver`. The test should call `output()` and assert that the expected exception is raised.
- [x] Update all driver unit tests to include `output` in the list of methods inherited from the base class.

# Description

This method will serve as the default behavior for when a subclass has not yet implemented its own `output()`. It will report to the user that the method hasn't yet been implemented and will raise a `UWNotImplemented` exception.

Consider implementing, in addition to the API-callable `output()` method, something like

```
@external
def  show_output(self):
    yield self.taskname("output")
    print(json.dumps(self.output, indent=2, sort_keys=True))
    yield asset(None, lambda: True)
```

that can be called from the CLI to show the `output()` results on `stdout`. This could then be called as part of the CLI documentation generation to show example output in the CLI docs. (The doc would should be done in per-driver tickets, not here.) The output might look something like

```
>>> import json
>>> show = lambda x: print(json.dumps(x, indent=2, sort_keys=True))
>>> show("/path/to/foo")
"/path/to/foo"
>>> show(["/path/to/foo1", "/path/to/foo2"])
[
  "/path/to/foo1",
  "/path/to/foo2"
]
>>> show({"foo": ["/path/to/foo1", "/path/to/foo2"], "bar": ["/path/to/bar1", "/path/to/bar2"]})
{
  "bar": [
    "/path/to/bar1",
    "/path/to/bar2"
  ],
  "foo": [
    "/path/to/foo1",
    "/path/to/foo2"
  ]
}
```

This way, users could use the CLI to see what the output of their driver invocation will look like, and examples can be programmatically included in our CLI docs.

maddenp-noaa:
**Synopsis**

Fixes #615.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/cli/drivers/chgres_cube/help.out
+++ b/docs/sections/user_guide/cli/drivers/chgres_cube/help.out
@@ -20,5 +20,7 @@ Positional arguments:
       A run
     runscript
       The runscript
+    show_output
+      Show the output to be created by this component
     validate
       Validate the UW driver config
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -396,7 +396,7 @@ class Driver(Assets):
     @external
     def show_output(self):
         """
-        Display the output to be created by this component.
+        Show the output to be created by this component.
         """
         yield self.taskname("expected output")
         print(json.dumps(self.output, indent=2, sort_keys=True))
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -26,7 +26,7 @@ from uwtools.config.validator import (
     validate_external,
     validate_internal,
 )
-from uwtools.exceptions import UWConfigError
+from uwtools.exceptions import UWConfigError, UWNotImplementedError
 from uwtools.logging import log
 from uwtools.scheduler import JobScheduler
 from uwtools.strings import STR
@@ -416,7 +416,16 @@ class Driver(Assets):
         cmd = "{x} >{x}.out 2>&1".format(x=self._runscript_path)
         run_shell_cmd(cmd=cmd, cwd=self.rundir, log_output=True)
 
-    # Private helper methods
+    # Public methods
+
+    @property
+    def output(self) -> dict[str, Union[str, list[str]]]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        raise UWNotImplementedError("The output() method is not yet implemented for this driver")
+
+    # Private methods
 
     @property
     def _run_resources(self) -> dict[str, Any]:
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -393,6 +393,15 @@ class Driver(Assets):
         yield None
         self._write_runscript(path)
 
+    @external
+    def show_output(self):
+        """
+        Display the output to be created by this component.
+        """
+        yield self.taskname("expected output")
+        print(json.dumps(self.output, indent=2, sort_keys=True))
+        yield asset(None, lambda: True)
+
     @task
     def _run_via_batch_submission(self):
         """
--- a/src/uwtools/exceptions.py
+++ b/src/uwtools/exceptions.py
@@ -21,6 +21,12 @@ class UWConfigRealizeError(UWConfigError):
     """
 
 
+class UWNotImplementedError(UWError):
+    """
+    Exception for signaling that uwtools has not (yet) implemented something.
+    """
+
+
 class UWTemplateRenderError(UWError):
     """
     Exception for issues arising from template rendering.
--- a/src/uwtools/tests/drivers/test_cdeps.py
+++ b/src/uwtools/tests/drivers/test_cdeps.py
@@ -17,6 +17,7 @@ from pytest import fixture, mark
 from uwtools.config.formats.nml import NMLConfig
 from uwtools.drivers import cdeps
 from uwtools.drivers.cdeps import CDEPS
+from uwtools.drivers.driver import AssetsCycleBased
 from uwtools.logging import log
 from uwtools.tests.support import logged
 from uwtools.tests.test_schemas import CDEPS_CONFIG
@@ -33,6 +34,14 @@ def driverobj(tmp_path):
 
 
 # Tests
+
+
+@mark.parametrize(
+    "method",
+    ["taskname", "_validate"],
+)
+def test_CDEPS(method):
+    assert getattr(CDEPS, method) is getattr(AssetsCycleBased, method)
 
 
 def test_CDEPS_atm(driverobj):
--- a/src/uwtools/tests/drivers/test_chgres_cube.py
+++ b/src/uwtools/tests/drivers/test_chgres_cube.py
@@ -102,6 +102,7 @@ def driverobj(config, cycle):
         "_scheduler",
         "_validate",
         "_write_runscript",
+        "output",
         "run",
     ],
 )
--- a/src/uwtools/tests/drivers/test_chgres_cube.py
+++ b/src/uwtools/tests/drivers/test_chgres_cube.py
@@ -10,10 +10,11 @@ from unittest.mock import patch
 
 import f90nml  # type: ignore
 from iotaa import refs
-from pytest import fixture, mark
+from pytest import fixture, mark, raises
 
 from uwtools.drivers.chgres_cube import ChgresCube
 from uwtools.drivers.driver import Driver
+from uwtools.exceptions import UWNotImplementedError
 from uwtools.logging import log
 from uwtools.scheduler import Slurm
 from uwtools.tests.support import logged, regex_logged
@@ -138,6 +139,12 @@ def test_ChgresCube_namelist_file_missing_base_file(caplog, driverobj):
     path = Path(refs(driverobj.namelist_file()))
     assert not path.exists()
     assert regex_logged(caplog, "missing.nml: State: Not Ready (external asset)")
+
+
+def test_ChgresCube_output(driverobj):
+    with raises(UWNotImplementedError) as e:
+        assert driverobj.output
+    assert str(e.value) == "The output() method is not yet implemented for this driver"
 
 
 def test_ChgresCube_provisioned_rundir(driverobj):
--- a/src/uwtools/tests/drivers/test_driver.py
+++ b/src/uwtools/tests/drivers/test_driver.py
@@ -68,7 +68,10 @@ class ConcreteDriverCycleLeadtimeBased(Common, driver.DriverCycleLeadtimeBased):
 
 
 class ConcreteDriverTimeInvariant(Common, driver.DriverTimeInvariant):
-    pass
+
+    @property
+    def output(self):
+        return {"foo": "/path/to/foo", "bar": ["/path/to/bar1", "/path/to/bar2"]}
 
 
 def write(path, x):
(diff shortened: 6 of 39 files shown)


docs/sections/user_guide/cli/drivers/chgres_cube/help.out
```diff
@@ -20,5 +20,7 @@ Positional arguments:
       A run
     runscript
       The runscript
+    show_output
+      Show the output to be created by this component
```
maddenp-noaa, line 24: Lots of these, automatically updated by `make docs`.

src/uwtools/drivers/driver.py
```diff
@@ -393,6 +393,15 @@ def runscript(self):
...
         self._write_runscript(path)
 
+    @external
+    def show_output(self):
+        """
+        Show the output to be created by this component.
+        """
+        yield self.taskname("expected output")
+        print(json.dumps(self.output, indent=2, sort_keys=True))
+        yield asset(None, lambda: True)
```
maddenp-noaa, line 403: This task method will allow a CLI invocation like
```
uw ungrib show_output --cycle 2024-10-22T12 --config-file config.yaml
```
to display the component-output information as JSON on `stdout`, for inspection or programmatic consumption by CLI users. See the unit tests for an example.

src/uwtools/drivers/driver.py
```diff
@@ -416,7 +425,16 @@ def _run_via_local_execution(self):
...
 
-    # Private helper methods
+    # Public methods
+
+    @property
+    def output(self) -> dict[str, Union[str, list[str]]]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        raise UWNotImplementedError("The output() method is not yet implemented for this driver")
```
maddenp-noaa, line 435: The default implementation, to be overridden with a component-specific implementation in each driver module.

After thinking over our discussion from yesterday, I've made an opinionated guess about the type of the return value: I think it probably makes sense to provide some kind of description for the files described by `output`, so the return type is a `dict` where the `str` keys can be descriptions. The values might be either scalar strings, or list of strings. This would support cases like either
```
{"analysis": "/path/to/atmanl.nc"}
```
where there's just a single file, or
```
{"analyses": ["/path/to/atmanl.nc", "/path/to/sfcanl.nc"]}
```
Is this a good enough idea for now?

src/uwtools/exceptions.py
```diff
@@ -21,6 +21,12 @@ class UWConfigRealizeError(UWConfigError):
     """
 
 
+class UWNotImplementedError(UWError):
+    """
+    Exception for signaling that uwtools has not (yet) implemented something.
+    """
```
maddenp-noaa, line 27: The new exception.

src/uwtools/tests/drivers/test_cdeps.py
```diff
@@ -35,6 +36,14 @@ def driverobj(tmp_path):
 # Tests
 
 
+@mark.parametrize(
+    "method",
+    ["taskname", "_validate"],
+)
+def test_CDEPS(method):
+    assert getattr(CDEPS, method) is getattr(AssetsCycleBased, method)
```
maddenp-noaa, line 44: The other `Assets` drivers provide this test, so I added a copy to the CDEPS test module.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -101,6 +102,7 @@ def driverobj(config, cycle):
         "_scheduler",
         "_validate",
         "_write_runscript",
+        "output",
```
maddenp-noaa, line 105: This is added in many of the driver test modules that follow. This states that the `output` method is inherited from `Driver`. When a driver module eventually overrides the default, exception-raising `output` inherited from `Driver`, this test will fail and this line can be removed to make it pass.

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -140,6 +142,12 @@ def test_ChgresCube_namelist_file_missing_base_file(caplog, driverobj):
     assert regex_logged(caplog, "missing.nml: State: Not Ready (external asset)")
 
 
+def test_ChgresCube_output(driverobj):
+    with raises(UWNotImplementedError) as e:
+        assert driverobj.output
+    assert str(e.value) == "The output() method is not yet implemented for this driver"
```
maddenp-noaa, line 148: A similar test is added to each driver test module. When a driver eventually provides a custom implementation, its test will need to be updated to test the new, accurate behavior.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -68,7 +68,10 @@ class ConcreteDriverCycleLeadtimeBased(Common, driver.DriverCycleLeadtimeBased):
 
 
 class ConcreteDriverTimeInvariant(Common, driver.DriverTimeInvariant):
-    pass
+
+    @property
+    def output(self):
+        return {"foo": "/path/to/foo", "bar": ["/path/to/bar1", "/path/to/bar2"]}
```
maddenp-noaa: So that the `show_output()` task method can be realistically tested, override the default, exception-raising `output` inherited from `Driver` for this test class.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -397,6 +400,22 @@ def test_Driver_namelist_schema_default_disable(driverobj):
         assert driverobj.namelist_schema() == {"type": "object"}
 
 
+def test_Driver_output(config):
+    driverobj = ConcreteDriverTimeInvariant(config)
+    assert driverobj.output == {"foo": "/path/to/foo", "bar": ["/path/to/bar1", "/path/to/bar2"]}
```
maddenp-noaa, line 406: Since `ConcreteDriverTimeInvariant` implements a custom `output` method, we can test its return value.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -397,6 +400,22 @@ def test_Driver_namelist_schema_default_disable(driverobj):
...
+
+@mark.parametrize("cls", [ConcreteDriverCycleBased, ConcreteDriverCycleLeadtimeBased])
+def test_Driver_output_not_implemented(cls, config):
+    kwargs = {"config": config, "cycle": dt.datetime(2024, 10, 22, 12)}
+    if cls == ConcreteDriverCycleLeadtimeBased:
+        kwargs["leadtime"] = 6
+    driverobj = cls(**kwargs)
+    with raises(UWNotImplementedError) as e:
+        assert driverobj.output
+    assert str(e.value) == "The output() method is not yet implemented for this driver"
```
maddenp-noaa, line 417: `ConcreteDriverCycleBased` and `ConcreteDriverCycleLeadtimeBased` are inheriting the default, exception-raising `output` from `Driver`.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -423,6 +442,45 @@ def test_Driver_runscript(arg, driverobj, type_):
...
+    expected = """
+    {
+      "bar": [
+        "/path/to/bar1",
+        "/path/to/bar2"
+      ],
+      "foo": "/path/to/foo"
+    }
+    """
+    assert capsys.readouterr().out.strip() == dedent(expected).strip()
```
maddenp-noaa, line 457: A test of the kind of output we'd expect to see on `stdout` if a CLI command like `uw <driver> show_output ...` were issued.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -68,7 +68,11 @@ class ConcreteDriverCycleLeadtimeBased(Common, driver.DriverCycleLeadtimeBased):
 
 
 class ConcreteDriverTimeInvariant(Common, driver.DriverTimeInvariant):
-    pass
+
+    @property
+    def output(self):
+        # The dict keys are intentionally out-of-alphabetical-order to test JSON sorting.
+        return {"bar": ["/path/to/bar1", "/path/to/bar2"], "foo": "/path/to/foo"}
```
maddenp-noaa, line 75: So that the `show_output()` task method can be realistically tested, override the default, exception-raising `output` inherited from `Driver` for this test class.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -423,6 +443,45 @@ def test_Driver_runscript(arg, driverobj, type_):
...
+    if not base_file:
+        del driverobj._config["base_file"]
+    if not update_values:
+        del driverobj._config["update_values"]
+    ConcreteDriverTimeInvariant._create_user_updated_config(
+        config_class=YAMLConfig, config_values=driverobj.config, path=path
+    )
+    with open(path, "r", encoding="utf-8") as f:
+        updated = yaml.safe_load(f)
+    assert updated == expected
```
maddenp-noaa, line 482: This was just incorrectly positioned in the module -- moved up from a bit lower.

src/uwtools/tests/drivers/test_esg_grid.py
```diff
@@ -116,6 +118,12 @@ def test_ESGGrid_namelist_file_missing_base_file(caplog, driverobj):
     assert regex_logged(caplog, "missing.nml: State: Not Ready (external asset)")
 
 
+def test_ESGGrid_output(driverobj):
+    with raises(UWNotImplementedError) as e:
+        assert driverobj.output
+    assert str(e.value) == "The output() method is not yet implemented for this driver"
```
maddenp-noaa, line 124: Lots of these follow, skim a bit...

src/uwtools/drivers/driver.py
christinaholtNOAA: I do like the idea that we'd get a dict no matter what, and that we'd be able to describe groups of output, even if the key is just `output` for the simple ones (not a requirement, just an example). With the added benefit that users can easily inquire via CLI or API what the dict keys are, I think this is a fine way forward.

christinaholtNOAA (approved):
LGTM!

NaureenBharwaniNOAA (approved):
Looks good!

src/uwtools/drivers/driver.py
elcarpenterNOAA: How often will it make sense to sort files by a description? I had previously assumed that in most instances keys would be directories and the values would be the contents within those directories, e.g.  `{"/path/to/": ["atmanl.nc", "sfcanl.nc"]}`.

elcarpenterNOAA (approved):
LGTM.

src/uwtools/drivers/driver.py
elcarpenterNOAA: I probably should have used the word "organize" or "group" instead of "sort".  With regard to the keys, I can't think of an instance where it would be helpful to group files by description, which is _not_ to say it's unhelpful, I'm just not familiar with a working example. I would assume, in some instances, files are already being categorically grouped by the directory they are in. For now, I don't have enough information to form an opinion. After UPP's implementation of `output()`, I'll likely have a better idea of how helpful it is.

maddenp-noaa: Thanks for the clarification. Considering the [FV3](https://ufs-weather-model.readthedocs.io/en/ufs-v1.0.0/InputsOutputs.html#output-files) output files, for example, there are both different kinds of files, as well as multiple files in what would be groups of one type of file. [chgres_cube](https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#program-inputs-and-outputs-for-global-applications) creates one group of atmosphere files, and another group of surface files. In both these cases -- and I bet there will be more -- I think it makes sense to be able to differentiate between different groups of files.

PR #639, opened 2024-10-31
Implement UPP output() method

### LINKED ISSUE #627: Output method for UPP driver
# AC

- [ ] Implementing the output method.
- [ ] Provide the output files in an appropriate data structure (dict) as a public helper method on the object.
- [ ] Update unit tests
- [ ] Add doc block to the method that shows what the output looks like (data structure with example values filled in)
- [ ] Add something to the CLI that prints out this information in a pretty-printed format (YAML). Can use it to generate docs.

# Description

Use the config from the UPP to understand the naming convention of the output.
Can have an "unknown" option.
Returns a dictionary.
Documentation of the configured output shows up in the CLI docs. 
Doc block that shows up in API shows an unconfigured example.

maddenp-noaa:
**Synopsis**

Fixes #627.

**Type**

- [x] Documentation

**Impact**

*NB* `control_file` is now a required and schema-checked item in the UPP config.

**Checklist**

### DIFF
--- a/src/uwtools/drivers/upp.py
+++ b/src/uwtools/drivers/upp.py
@@ -2,13 +2,15 @@
 A driver for UPP.
 """
 
+import math
 from pathlib import Path
 
 from iotaa import asset, task, tasks
 
 from uwtools.config.formats.nml import NMLConfig
-from uwtools.drivers.driver import DriverCycleLeadtimeBased
+from uwtools.drivers.driver import DriverCycleLeadtimeBased, OutputT
 from uwtools.drivers.support import set_driver_docstring
+from uwtools.exceptions import UWConfigError
 from uwtools.strings import STR
 from uwtools.utils.tasks import file, filecopy, symlink
 
@@ -91,6 +93,40 @@ class UPP(DriverCycleLeadtimeBased):
         The name of this driver.
         """
         return STR.upp
+
+    @property
+    def output(self) -> OutputT:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        # Define facts specific to the supported UPP version.
+        FIELDS_PER_BLOCK = 16
+        GEN_PROC_TYPE_IDX = 8
+        PARAMS_PER_VAR = 42
+        # Derive values from the current driver config. GRIB output filename suffixes include the
+        # forecast leadtime, zero-padded to at least 2 digits (more if necessary). Avoid taking the
+        # log of zero.
+        cf = self.config["control_file"]
+        leadtime = int(self.leadtime.total_seconds() / 3600)
+        suffix = ".GrbF%0{}d".format(max(2, int(math.log10(leadtime or 1)) + 1)) % leadtime
+        # Read the control file into an array of lines. Get the number of blocks (one per output
+        # GRIB file) and the number of variables per block. For each block, construct a filename
+        # from the block's identifier and the suffix defined above.
+        try:
+            with open(cf, "r", encoding="utf-8") as f:
+                cfg = f.read().split("\n")
+        except (FileNotFoundError, PermissionError) as e:
+            raise UWConfigError(f"Could not open UPP control file {cf}") from e
+        nblocks, cfg = int(cfg[0]), cfg[1:]
+        nvars, cfg = list(map(int, cfg[:nblocks])), cfg[nblocks:]
+        paths = []
+        for _ in range(nblocks):
+            identifier = cfg[0]
+            paths.append(str(self.rundir / (identifier + suffix)))
+            fields, cfg = cfg[:FIELDS_PER_BLOCK], cfg[FIELDS_PER_BLOCK:]
+            _, cfg = (cfg[0], cfg[1:]) if fields[GEN_PROC_TYPE_IDX] == "ens_fcst" else (None, cfg)
+            cfg = cfg[PARAMS_PER_VAR * nvars.pop() :]
+        return {"gribfiles": paths}
 
     # Private helper methods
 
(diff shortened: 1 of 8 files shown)


src/uwtools/tests/drivers/test_upp.py
```diff
@@ -160,10 +160,34 @@ def test_UPP_namelist_file_missing_base_file(caplog, driverobj):
...
+        "2",                # number of blocks
+        "1",                # number variables in 2nd block
+        "2",                # number variables in 1st block
+        "FOO",              # identifier of 1st block
+        *fields,            # fields of 1st block
+        *(parameters * 2) , # variable parameters of 1st block
+        "BAR",              # identifier of 2nd block
+        *fields,            # fields of 2nd block
+        *parameters,        # variable parameters of 2nd block
+    ]
```
maddenp-noaa, line 177: This is a mock-up of the lines we'd find in a `postxconfig-NT.txt` file.

WeirAE: The commenting of the data is particularly appreciated here

src/uwtools/drivers/upp.py
```diff
@@ -81,6 +100,40 @@ def driver_name(cls) -> str:
...
+            identifier = lines[0]
+            paths.append(str(self.rundir / (identifier + suffix)))
+            fields, lines = lines[: self.NFIELDS], lines[self.NFIELDS :]
+            _, lines = (
+                (lines[0], lines[1:])
+                if fields[self.GENPROCTYPE_IDX] == "ens_fcst"
+                else (None, lines)
+            )
+            lines = lines[self.NPARAMS * nvars.pop() :]
+        return {"gribfiles": paths}
```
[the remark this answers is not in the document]
WeirAE: The amount that this is clearer than the source Fortran code cannot be understated

src/uwtools/drivers/upp.py
```diff
@@ -81,6 +100,40 @@ def driver_name(cls) -> str:
...
+    def output(self) -> OutputT:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        # Derive values from the current driver config. GRIB output filename suffixes include the
+        # forecast leadtime, zero-padded to at least 2 digits (more if necessary). Avoid taking the
+        # log of zero.
+        cf = self.config["control_file"]
+        leadtime = int(self.leadtime.total_seconds() / 3600)
+        suffix = ".GrbF%0{}d".format(max(2, int(log10(leadtime or 1)) + 1)) % leadtime
```
christinaholtNOAA, line 113: Since it's only zero-padded to 2 digits (not 3), will this (or something like it) not work in Python?
```
suffix = f".GrbF{leadtime:02d}"
```
For example:
```
>>> leadtime = 1
>>> print(f".GrbF{leadtime:02d}")
.GrbF01
>>> leadtime = 10
>>> print(f".GrbF{leadtime:02d}")
.GrbF10
>>> leadtime = 100
>>> print(f".GrbF{leadtime:02d}")
.GrbF100
```

src/uwtools/drivers/upp.py
```diff
@@ -81,6 +100,40 @@ def driver_name(cls) -> str:
...
+        # Read the control file into an array of lines. Get the number of blocks (one per output
+        # GRIB file) and the number of variables per block. For each block, construct a filename
+        # from the block's identifier and the suffix defined above.
+        try:
+            with open(cf, "r", encoding="utf-8") as f:
+                lines = f.read().split("\n")
```
christinaholtNOAA, line 119: Will this file be long enough to slow the script down if we load the whole thing?

Also, is there any benefit to `f.read().split("\n")` vs `f.readlines()`?

src/uwtools/drivers/upp.py
```diff
@@ -81,6 +100,40 @@ def driver_name(cls) -> str:
...
+        except (FileNotFoundError, PermissionError) as e:
+            raise UWConfigError(f"Could not open UPP control file {cf}") from e
+        nblocks, lines = int(lines[0]), lines[1:]
```
christinaholtNOAA, line 122: Wouldn't it be nice to `[h | rest]` this?

christinaholtNOAA (approved):
This looks good to me. 

Just a few comments below.

src/uwtools/drivers/upp.py
maddenp-noaa: Ooh, nice observation: We'd never have more than one leading zero. Let me give that a try.

src/uwtools/drivers/upp.py
maddenp-noaa: I don't think speed will be an issue.

`.readlines()` leaves the newline character on each line, which then has to be dealt with separately; `.read().split()` discards the newlines at once.

src/uwtools/drivers/upp.py
maddenp-noaa: Yes! I went looking for something suitable in the standard library and was momentarily excited about `isplit()` from `itertools`, but it doesn't do what I wanted. So I temporarily had
```
split = lambda x, i: (x[:i], x[i:])
```
but that returns two `list` objects, which I want in some places, but in others I want the scalar head value, and sometimes I want to apply a function to the result... so extra abstraction didn't seem worth it. But, someday: https://github.com/pytoolz/toolz

src/uwtools/drivers/upp.py
maddenp-noaa: I had to test it to find out whether I was using an old bad habit or an old good habit. Fortunately it was the latter.

Byrnetp (approved):
LGTM

elcarpenterNOAA (approved):

WeirAE (approved):

NaureenBharwaniNOAA (approved):

PR #641, opened 2024-11-01
Add a leadtime argument to the chgres_cube driver.

### LINKED ISSUE #640: The `chgres_cube` driver needs a `--leadtime` argument.
If this is a bug, please consider using the [Bug Report](./bug_report.md) template.
Your issue may already have been created!

Please look through the existing issues on the [Issue tracker](https://github.com/ufs-community/uwtools/issues) to see if this issue has been created.
If so, please consider using that Issue to add any additional information.

**Description**

When running the `chgres_cube` driver to create lateral boundary conditions, the namelist expects the date information to be consistent with the valid-time of the forecast file being processed, however, these forecast times are prepared for a given `cycle`, so we will need to pass both pieces of information to get the expected behavior -- the `rundir` has `cycle` information, but the namelist has `cycle + leadtime` information.

**Requirements**

**Describe alternatives you've considered**

This could be handled as an environment variable, but would be horribly inconsistent with the other drivers that do have an optional lead time argument.

**Acceptance Criteria (Definition of Done)**

- [ ] The chgres_cube driver can be appropriately configured for both initial and lateral boundary condition tasks with output behavior similar to that in existing apps.
- [ ] Docs and tests updated to support the new argument.

christinaholtNOAA:
**Synopsis**

Fixes #640 

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/drivers/test_chgres_cube.py
+++ b/src/uwtools/tests/drivers/test_chgres_cube.py
@@ -81,8 +81,13 @@ def config(tmp_path):
 
 
 @fixture
-def driverobj(config, cycle):
-    return ChgresCube(config=config, cycle=cycle, batch=True)
+def driverobj(config, cycle, leadtime):
+    return ChgresCube(config=config, cycle=cycle, leadtime=leadtime, batch=True)
+
+
+@fixture
+def leadtime():
+    return dt.timedelta(hours=24)
 
 
 # Tests
@@ -157,4 +162,4 @@ def test_ChgresCube_runscript(driverobj):
 
 
 def test_ChgresCube_taskname(driverobj):
-    assert driverobj.taskname("foo") == "20240201 18Z chgres_cube foo"
+    assert driverobj.taskname("foo") == "20240202 18:00:00 chgres_cube foo"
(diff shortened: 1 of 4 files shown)


elcarpenterNOAA (approved):

WeirAE (approved):
Looks clean and should help to smooth out the SRW code too, LGTM

nbharwani-cu (approved):

src/uwtools/tests/drivers/test_chgres_cube.py
```diff
@@ -165,4 +170,4 @@ def test_ChgresCube_runscript(driverobj):
 
 
 def test_ChgresCube_taskname(driverobj):
-    assert driverobj.taskname("foo") == "20240201 18Z chgres_cube foo"
+    assert driverobj.taskname("foo") == "20240202 18:00:00 chgres_cube foo"
```
maddenp-noaa, line 173: I know that this output is from existing behavior, but seeing it now I wonder if it's what we really want for a cycle + leadtime driver. This gives a valid time for the data, but the connection to the cycle is lost. I mean, alternatively, we could have `20240201 18Z 24:00:00`, i.e. this is the Feb 1 18Z cycle, with a 24-hour leadtime. On the other hand it _does_ seem handy to not have to calculate the valid time -- like what time does the hour-97 forecast of some cycle actually correspond to? I can see it both ways, and am not asking for a change, just musing.

maddenp-cu (approved):

Byrnetp (approved):

src/uwtools/tests/drivers/test_chgres_cube.py
christinaholtNOAA: I could see reporting `20240201 18Z 24:00:00` as an improvement. We typically don't care about valid time in these logs. I think the only tasks that really care about valid time for user-facing aspects is verification.

PR #649, opened 2024-11-13
Relax jsonschema pin

maddenp-noaa:
**Synopsis**

Relax the pin on `jsonschema` to permit the range 4.18 through 4.23, inclusive. This supports the installation of `uwtools` into environments with packages that also require `jsonschema`, pinned to a version in this range. (Specifically, a user wanted to install `uwtools` alongside the Globus Compute package, which pins to `jsonschema` 4.19.) Each minor release version was manually tested, and unit tests were updated to account for the different error messages emitted by `jsonschema` for different subsets of the range.

Probably other packages could get similar treatment (relaxed pinning), but more testing would be needed, and this this addresses the immediate blocker for our user.

I plan to do a same-version, incremented-build release based on this change.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/test_schemas.py
+++ b/src/uwtools/tests/test_schemas.py
@@ -300,20 +300,20 @@ def ww3_prop():
 # Helpers
 
 
-def non_empty_dict(s: str, errors: list[str]) -> bool:
+def non_empty_dict(errors: list[str]) -> bool:
     for msg in [
-        f"{s} does not have enough properties",  # jsonschema [???,4.20.*]
-        f"{s} should be non-empty",  # jsonschema [4.21.*,???]
+        "{} does not have enough properties",  # jsonschema [???,4.20.*]
+        "{} should be non-empty",  # jsonschema [4.21.*,???]
     ]:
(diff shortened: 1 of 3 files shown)


src/uwtools/tests/test_schemas.py
```diff
@@ -297,6 +297,29 @@ def ww3_prop():
...
+
+
+def non_empty_list(errors: list[str]) -> bool:
+    for msg in [
+        "[] is too short",  # jsonschema [4.18.*,4.20.*]
+        "[] should be non-empty",  # jsonschema [4.21.*,?]
+    ]:
+        if msg in errors:
+            return True
+    return False
```
maddenp-noaa, line 320: If I'd found a way to detect the version of `jsonschema` currently installed, I would have made the expected error message conditional on that. But I couldn't find a straightforward way to detect the version at run time.

christinaholtNOAA (approved):

NaureenBharwaniNOAA (approved):

Byrnetp (approved):

PR #652, opened 2024-11-13
Update jsonschema pin range and increment build number

maddenp-noaa:
**Synopsis**

Update build number to 1 for version 2.4.2, with a relaxed pin range for `jsonschema`.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/docs/environment.yml
+++ b/docs/environment.yml
@@ -2,6 +2,7 @@ name: readthedocs
 channels:
   - conda-forge
 dependencies:
+  - python<3.13
   - sphinx_rtd_theme=2.0.*
   - sphinxcontrib-bibtex=2.6.*
   - tree
--- a/recipe/meta.json
+++ b/recipe/meta.json
@@ -1,6 +1,6 @@
 {
-  "build": "py_0",
-  "buildnum": "0",
+  "build": "py_1",
+  "buildnum": "1",
   "name": "uwtools",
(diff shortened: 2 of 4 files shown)


docs/environment.yml
```diff
@@ -2,6 +2,7 @@ name: readthedocs
 channels:
   - conda-forge
 dependencies:
+  - python<3.13
```
maddenp-noaa, line 5: Some bit of Sphinx/RtD code uses the Python standard-library `imghdr` module, which was deprecated by [this PEP](https://peps.python.org/pep-0594/#imghdr) and has been removed in Python 3.13, which RtD is using for its builds by default. It's possible that updating/relaxing the pins on `sphinx_rtd_theme` and `sphinxcontrib-bibtex`, below, might also fix this, as newer versions of those packages might deal with the `imghdr` issue. But I think this is the simplest immediate fix for the issue. We will have a number of other tasks to attend to (probably including relaxing some dependency-package pins in `meta.yaml`) when we extend `uwtools` support to include Python 3.13, so I propose that we deal with them all at once, maybe in our next `uwtools` release.

christinaholtNOAA (approved):

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

PR #663, opened 2024-11-27
Omnibus 2024-11-27

maddenp-noaa:
**Synopsis**

- Update `.github` such that the _Test_ workflow will be run for PRs or pushes to all branches. This means that tests will run for updates on release branches, which seems like it would have been a good idea to always be doing.
- Add `setuptools` to conda build requirements, as it appears not to be automatically installed with Python 3.13 now in CI jobs, but is required for the `uwtools` build.
- Some updated documentation about the `!include` tag.
- Some trivial doc changes per `make docs`.
- Fix a bug whereby Jinja2 expressions referencing tagged YAML values would render to strings including the tag as part of the string. The fix involves detecting tagged values and raising an exception to delay rendering the value until a later iteration of the dereferencing process, when the tagged value will have been converted to its simple type (`int`, `bool`, etc.). See below (and the new unit tests) for examples.

Given `config.yaml`
```
a: 2
b: 7
foo: !int "{{ a * b }}"
bar: !int "{{ foo }}"
```
Old behavior:
```
$ uw config realize -i config.yaml --output-format yaml
a: 2
b: 7
foo: 14
bar: !int '!int 14'
```
New behavior:
```
$ uw config realize -i config.yaml --output-format yaml
a: 2
b: 7
foo: 14
bar: 14
```

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/config/formats/yaml.py
```diff
@@ -157,13 +156,10 @@ def _yaml_include(self, loader: yaml.Loader, node: yaml.SequenceNode) -> dict:
     @property
     def _yaml_loader(self) -> type[yaml.SafeLoader]:
         """
-        The loader, with appropriate constructors added.
+        A loader with all UW constructors added.
         """
-        loader = yaml.SafeLoader
+        loader = uw_yaml_loader()
```
maddenp-noaa, line 161: The basic loader, handling tags `!bool`, `!datetime`, `!float`, and `!int`, has been factored out to `uwtools.config.support` so that it can be used by code outside `YAMLConfig`. This method obtains that loader, then adds `!include` support. It seemed far from straightforward to factor `!include` support out of `YAMLConfig`, and I think there are open questions about how the `!include` tag would interact with other features (e.g. an `!include`-tagged value that loads values tagged with other tags like `!int` et al, how these interact with Jinja2 value rendering, etc.) I think we'll have to cross that bridge when we come to it.

src/uwtools/config/jinja2.py
```diff
@@ -266,6 +269,9 @@ def _deref_render(val: str, context: dict, local: Optional[dict] = None) -> str:
     context = {**(local or {}), **context}
     try:
         rendered = _register_filters(env).from_string(val).render(context)
+        if isinstance(yaml.load(rendered, Loader=uw_yaml_loader()), UWYAMLConvert):
+            _deref_debug("Held", rendered)
+            raise UWConfigRealizeError()
```
maddenp-noaa, line 274: Here's the crux of the bugfix: If the rendered value turns out to be something (e.g. `!int '{{ foo }}'`) that loads as a `UWYAMLConvert` object, reject the rendered value, and hold the old value, by raising an exception. The value may be rendered by a future iteration of the dereferencing loop when the value it refers to has been reduced to a basic value (a `bool`, `int`, etc.).

src/uwtools/config/formats/yaml.py
maddenp-noaa: I did a basic test involving tagged values and `!include` and things seem to work as I'd expect:

`numbers.yaml`
```
e_plus_half_pi: !float '{{ values.e + half_pi }}'
half_pi: !float '{{ values.pi / 2 }}'
values: !include [e.yaml, pi.yaml]
```
`e.yaml`
```
e: 2.718
```
`pi.yaml`
```
pi: 3.141
```
And
```
$ uw config realize --input-file numbers.yaml --output-format yaml
e_plus_half_pi: 4.2885
half_pi: 1.5705
values:
  e: 2.718
  pi: 3.141
```

docs/sections/user_guide/yaml/tags.rst
```diff
@@ -69,49 +69,76 @@ Converts the tagged node to a Python ``float`` value. For example, given ``input
...
-   f2: 11
-   f2: 140
+   $ uw config realize --input-file numbers.yaml --output-format yaml
+   values:
+     e: 2.718
+     pi: 3.141
 
-``!include``
-^^^^^^^^^^^^
+Values from files later in the sequence overwrite their predecessors, and full-value replacement, not structural merging, is performed. For example, giben ``numbers.yaml``:
```
elcarpenterNOAA, line 98:
```suggestion
Values from files later in the sequence overwrite their predecessors, and full-value replacement, not structural merging, is performed. For example, given ``numbers.yaml``:
```

elcarpenterNOAA (approved):

src/uwtools/config/formats/yaml.py
christinaholtNOAA: I'm not sure this is consistent with what is documented. In the docs from this PR, I see:

```
$ uw config realize --input-file numbers.yaml --output-format yaml
values:
  constants:
    pi: 3.141
```

As the result from `values: !include [e.yaml, pi.yaml]`.

docs/sections/user_guide/yaml/tags.rst
```diff
@@ -69,49 +69,76 @@ Converts the tagged node to a Python ``float`` value. For example, given ``input
...
 
-and ``supplemental.yaml``:
+and ``pi.yaml``:
 
 .. code-block:: yaml
 
-   e: 2.718
-   pi: 3.141
+   constants:
+     pi: 3.141
```
christinaholtNOAA, line 116: Should this have `e: 2.718` included based on PR comments below?

christinaholtNOAA (approved):
LGTM. This makes for a much cleaner user experience than my workaround.

Maybe a small docs change is needed?

src/uwtools/config/formats/yaml.py
maddenp-noaa: I see what you mean, and it took me a minute, but I think it's consistent:

The situation in the documented case is that, under `values:`, there's a repeated `constants:` key in both `e.yaml` and `pi.yaml`, so `constants: e: 2.718` gets clobbered by `constants: pi: 3.141`, leaving only the latter.

The situation in my comment above is that there is no repeated key under `values:`, so `e` and `pi` can coexist.

PR #671, opened 2024-12-11
Render fix

maddenp-noaa:
**Synopsis**

Fixes a bug whereby a value like `&XMLENTITY;` that cannot be parsed as YAML on its own would lead to a dereferencing failure. See the new unit test for an example.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/config/test_jinja2.py
+++ b/src/uwtools/tests/config/test_jinja2.py
@@ -339,6 +339,14 @@ def test__deref_render_ok(caplog, deref_render_assets):
     assert not regex_logged(caplog, "Rendering failed")
 
 
+def test__deref_render_unloadable_val(caplog):
+    log.setLevel(logging.DEBUG)
+    val = "&PARTITION_DEFAULT;"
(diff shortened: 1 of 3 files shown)


src/uwtools/tests/config/test_jinja2.py
```diff
@@ -309,40 +315,47 @@ def test__deref_convert_ok(caplog, converted, tag, value):
...
     assert regex_logged(caplog, "Rendered")
     assert not regex_logged(caplog, "Rendering failed")
 
 
+def test__deref_render_unloadable_val(caplog):
+    log.setLevel(logging.DEBUG)
+    val = "&XMLENTITY;"
+    assert jinja2._deref_render(val='{{ "%s" if True }}' % val, context={}) == val
+    assert regex_logged(caplog, "Rendered")
+    assert not regex_logged(caplog, "Rendering failed")
```
maddenp-noaa, line 347: A new unit test to cover the previously failing case.

christinaholtNOAA (approved):
Thanks so much for the quick fix!!

PR #674, opened 2024-12-12
Add an output method for the chgres_cube driver

### LINKED ISSUE #636: Output method for chgres_cube driver.
# AC

# Description

Best guess at appropriate data structures for the chgres_cube driver.

No need to hook into CLI.

Byrnetp:
**Synopsis**

Fixes #636.

**Type**

**Impact**

**Checklist**


src/uwtools/drivers/chgres_cube.py
```diff
@@ -110,5 +110,14 @@ def driver_name(cls) -> str:
...
         return STR.chgrescube
 
+    @property
+    def output(self) -> dict[str, list[str]]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        output_types = ["atm", "sfc"]
+        paths = [f"{self.rundir}/out.{t}.tile7.nc" for t in output_types]
+        return {"NetCDFfiles": paths}
```
maddenp-noaa, line 120:
```suggestion
        return {"netcdffiles": paths}
```
We generally don't use mixed case in keys where possible. This would be more like the `gribfiles` key used by `UPP.output()`, though now I wonder if they should be `grib_files` and `netcdf_files`, or just `grib` and `netcdf`, or... In any case, if we at least follow our naming convention here, we can circle back to all these `output()` methods once we have a bunch of them -- and before we start calling them from other code -- and see if a good pattern/convention has become obvious.

maddenp-noaa (approved):
Looks good. I made one small suggestion.

PR #682, opened 2024-12-27
Add !dict and !list UW YAML tags 

### LINKED ISSUE #676: Add !list and !dict UW YAML tags
# Description

Add support for `!list` and `!dict` UW YAML tags to convert quoted values, e.g. Jinja2 expressions, into Python `list` and `dict` objects. Among other use cases, this may help avoid the use of wildcards in file paths in e.g. `fs` mode configs, while keeping them DRY.

For example `config.yaml`
```
as: !list "[{% for n in range(3) %} a{{ n }},{% endfor %} ]"
bs: !dict "{{ '{' }}{% for n in range(3) %} b{{ n }}: {{ n }},{% endfor %}{{ '}' }}"
cs: !dict "[{% for n in range(3) %} ['c{{ n }}',{{ n }}],{% endfor %} ]"
```
would currently be rendered as
```
$ uw template render --input-file config.yaml
as: !list "[ a0, a1, a2, ]"
bs: !dict "{ b0: 0, b1: 1, b2: 2,}"
cs: !dict "[ ['c0',0], ['c1',1], ['c2',2], ]"
```
Recall that, since `config realize` renders values in key-value configs on their own, this is a realistic view of a potential intermediate state of the config during dereferencing. After implementation of the new tags, these values would then be converted to primitive Python values like
```
>>> list(yaml.safe_load("[ a0, a1, a2, ]"))
['a0', 'a1', 'a2']
>>> dict(yaml.safe_load("{ b0: 0, b1: 1, b2: 2,}"))
{'b0': 0, 'b1': 1, 'b2': 2}
>>> dict(yaml.safe_load("[ ['c0',0], ['c1',1], ['c2',2], ]"))
{'c0': 0, 'c1': 1, 'c2': 2}
```
Note that the final values of `bs` and `cs` are equivalent, but the list-of-lists representation in the original Jinja2 expression was more concise, due to not having to escape braces.

# AC

- [ ] Implement tags.
- [ ] Add/update unit tests.
- [ ] Update documentation.

elcarpenterNOAA:
**Synopsis**

Fixes #676. 

- Adds support for `!dict` and `!list` UW YAML tags. 
- Expands evaluation of `!bool` tag to Python's truthiness standards.  

**Type**

**Impact**

**Checklist**


src/uwtools/tests/config/test_jinja2.py
```diff
@@ -287,11 +287,20 @@ def test_unrendered(s, status):
...
-def test__deref_convert_no(caplog, tag):
+@mark.parametrize(
+    "tag,value",
+    [
+        ("!datetime", "foo"),
+        ("!dict", "foo"),
+        ("!float", "foo"),
+        ("!int", "foo"),
+        ("!list", "null"),
+    ],
```
elcarpenterNOAA, line 298: `foo` _is_ an acceptable value using the `!list` tag because it gets converted into the list `['f', 'o', 'o']`, so I used `@mark.parametrize` to include an alternative bad case. The `!bool` tag was removed from this test because all objects in Python can be converted into `bool` values.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -301,9 +310,15 @@ def test__deref_convert_no(caplog, tag):
...
         (True, "!bool", "True"),
+        (False, "!bool", "0"),
         (datetime(2024, 9, 9, 0, 0), "!datetime", "2024-09-09 00:00:00"),
+        ({"a": 0, "b": 1}, "!dict", "{a: 0, b: 1}"),
+        ({"a": 0, "b": 1}, "!dict", "[[a, 0], [b, 1]]"),
         (3.14, "!float", "3.14"),
         (42, "!int", "42"),
+        ([0, 1, 2], "!list", "[0, 1, 2]"),
+        (["f", "o", "o"], "!list", "foo"),
+        ([0, 1, 2], "!list", "{0: a, 1: b, 2: c}"),
```
elcarpenterNOAA, line 321: These examples were added to document the functionality of the `!list` tag.

src/uwtools/tests/config/test_support.py
```diff
@@ -85,14 +85,9 @@ def loader(self):
...
     # These tests bypass YAML parsing, constructing nodes with explicit string values. They then
-    # demonstrate that those nodes' convert() methods return representations in type type specified
+    # demonstrate that those nodes' convert() methods return representations in the type specified
     # by the tag.
 
-    def test_bool_bad(self, loader):
-        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!bool", value="foo"))
-        with raises(KeyError):
-            ts.convert()
-
```
elcarpenterNOAA, line 95: Since there are no values in Python that can't be converted into a `bool` object, I removed this test.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -213,16 +213,25 @@ def test_dereference(tmp_path):
...
 e:
-  - !int '42'
-  - !float '3.14'
-  - !datetime '{{ D }}'
   - !bool "False"
+  - !datetime '{{ D }}'
+  - !float '3.14'
+  - !int '42'
+  - !dict "{ b0: 0, b1: 1, b2: 2,}"
+  - !dict "[ ['c0',0], ['c1',1], ['c2',2], ]"
```
NaureenBharwaniNOAA, line 221: Alphabetical ordering to match other sections. Unless there's more of a reason it has to be in this order. In that case, I'm curious to why that is needed.
```suggestion
  - !dict "{ b0: 0, b1: 1, b2: 2,}"
  - !dict "[ ['c0',0], ['c1',1], ['c2',2], ]"
    - !int '42'
```

NaureenBharwaniNOAA (commented):
Looking good! Just one comment on ordering.

docs/sections/user_guide/yaml/tags.rst
```diff
@@ -58,6 +60,24 @@ Converts the tagged node to a Python ``datetime`` object. For example, given ``i
...
+   d1: {'a0': 0, 'b1': 1, 'c2': 2}
+   d2: !dict "{{ '{' }}{% for n in range(3) %} b{{ n }}: {{ n }},{% endfor %}{{ '}' }}"
+   d3: !dict "{ b0: 0, b1: 1, b2: 2,}"
+
+.. code-block:: text
+
+   $ uw config realize --input-file input.yaml --output-format yaml
+   d1: {'a0': 0, 'b1': 1, 'c2': 2}
+   d2: {'b0': 0, 'b1': 1, 'b2': 2}
+   d3: {'c0': 0, 'c1': 1, 'c2': 2}
```
maddenp-noaa, line 79:
```suggestion
   d1: {'k0': 0, 'k1': 1, 'k2': 2}
   d2: !dict "{ k0: 0, k1: 1, k2: 2 }"
   d3: !dict "{{ '{' }}{% for n in range(3) %} k{{ n }}:{{ n }},{% endfor %}{{ '}' }}"
   d4: !dict "[{% for n in range(3) %}[k{{ n }},{{ n }}],{% endfor %}]"

.. code-block:: text

   $ uw config realize --input-file input.yaml --output-format yaml
   d1: {'k0': 0, 'k1': 1, 'k2': 2}
   d2: {'k0': 0, 'k1': 1, 'k2': 2}
   d3: {'k0': 0, 'k1': 1, 'k2': 2}
   d4: {'k0': 0, 'k1': 1, 'k2': 2}
```
I thought it might be nice to have the tuple example. And I thought it might be clearer to show that this is just four ways of expressing the same `dict` value, by using the same keys for each `dict`. I got slightly visually confused seeing `a`, `b`, and `c` keys for `d1` but only `b` keys for `d2` and only `c` keys for `d3`.

docs/sections/user_guide/yaml/tags.rst
```diff
@@ -140,6 +160,24 @@ Converts the tagged node to a Python ``int`` value. For example, given ``input.y
...
+   l1: [1, 2, 3]
+   l2: !list "[{% for n in range(3) %} a{{ n }},{% endfor %} ]"
+   l3: !list "[ a, b, c, ]"
+
+.. code-block:: text
+
+   $ uw config realize --input-file input.yaml --output-format yaml
+   l1: [1, 2, 3]
+   l2: ['a0', 'a1', 'a2']
+   l3: ['a', 'b', 'c']
```
maddenp-noaa, line 179: As above, I wonder if it would be clearer to produce the same list three different ways to make it obvious that these are equivalent. Your call on both.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -237,10 +246,28 @@ def test_dereference(tmp_path):
...
+            "f3": {"b0": 0, "b1": 1, "b2": 2},
+            "f4": {"c0": 0, "c1": 1, "c2": 2},
+            "f5": 42,
+            "f6": [0, 1, 2],
+        },
         "g": True,
+        "h": False,
         "D": datetime.fromisoformat("2024-10-10 00:19:00"),
+        "I": {"b0": 0, "b1": 1, "b2": 2},
+        "L": ["a0", "a1", "a2"],
```
maddenp-noaa, line 270: I'm not sure what happened prior to this PR with the key names: I think they were probably initially meant to be a just meaningless march through the alphabet, but then `D` and `N` showed up with apparent meaning and now I'm not sure it'll be clear to maintainers why to choose a particular key name. Would it make sense to change `D` to `i`, `I` to `j`, and `L`, to `k` to restore the `a` -> `z` progression?

src/uwtools/tests/config/test_support.py
```diff
@@ -110,6 +105,16 @@ def test_datetime_ok(self, loader):
         assert ts.convert() == datetime(2024, 8, 9, 12, 22, 42)
         self.comp(ts, "!datetime '2024-08-09 12:22:42'")
 
+    def test_dict_no(self, loader):
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!dict", value="42"))
```
maddenp-noaa, line 109: I'm puzzled about why the existing tests (and, following convention, this one) use `ts` as a variable name for the `UWYAMLConvert` value. Any idea?

src/uwtools/tests/config/test_support.py
```diff
@@ -130,6 +135,16 @@ def test_int_ok(self, loader):
...
 
+    def test_list_no(self, loader):
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!list", value="null"))
+        with raises(TypeError):
+            ts.convert()
+
+    def test_list_ok(self, loader):
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!list", value="[1, 2, 3,]"))
+        assert ts.convert() == [1, 2, 3]
+        self.comp(ts, "!list '[1, 2, 3,]'")
```
maddenp-noaa, line 146: Just a curiosity: I'm surprised, seeing this test, that when we re-serialize this `UWYAMLConvert` value to YAML, we retain the trailing comma in `3,`. I would have thought that would be normalized to the simplest possible form, which wouldn't include the comma. I'm sure this has nothing to do with this PR, but is probably something that could be improved in the `UWYAMLConvert` class. I put it on my todo list to take a look someday.

maddenp-noaa (approved):
Looks great. Just a handful of minor comments.

src/uwtools/tests/config/test_support.py
elcarpenterNOAA: Hmm. My best guess would be "tagged scalars", but in that case we should be using `MappingNode` and `SequenceNode` for `!dict` and `!list` tags, right? The `UWYAMLTag` class claims to only deal with `yaml.nodes.ScalarNode`, but that's no longer true.

docs/sections/user_guide/yaml/tags.rst
elcarpenterNOAA: I updated `l3` to match `l2`, but left `l1` as-is to show that it handles `int` values as well.

src/uwtools/tests/config/test_support.py
maddenp-noaa: Interesting. In the context of the tests, it looks like constructing `ScalarNode` objects is still correct, because as far as I can tell all the values are `str`s, which are scalars from YAML's perspective.

But maybe the `yaml.nodes.ScalarNode` types used in the constructor signature and `.represent()` return type in class `UWYAMLTag` are too specific. For example, [the doc update](https://github.com/ufs-community/uwtools/pull/682/files#diff-666152b3246d4e22e7eb4f6fae168820d066d542db5ce3566490d8295e64c690R35) shows an `int` value tagged  `!bool`, and although this is still a `ScalarNode`, I suppose that instead of `!bool 0` we could have `!bool []` that should end up as `False`, or `!bool [1, 2, 3]` that should end up as `True`, and I guess those YAML sequence / Python `list` values would represented by `SequenceNode` as you say, or similarly `MappingNode` for a `dict`-like value.

But... I just tried `!bool []` on your branch and it doesn't work. I'm looking to find out why, but I don't think it's necessarily a bad idea if we restricted the use of these tags to `str` values -- which is kind of the point: You've got a thing that unfortunately has to be a `str` because e.g. it contained a `{{ ... }}` Jinja2 expression that had to be quoted, and you want to turn it into something other than a `str`. But, if we want to enforce this, we should probably not show a `!bool 0` example even though it works, but rather `!bool "0"`, and we should probably raise an exception somewhere if users tag a non-`str` value, to avoid more difficult to debug issues that might happen later.

I'll return here as soon as I know a little more, but please share any thoughts you have in the meantime.

src/uwtools/config/formats/base.py
```diff
@@ -87,8 +86,8 @@ def _compare_config_get_lines(d: dict) -> list[str]:
         :param d: A dict object.
         """
         sio = StringIO()
-        yaml.safe_dump(d, stream=sio, default_flow_style=False, indent=2, width=inf)
-        return sio.getvalue().splitlines(keepends=True)
+        sio.write(dict_to_yaml_str(d, sort=True))
```
christinaholtNOAA, line 89: Nice.

src/uwtools/config/support.py
```diff
@@ -119,23 +121,45 @@ class UWYAMLConvert(UWYAMLTag):
     method. See the pyyaml documentation for details.
     """
 
-    TAGS = ("!bool", "!datetime", "!float", "!int")
+    TAGS = ("!bool", "!datetime", "!dict", "!float", "!int", "!list")
+    ValT = Union[bool, datetime, dict, float, int, list]
```
christinaholtNOAA, line 125: I'm curious if there might be a better name than `ValT`.

src/uwtools/tests/config/test_support.py
```diff
@@ -84,63 +84,98 @@ def loader(self):
...
+    def test_UWYAMLConvert_dict_ok(self, loader):
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!dict", value="{a0: 0,a1: 1,}"))
+        assert ts.converted == {"a0": 0, "a1": 1}
+        self.comp(ts, "!dict '{a0: 0,a1: 1,}'")
+
+    def test_UWYAMLConvert_float_no(self, loader):
         ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!float", value="foo"))
         with raises(ValueError):
-            ts.convert()
+            assert ts.converted
```
christinaholtNOAA, line 137: This syntax is new to me. The `assert` checks that the expression returns `True`, no? Inside the `with raises` context, I don't know that it would ever get there to return a boolean value or that we'd expect it to.

Just a bit curious on your reasoning here.

christinaholtNOAA (approved):
Just a few comments below. No changes necessary, just some curiosities.

src/uwtools/config/support.py
maddenp-noaa: Replying because I wrote this in a PR to @elcarpenterNOAA's PR...

Something like `TaggedValueT`? The `T` part is idiomatic for types, but the prefix could be more descriptive.

src/uwtools/tests/config/test_support.py
maddenp-noaa: Replying because I wrote this in a PR to @elcarpenterNOAA's PR...

This needs to be a complete statement, which `ts.converted` isn't (it's just an expression). This could be anything that makes a complete statement, e.g. `print(ts.converted)`, but has to be _something_, even though it's expected next to execute. `pass` doesn't take arguments, which is too bad as it'd be nice for a no-op statement. Any ideas?

src/uwtools/config/support.py
christinaholtNOAA: Yeah, I just look at this through a meteorology lens and see "the value of temperature" and it could be nice to just have a bit of clarity.

src/uwtools/tests/config/test_support.py
christinaholtNOAA: Ahh. Thanks for the clarification. I am cool with that. I just didn't understand why.

PR #683, opened 2025-01-08
Validate configs based on most primitive JSON types

maddenp-noaa:
**Synopsis**

Support validation (via CLI or API) of configs of `bool`, `dict`, `float`, `int`, `list`, or `str` types -- basically, all primitive JSON types except for `null`. Previously, validation of only `dict`-based configs was supported. Internally, driver and `fs`-mode configs remain `dict`-based, and for now `YAMLConfig` continues to support only `dict`-based configs, though this could be similarly generalized later.

A config value with `str` type was perviously interpreted as a path to a file containing the config data to validate, so supporting `str`-based configs required splitting the former `config` argument into `config_data` and `config_path`, where a `str` value passed in to the former represents the config data itself, and a `str` value passed in to the latter continues to represent a path to a file.

**Type**

- [x] Documentation

**Impact**

-
**Checklist**


docs/conf.py
```diff
@@ -21,7 +21,10 @@
 html_static_path = ["static"]
 html_theme = "sphinx_rtd_theme"
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
-nitpick_ignore_regex = [("py:class", r"^uwtools\..*"), ("py:class", "f90nml.Namelist")]
+nitpick_ignore = [
+    ("py:class", "Path"),
+    ("py:class", "f90nml.Namelist"),
+]
```
maddenp-noaa, line 27: We don' seem to need regular expressions here, so simplify. I'm still not sure why this is necessary: Possibly `f90nml.Namelist` does not provide documentation that Sphinx can link to, but presumably `Path` (from stdlib `pathlib`) does. Maybe one or both of these could be resolved by adding _something_ to the `intersphinx_mapping` variable above, but my efforts so far haven't yielded a solution. I'd be grateful for ideas on how to improve this. (If you'd like to try, just remove one or both of these lines, try `make html` in the `docs/` directory, and try to correct the errors.)

docs/sections/user_guide/api/config.rst
```diff
@@ -5,5 +5,5 @@
    :target: https://mybinder.org/v2/gh/ufs-community/uwtools/main?labpath=notebooks%2Fconfig.ipynb
 
 .. automodule:: uwtools.api.config
-   :inherited-members: UserDict
    :members:
+   :show-inheritance:
```
maddenp-noaa, line 9: This documents the parent class(es) of documented classes. For example, the documentatino for `YAMLConfig` now shows that it inherits from `Config`. (`Config` itself is also now documented so that readers can see the methods that are inherited by subclasses.)

src/uwtools/api/config.py
```diff
@@ -6,7 +6,7 @@
 from pathlib import Path
 from typing import Optional, Union
 
-from uwtools.config.formats.base import Config as _Config
+from uwtools.config.formats.base import Config
```
maddenp-noaa, line 9: Don't hide this, and add it to `__all__`, below, so that Sphinx can produce documentation for it.

src/uwtools/cli.py
```diff
@@ -252,7 +252,7 @@ def _dispatch_config_validate(args: Args) -> bool:
     """
     return uwtools.api.config.validate(
         schema_file=args[STR.schemafile],
-        config=args[STR.infile],
+        config_path=args[STR.infile],
```
maddenp-noaa, line 255: The CLI always provides a path to a file containing a config to validate.

src/uwtools/config/formats/base.py
```diff
@@ -23,14 +25,17 @@ class Config(ABC, UserDict):
...
         """
         :param config: Config file to load (None => read from stdin), or initial dict.
         """
         super().__init__()
         if isinstance(config, dict):
             self._config_file = None
             self.update(config)
+        elif isinstance(config, Config):
+            self._config_file = config._config_file
+            self.update(config.data)
```
maddenp-noaa, line 38: Permit a `Config` object to be instantiated from another `Config` object. Among other use cases, this makes it safe to say e.g. `c = YAMLConfig(c)` where `c` might already be a `YAMLConfig` (or a `dict` representing data, or a `Path` or `str` representing a path to a file, or `None` to read from `stdin`). Used below.

src/uwtools/config/formats/base.py
```diff
@@ -234,6 +239,7 @@ def logstate(state: str) -> None:
                 break
             self.data = new
         logstate("final")
+        return self
```
maddenp-noaa, line 242: Have `dereference()` return the dereferenced config, so make it possible to say e.g. `d = YAMLConfig(config).dereference().data`, used below.

src/uwtools/config/validator.py
```diff
@@ -19,6 +19,10 @@
 
 # Public functions
 
+JSONValueT = Union[bool, dict, float, int, list, str]
```
maddenp-noaa, line 22: The types of expected values for a config-to-validate.

src/uwtools/config/validator.py
```diff
@@ -19,6 +19,10 @@
...
+ConfigDataT = Union[JSONValueT, YAMLConfig]
```
maddenp-noaa, line 23: The types of expected values for a config-to-validate -- `YAMLConfig` will be converted to a `dict` before validation.

src/uwtools/config/validator.py
```diff
@@ -19,6 +19,10 @@
...
+ConfigPathT = Union[str, Path]
```
maddenp-noaa, line 24: The types of expected values for paths to files containing configs.

src/uwtools/config/validator.py
```diff
@@ -57,7 +61,7 @@ def internal_schema_file(schema_name: str) -> Path:
     return resource_path("jsonschema") / f"{schema_name}.jsonschema"
 
 
-def validate(schema: dict, desc: str, config: dict) -> bool:
+def validate(schema: dict, desc: str, config: JSONValueT) -> bool:
```
maddenp-noaa, line 64: The fundamental change of this PR.

src/uwtools/tests/api/test_config.py
```diff
@@ -117,28 +117,29 @@ def test_realize_update_config_none():
     )
 
 
-@mark.parametrize("cfg", [{"foo": "bar"}, YAMLConfig(config={})])
-def test_validate(cfg):
-    kwargs: dict = {"schema_file": "schema-file", "config": cfg}
+@mark.parametrize("cfg", [{"foo": "bar"}, YAMLConfig(config={"foo": "bar"})])
+def test_validate_config_data(cfg):
+    kwargs: dict = {"schema_file": "schema-file", "config_data": cfg}
```
maddenp-noaa, line 122: Here and in the test below, use either the `config_data` or the `config_path` argument, as needed.

src/uwtools/tests/config/test_validator.py
```diff
@@ -150,32 +150,46 @@ def test_bundle(caplog):
...
+    [
+        ({"type": "boolean"}, True),  # bool
+        ({"type": "number"}, 3.14),  # float
+        ({"type": "integer"}, 42),  # int
+        ({"type": "array"}, [1, 2, 3]),  # list
+        ({"type": "string"}, "foo"),  # str
+    ],
+)
+def test_config_validator_validate_alt_types(schema, config):
+    assert validator.validate(schema=schema, desc="test", config=config) is True
```
maddenp-noaa, line 169: Simple tests validating non-`dict` configs.

src/uwtools/tests/config/test_validator.py
```diff
@@ -125,7 +125,7 @@ def write_as_json(data: dict[str, Any], path: Path) -> Path:
 # Test functions
 
 
-def test_bundle(caplog):
+def test_config_validator_bundle(caplog):
```
maddenp-noaa, line 128: I made the test-function names in this module more descriptive, for easier selection with `pytest -k`.

NaureenBharwaniNOAA (approved):
This looks great!

PR #695, opened 2025-02-14
Remove tag fix

maddenp-noaa:
**Synopsis**

Fix a bug whereby YAML values tagged `!remove` were serialized incorrectly.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/support.py
+++ b/src/uwtools/config/support.py
@@ -74,7 +74,7 @@ def uw_yaml_loader() -> type[yaml.SafeLoader]:
     A loader with basic UW constructors added.
     """
     loader = yaml.SafeLoader
-    for tag_class in (UWYAMLConvert, UWYAMLRemove):
+    for tag_class in (UWYAMLConvert, UWYAMLGlob, UWYAMLRemove):
         for tag in getattr(tag_class, "TAGS"):
             loader.add_constructor(tag, tag_class)
     return loader
@@ -113,16 +113,10 @@ class UWYAMLTag:
         return dumper.represent_scalar(data.tag, data.value)
 
 
-class UWYAMLConvert(UWYAMLTag):
+class UWYAMLTaggedStr(UWYAMLTag):
     """
-    A class supporting custom YAML tags specifying type conversions.
-
-    The constructor implements the interface required by a pyyaml Loader object's add_consructor()
-    method. See the pyyaml documentation for details.
+    Support for YAML tags targeting str values.
     """
-
-    TAGS = ("!bool", "!datetime", "!dict", "!float", "!int", "!list")
-    TaggedValT = Union[bool, datetime, dict, float, int, list]
 
     def __init__(self, loader: yaml.SafeLoader, node: yaml.nodes.ScalarNode) -> None:
         super().__init__(loader, node)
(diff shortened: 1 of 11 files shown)


src/uwtools/tests/config/test_support.py
```diff
@@ -170,12 +176,34 @@ def test_UWYAMLConvert___str__(self, loader):
...
         assert str(node) == "!remove"
+
+    def test_UWYAMLRemove_dump(self, tmp_path):
+        expected = "ns: !remove [1, 2, 3]"
+        old = tmp_path / "old.yaml"
+        old.write_text(expected)
+        c = YAMLConfig(old)
+        new = tmp_path / "new.yaml"
+        c.dump(new)
+        assert new.read_text().strip() == expected
```
maddenp-noaa, line 209: Expanded tests for `UWYAMLRemove`, which now checks that a so-tagged value serializes correctly. This test fails on `main` and passes here.

src/uwtools/config/support.py
```diff
@@ -95,35 +95,33 @@ class UWYAMLTag:
     A base class for custom UW YAML tags.
     """
 
-    def __init__(self, _: yaml.SafeLoader, node: yaml.nodes.ScalarNode) -> None:
+    def __init__(self, _: yaml.SafeLoader, node: yaml.nodes.Node) -> None:
+        self.node = node
```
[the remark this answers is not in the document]
NaureenBharwaniNOAA: Just curious, what was the need to replace `ScalarNode` with `Node`? Does this have to do with the reasoning to want to return the original node unchanged?

src/uwtools/tests/config/test_support.py
```diff
@@ -170,12 +176,34 @@ def test_UWYAMLConvert___str__(self, loader):
...
 
 
+class Test_UWYAMLGlob:
+    """
+    Tests for class uwtools.config.support.UWYAMLGlob.
+    """
+
+    def test_UWYAMLGlob(self, loader):
+        yaml.add_representer(support.UWYAMLGlob, support.UWYAMLTag.represent)
+        node = support.UWYAMLGlob(loader, yaml.ScalarNode(tag="!glob", value="/path/to/*"))
```
NaureenBharwaniNOAA, line 186: Seeing the `ScalarNode` reference here vs. `Node`.

christinaholtNOAA (approved):
Looks good. Thanks for the quick turnaround!

NaureenBharwaniNOAA (approved):
Looks good! Just a question for curiosity.

src/uwtools/tests/config/test_support.py
maddenp-noaa: Thanks for pointing out the possible discrepancy. I think this is correct, though: `ScalarNode` is a subclass of `Node` that can represent YAML values like integers, booleans, and -- here -- strings. There are other `Node` subclasses, `MappingNode` for YAML mappings (aka Python `dict`s) and `SequenceNode` for YAML sequences (aka Python `list`s). In type annotations elsewhere I've used the more-generic parent class `Node` to specify that instances of any of the subclasses are acceptable. In this case, though, we really want to instantiate a `ScalarNode` since the `value` is a string.

NaureenBharwaniNOAA: Thanks for the clarification! Makes sense to me.

src/uwtools/config/support.py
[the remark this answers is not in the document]
maddenp-noaa: I think the previous annotation was actually incorrect: The `node` values that might be supplied here were sometimes scalars, e.g. string values tagged with `!int` like `!int "42"`. But the `!remove` tag can be applied to any YAML value, to the generic `Node` type if, I think, appropriate here.

Issue #677, opened 2024-12-12
Add wildcard support to fs copy and link actions


maddenp-noaa:
# Description

Add support for glob and regex wildcards in `fs` `copy` and `link` actions.

Currently, the `fs` config key-value pair
```
/dst/<foo>: /src/foo.*
```
would be interpreted, when fed to `uw fs copy`, to mean:

> Copy the file literally named `foo.*` in the `/src/` directory as a file named `<foo>` in the `/dst/` directory.

With support for a `!glob` UW YAML tag, key-value pair
```
/dst/<foo>: !glob /src/foo.*
```
would be interpreted to mean

> Find all files in the `/src/` directory matching glob pattern `foo.*` and copy them, retaining their names, to the directory `/dst`/.

Note that the `<foo>` element of the destination path is discarded. It is just a placeholder and can be anything, but is necessary to support multiple key-value pairs targeting the `/dst/` directory, since keys must be unique.

Source paths tagged `!regex` would receive the same treatment, except that the patterns would be treated as regular expressions.

The `uw fs link` action would behave similarly, but create symlinks instead of copies.

# AC

- [ ] Implement the tags.
- [ ] Add/update unit tests.
- [ ] Update docs, both the UW YAML tags page and the `fs` mode page.
- [ ] Update Jupyter `fs` notebook with examples.

PR #697, opened 2025-02-15
GH 677 First Pass: Glob-Style Wildcards

maddenp-noaa:
**Synopsis**

This PR provides an implementation of [glob](https://docs.python.org/3/library/glob.html)-based wildcards for use in `fs copy` and `fs link` modes. #677 calls for `!regex` support, too, but I'd like to PR this first part to make sure we all agree on the direction. Please see the rendered RtD docs for examples of the new feature in action.

I overhauled the docs under `docs/sections/user_guide/cli/tools/fs` to simplify names and improve the overall structure, so there are a lot of fairly trivial diffs there. By far most of the PR diffs, in fact, so don't panic.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -27,124 +27,91 @@ Source paths prefixed with ``http://`` or ``https://`` will be copied from their
 Examples
 ^^^^^^^^
 
-Given ``copy-config.yaml`` containing a mapping from local-filesystem destination paths to source paths
+Given a config containing a mapping from local-filesystem destination paths to source paths
 
-.. literalinclude:: fs/copy-config.yaml
+.. literalinclude:: fs/copy.yaml
```
maddenp-noaa, line 32: Here and below, many filenames were simplified. For example, `config` was removed from filenames ending in `.yaml` because of course they're configs. Similar, `.cmd` and `.out` files had `exec` removed from their names because they're always about `exec`ution.

docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -27,124 +27,91 @@ Source paths prefixed with ``http://`` or ``https://`` will be copied from their
...
 
-.. literalinclude:: fs/copy-exec-report-jq.cmd
+.. literalinclude:: fs/copy-report-jq.cmd
    :language: text
    :emphasize-lines: 2
-.. literalinclude:: fs/copy-exec-report-jq.out
+.. literalinclude:: fs/copy-report-jq.out
    :language: text
 
-``link``
```
maddenp-noaa, line 80: The `link` section was nearly identical to the `copy` section, and making duplicate parallel changes was getting tiresome. So I simplified that section, below, to basically say "This is like `copy`, except for...". This will reduce maintenance burden.

docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -27,124 +27,91 @@ Source paths prefixed with ``http://`` or ``https://`` will be copied from their
...
-   :language: text
-   :emphasize-lines: 1
-.. literalinclude:: fs/link-help.out
-   :language: text
-
-Examples
-^^^^^^^^
-
-Given ``link-config.yaml`` containing
+Use the ``!glob`` tag to specify that a source-path value should be treated as a glob pattern:
```
maddenp-noaa, line 80: This is the start of a new block about the `!glob` functionality. There's an example (see the rendered RtD docs) and some commentary below.

In the commentary I've tried to anticipate "Why can't it..." questions and give a rationale for why recursion is not supported. I think the currently implemented functionality is a good 80/20 (or better) solution that covers many useful, if simple, use cases. There are better existing tools, suggested below, for more complex use cases, and IMO trying to re-implement them in `uwtools` would lead to a confusing and less-robust tool.

docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -27,124 +27,91 @@ Source paths prefixed with ``http://`` or ``https://`` will be copied from their
...
-   :language: yaml
-.. literalinclude:: fs/link-exec-no-target-dir-err.cmd
-   :language: text
-   :emphasize-lines: 1
-.. literalinclude:: fs/link-exec-no-target-dir-err.out
-   :language: text
+See :ref:`files_yaml` for more information on the semantics of the ``!glob`` tag and wildcard copies.
 
-When the ``--report`` option is specified, a report of files not linked ("not-ready") and linked ("ready") will be printed to ``stdout`` as machine-readable JSON. For example, using a config specifying both available and unavailable source files:
+``link``
```
maddenp-noaa, line 96: Here's the simplified `link` section.

docs/sections/user_guide/cli/tools/fs/.gitignore
```diff
@@ -1,7 +1 @@
-copy-dst
-copy-dst-report-jq
-copy-dst-timedep
-link-dst
-link-dst-report
-link-dst-report-jq
-link-dst-timedep
+dst/
```
maddenp-noaa, line 1: Simplify by nesting all temporary output files under a `dst/` directory.

docs/sections/user_guide/cli/tools/fs/copy-no-target-dir-err.cmd
```diff
@@ -0,0 +1 @@
+uw fs copy --config-file copy.yaml --key-path config.files
```
maddenp-noaa, line 1: Here and below, many apparently new files are really old files renamed with simpler names, some of which refer to _other_ renamed files, which I believe is why git sees them as new files rather than just renames.

docs/sections/user_guide/yaml/files.rst
```diff
@@ -3,6 +3,9 @@
 File Blocks
 ===========
 
+Basics
+------
```
maddenp-noaa, line 7: I added section headings to this doc, and added HTTP and wildcard sections.

src/uwtools/config/validator.py
```diff
@@ -184,5 +185,10 @@ def _validation_errors(config: JSONValueT, schema: dict) -> list[ValidationError
     :param schema: JSON Schema to validate the config against.
     :return: Any validation errors.
     """
-    validator = Draft202012Validator(schema, registry=_registry())
+    base = Draft202012Validator
+    type_checker = base.TYPE_CHECKER.redefine(
+        "fs_src", lambda _, x: any(isinstance(x, t) for t in [str, UWYAMLGlob])
+    )
+    UWValidator = validators.extend(base, type_checker=type_checker)
+    validator = UWValidator(schema, registry=_registry())
```
maddenp-noaa, line 193: The new `!glob` tag leads to objects of (new in this PR) type `UWYAMLGlob` appearing in `dict` objects that then need to be validated by JSON Schema. For example, a YAML entry like
```
/dst/<f>: !glob /src/*
```
would lead to a dict like
```
{"/dst/<f>": UWYAMLGlob(...)}
```
Previously, the schema defined in `files-to-stage.jsonschema`, used to validate `uw file` and `uw link` configs, specified `"type": "string"` for the values in such a `dict`, so a config with `!glob` values would fail validation because a `UWYAMLGlob` object is not a `str`.

This update adds a custom JSON Schema type, called `fs_src` here, which accepts either a `str` as before *or* a `UWYAMLGlob` object, so that configs with `!glob` values can pass validation.

src/uwtools/fs.py
```diff
@@ -136,6 +138,27 @@ def _dst_paths(self) -> list[str]:
...
+                if attrs.scheme == "file":
+                    pattern = attrs.path
+                for path in map(Path, glob(pattern)):
+                    if not path.is_dir():
+                        items.append((str(Path(dst).parent / path.name), str(path)))
+                    else:
+                        log.warning("Ignoring directory %s", path)
+            else:
+                items.append((dst, src))
+        return items
```
maddenp-noaa, line 160: This is the crux of the new feature. Previously, the `Copier` and `Linker` `.go()` methods would just iterate over the key-value pairs in their configs and yield `filecopy()` task instances for each. Now, since a value might include a wildcard, one key-value pair might need to expand into many `filecopy()` tasks, one per file matching the wildcard. This method expands as necessary, and passes through each non-wildcard key-value pair as-is. The caller then generates a sequence of `filecopy()` tasks from these.

A clause to handle `!regex` will be added later, but much of the logic here can be shared between `UWYAMLGlob` and future `UWYAMLRegex` handlers.

src/uwtools/resources/jsonschema/files-to-stage.jsonschema
```diff
@@ -1,6 +1,6 @@
 {
   "additionalProperties": {
-    "type": "string"
+    "type": "fs_src"
```
maddenp-noaa, line 3: If anybody has a better idea for the name of this custom type, please let me know.

src/uwtools/tests/test_fs.py
```diff
@@ -53,12 +58,13 @@ def _schema(self):
...
-def test_fs_Copier_go(src_fn, dst_fn, td_fn):
-    src, td, dst = src_fn("/src/file"), td_fn("/dst"), dst_fn("file")
-    obj = Mock(_config={dst: src}, _simple=fs.Copier._simple, _target_dir=td)
+@mark.parametrize("src_func", [str, Path])
+@mark.parametrize("dst_func", [str, Path])
+@mark.parametrize("tgt_func", [str, Path])
+def test_fs_Copier_go(src_func, dst_func, tgt_func):
+    src, dst, tgt = src_func("/src/file"), dst_func("file"), tgt_func("/dst")
+    obj = Mock(_simple=fs.Copier._simple, _target_dir=tgt)
+    obj._expand_wildcards.return_value = [(dst, src)]
```
maddenp-noaa, line 67: This existing test needs to account for the presence of `_expand_wildcards()` now.

docs/sections/user_guide/cli/tools/fs.rst
maddenp-noaa: While reading through the rest of this PR, I thought of a way to maybe handle recursive wildcards elegantly. I'll think about it more and, if it seems workable, see about implementing it before the next release.

maddenp-noaa:
There's currently a problem with the Binder website that's causing the documentation build to fail during link checking. It's probably a transient issue, but we'll see.

NaureenBharwaniNOAA (approved):
LGTM!

PR #698, opened 2025-02-18
Bugfix config run duration

NaureenBharwaniNOAA:
**Synopsis**

Currently there is a known issue when a user changes the length of a forecast to run for 24 hours. The `config_run_duration` does not replace `days` to the correct format of `[DDD]_hh:mm:ss` that is acceptable to run the forecast. This causes `mpassit` to fail further down the workflow.  

Bug below:
```
        duration = timedelta(hours=self.config["length"])
        str_duration = str(duration).replace(" days, ", "_")
```

Example `config_run_duration` format not being generated correctly:
```
    config_start_time = '2024-11-20_00:00:00'
    config_run_duration = '1 day, 0:00:00'
```

Desired `config_run_duration` format:
```
    config_start_time = '2024-11-20_00:00:00'
    config_run_duration = '001_00:00:00'
```

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/resources/jsonschema/mpas.jsonschema
+++ b/src/uwtools/resources/jsonschema/mpas.jsonschema
@@ -58,8 +58,8 @@
           "type": "object"
         },
         "length": {
-          "minimum": 1,
-          "type": "integer"
+          "minimum": 0.25,
+          "type": "number"
         },
         "namelist": {
           "additionalProperties": false,
--- a/src/uwtools/tests/drivers/test_mpas.py
+++ b/src/uwtools/tests/drivers/test_mpas.py
@@ -215,7 +215,7 @@ def test_MPAS_namelist_file_long_duration(caplog, config, cycle):
     assert logged(caplog, f"Wrote config to {path}")
     nml = f90nml.read(dst)
     assert isinstance(nml, f90nml.Namelist)
-    assert nml["nhyd_model"]["config_run_duration"] == "5_0:00:00"
+    assert nml["nhyd_model"]["config_run_duration"] == "005_00:00:00"
 
 
 def test_MPAS_namelist_file_missing_base_file(caplog, driverobj):
--- a/src/uwtools/tests/drivers/test_mpas.py
+++ b/src/uwtools/tests/drivers/test_mpas.py
@@ -204,6 +204,20 @@ def test_MPAS_namelist_file_fails_validation(caplog, driverobj):
     assert logged(caplog, "  None is not of type 'array', 'boolean', 'number', 'string'")
 
 
+def test_MPAS_namelist_file_short_duration(caplog, config, cycle):
+    log.setLevel(logging.DEBUG)
+    config["mpas"]["length"] = 0.25
+    driverobj = MPAS(config=config, cycle=cycle)
+    dst = driverobj.rundir / "namelist.atmosphere"
+    assert not dst.is_file()
+    path = Path(iotaa.refs(driverobj.namelist_file()))
+    assert dst.is_file()
+    assert logged(caplog, f"Wrote config to {path}")
+    nml = f90nml.read(dst)
+    assert isinstance(nml, f90nml.Namelist)
+    assert nml["nhyd_model"]["config_run_duration"] == "00:15:00"
+
+
 def test_MPAS_namelist_file_long_duration(caplog, config, cycle):
     log.setLevel(logging.DEBUG)
     config["mpas"]["length"] = 120
--- a/src/uwtools/drivers/mpas.py
+++ b/src/uwtools/drivers/mpas.py
@@ -49,7 +49,8 @@ class MPAS(MPASBase):
         base_file = self.config[STR.namelist].get(STR.basefile)
         yield file(Path(base_file)) if base_file else None
         duration = timedelta(hours=self.config["length"])
-        str_duration = str(duration).replace(" days, ", "_")
+        hhmmss = ":".join(f"{x:02}" for x in str(timedelta(seconds=duration.seconds)).split(":"))
+        str_duration = "%s%s" % (f"{duration.days:03}_" if duration.days else "", hhmmss)
         namelist = self.config[STR.namelist]
         update_values = namelist.get(STR.updatevalues, {})
         update_values.setdefault("nhyd_model", {}).update(


src/uwtools/resources/jsonschema/mpas.jsonschema
```diff
@@ -58,8 +58,8 @@
           "type": "object"
         },
         "length": {
-          "minimum": 1,
-          "type": "integer"
+          "minimum": 0.25,
+          "type": "number"
```
NaureenBharwaniNOAA, line 62: I added a test case for a 15 min run with `config_run_duration` format generated as '00:15:00'. Since the minimum `length` here is `1` I couldn't generate a forecast that was less than an hour. I had to change this to `0.25` for a 15 min run, and change the type to a number that accepts decimals like this case.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -204,6 +204,20 @@ def test_MPAS_namelist_file_fails_validation(caplog, driverobj):
     assert logged(caplog, "  None is not of type 'array', 'boolean', 'number', 'string'")
 
 
+def test_MPAS_namelist_file_short_duration(caplog, config, cycle):
```
NaureenBharwaniNOAA, line 207: New test case for a 15 min forecast length.

src/uwtools/tests/drivers/test_mpas.py
```diff
@@ -215,7 +229,7 @@ def test_MPAS_namelist_file_long_duration(caplog, config, cycle):
     assert logged(caplog, f"Wrote config to {path}")
     nml = f90nml.read(dst)
     assert isinstance(nml, f90nml.Namelist)
-    assert nml["nhyd_model"]["config_run_duration"] == "5_0:00:00"
+    assert nml["nhyd_model"]["config_run_duration"] == "005_00:00:00"
```
NaureenBharwaniNOAA, line 232: Updated format for `config_run_duration` that matches what is expected from the mpas user guide.

src/uwtools/resources/jsonschema/mpas.jsonschema
maddenp-noaa: Interesting! This looks good to me, but maybe @christinaholtNOAA will have an opinion on a minimum realistic duration. We might consider using [`exclusiveMinimum`](https://json-schema.org/understanding-json-schema/reference/numeric#range) (which I just learned about) with a value of `0` to say that any value > 0 is ok.

src/uwtools/tests/drivers/test_mpas.py
maddenp-noaa: It might be worth combining these two tests with `@mark.parametrize` since they're so similar. We might do something like
```
@mark.parametrize(("hours", "expected"), [(0.25, "00:15:00"), (120, "005_00:00:00")])
def test_MPAS_namelist_file_durations(caplog, config, cycle, hours, expected):
    ...
```
and then make use of the `hours` and `expected` variables in the body. See the [pytest docs](https://docs.pytest.org/en/stable/how-to/parametrize.html) for more information of parametrization.

src/uwtools/resources/jsonschema/mpas.jsonschema
maddenp-noaa: If you could try this option out (`"exclusiveMinimum": 0`) and it works, let's just go with this to be as general as possible. Then we can merge this PR, when the tests pass, and do the bugfix releases.

NaureenBharwaniNOAA: I left the `length` type as a `number` since I received this error message when I tried to change it to `integer`.

```
ERROR    root:validator.py:79 Error at mpas.length:
ERROR    root:validator.py:80   0.25 is not of type 'integer'
```

I think this would be the best `type` seeing as if a user wanted to run for 30 mins they would also have to use 0.5 which would fail as well. I haven't seen any better options yet, but [these are the options](https://json-schema.org/understanding-json-schema/reference/numeric#number) for what counts as a `number`. It accepts exponential notation as an option, but I'm not sure we would want to give users this ability as they could put in pretty large numbers. 

For example, using the jsonschema resource as an example, if a user used an exponential number 
`forecast length = 1e1`, it would pass our `mpas` jsonschema. 

It would output `assert '10:00:00'`, this would generate a 10 hour forecast. Thoughts? Should we be giving a maximum number to the forecast length?

maddenp-noaa: I think `"number"` is the right type here, and that we should not place an upper limit on it.

There might actually be a use case for a zero-duration forecast, which is one that would just initialize the model and dump the initial state as a 0-hour "forecast". But we already had a lower bound higher than zero, so let's go with zero as the `exclusiveMinimum` until we get an actual use case.

NaureenBharwaniNOAA:
A couple of failing tests two specifically on the format of the solution being implemented
```
DEBUG    root:__init__.py:241 20240322 06Z mpas /tmp/pytest-of-runner/pytest-0/popen-gw3/test_MPAS_namelist_file_durati1/namelist.atmosphere: Task failed:     hhmmss = ":".join(f"{x:02}" for x in str(timedelta(seconds=duration.seconds)).split(":"))
DEBUG    root:__init__.py:241 20240322 06Z mpas /tmp/pytest-of-runner/pytest-0/popen-gw3/test_MPAS_namelist_file_durati1/namelist.atmosphere: Task failed: ValueError: '=' alignment not allowed in string format specifier
```

I'm thinking this is the part of the solution `seconds=duration.seconds` that causes the error.

And another two on the file not existing
```
E       AssertionError: assert False
E        +  where False = is_file()
E        +    where is_file = PosixPath('/tmp/pytest-of-runner/pytest-0/popen-gw3/test_MPAS_namelist_file0/namelist.atmosphere').is_file
```

Here I think these files are created locally and they don't actually exist as a tmp file when the workflow runs.

Interestingly, I don't get an errors on the first error locally, which I would've expected.

maddenp-noaa:
It looks like something changed w.r.t. string formatting between Python 3.9 (which I see is the Python version being selected by the CI tests) and 3.12, which is what's installed in my local devshell. (I'll make an update later that will hopefully keep our devshells tied to the oldest-support Python version, so that we don't write code that passes locally and fails in CI.)

Assuming these tests don't fail for you locally, in your devshell do `conda install python=3.9` and, when that completes, re-run the tests and you should see failures. That should make it easier to commit a fix that then passes in CI. I did that, then could reproduce the test failure. For me changing `f"{x:02}"` to `f"{int(x):02}"` in the `mpas.py` code fixed the issue.

Here's a discussion about the issue: https://stackoverflow.com/questions/36044676/text-formatting-error-alignment-not-allowed-in-string-format-specifier

maddenp-noaa (approved):

elcarpenterNOAA (approved):

PR #706, opened 2025-02-21
fs copy/link !glob recursion

maddenp-noaa:
**Synopsis**

Add recursion support to `fs copy` / `fs link` when using `!glob`-tagged sources. See the updated docs and unit tests for detail but, briefly, a source-path value like `!glob /src/**/*.nc` would now identify, and copy or link depending on the mode, `.nc` files in _any subdirectory_ of `/src/`. The relative subdirectory structure of the glob-identified files would be kept intact when they're copied/linked to the destination.

One interesting question was how to handle glob-identified directories. I made the provisional choice -- certainly open for discussion -- to link directories but not to copy them. I think this is consistent with the non-glob semantics of the fs copy / fs link mode/actions, where we support linking, but not copying directories.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -87,20 +87,19 @@ Use the ``!glob`` tag to specify that a source-path value should be treated as a
...
+The ``--report`` output can be especially useful in combination with glob patterns to allow downstream logic to process a set of copied files whose identity is not known in advance.
 
-Note that directories are excluded, and recursive copies are not supported. To recursively copy from a shell script, where you might otherwise use ``uw fs copy``, consider using ``cp -r``; and from Python code where you might otherwise call ``uwtools.api.fs.copy()``, consider using the standard-library :python:`shutil.copytree <shutil.html#shutil.copytree>`. For more control, including file-grained include and exclude, consider the unrivaled `rsync <https://rsync.samba.org/>`_, which can be installed from conda in case your system does not already provide it. It can be called from shell scripts, or via :python:`subprocess <subprocess.html>` from Python.
-
-See :ref:`files_yaml` for more information on the semantics of the ``!glob`` tag and wildcard copies.
+See :ref:`files_yaml` for more examples and information on the semantics of the ``!glob`` tag.
```
maddenp-noaa, line 92: Since recursion now _is_ supported, I deleted some of this text and moved the rest, about `rsync` (which is an amazing tool that people should use when they need its power, and whose versatility I don't think we should try to duplicate in `uwtools`), to the linked-to document.

docs/sections/user_guide/cli/tools/fs.rst
```diff
@@ -87,20 +87,19 @@ Use the ``!glob`` tag to specify that a source-path value should be treated as a
 .. literalinclude:: fs/copy-glob.out
    :language: text
 
-The ``--report`` output can be especially useful in combination with wildcards to allow downstream logic to process a set of copied files whose identity is not known in advance.
+The ``--report`` output can be especially useful in combination with glob patterns to allow downstream logic to process a set of copied files whose identity is not known in advance.
```
maddenp-noaa, line 90: Since I don't think we can implement regex-based wildcards, I changed generic references to "wildcards" to "glob patterns", since that's what we're actually implementing here.

src/uwtools/fs.py
```diff
@@ -138,25 +140,27 @@ def _dst_paths(self) -> list[str]:
...
+                attrs = urlparse(src.value)
+                if attrs.scheme in ["", "file"]:
+                    for path in iglob(attrs.path, recursive=True):
+                        if Path(path).is_dir() and not isinstance(self, Linker):
+                            log.warning("Ignoring directory %s", path)
+                        else:
+                            parts = zip_longest(*[Path(x).parts for x in (path, attrs.path)])
+                            pairs = dropwhile(lambda x: eq(*x), parts)
+                            unique = Path(*[pair[0] for pair in pairs if pair[0]])
+                            items.append((str(Path(dst).parent / unique), path))
```
maddenp-noaa, line 159: The logic in this inner `else` block boils down to: Identify the unique path suffix of a glob-identified file vs the glob pattern, so that it can be appended to the destination path to create a mirror directory structure. For example, given a source-path glob pattern like `/path/to/some/**/*.nc` and an actual path like `/path/to/some/20250220/18/foo.nc` matching the glob pattern, `parts` would look like
```
[("path", "path"), ("to", "to"), ("some", "some"), ("20250220", "**"), ("18", "*.nc"), ("foo.nc", None)]
```
(`zip_longest()`, unlike `zip()`, doesn't terminate when one iterable runs out of items, it keeps going and supplies, by default, `None` values for the exhausted iterable.)

`dropwhile()`, with the lambda supplied to it, discards entries from `pairs` as long as the items in the pairs are equal, reducing `parts` to this `pairs`:
```
[("20250220", "**"), ("18", "*.nc"), ("foo.nc", None)]
```
The RHS of the `unique =` statement discards the second element of each pair, so
```
[("20250220",), ("18",), ("foo.nc",)]
```
which is then concatenated into a `Path`:
```
Path("20250220/18/foo.nc")
```
This can then be appended to the destination directory to make the full destination path to create as a copy or link.

(The `if pair[0]` clause in the `unique` expression handles an edge case where the parts of the glob-identified path run out before the parts of the pattern.)

If anybody can think of a simpler way to do this, please let me know! I thought about using something like `.relative_to()` on `Path` objects, but it doesn't seem quite right.

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

Issue #678, opened 2024-12-12
Add HPSS support to fs copy action and related util tasks


maddenp-noaa:
# Description

Support `hsi://` and `htar://` schemes in the `fs copy` mode/action, and in `uwtools.utils.tasks`.

Semantics are as follows:

- `/dst/b.dat: hsi:///src/a.dat`: Call `hsi` to fetch HPSS file `/src/a.dat` to local file `/dst/b.dat`.
- `/dst/<files>: !glob hsi:///src/file.*`: Call `hsi` to fetch all files from `/src/` that match glob pattern `file.*`, placing them with their original names in `/dst/`.
- `/dst/b.dat: htar:///src/archive.tar?a.dat`: Call `htar` to extract file `a.dat` from archive `/src/archive.tar` and save as `/dst/b.dat`.
- ~`/dst/<files>: htar:///src/archive.tar?a.dat&b.dat`: Call `htar` to extract files `a.dat` and `b.dat` from `/src/archive.tar` and save with their original names in `/dst/`.~
- `/dst/<files>: !glob htar:///src/archive.tar?*.dat`: Call `htar` to extract all `.dat` files from `/src/archive.tar`, saving them with their original names in `/dst/`.
- `/dst/<files>: !glob htar:///src/*.tar?*.dat`: Call `htar` to extract all `.dat` files from all `.tar` files in `/src/`, saving all files with their original names in `/dst/`.

~The `!regex` tag could be used in place of `!glob` for finer control.~

Use `uwtools.utils.tasks.executable()` to ensure that `hsi` or `htar`, as appropriate are available. It's the caller's responsibility to make sure that these utilities are on their `PATH`.

Consider whether to extend `filecopy()` and `existing()` in `uwtools.utils.tasks` to support `hsi://` and `htar://` URLs. For simple cases, these might be handy but, for wildcard support, it may be unacceptably inefficient to repeatedly call `hsi ls` or `htar -tf` to establish existence of source files, when a single call could be made and the results reused for a set of source files. Similarly, where glob patterns are used, a single `hsi cp` command with a wildcard, or an `htar` command with a wildcard (and possibly a `-T` option to specify multiple threads (Hera provides an `htar` alias specifying `-T 4`, so that might be a useful value) could be more efficient. In particular, if files are on tape and need to be staged to disk, a wildcard command might stage all files at once and so be _much_ faster than a series of individual requests, where each might have to wait for one file to be staged and transferred, then another, then another, etc.

# AC

[...]

PR #715, opened 2025-03-21
GH-678 Simple HSI Fetch

maddenp-noaa:
**Synopsis**

First work on #678: Single full-file copy from HPSS to local filesystem with the `hsi` executable.

Here's an example:
`config.yaml`
```
psc.tgz: hsi:///BMC/fim/madden/prepsourceschem.tar.gz
```
```
$ uw fs copy -c config.yaml --target-dir . --report
[2025-03-21T14:02:58]     INFO Validating config against internal schema: files-to-stage
[2025-03-21T14:02:58]     INFO 0 schema-validation errors found in fs config
[2025-03-21T14:02:58]     INFO HPSS file /BMC/fim/madden/prepsourceschem.tar.gz: Executing
[2025-03-21T14:02:58]     INFO Running: hsi ls /BMC/fim/madden/prepsourceschem.tar.gz
[2025-03-21T14:03:03]     INFO HPSS file /BMC/fim/madden/prepsourceschem.tar.gz: Ready
[2025-03-21T14:03:03]     INFO Copy hsi:///BMC/fim/madden/prepsourceschem.tar.gz -> psc.tgz: Executing
[2025-03-21T14:03:03]     INFO Running: hsi -q get psc.tgz : /BMC/fim/madden/prepsourceschem.tar.gz
[2025-03-21T14:03:09]     INFO => [connecting to hpsscore1.fairmont.rdhpcs.noaa.gov/1217]
[2025-03-21T14:03:09]     INFO => get  '/home/Paul.Madden/git/uwtools/psc.tgz' : '/BMC/fim/madden/prepsourceschem.tar.gz' (2011/07/22 16:55:04 2209010 bytes, 96223.4 KBS )
[2025-03-21T14:03:09]     INFO Copy hsi:///BMC/fim/madden/prepsourceschem.tar.gz -> psc.tgz: Ready
[2025-03-21T14:03:09]     INFO File copies: Ready
{
  "not-ready": [],
  "ready": [
    "psc.tgz"
  ]
}
```

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/Makefile.outputs
```diff
@@ -7,3 +7,6 @@ all: $(OUTPUTS)
 
 $(OUTPUTS):
 	@bash $(basename $@).cmd >$@ 2>&1 | true
+
+%.out: %.txt %.yaml
+	@bash $< >$@ 2>&1 | true
```
maddenp-noaa, line 12: See the comment in `fs.rst`, below. This is a new `make` rule to manually create `.out` files from `.txt` files. The convention being introduced here is that `.cmd` files will be run _automatically_ to create `.out` files when running `make docs`, but creation of `.out` files from `.txt` files requires a manual step. In the current case, this is necessary because the HPSS example using `hsi` can only be run on certain platforms, and not from personal systems or from CI. So, `make docs` will not update, in this case (see below) `copy-hpss-single.out`, but there is a mechanism to manually run it on Hera, Jet, etc., after which one can `git commit` the changes.

docs/sections/user_guide/cli/tools/fs/copy-glob.out
```diff
@@ -1,13 +1,11 @@
...
-[2025-02-15T18:52:47]     INFO Copy src/bar -> dst/copy-glob/dst/bar: Ready
-[2025-02-15T18:52:47]     INFO File copies: Ready
+[2025-03-21T02:57:06]     INFO Validating config against internal schema: files-to-stage
+[2025-03-21T02:57:06]     INFO 0 schema-validation errors found in fs config
+[2025-03-21T02:57:08]  WARNING Ignoring directory src/20240529
+[2025-03-21T02:57:08]     INFO Copy src/foo -> dst/copy-glob/dst/foo: Executing
+[2025-03-21T02:57:08]     INFO Copy src/foo -> dst/copy-glob/dst/foo: Ready
+[2025-03-21T02:57:08]     INFO Copy src/bar -> dst/copy-glob/dst/bar: Executing
+[2025-03-21T02:57:08]     INFO Copy src/bar -> dst/copy-glob/dst/bar: Ready
+[2025-03-21T02:57:08]     INFO File copies: Ready
```
maddenp-noaa, line 8: Unrelated update probably missed in a previous PR.

docs/sections/user_guide/cli/tools/fs/copy-report.out
```diff
@@ -1,16 +1,15 @@
...
+[2025-03-21T02:57:08]     INFO Copy src/foo -> dst/copy-report-jq/foo: Executing
+[2025-03-21T02:57:08]     INFO Copy src/foo -> dst/copy-report-jq/foo: Ready
+[2025-03-21T02:57:08]  WARNING File src/qux: Not ready [external asset]
+[2025-03-21T02:57:08]  WARNING Copy src/qux -> dst/copy-report-jq/qux: Not ready
+[2025-03-21T02:57:08]  WARNING Copy src/qux -> dst/copy-report-jq/qux: Requires:
+[2025-03-21T02:57:08]  WARNING Copy src/qux -> dst/copy-report-jq/qux: ✖ File src/qux
+[2025-03-21T02:57:08]  WARNING File copies: Not ready
+[2025-03-21T02:57:08]  WARNING File copies: Requires:
+[2025-03-21T02:57:08]  WARNING File copies: ✔ Copy src/foo -> dst/copy-report-jq/foo
+[2025-03-21T02:57:08]  WARNING File copies: ✖ Copy src/qux -> dst/copy-report-jq/qux
```
maddenp-noaa, line 12: Unrelated update probably missed in a previous PR.

docs/sections/user_guide/cli/tools/fs/copy-timedep.out
```diff
@@ -1,9 +1,8 @@
...
-[2025-02-15T17:50:54]     INFO 0 schema-validation errors found in fs config
-[2025-02-15T17:50:56]     INFO File src/20240529/12/006/baz: Ready
-[2025-02-15T17:50:56]     INFO Copy src/20240529/12/006/baz -> dst/copy-timedep/baz-2024-05-29T18: Executing
-[2025-02-15T17:50:56]     INFO Copy src/20240529/12/006/baz -> dst/copy-timedep/baz-2024-05-29T18: Ready
-[2025-02-15T17:50:56]     INFO File copies: Ready
+[2025-03-21T02:57:07]     INFO Validating config against internal schema: files-to-stage
+[2025-03-21T02:57:07]     INFO 0 schema-validation errors found in fs config
+[2025-03-21T02:57:08]     INFO Copy src/20240529/12/006/baz -> dst/copy-timedep/baz-2024-05-29T18: Executing
+[2025-03-21T02:57:08]     INFO Copy src/20240529/12/006/baz -> dst/copy-timedep/baz-2024-05-29T18: Ready
+[2025-03-21T02:57:08]     INFO File copies: Ready
```
maddenp-noaa, line 5: Unrelated update probably missed in a previous PR.

docs/sections/user_guide/cli/tools/fs/copy.out
```diff
@@ -1,13 +1,10 @@
...
-[2025-02-15T19:08:21]     INFO Copy src/bar -> dst/copy/subdir/bar: Executing
-[2025-02-15T19:08:21]     INFO Copy src/bar -> dst/copy/subdir/bar: Ready
-[2025-02-15T19:08:21]     INFO File copies: Ready
+[2025-03-21T02:57:06]     INFO Validating config against internal schema: files-to-stage
+[2025-03-21T02:57:06]     INFO 0 schema-validation errors found in fs config
+[2025-03-21T02:57:07]     INFO Copy https://www.gnu.org/licenses/gpl-3.0.txt -> dst/copy/licenses/gpl: Executing
+[2025-03-21T02:57:08]     INFO Copy https://www.gnu.org/licenses/gpl-3.0.txt -> dst/copy/licenses/gpl: Ready
+[2025-03-21T02:57:08]     INFO Copy src/bar -> dst/copy/subdir/bar: Executing
+[2025-03-21T02:57:08]     INFO Copy src/bar -> dst/copy/subdir/bar: Ready
+[2025-03-21T02:57:08]     INFO File copies: Ready
```
maddenp-noaa, line 7: Unrelated update probably missed in a previous PR.

src/uwtools/tests/utils/test_tasks.py
```diff
@@ -28,14 +29,14 @@ def exists(x):
 # Tests
 
 
-def test_tasks_directory(tmp_path):
+def test_utils_tasks_directory(tmp_path):
```
maddenp-noaa, line 32: Added some more detail to the test-function names to make it easier to run them selectively from the command line with `pytest -k`.

src/uwtools/tests/utils/test_tasks.py
```diff
@@ -58,35 +59,42 @@ def test_tasks_executable(tmp_path):
         assert iotaa.ready(tasks.executable(program=p))
 
 
+def test_utils_tasks_existing__bad_scheme():
+    path = "foo://bucket/a/b"
+    with raises(UWConfigError) as e:
+        tasks.existing(path=path)
+    assert str(e.value) == f"Scheme 'foo' in '{path}' not supported"
```
maddenp-noaa, line 66: Just relocated in the module.

src/uwtools/utils/tasks.py
```diff
@@ -82,6 +88,21 @@ def file(path: Union[Path, str], context: str = ""):
...
+    An existing file in HPSS.
+
+    :param path: HPSS path to the file.
+    """
+    yield "HPSS file %s" % path
+    available: list[Union[bool, Path, str]] = [False]
+    yield asset(path, lambda: bool(available[0]))
+    yield executable(STR.hsi)
+    success, _ = run_shell_cmd(f"{STR.hsi} ls {str(path)}")
+    available[0] = path if success else False
```
maddenp-noaa, line 103: This task ensures that the upstream file is available before an attempt to download it is made.

src/uwtools/utils/tasks.py
```diff
@@ -94,22 +115,17 @@ def filecopy(src: Union[Path, str], dst: Union[Path, str]):
...
-        response = requests.get(src, allow_redirects=True, timeout=3)
-        if (code := response.status_code) == 200:
-            with open(dst, "wb") as f:
-                f.write(response.content)
-        else:
-            log.error("Could not get '%s', HTTP status was: %s", src, code)
+        _filecopy_http(str(src), dst)
+    elif src_scheme in SCHEMES.local:
+        yield file(src)
+        _filecopy_local(_local_path(src), dst)
```
maddenp-noaa, line 128: The logic for copying local and HTTP files was factored out into separate functions, below. This function would have become uncomfortably long without a refactor. Smaller, single-purpose functions make writing tests easier, too.

christinaholtNOAA (commented):
I have questions...

src/uwtools/utils/tasks.py
maddenp-noaa: Excellent question about a nuanced technique! There's a description [here](https://github.com/maddenp/iotaa#in-memory-asset) of a different use case, and I'd be interested to know whether or not it explains satisfactorily what's happening here -- because I can try to improve those docs if not.

> Won't it go out of scope?

It will only go out of scope when this function returns, which will only happen after `iotaa` has called `next()` on the function 4 times: First to get the task name from the first `yield`, then to get its `asset` from the second `yield`, then to get its dependencies from the third `yield`, and a fourth and final time to let the action code run. Even then, though, `available` may not be eligible for garbage collection yet, as other tasks may hold a reference to this `asset`, which is captured in the closure of the asset's `lambda`; in other words, it may go out of scope relative to this function, but it may still be in-scope elsewhere.

When `iotaa` gets the `asset`, it will call its `.ready()` to see if anything needs to be done for this task. If the asset is ready, then this task will not be added to the task graph, and nothing past the second `yield` will be executed. This is the **first time** `available` gets inspected.

If the asset is not ready, then `iotaa` will then request the task's dependencies and check if those are ready. If they're not, the task will not be added to the task graph and nothing past the third `yield` will be executed. If they are, then the task gets added to the graph and _eventually_ (depending on the execution order implied by dependencies) the action code following the final `yield` will be executed. At that point, `available` will be updated. We can't just re-assign `available` with e.g. `available = path`, because that would just reassign the local variable but would leave the `available` referred to in the asset's `lambda` pointing to the old value of `available`, such that any other piece of code that calls `.ready()` on the asset will always see the original value and won't know that the task now _is_ ready. By not changing the _identity_ of `available`, but the contents of the object it names (a single-element `list`), the `available` captured in the `lambda` sees the change, and any external code that calls `.ready()` on the asset sees the change, too.

Later, after `iotaa` has assembled the task graph, it executes nodes in an order that honors their declared dependencies. Each `@task`-type task, when executed, [checks the readiness of all its dependencies](https://github.com/maddenp/iotaa/blob/main/src/iotaa/__init__.py#L296) before running its action code. (That line of code says: "If _this_ task is not ready, but if all its requirements (aka dependencies) _are_ ready then, if this isn't a dry-run run, execute its action code, aka continuation") This check on readiness of required tasks is the **second time** `available` gets inspected (where this `file_hpss` task is required by another task), and happens outside the scope of the `file_hpss` function, so `available` has to be the same object in memory in every scope, though that object's content may change.

I hope this makes sense. It's tricky. The nuance is mostly hidden in `iotaa`, but this case where the asset in question is an in-memory Python object and not some external asset like a file requires slightly special treatment.

If you're curious, try playing with that `location1.py` example from the `iotaa` docs and see what happens if you change `val` from a `list` to a `str`, or if you replace the action code with `pass` in the `json()` function.

> My mental model assumed the entire task was re-run to evaluate the readiness after the action code was run...

The only thing that runs multiple times is the readiness function of an `asset`. That will be run multiple times, and needs to tell the current truth each time, which is why it needs to capture ("close over") a mutable object. In a language like C, this would be the difference between a "stack-allocated object", i.e. one that goes out of scope when the function ends, and a "heap-allocated object", i.e. one that survives until explicitly deleted.

maddenp-noaa: Two more related things:

1. When an asset represents something like a file on disk, things are simpler because a file doesn't "go out of scope", and defining the asset's readiness function as something like `lambda: path.is_file` always tells the current truth, wherever/whenever it's called. (To define "ready" as "path is a file AND it is completely written" is slightly trickier, but we have ways.)
2. I think I can rewrite this function as a simpler `@external` task. I'll try.

christinaholtNOAA: Thanks for the very thorough response and the update to my mental model. :) 

It could be useful in the Tasks (or @tasks) section of the `iotaa` docs to direct users to the Cookbook below for more information on how to deal with assets, especially in-memory ones. Or maybe even more generically a table of contents at the top. I was a bad user and didn't keep scrolling past the Demo section after reading the relevant `Tasks` sections because it seemed like I was past the relevant info.

christinaholtNOAA (approved):

src/uwtools/utils/tasks.py
maddenp-noaa: I will add a link to the _Cookbook_ section, good idea.

I recently learned that GitHub automatically makes a table of contents for the README.md -- its a little bullet-list icon in the header line:

image

christinaholtNOAA: What!? TIL! Thanks!

PR #718, opened 2025-03-31
Reduce documentation churn

maddenp-noaa:
**Synopsis**

I think it was initially helpful to have `make docs` always generate some diffs to demonstrate that it was correctly generating dynamic content, but sifting through those diffs to find the meaningful ones and reverting unwanted changes has become toilsome. This PR reduces documentation churn by:

1. Ensuring a constant timestamp in log messages that appear in docs,
2. Using a generic example cycle timestamp in CLI help messages, and
3. Updating example external-driver code used in the docs to use a fixed, rather than a random, number.

Given these changes, `make docs` should only update documentation files related to meaningful code or content changes.

Don't Panic: Most of the changed files are the results of the three changes described above and can mostly be ignored.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/tools/fs/copy-report.cmd
```diff
@@ -1,2 +1,2 @@
-rm -rf dst/copy-report-jq
-uw fs copy --report --target-dir dst/copy-report-jq --config-file copy-report.yaml
+rm -rf dst/copy-report
+uw fs copy --report --target-dir dst/copy-report --config-file copy-report.yaml
```
maddenp-noaa, line 2: There was a race-condition bug here, with two doc-generation commands targeting the same output directory.

elcarpenterNOAA (approved):
Thank you!

PR #720, opened 2025-04-02
Tooling Updates

maddenp-cu:
**Synopsis**

- Replace `black`, `isort`, and `pylint` with `ruff`. The formatting behaviors are essentially identical to those of `black` and `isort`, and the linter is _much_ faster. `ruff` mostly follows [flake8](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-flake8) linting conventions, which are different in some ways to those of `pylint`. I have made the appropriate updates, either to the code where it seemed either like a good idea to follow the new linter advice (there were many cases of this), or by disabling some rules in `pyproject.toml`. Overall, I'm happy with the state of the code after the updates, and very happy with the linting speed.
- Update to the latest version of `iotaa`, which provides some new methods on the `Node` object returned by task-function calls that in turn simplify application code. I made the corresponding simplifications here.

Note that the changes to the unit tests relate more to style than substance: I hope that the fact that I had to make few meaningful changes to the unit tests is evidence that there are no breaking changes here.

There's a lot in this PR but, as is often the case, there are repeated patterns. I'll add inline comments about changes I thought were interesting, but will ignore changes prompted by `ruff` advice that I think are trivial or self-explanatory.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/config/formats/base.py
```diff
@@ -34,16 +32,18 @@ def __init__(self, config: Optional[Union[dict, str, Config, Path]] = None) -> N
...
         if self._get_depth_threshold() and self._depth != self._get_depth_threshold():
-            raise UWConfigError(
-                "Cannot instantiate depth-%s %s with depth-%s config"
-                % (self._get_depth_threshold(), type(self).__name__, self._depth)
+            msg = "Cannot instantiate depth-%s %s with depth-%s config" % (
+                self._get_depth_threshold(),
+                type(self).__name__,
+                self._depth,
             )
+            raise UWConfigError(msg)
```
maddenp-noaa, line 46: `ruff` prefers the message passed to an exception constructor to be defined separately instead of inline. Its docs claim that this makes stacktraces easier to read, but I haven't verified; I just did what it asked in this case.

src/uwtools/config/formats/base.py
```diff
@@ -152,17 +152,18 @@ def _load_paths(self, config_files: list[Path]) -> dict:
         """
         cfg = {}
         for config_file in config_files:
-            if not os.path.isabs(config_file):
+            cf = config_file
```
maddenp-noaa, line 155: `ruff` complains about overwriting the value of the loop variable inside the loop, so I renamed this one, as it is changed a few lines down. At first I thought this was maybe worth suppressing, but I found that it only happened once in the `uwtools` codebase, so maybe it's not that great an imposition.

src/uwtools/config/formats/nml.py
```diff
@@ -1,9 +1,10 @@
+from __future__ import annotations
+
 from collections import OrderedDict
 from io import StringIO
-from pathlib import Path
-from typing import Optional, Union
+from typing import TYPE_CHECKING
 
-import f90nml  # type: ignore
+import f90nml  # type: ignore[import-untyped]
```
maddenp-noaa, line 7: `ruff` prefers that `# type: ignore` directives be specific about the typing problem they're ignoring, which seems like a good idea to me.

src/uwtools/config/support.py
```diff
@@ -43,13 +45,13 @@ def format_to_config(fmt: str) -> Type:
         FORMAT.sh: "SHConfig",
         FORMAT.yaml: "YAMLConfig",
     }
-    if not fmt in lookup:
+    if fmt not in lookup:
```
maddenp-noaa, line 48: `ruff` prefers this syntax, which I also like now that I notice it.

src/uwtools/config/support.py
```diff
@@ -105,7 +107,7 @@ def __repr__(self) -> str:
 
     @staticmethod
     def represent(
-        dumper: yaml.Dumper,  # pylint: disable=unused-argument
+        _dumper: yaml.Dumper,
```
maddenp-noaa, line 110: `ruff` didn't like this line any more than `pylint`, but the leading underscore tells it that this variable is intentionally unused.

src/uwtools/config/tools.py
```diff
@@ -148,18 +153,19 @@ def _ensure_format(
     :raises: UWError if the format cannot be determined.
     """
     if isinstance(config, Config):
-        return config._get_format()  # pylint: disable=protected-access
+        return config._get_format()  # noqa: SLF001
```
maddenp-noaa, line 156: I can't deal with doing any more in this PR, but someday it might be a good idea to reconsider whether some of the private attributes for which we are currently adding protected-access suppressors truly need to be private: Maybe we should make them public and remove the associated suppressors, at least where we own the code.

src/uwtools/config/validator.py
```diff
@@ -189,6 +196,6 @@ def _validation_errors(config: JSONValueT, schema: dict) -> list[ValidationError
     type_checker = base.TYPE_CHECKER.redefine(
         "fs_src", lambda _, x: any(isinstance(x, t) for t in [str, UWYAMLGlob])
     )
-    UWValidator = validators.extend(base, type_checker=type_checker)
-    validator = UWValidator(schema, registry=_registry())
+    uwvalidator = validators.extend(base, type_checker=type_checker)
```
maddenp-noaa, line 199: `ruff` doesn't like camelcase variable names.

src/uwtools/drivers/driver.py
```diff
@@ -550,7 +559,7 @@ def _write_runscript(self, path: Path, envvars: Optional[dict[str, str]] = None)
         )
         with writable(path) as f:
             print(rs, file=f)
-        os.chmod(path, os.stat(path).st_mode | stat.S_IEXEC)
+        path.chmod(path.stat().st_mode | stat.S_IEXEC)
```
maddenp-noaa, line 562: Here's another example of calling a method on a `Path` object instead of calling into a separate library.

src/uwtools/drivers/driver.py
```diff
@@ -670,14 +679,10 @@ def _add_docstring(class_: type, omit: Optional[list[str]] = None) -> None:
...
-        "__doc__",
-        "\n".join(
-            line
-            for line in dedent(base).strip().split("\n")
-            if not any(line.startswith(f":param {o}:") for o in omit or [])
-        ),
+    class_.__doc__ = "\n".join(
+        line
+        for line in dedent(base).strip().split("\n")
+        if not any(line.startswith(f":param {o}:") for o in omit or [])
```
maddenp-noaa, line 685: `ruff` pointed out that `setattr()` is unnecessary here, which simplified the code a little.

src/uwtools/drivers/filter_topo.py
```diff
@@ -27,7 +27,7 @@ def input_grid_file(self):
         """
         src = Path(self.config["config"]["input_grid_file"])
         dst = Path(self.config[STR.rundir], src.name)
-        yield self.taskname(f"Input grid {str(src)}")
+        yield self.taskname(f"Input grid {src!s}")
```
maddenp-noaa, line 30: `ruff` prefers this idiom, which I've seen but haven't used myself. The `!s` says to call the object's `__str__()` method (what `str()` also does). `!r` would call `__repr__()`. Nice in f-strings, especially, as the syntax is less noisy.

src/uwtools/drivers/fv3.py
```diff
@@ -40,7 +40,7 @@ def boundary_files(self):
                     self.rundir / "INPUT" / f"gfs_bndy.tile{n}.{(boundary_hour - offset):03d}.nc"
                 )
                 symlinks[target] = linkname
-        yield [symlink(target=t, linkname=l) for t, l in symlinks.items()]
+        yield [symlink(target=tgt, linkname=lnk) for tgt, lnk in symlinks.items()]
```
maddenp-noaa, line 43: `ruff` did not appreciate `l` as a variable name, and I can't argue with that.

src/uwtools/drivers/mpas.py
```diff
@@ -34,9 +34,10 @@ def boundary_files(self):
         for boundary_hour in range(0, endhour + 1, interval):
             file_date = self._cycle + timedelta(hours=boundary_hour)
             fn = f"lbc.{file_date.strftime('%Y-%m-%d_%H.%M.%S')}.nc"
+            target = Path(lbcs["path"], fn)
             linkname = self.rundir / fn
-            symlinks[linkname] = Path(lbcs["path"], fn)
-        yield [symlink(target=t, linkname=l) for l, t in symlinks.items()]
+            symlinks[target] = linkname
```
maddenp-noaa, line 39: There are several other places where a `symlinks` `dict` like this is built up, and this one was the only case where the link names were the keys and the targets were the values. So I reversed them so that they matched e.g. the way it's done in `mpas_init.py`.

src/uwtools/tests/api/test_config.py
```diff
@@ -32,7 +30,7 @@ def test_compare():
 
 
 @mark.parametrize(
-    "classname,f",
+    ("classname", "f"),
```
maddenp-noaa, line 33: `ruff` prefers the `tuple`-based syntax for defining the parametrization arguments, which I'd been migrating to myself anyway, so it was nice to see tooling asking for it.

src/uwtools/tests/api/test_drivers.py
```diff
@@ -56,7 +54,7 @@
 
 
 @mark.parametrize("module", modules)
-def test_api_execute(module):
+def test_api_execute(module, utc):
```
maddenp-noaa, line 57: I borrowed and extended the `utc` fixture from the `wxvx` tests, which simplifies the code in light of the new-ish requirement to specify timezone information when constructing `datetime` objects. With no arguments `utc()` returns a "now" `datetime` object; with arguments it returns a `datetime` object matching the year, month, date, etc. argument.

src/uwtools/tests/api/test_execute.py
```diff
@@ -42,35 +38,45 @@ def kwargs(args):
...
     kwargs["stdin_ok"] = False
     with raises(UWError) as e:
         execute.execute(**kwargs)
     assert str(e.value) == "Set stdin_ok=True to permit read from stdin"
 
 
-@mark.parametrize("remove", ([], ["schema_file"]))
-def test_execute_pass(kwargs, logged, remove, tmp_path):
+@mark.parametrize("remove", [[], ["schema_file"]])
+def test_execute_pass(kwargs, logged, remove, tmp_path, utc):
```
maddenp-noaa, line 58: This is one of the few unit tests that changed significantly. The change was necessitated by a change, in the tested-function code, from something like `iotaa.graph(node)` to `node.graph`, taking advantage of the new `iotaa` versions' simpler syntax. Previously, it was enough to mock `iotaa.graph()`; now a node object with a correctly functioning `.graph` property has to be mocked up.

src/uwtools/tests/config/formats/test_yaml.py
```diff
@@ -161,27 +159,28 @@ def test_yaml_constructor_error_no_quotes(tmp_path):
...
 def test_yaml_constructor_error_not_dict_from_stdin():
     # Test that a useful exception is raised if the YAML stdin input is a non-dict value.
-    with StringIO("42") as sio, patch.object(sys, "stdin", new=sio):
-        with raises(exceptions.UWConfigError) as e:
-            YAMLConfig()
+    with (
+        StringIO("42") as sio,
+        patch.object(sys, "stdin", new=sio),
+        raises(exceptions.UWConfigError) as e,
+    ):
```
maddenp-noaa, line 174: `rufff` prefers combining context managers in a single `with` statement instead of nesting.

src/uwtools/tests/config/test_jinja2.py
```diff
@@ -477,7 +468,7 @@ def test__write_template_stdout(capsys):
     assert actual.strip() == expected
 
 
-class Test_J2Template:
+class TestJ2Template:
```
maddenp-noaa, line 471: `ruff` prefers camelcase names for test classes.

src/uwtools/tests/config/test_tools.py
```diff
@@ -598,7 +610,7 @@ def test_realize_config_values_needed_yaml(logged):
       FV3GFS.nomads.url: https://nomads.ncep.noaa.gov/pub/data/nccf/com/gfs/prod/gfs.{{ yyyymmdd }}/{{ hh }}/atmos
       FV3GFS.nomads.file_names.grib2.anl: ['gfs.t{{ hh }}z.atmanl.nemsio', 'gfs.t{{ hh }}z.sfcanl.nemsio']
       FV3GFS.nomads.file_names.grib2.fcst: ['gfs.t{{ hh }}z.pgrb2.0p25.f{{ fcst_hr03d }}']
-    """
+    """  # noqa: E501
```
maddenp-noaa, line 613: Some lines were too long.

src/uwtools/tests/config/test_tools.py
```diff
@@ -639,7 +651,7 @@ def test__ensure_format_dict_implicit():
 
 
 def test__ensure_format_deduced():
-    assert tools._ensure_format(desc="foo", config=Path("/tmp/config.nml")) == FORMAT.nml
+    assert tools._ensure_format(desc="foo", config=Path("/some/config.nml")) == FORMAT.nml
```
maddenp-noaa, line 654: `ruff` doesn't like seeing `/tmp` in code, as actually using `/tmp` can be insecure. It doesn't know that this path isn't actually used to reading from or writing to disk, but I changed paths it complained about as the simplest solution.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -1,15 +1,13 @@
-# pylint: disable=missing-class-docstring
-# pylint: disable=missing-function-docstring
-# pylint: disable=protected-access
-# pylint: disable=redefined-outer-name
 """
 Tests for uwtools.drivers.driver module.
 """
+
+from __future__ import annotations
```
maddenp-noaa, line 5: BTW, once Python 3.9 is EOL, these imports can go since the `|` constructor for `Union` types is standard in Python 3.10 and above.

src/uwtools/tests/drivers/test_orog.py
```diff
@@ -87,22 +87,22 @@ def test_Orog_driver_name(driverobj):
 
 
 def test_Orog_files_linked(driverobj):
-    for _, src in driverobj.config["files_to_link"].items():
+    for src in driverobj.config["files_to_link"].values():
         Path(src).touch()
-    for dst, _ in driverobj.config["files_to_link"].items():
+    for dst in driverobj.config["files_to_link"]:
```
maddenp-noaa, line 92: `ruff` pointed out the simpler syntax adopted here.

src/uwtools/tests/test_cli.py
```diff
@@ -295,9 +289,9 @@ def test__check_verbosity_fail(capsys):
     assert "--quiet may not be used with --verbose" in capsys.readouterr().err
 
 
-@mark.parametrize("flags", ([STR.quiet], [STR.verbose]))
+@mark.parametrize("flags", [[STR.quiet], [STR.verbose]])
 def test__check_verbosity_ok(flags):
-    args = {flag: True for flag in flags}
+    args = dict.fromkeys(flags, True)
```
maddenp-noaa, line 294: `ruff` suggested this.

src/uwtools/tests/test_schemas.py
```diff
@@ -311,13 +310,7 @@ def non_empty_dict(errors: list[str]) -> bool:
...
 
 def non_empty_list(errors: list[str]) -> bool:
-    for msg in [
-        "[] is too short",  # jsonschema [4.18.0,4.20.*]
-        "[] should be non-empty",  # jsonschema [4.21.0,?]
-    ]:
-        if msg in errors:
-            return True
-    return False
+    return any(msg in errors for msg in ["[] is too short", "[] should be non-empty"])
```
maddenp-noaa, line 313: `ruff` pointed out this simpler boolean expression.

src/uwtools/utils/tasks.py
```diff
@@ -77,11 +80,14 @@ def existing_http(url: str):
     :param url: URL of the HTTP resource.
     """
     yield "Remote HTTP resource %s" % url
-    yield asset(url, lambda: requests.head(url, allow_redirects=True, timeout=3).status_code == 200)
+    yield asset(
+        url,
+        lambda: requests.head(url, allow_redirects=True, timeout=3).status_code == HTTPStatus.OK,
```
maddenp-noaa, line 85: `ruff` complained about the `200` "magic number", which prompted me to look for, and find, the (of course) `HTTPStatus` module.

.vscode/settings.json
```diff
@@ -1,5 +1,4 @@
 {
-  "python.linting.flake8Enabled": false,
-  "python.linting.pylintEnabled": true,
+  "python.linting.flake8Enabled": true,
```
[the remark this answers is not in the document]
elcarpenterNOAA: Both lines 2 and 3 are being [deprecated](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions) in the future. I think we can simply replace these two lines with `"flake8.enabled": true`. That's working for me locally without warning.

src/uwtools/api/config.py
```diff
@@ -23,14 +24,17 @@
...
 
 def compare(
-    config_1_path: Union[Path, str],
-    config_2_path: Union[Path, str],
-    config_1_format: Optional[str] = None,
-    config_2_format: Optional[str] = None,
+    config_1_path: Path | str,
+    config_2_path: Path | str,
+    config_1_format: str | None = None,
+    config_2_format: str | None = None,
```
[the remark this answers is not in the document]
christinaholtNOAA: I like this a lot!

src/uwtools/api/execute.py
```diff
@@ -83,12 +88,11 @@ def execute(
     log.debug("Instantiated %s with: %s", classname, kwargs)
     node: Node = getattr(driverobj, task)(dry_run=dry_run)
     if graph_file:
-        with open(graph_file, "w", encoding="utf-8") as f:
-            print(graph(node), file=f)
+        Path(graph_file).write_text(f"{node.graph}\n")
```
christinaholtNOAA, line 91: This is also pretty sweet!

src/uwtools/tests/drivers/test_orog.py
```diff
@@ -87,22 +87,22 @@ def test_Orog_driver_name(driverobj):
...
         assert (driverobj.rundir / dst).is_symlink()
 
 
 @mark.parametrize("exist", [True, False])
 def test_Orog_grid_file_existence(driverobj, logged, exist):
     grid_file = Path(driverobj.config["grid_file"])
-    status = f"Input grid file {str(grid_file)}: Not ready [external asset]"
+    status = f"Input grid file {grid_file!s}: Not ready [external asset]"
```
christinaholtNOAA, line 102: This is an interesting way to go about this—yet another way to format strings.

christinaholtNOAA (approved):
LGTM.

I especially like the `Union` -> `|` syntax and the transition from using context managers to `Path` methods. Also, it seems that `ruff` is finding more ways to be helpful than `pylint`.

There were lots of other solid changes I appreciated in there, too.

nbharwani-cu (approved):
LGTM! All the stylistic updates are a great addition and make the code much easier to read!

elcarpenterNOAA (approved):

Issue #634, opened 2024-10-28
Output methods for make_orog drivers


christinaholtNOAA:
# AC

- [ ] Add an interface for a public method task (if not already done)
- [ ] Provide the output files in an appropriate data structure (dict, list) as a public helper method on the object. 

# Description

Best guess at appropriate data structures for output of drivers:
* orog
* orog_gsl
* filter_topo

No need to hook into CLI.
shave is handled separately.

PR #728, opened 2025-04-28
Output methods for shave, make_orog, sfc_climo components

### LINKED ISSUE #633: Output method for sfc_climo driver
# AC

# Description

Best guess at appropriate data structures for output of sfc_climo_gen.

No need to hook into CLI.

maddenp-cu:
**Synopsis**

Implement `.output` property methods in `filter_topo`, `orog`, `orog_gsl`, `sfc_climo_gen`, and `shave` drivers.

Fixes #633 #634 #635 

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,7 +57,6 @@ extlinks = {
     "rst": ("https://www.sphinx-doc.org/en/master/usage/restructuredtext/%s", "%s"),
     "rtd": ("https://readthedocs.org/projects/uwtools/%s", "%s"),
     "schism": ("https://schism-dev.github.io/schism/master/%s", "%s"),
-    "sfc-climo-gen": ("https://ufs-community.github.io/UFS_UTILS/sfc_climo_gen/%s", "%s"),
     "ufs": ("https://ufs.epic.noaa.gov/%s", "%s"),
     "ufs-utils": ("https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#%s", "%s"),
     "ufs-weather-model": ("https://github.com/ufs-community/ufs-weather-model/%s", "%s"),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -57,6 +57,7 @@ extlinks = {
     "rst": ("https://www.sphinx-doc.org/en/master/usage/restructuredtext/%s", "%s"),
     "rtd": ("https://readthedocs.org/projects/uwtools/%s", "%s"),
     "schism": ("https://schism-dev.github.io/schism/master/%s", "%s"),
+    "sfc-climo-gen": ("https://ufs-community.github.io/UFS_UTILS/sfc_climo_gen/%s", "%s"),
     "ufs": ("https://ufs.epic.noaa.gov/%s", "%s"),
     "ufs-utils": ("https://noaa-emcufs-utils.readthedocs.io/en/latest/ufs_utils.html#%s", "%s"),
     "ufs-weather-model": ("https://github.com/ufs-community/ufs-weather-model/%s", "%s"),
--- a/docs/sections/user_guide/cli/drivers/mpas_init/show-schema.out
+++ b/docs/sections/user_guide/cli/drivers/mpas_init/show-schema.out
@@ -11,7 +11,7 @@
               "type": "integer"
             },
             "length": {
-              "minimum": 1,
+              "minimum": 0,
               "type": "integer"
             },
             "offset": {
--- a/docs/sections/user_guide/cli/drivers/orog/index.rst
+++ b/docs/sections/user_guide/cli/drivers/orog/index.rst
@@ -3,7 +3,7 @@
 
 .. include:: ../shared/idempotent.rst
 
-The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``orog``.
+The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``orog``. Documentation for this UFS Utils component is :ufs-utils:`here <orog>`.
 
 .. include:: help.rst
 
--- a/docs/sections/user_guide/cli/drivers/sfc_climo_gen/index.rst
+++ b/docs/sections/user_guide/cli/drivers/sfc_climo_gen/index.rst
@@ -3,7 +3,7 @@
 
 .. include:: ../shared/idempotent.rst
 
-The ``uw`` mode for configuring and running the :sfc-climo-gen:`sfc_climo_gen<>` component.
+The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``sfc_climo_gen``. Documentation for this UFS Utils component is :ufs-utils:`here <sfc-climo-gen>`.
 
 .. include:: help.rst
 
--- a/src/uwtools/drivers/chgres_cube.py
+++ b/src/uwtools/drivers/chgres_cube.py
@@ -114,12 +114,12 @@ class ChgresCube(DriverCycleLeadtimeBased):
         return STR.chgrescube
 
     @property
-    def output(self) -> dict[str, Path]:
+    def output(self) -> dict[str, list[Path]]:
         """
         Returns a description of the file(s) created when this component runs.
         """
         outfile = lambda x: self.rundir / f"out.{x}.tile7.nc"
-        return {"atm": outfile("atm"), "sfc": outfile("sfc")}
+        return {"atm": [outfile("atm")], "sfc": [outfile("sfc")]}
 
 
 set_driver_docstring(ChgresCube)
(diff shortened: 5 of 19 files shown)


docs/conf.py
```diff
@@ -57,7 +57,6 @@
     "rst": ("https://www.sphinx-doc.org/en/master/usage/restructuredtext/%s", "%s"),
     "rtd": ("https://readthedocs.org/projects/uwtools/%s", "%s"),
     "schism": ("https://schism-dev.github.io/schism/master/%s", "%s"),
-    "sfc-climo-gen": ("https://ufs-community.github.io/UFS_UTILS/sfc_climo_gen/%s", "%s"),
```
maddenp-cu, line 60: Updated the documentation so that this is no longer needed.

docs/sections/user_guide/cli/drivers/mpas_init/show-schema.out
```diff
@@ -11,7 +11,7 @@
               "type": "integer"
             },
             "length": {
-              "minimum": 1,
+              "minimum": 0,
```
maddenp-cu, line 14: Updated from `make docs` stemming from a previous PR.

docs/sections/user_guide/cli/drivers/orog/index.rst
```diff
@@ -3,7 +3,7 @@
 
 .. include:: ../shared/idempotent.rst
 
-The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``orog``.
+The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``orog``. Documentation for this UFS Utils component is :ufs-utils:`here <orog>`.
```
maddenp-cu, line 6: Add a missing link to upstream docs.

docs/sections/user_guide/cli/drivers/sfc_climo_gen/index.rst
```diff
@@ -3,7 +3,7 @@
 
 .. include:: ../shared/idempotent.rst
 
-The ``uw`` mode for configuring and running the :sfc-climo-gen:`sfc_climo_gen<>` component.
+The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``sfc_climo_gen``. Documentation for this UFS Utils component is :ufs-utils:`here <sfc-climo-gen>`.
```
maddenp-cu, line 6: For consistency with upstream-docs links for other components.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -114,13 +114,12 @@ def driver_name(cls) -> str:
         return STR.chgrescube
 
     @property
-    def output(self) -> dict[str, list[str]]:
+    def output(self) -> dict[str, Path]:
```
maddenp-cu, line 117: I'm not sure why we went with `str` instead of `Path` previously, as we generally prefer the latter wherever possible. Please let me know if anyone thinks `Path` isn't the right choice here.

src/uwtools/drivers/chgres_cube.py
```diff
@@ -114,13 +114,12 @@ def driver_name(cls) -> str:
...
         """
         Returns a description of the file(s) created when this component runs.
         """
-        output_types = ["atm", "sfc"]
-        paths = [f"{self.rundir}/out.{t}.tile7.nc" for t in output_types]
-        return {"netcdffiles": paths}
```
maddenp-cu, line 123: Simplified slightly.

src/uwtools/drivers/upp.py
```diff
@@ -119,15 +119,15 @@ def output(self) -> dict[str, list[str]]:
...
+            paths.append(self.rundir / (identifier + suffix))
             fields, lines = lines[: self.NFIELDS], lines[self.NFIELDS :]
             _, lines = (
                 (lines[0], lines[1:])
                 if fields[self.GENPROCTYPE_IDX] == "ens_fcst"
                 else (None, lines)
             )
             lines = lines[self.NPARAMS * nvars.pop() :]
-        return {"gribfiles": paths}
+        return {"paths": paths}
```
maddenp-cu, line 130: I'm really not sure what's best for the keys in these values, but started to think that lots of different names might not be useful, and might be a least-surprise violation. So, where it seemed like there were meaningful keys to use, I've used those; elsewhere I've just used `path` for a single item and `paths` for a list. Happy to hear other ideas!

elcarpenterNOAA (approved):

src/uwtools/drivers/orog_gsl.py
```diff
@@ -94,6 +94,16 @@ def driver_name(cls) -> str:
         """
         return STR.oroggsl
 
+    @property
+    def output(self) -> dict[str, Path]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        vals = tuple(self.config["config"][k] for k in ["resolution", "tile", "halo"])
```
christinaholtNOAA, line 102: TIL that Python doesn't have explicit tuple comprehensions.

christinaholtNOAA (commented):
This looks great.

I think any time we see `tile7` show up in one of these output files, we will expect size output files there for the global configuration (one each for tiles 1-6). How big a breaking change would we see in the future if we needed to create lists where the single items currently exist? Would we want to consider extending that now to be closer to futureproof for global support?

maddenp-cu:
> I think any time we see `tile7` show up in one of these output files, we will expect size output files there for the global configuration (one each for tiles 1-6). How big a breaking change would we see in the future if we needed to create lists where the single items currently exist? Would we want to consider extending that now to be closer to futureproof for global support?

@christinaholtNOAA I think it should be generalized, but also that the limitation is currently widespread in `uwtools` drivers and needs to be dealt with holistically, in tandem with a global-model use case that lets us validate what we're doing. We knew that we were making assumptions/guesses in the initial implementation of the drivers, but IIRC intended to integrate them into SRW and then fix any issues that arose; then, once we had the regional case locked down with SRW as a sort of integration test for the drivers in regional mode, implement global behavior -- with _something_ as a use case, but I'm not sure what. I don't think it makes much sense to try to implement global behavior that we have no way to test, when we have only partial integration with SRW to even test regional behavior. I'm not happy about any of this, but feel like this piecemeal approach is the only sane way forward. Is any of that inaccurate?

That said, I do think that I should go back and make sure that the values are _lists_ where they should be, even if currently they're lists of 1 item with `time7` hardcoded. (And I see that `sfc_climo_gen` actually creates `halo0` files _plus_ files for whatever `halo` is specified, if it's specified, which it doesn't have to be but _can_ be for regional mode... Ugh. So much additional validation should be done that isn't implemented in JSON Schema at the moment, and maybe can't reasonably be given how complex it could be.)

src/uwtools/drivers/chgres_cube.py
```diff
@@ -114,13 +114,12 @@ def driver_name(cls) -> str:
...
-    def output(self) -> dict[str, list[str]]:
+    def output(self) -> dict[str, list[Path]]:
         """
         Returns a description of the file(s) created when this component runs.
         """
-        output_types = ["atm", "sfc"]
-        paths = [f"{self.rundir}/out.{t}.tile7.nc" for t in output_types]
-        return {"netcdffiles": paths}
+        outfile = lambda x: self.rundir / f"out.{x}.tile7.nc"
+        return {"atm": [outfile("atm")], "sfc": [outfile("sfc")]}
```
maddenp-cu, line 122: Updated so that the values are `list[Path]` instead of just `Path`, to support multiple values for global configurations later without requiring breaking downstream changes.

src/uwtools/drivers/sfc_climo_gen.py
```diff
@@ -52,6 +53,17 @@ def provisioned_rundir(self):
...
+    @property
+    def output(self) -> dict[str, list[Path]]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        cfg = self.config["namelist"]["update_values"]["config"]
+        halo = cfg.get("halo")
+        ns = [0, halo] if halo else [0]
+        keys = [m[1] for key in cfg if (m := re.match(r"^input_(.*)_file$", key))]
+        return {key: [self.rundir / f"{key}.tile7.halo{n}.nc" for n in ns] for key in keys}
```
maddenp-cu, line 65: Updated so that the values are `list[Path]` instead of just `Path`, to support multiple values for global configurations later without requiring breaking downstream changes, and accounted for the fact that both `halo0` and e.g. `halo4` when `halo` is `4` output files are created.

maddenp-cu:
I'm going to go ahead and merge this, as I have another PR waiting in the wings. We can and should circle back on the regional-vs-global treatment of tiled outputs soon.

PR #729, opened 2025-04-30
Output methods for make_grid component drivers

### LINKED ISSUE #632: Output methods for SRW make_grid drivers
# AC

# Description

Best guess at appropriate data structures for output of drivers:
* make_hgrid
* esg_grid
* global_equiv_resol
* make_solo_mosaic

No need to hook into CLI.
shave is handled separately.

maddenp-cu:
**Synopsis**

Fixes #632.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/esg_grid.py
+++ b/src/uwtools/drivers/esg_grid.py
@@ -58,5 +58,12 @@ class ESGGrid(DriverTimeInvariant):
         """
         return STR.esggrid
 
+    @property
+    def output(self) -> dict[str, Path]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        return {"path": self.rundir / "regional_grid.nc"}
+
 
 set_driver_docstring(ESGGrid)
--- a/docs/sections/user_guide/cli/drivers/esg_grid/index.rst
+++ b/docs/sections/user_guide/cli/drivers/esg_grid/index.rst
@@ -3,7 +3,7 @@
 
 .. include:: ../shared/idempotent.rst
 
-The ``uw`` mode for configuring and running the :ufs-utils:`regional_esg_grid<regional-esg-grid>` component.
+The ``uw`` mode for configuring and running the UFS Utils preprocessing component ``regional_esg_grid``. Documentation for this UFS Utils component is :ufs-utils:`here <regional-esg-grid>`.
 
 .. include:: help.rst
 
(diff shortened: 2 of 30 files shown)


src/uwtools/drivers/esg_grid.py
```diff
@@ -58,5 +58,12 @@ def driver_name(cls) -> str:
         """
         return STR.esggrid
 
+    @property
+    def output(self) -> dict[str, Path]:
+        """
+        Returns a description of the file(s) created when this component runs.
+        """
+        return {"path": self.rundir / "regional_grid.nc"}
```
maddenp-cu, line 66: I hope I haven't missed something, but: The docs (e.g. [here](https://github.com/settings/profile)) for many of these components claim that output filenames include a CRES component, for example:

> File naming convention: CRES_grid.tile7.nc

However, looking at [the code](File naming convention: CRES_grid.tile7.nc), this does not seem to be the case, and even the [SRW script](https://github.com/ufs-community/ufs-srweather-app/blob/develop/scripts/exregional_make_grid.sh#L256) seem to indicate an output filename, `regional_grid.nc`, that matches the Fortran code. I think maybe the docs refer to a filename that SRW _renames_ the file to, which is IMO confusing and, to make matters worse, the SRW scripts do things like
```
grid_fn="regional_grid.nc"
...
grid_fp="$DATA/${grid_fn}"
...
grid_fp_orig="${grid_fp}"
grid_fn="${CRES}${DOT_OR_USCORE}grid.tile${TILE_RGNL}.halo${NHW}.nc"
grid_fp="${GRID_DIR}/${grid_fn}"
mv "${grid_fp_orig}" "${grid_fp}"
```
So the original output file just disappears and there's no trace of what happened.

Note to self/readers: Whenever possible, please use the `-v` switch on Linux utilities like `mv`, `cp`, `ln`, `mkdir`, etc. so that there is at least a record in logs of what happened to avoid mystifying people.

So, across these `.output` property methods, I've done my best from looking at the docs, the source code, the scripts, the logs, etc. to try to get the facts right, but I bet  we'll need to fix something eventually.

Last but not least, it looks like maybe the name of the compiled executable is `regional_esg_grd`, but our [example config](https://github.com/ufs-community/uwtools/blob/main/docs/shared/esg_grid.yaml#L20) has it as `esg_grid`. Should we change this in `uwtools`? Should we change the driver name?

elcarpenterNOAA: Do you still have the link to the code available to share?

maddenp-cu: Sorry about that -- I think I fixed it. The first link in my comment was also wrong, and I think I've fixed that now, too.

christinaholtNOAA (approved):

Issue #625, opened 2024-10-28
Output methods for ungrib, MPAS, and MPAS Init.


christinaholtNOAA:
# AC

- [ ] Add an interface for a public method task (if not already done)
- [ ] Provide the output files in an appropriate data structure (dict, list) as a public helper method on the object. 

# Description

Use stream files and other config settings to determine expected output for MPAS and MPAS Init and output in a dict with labels like those in stream file. Return a list (probably) of the ungribbed files.

No need to hook into CLI.

christinaholtNOAA:
## Ungrib 

Output files are named from namelist entries like this pseudo code:

for file_date in share.start_date to share.end_data in share.interval_seconds:
    file_name = f"{ ungrib.prefix }:{ file_date.strfmt("%Y-%m-%d_%h") }"

## MPAS Init

Output file is defined from the streams.init_atmosphere for any entries matching the type=output. The filename_template entry defines the name of the output files. They use templates that look like:
```
filename_template="lbc.$Y-$M-$D_$h.$m.$s.nc"
```

## MPAS

Same approach as MPAS, but the name of the streams file is streams.atmosphere. Both can probably use the values defined in the UW YAML.

PR #735, opened 2025-05-07
Implement output() methods on driver classes Ungrib and MPASInit

maddenp-cu:
**Synopsis**

Implements `output()` methods on driver classes `Ungrib` and `MPASInit`, per #625. Work on class `MPAS` will follow, but there's enough here to discuss for now, I think. Hopefully much of the work for `MPAS` will involve refactoring, as much of the logic will be similar.

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/mpas_init.py
+++ b/src/uwtools/drivers/mpas_init.py
@@ -132,11 +132,11 @@ class MPASInit(MPASBase):
 
     @staticmethod
     def _decode_interval(interval: str) -> dict[str, int]:
-        # See MPAS User Guide section 5 for a description of the interval format and semantics, but
-        # the general form is years-months-days_hours:minutes:seconds, e.g. 1-2-3_4:5:6 means 1
-        # year, 2 months, 3 days, 4 hours, 5 minutes, 6 seconds. Leading components can be omitted,
-        # so reverse for an ascending seconds -> years order, pad with trailing zeros as needed,
-        # then reverse again for a natural descending years -> seconds order.
+        # See MPAS User Guide section 5 in re: interval format and semantics, but the general form
+        # is years-months-days_hours:minutes:seconds, e.g. 1-2-3_4:5:6 means 1 year, 2 months, 3
+        # days, 4 hours, 5 minutes, 6 seconds. Leading components can be omitted, so reverse for an
+        # ascending seconds -> years order, pad with trailing zeros as needed, then reverse again
+        # for a natural descending years -> seconds order.
         keys = ["years", "months", "days", "hours", "minutes", "seconds"]
         components = re.sub(r"[-_:]", " ", interval).split()[::-1]
         vals = [next(islice(components, i, i + 1), 0) for i in range(len(keys))][::-1]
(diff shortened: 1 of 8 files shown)


src/uwtools/drivers/mpas_init.py
```diff
@@ -93,8 +100,107 @@ def driver_name(cls) -> str:
...
+        # See MPAS User Guide section 5.1 in re: filename_template logic.
+        kvs = [
+            ("$Y", "%Y"),
+            ("$M", "%m"),
+            ("$D", "%d"),
+            ("$d", "%j"),
+            ("$h", "%H"),
+            ("$m", "%M"),
+            ("$s", "%S"),
+        ]
```
maddenp-cu, line 200: I wonder if MPAS had a good reason for not adopting the 1989 C codes [like Python did](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes).

src/uwtools/drivers/ungrib.py
```diff
@@ -46,23 +49,18 @@ def namelist_file(self):
         """
         The namelist file.
         """
-        # Do not use offset here. It's relative to the MPAS fcst to run.
-        gribfiles = self.config["gribfiles"]
-        endhour = gribfiles["max_leadtime"]
-        end_date = self._cycle + timedelta(hours=endhour)
-        interval = int(gribfiles["interval_hours"]) * 3600  # hour to sec
```
maddenp-cu, line 53: Factored out to helper methods so that `output()` can use the same logic to get the end date and interval.

src/uwtools/drivers/ungrib.py
```diff
@@ -19,6 +19,9 @@ class Ungrib(DriverCycleBased):
     A driver for ungrib.
     """
 
+    PREFIX = "FILE"
```
christinaholtNOAA, line 22: We're already seeing that the GSL folks are renaming this prefix. Not sure if that sways the choice here.

christinaholtNOAA (commented):
This looks very thorough for the cases that we should need to support. A lot is going on here, and I am hopeful that it can be reused to some degree for the MPAS output.

src/uwtools/drivers/ungrib.py
maddenp-cu: It might have been obvious, but just in case: The driver was already hardwired to use `FILE`; this update just factors it out to a variable to keep the code DRY, as the value is now needed in two places.

In later work, we could factor this out further to the YAML config (possibly keeping, and documenting, a default value of `FILE`), in which case we should probably do the same in `mpas_init`, which hardcodes the same `FILE` value. The `ungrib` namelists are so simple that we don't even support a `base_file` option, so there's no chance that we're clobbering a user-specified `prefix` value, but that's a risk with the `mpas_init` namelist file, as we discussed in standup this morning.

For this and similar configuration values, a practical way forward might be to define a UW YAML key for setting the value. In a sense, this already exists for the `FILE` prefix in `mpas_init` (but not in `ungrib`), in that the user could provide an explicit config value `mpas_init.namelist.update_values.data_sources.config_met_prefix` -- except that the driver doesn't _look at_ that variable and just passes it through to the namelist, such that there could be a run-time mismatch error. We may or may not want to have users customize the value via this namelist item; maybe we accept it at a higher level, use it how we need it, and the _propagate_ it to the namelist to ensure consistency. Then we could try to detect cases where we're overriding a value provided by the user via `base_file` or `update_values` (possible with `mpas_init`, not currently possible with `ungrib`) and warn the user about it.

christinaholtNOAA: Ahhh. I didn't catch that it was just a refactoring of the already hard-coded value. In that case, this is totally fine. 

We might want to consider the other options in the future for using a different name. I'm not sure _why_ folks might be changing it since it's just this intermediate step, but I suppose we should give them the option.

PR #743, opened 2025-05-20
Improve UWYAMLTag value semantics

### LINKED ISSUE #742: UWYAMLTag value semantics are not useful
`UWYAMLTag` objects' value (equality) semantics are currently not useful, as they rely on the default `id()` check inherited from `object`. Useful `__eq__` methods should be defined on `UWYAMLTag` using its `tag` and `value` attributes.

**Current Behavior**

```
>>> import io, yaml
>>> from uwtools.config.support import uw_yaml_loader
>>> c0 = yaml.load(io.StringIO("!int '42'"), Loader=uw_yaml_loader())
>>> c1 = yaml.load(io.StringIO("!int '42'"), Loader=uw_yaml_loader())
>>> c0 == c1
False
```

**Expected Behavior**

```
>>> c0 == c1
True
```

**To Reproduce**

See above.

**Current Version**

All released versions < 2.8.

maddenp-cu:
**Synopsis**

Fixes #742.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/tests/config/test_support.py
+++ b/src/uwtools/tests/config/test_support.py
@@ -54,6 +54,17 @@ def test_config_support_log_and_error(logged):
     assert logged(msg)
 
 
+def test_config_support_UWYAMLTag__eq():
+    tag0, tag1 = [
+        support.UWYAMLTag(
+            yaml.SafeLoader("data"),
+            node=yaml.nodes.Node(tag="!foo", value="bar", start_mark=None, end_mark=None),
+        )
+        for _ in range(2)
+    ]
+    assert tag0 == tag1
+
+
 class TestUWYAMLConvert:
     """
     Tests for class uwtools.config.support.UWYAMLConvert.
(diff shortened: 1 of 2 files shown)


src/uwtools/tests/config/test_support.py
```diff
@@ -54,6 +54,17 @@ def test_config_support_log_and_error(logged):
...
 
+def test_config_support_UWYAMLTag__eq():
+    tag0, tag1 = [
+        support.UWYAMLTag(
+            yaml.SafeLoader("data"),
+            node=yaml.nodes.Node(tag="!foo", value="bar", start_mark=None, end_mark=None),
+        )
+        for _ in range(2)
+    ]
+    assert tag0 == tag1
```
maddenp-cu, line 65: Passes with the `__eq__` implemented above, fails without it.

elcarpenterNOAA (approved):

NaureenBharwaniNOAA (approved):

PR #751, opened 2025-06-11
Update to iotaa 1.3.0

maddenp-cu:
**Synopsis**

`iotaa` `1.3.0` deprecates the `.refs` property on task-node objects, replacing it with the singular `.ref`. (And similarly deprecates the `iotaa.refs()` helper function, replacing it with `.ref()`, though `uwtools` doesn't use this function.) It turns out to be most common for a task to yield a single asset, such that accessing the object (e.g. filesystem file) it references with singular terminology makes more sense. Even when a task yields a `list` or `dict` of assets, accessing them with e.g. `task.ref[0]` ("the task's 0 ref") or `task.ref["foo"]` ("the task's 'foo' ref") read better, too.

- Update the iotaa package range's lower bound to `1.3.0` to ensure that the `.refs` property is available.
- Update `.refs` accesses to `.ref`.

**Type**

**Impact**

**Checklist**


src/pyproject.toml
```diff
@@ -21,9 +21,6 @@ check_untyped_defs = true
 pretty = true
 warn_return_any = true
 
-[tool.pytest.ini_options]
-filterwarnings = ["ignore:The 'refs' property is deprecated:DeprecationWarning"]
```
maddenp-cu, line 25: No longer needed, as there are no more uses of the deprecated `.refs`.

elcarpenterNOAA (approved):

PR #756, opened 2025-06-25
Bugfix: Do not modify user-provided configs via drivers

maddenp-cu:
**Synopsis**

Fixes a bug whereby dereferencing a config in a driver's initializer affected a `YAMLConfig` object passed to the driver. Ensure that drivers have an independent deep copy of the user-provided config, insulating both parties from unexpected changes.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -55,15 +55,15 @@ class Assets(ABC):
         schema_file: Path | None = None,
         controller: list[YAMLKey] | None = None,
     ) -> None:
-        config_input = config if isinstance(config, YAMLConfig) else YAMLConfig(config=config)
-        config_input.dereference(
+        config_copy = YAMLConfig(config)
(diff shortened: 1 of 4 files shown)


src/uwtools/drivers/driver.py
```diff
@@ -55,15 +55,15 @@ def __init__(
         schema_file: Path | None = None,
         controller: list[YAMLKey] | None = None,
     ) -> None:
-        config_input = config if isinstance(config, YAMLConfig) else YAMLConfig(config=config)
-        config_input.dereference(
+        config_copy = YAMLConfig(config)
```
maddenp-cu, line 58: Now that `Config` classes, including `YAMLConfig`, deep-copy passed-in configs (see above), we can rely on the `YAMLConfig` initializer to disassociate the user-provided config from the config internal to the driver.

elcarpenterNOAA (approved):

PR #761, opened 2025-07-11
Chunked HTTP get to avoid memory issues

maddenp-cu:
**Synopsis**

To avoid out-of-memory issues when fetching large remote files via HTTP, use streaming, chunked reads/writes instead of reading the entire file into memory and _then_ writing it to disk. This should be an optimization invisible to users.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/utils/tasks.py
+++ b/src/uwtools/utils/tasks.py
@@ -182,9 +182,11 @@ def filecopy_http(url: str, dst: Path, check: bool = True):
     yield asset(dst, dst.is_file)
     yield existing_http(url) if check else None
     dst.parent.mkdir(parents=True, exist_ok=True)
-    response = requests.get(url, allow_redirects=True, timeout=3)
+    response = requests.get(url, allow_redirects=True, stream=True, timeout=3)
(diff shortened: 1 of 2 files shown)


src/uwtools/utils/tasks.py
```diff
@@ -182,9 +182,11 @@ def filecopy_http(url: str, dst: Path, check: bool = True):
     yield asset(dst, dst.is_file)
     yield existing_http(url) if check else None
     dst.parent.mkdir(parents=True, exist_ok=True)
-    response = requests.get(url, allow_redirects=True, timeout=3)
+    response = requests.get(url, allow_redirects=True, stream=True, timeout=3)
     if (code := response.status_code) == HTTPStatus.OK:
-        dst.write_bytes(response.content)
+        with dst.open(mode="wb") as f:
+            for chunk in response.iter_content(chunk_size=8192):
```
maddenp-cu, line 188: I tried doubling the chunk size, but download was no faster, so this value, taken from the `requests` docs, seems fine.

PR #762, opened 2025-07-12
Rocoto sh dependency attrs are optional

maddenp-cu:
**Synopsis**

I believe this fixes a bug whereby Python code mistakenly assumes that an `attrs:` block exists under a `sh:` dependency, while this is not required by the schema and appears not to be required by Rocoto.

**Type**

**Impact**

**Checklist**


christinaholtNOAA (approved):
Nice.

PR #766, opened 2025-07-21
Add a uw rocoto iterate action

### LINKED ISSUE #629: Consider adding a Rocoto action to monitor jobs.
@mkavulich has introduced code to the SRW E2E testing framework that runs and monitors a Rocoto workflow through completion (or various forms of failure) and logs the results.

This functionality would be useful for any app running a Rocoto workflow and could be foundational for building testing frameworks for those apps. It also removes the need for a user to have cron installed, and its use is restricted on some platforms.

The idea is that this code would fit into the `uw rocoto` tool under a new action named something like `monitor` (not a requirement) and would take arguments indicating the paths to an XML, a database file, and a log file, and would issue `rocotorun` and `rocotostat` alternatively until the workflow fails or completes, logging the status with each iteration.

Mike also added the ability to restart a monitoring session from an existing log file, so we'd likely want to take that into account.

maddenp-cu:
**Synopsis**

Add a `iterate` action to mode `rocoto` to iterate a Rocoto workflow to completion (successful or unsuccessful).

Fixes #629.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/rocoto.py
```diff
@@ -77,6 +91,94 @@ def validate_rocoto_xml_string(xml: str) -> bool:
...
+        self._con: sqlite3.Connection | None = None
+        self._cur: sqlite3.Cursor | None = None
+
+    def __del__(self):
+        if self._con:
+            self._con.close()
+
+    def run(self) -> bool:
+        state = self._state
+        while state not in self._states["inactive"]:
```
maddenp-cu, line 110: This initial checks provides an opportunity to _not_ iterate the workflow at all if the task in question is already in a terminal state.

src/uwtools/rocoto.py
```diff
@@ -77,6 +91,94 @@ def validate_rocoto_xml_string(xml: str) -> bool:
...
+            if not self._iterate():
+                return False
```
maddenp-cu, line 112: Bail out if there are system-related problems (e.g. `rocotorun` not being on `PATH`, which is the user's responsibility to ensure) that preclude iteration.

src/uwtools/rocoto.py
```diff
@@ -77,6 +91,94 @@ def validate_rocoto_xml_string(xml: str) -> bool:
...
+            state = self._state
+            if not state or state in self._states["active"]:
+                self._report()
+                log.debug("Sleeping %s seconds", self._rate)
+                sleep(self._rate)
```
maddenp-cu, line 117: After iteration, if the state is one of the inactive or transient states, just return to the top of the loop and either exit or iterate again, respectively. If the state is an active one, report on the workflow status, sleep, then iterate.

src/uwtools/rocoto.py
```diff
@@ -77,6 +91,94 @@ def validate_rocoto_xml_string(xml: str) -> bool:
...
+    @property
+    def _query_data(self) -> dict:
+        return {"taskname": self._task, "cycle": int(self._cycle.timestamp())}
+
+    @property
+    def _query_stmt(self) -> str:
+        return "select state from jobs where taskname=:taskname and cycle=:cycle"
+
+    def _report(self) -> None:
+        cmd = "rocotostat -d %s -w %s" % (self._database, self._workflow)
```
maddenp-cu, line 151: Unfortunately, Rocoto apparently does not create entries in its database for tasks that have not run at all yet. So, for example, if task `bar` depends on task `foo`, after one iteration of `rocotorun`, `bar` will not be in the database at all, even though `rocotostat` reports on it. That is to say, Rocoto has some internal knowledge about tasks that is not reflected in the database.

To help users understand what's going on, call `rocotostat` to show the state of the workflow. In the above example, the user would see that task `foo` is running and that task `bar` is known but not yet running, something we can't deduce just from looking at the database. After `foo` runs, `bar` will run, and _then_ it will appear in the database where we can track it.

We could just scrape the `rocotostat` output, but I think this is a clean solution: Use knowledge from the database as it becomes available, and show the user the `rocotostat` output for context.

docs/sections/user_guide/cli/Makefile.outputs
```diff
@@ -10,6 +10,12 @@ all: $(OUTPUTS)
 $(OUTPUTS):
 	@bash $(basename $@).cmd >$@ 2>&1 | true
 
+# The following targets support semi-automated output generation: They will not run automatically,
+# but can be manually invoked (e.g. "make foo.out" given a "foo.txt" command file) to upate .out
+# files. They must be invoked in a context where all commands in the .txt file are available, e.g.
+# on an HPC where the "hsi" or "rocotorun" commands are on PATH, if those are ultimately called by
+# the recipe.
```
maddenp-cu, line 17: Added some commentary as a reminder for what these targets are supposed to be used for.

src/uwtools/rocoto.py
```diff
@@ -85,6 +187,7 @@ class _RocotoXML:
     def __init__(self, config: dict | YAMLConfig | Path | None = None) -> None:
         self._config_validate(config)
         cfgobj = config if isinstance(config, YAMLConfig) else YAMLConfig(config)
+        cfgobj.dereference()
```
maddenp-cu, line 190: This fixed an issue with some Jinja2 expressions in a UW YAML for Rocoto config not being rendered.

docs/sections/user_guide/cli/Makefile.outputs
```diff
@@ -10,6 +10,12 @@ all: $(OUTPUTS)
 $(OUTPUTS):
 	@bash $(basename $@).cmd >$@ 2>&1 | true
 
+# The following targets support semi-automated output generation: They will not run automatically,
+# but can be manually invoked (e.g. "make foo.out" given a "foo.txt" command file) to upate .out
```
elcarpenterNOAA, line 14:
```suggestion
# but can be manually invoked (e.g. "make foo.out" given a "foo.txt" command file) to update .out
```

elcarpenterNOAA (approved):

PR #791, opened 2025-09-09
Enable !glob staging for drivers.

christinaholtNOAA:
**Synopsis**

Use the Copier and Linker tools to enable the use of the !glob tag in the files_to_copy and files_to_link sections of relevant Driver YAMLs.
Also adds support for files_to_hardlink section in those same drivers.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/stager.py
+++ b/src/uwtools/drivers/stager.py
@@ -1,5 +1,5 @@
 """
-A mixin class for tasks that stage files.
+A mixin class providing tasks that stage files.
 """
 
 from iotaa import tasks


src/uwtools/config/jinja2.py
```diff
@@ -152,6 +152,10 @@ def dereference(
         deref_debug("Rendering", val.value)
         val.value = _deref_render(val.value, context, local)
         rendered = _deref_convert(val)
+    elif isinstance(val, UWYAMLGlob):
+        deref_debug("Rendering", val.value)
+        val.value = _deref_render(val.value, context, local)
+        rendered = val
```
christinaholtNOAA, line 158: This allows me to use a Jinja expression in combination with a `!glob` tag.

src/uwtools/drivers/fv3.py
```diff
@@ -79,9 +80,21 @@ def files_copied(self):
         Files copied for run.
         """
         yield self.taskname("files copied")
+        yield [Copier(config=self.config.get("files_to_copy", {}), target_dir=self.rundir).go()]
```
christinaholtNOAA, line 83: This got a bit mangled in my GitHub view, but this replaces: 

```
            filecopy(src=Path(src), dst=self.rundir / dst)
            for dst, src in self.config.get("files_to_copy", {}).items()
```

And is similarly done for the Linker, and the new hardlinked call to the Linker.

src/uwtools/drivers/jedi_base.py
```diff
@@ -43,9 +44,21 @@ def files_copied(self):
         Files copied for run.
         """
         yield self.taskname("files copied")
+        yield [Copier(config=self.config.get("files_to_copy", {}), target_dir=self.rundir).go()]
+
+    @tasks
+    def files_hardlinked(self):
```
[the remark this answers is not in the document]
maddenp-cu: I was thinking something like
```
class FV3(DriverCycleBased, CopiesFiles):
    ...
```
where we'd define something like
```
class CopiesFiles:
    @tasks
    def files_copied(self):
        ... factored out task code ...
```
Then `FV3` would inherit the `files_copied` task method. It might be necessary to suppress some linter etc. complaints, as this class will necessarily reference instance variables that it doesn't actually have, but that it _will_ have when it's mixed in with a class like `FV3`. I'm not sure what the code-quality tools will say about this, but I think it's a valid technique.

I think that multiple inheritance has a somewhat negative reputation in Python, but I think the mixins are considered an acceptable use case. _Fluent Python_ has some material on it, but even just the Google AI overview for "python mixin class" gives a decent overview.

src/uwtools/drivers/fv3.py
```diff
@@ -79,9 +80,21 @@ def files_copied(self):
...
+        """
+        yield self.taskname("files hard-linked")
         yield [
-            filecopy(src=Path(src), dst=self.rundir / dst)
-            for dst, src in self.config.get("files_to_copy", {}).items()
+            Linker(
+                config=self.config.get("files_to_hardlink", {}),
+                target_dir=self.rundir,
+                hardlink=True,
+                symlink_fallback=True,
```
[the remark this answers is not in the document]
maddenp-cu: Interesting. Maybe instead of a `bool`-valued `symlink_fallback`, we should have an argument `fallback: str | None`so that one could instantiate e.g. `Linker(..., hardlink=True, fallback="copy")` or `Linker(..., hardlink=True, fallback="symlink")`, and in `uwtools.utils.tasks.hardlink()` we could do the specified fallback operation, or continue to intentionally fail if no fallback was specified. It seems like that'd be a pretty straightforward extension, and quite useful based on what you've pointed out here. (We should probably also check for values other than `"copy"` and `"symlink"` and raise a `UWError` for bad values.)

docs/sections/user_guide/cli/drivers/orog/help.out
```diff
@@ -12,6 +12,10 @@ Optional arguments:
 
 Positional arguments:
   TASK
+    files_copied
+      Files copied for run
+    files_hardlinked
+      Files hard-linked for run
```
christinaholtNOAA, line 18: For some reason, orog only supported `files_linked` previously, and I _think_ that should be up to the user on how to stage the files.

docs/sections/user_guide/cli/tools/fs/hardlink-help.out
```diff
@@ -2,7 +2,7 @@ usage: uw fs hardlink [-h] [--version] [--config-file PATH]
                       [--target-dir PATH] [--cycle CYCLE]
                       [--leadtime LEADTIME] [--dry-run]
                       [--key-path KEY[.KEY...]] [--report] [--quiet]
-                      [--verbose] [--symlink-fallback]
+                      [--verbose] [--fallback {copy,symlink}]
```
christinaholtNOAA, line 5: I also moved forward with this change, as suggested in an earlier review.

docs/sections/user_guide/yaml/components/fv3.rst
```diff
@@ -35,15 +35,7 @@ field_table:
...
-files_to_copy:
-^^^^^^^^^^^^^^
-
-See :ref:`this page <files_yaml>` for details.
-
-files_to_link:
-^^^^^^^^^^^^^^
-
-Identical to ``files_to_copy:`` except that symbolic links will be created in the run directory instead of copies.
+.. include:: /shared/stager.rst
```
christinaholtNOAA, line 38: I figured I might as well factor these out in docs, too.

src/uwtools/drivers/stager.py
```diff
@@ -0,0 +1,56 @@
+"""
```
christinaholtNOAA, line 1: I'm happy to change the name of the module or class, if this isn't a popular choice.

src/uwtools/tests/drivers/test_jedi.py
```diff
@@ -98,35 +98,6 @@ def test_JEDI_driver_name(driverobj):
     assert driverobj.driver_name() == JEDI.driver_name() == "jedi"
 
 
-def test_JEDI_files_copied(driverobj):
```
christinaholtNOAA, line 101: This seemed to be the only pre-existing tests of these tasks that failed given they were patching objects in this way. I left all the other passing ones in place. Does it make sense to replace these two in JEDI with tests similar to the other drivers or will the test of the FileStager methods suffice?

docs/sections/user_guide/yaml/components/fv3.rst
maddenp-cu: Nice.

src/uwtools/api/fs.py
```diff
@@ -59,14 +59,14 @@ def link(
...
-    symlink_fallback: bool = False,
+    fallback: str | None = None,
 ) -> dict[str, list[str]]:
     """
     Create links to filesystem items.
 
     When ``hardlink`` is ``False`` (the default), links may target files, hardlinks, symlinks, and
     directories; when ``True``, links may not be made across filesystems, or to directories. When
-    ``symlink_fallback`` is ``True``, a symlink will be created, if possible, if a hardlink cannot
+    ``fallback`` is set, a copy or symlink will be created, if possible, if a hardlink cannot
```
maddenp-cu, line 69: GitHub won't let me make the multi-line suggestion I want, but I suggest, as an alternative to the sentence starting with `When`:
```
Alternative if hardlink fails (choices: ``copy``, ``symlink``).
```
Since this will be in the rendered API docs, I think users may need to know that `copy` and `symlink` are the two words they can use.

src/uwtools/api/fs.py
```diff
@@ -77,7 +77,7 @@ def link(
     :param key_path: Path of keys to config block to use.
     :param dry_run: Do not link files.
     :param stdin_ok: OK to read from ``stdin``?
-    :param symlink_fallback: Symlink if hardlink fails when hardlink=True?
+    :param fallback: copy or symlink if hardlink fails when hardlink=True?
```
maddenp-cu, line 80:
```suggestion
    :param fallback: Alternative if hardlink fails (choices: ``copy``, ``symlink``).
```

src/uwtools/drivers/stager.py
```diff
@@ -0,0 +1,56 @@
...
+            Copier(
+                config=self.config.get("files_to_copy", {}),
+                target_dir=self.rundir,
+            ).go()
+        ]
+
+    @tasks
+    def files_hardlinked(self):
+        """
+        Files hard-linked for run.
```
maddenp-cu, line 33:
```suggestion
        Files hardlinked for run.
```
Can I propose that we treat "hardlink" as a single word the same way we treat "symlink", for consistency and to avoid additional punctuation? What do you think? (Fixing it here and re-running `make docs` should update all the occurrences in the RST docs.)

src/uwtools/drivers/stager.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        yield self.taskname("files hard-linked")
```
maddenp-cu, line 35:
```suggestion
        yield self.taskname("files hardlinked")
```

src/uwtools/drivers/stager.py
maddenp-cu: Works for me. It's not exposed via the API, so it'll be easy to change it later if we change our minds.

src/uwtools/drivers/stager.py
```diff
@@ -0,0 +1,56 @@
+"""
+A base class for tasks that stage files.
```
maddenp-cu, line 2:
```suggestion
A mixin class for tasks that stage files.
```
Just to be clear about our intent.

src/uwtools/utils/tasks.py
```diff
@@ -209,15 +209,15 @@ def filecopy_local(src: Path, dst: Path, check: bool = True):
...
+    target: Path | str, linkname: Path | str, check: bool = True, fallback: str | None = None
 ):
     """
     A hardlink.
 
     :param target: The existing file or directory.
     :param linkname: The symlink to create.
     :param check: Check existence of source before trying to link.
-    :param symlink_fallback: Symlink if hardlink fails?
+    :param fallback: copy or symlink if hardlink fails?
```
maddenp-cu, line 220:
```suggestion
    :param fallback: Alternative if hardlink fails (choices: 'copy', 'symlink').
```

src/uwtools/utils/tasks.py
```diff
@@ -229,8 +229,12 @@ def hardlink(
     try:
         os.link(src, dst)
     except Exception as e:
-        if symlink_fallback:
+        if fallback == STR.symlink:
             os.symlink(src, dst)
+            log.info("Could not hardlink %s -> %s, symlink instead" % (dst, src))
```
maddenp-cu, line 234:
```suggestion
            log.info("Could not hardlink %s -> %s, symlinked instead" % (dst, src))
```
Since it has already happened by the time this is logged...

src/uwtools/utils/tasks.py
```diff
@@ -229,8 +229,12 @@ def hardlink(
...
+        elif fallback == STR.copy:
+            copy(src, dst)
+            log.info("Could not hardlink %s -> %s, copy instead" % (dst, src))
```
maddenp-cu, line 237:
```suggestion
            log.info("Could not hardlink %s -> %s, copied instead" % (dst, src))
```

maddenp-cu (approved):
Nice! I made a few suggestions for your consideration.

PR #793, opened 2025-09-10
Input/Output Formats Default to YAML

### LINKED ISSUE #790: Consider making --input-format / --output-format default to "yaml"
**Description**

Manipulating YAML configs seems to be the most common thing done via `uw config`, and when piping together `uw` invocations into a multi-step transform, repeating the config format can become tedious. Consider making "yaml" the default when `--input-format` or `--output-format` are not specified.

This would have do be done with some care. For example, if `-i some.nml` is specified, behaving as if `--input-format yaml` was specified would be incorrect, as "Fortran namelist" should have been deduced from the input-file extension. Defaulting to YAML should only be done if `--input-format` / `--output-format` (highest priority, always honored) was not specified, and if the format could not be deduced from an extension (which is always the case when reading from `stdin` or writing to `stdout`).

Also, if YAML format was assumed and a parse error occurred because the assumption was wrong, users should get something better than a stacktrace. At least, we could log that we are defaulting to YAML because the format was neither specified or deduced.

**Acceptance Criteria**

- `uw config realize -i some.yaml --output-format yaml` could be simplified to `uw config realize -i some.yaml`.
- `cat some.yaml | uw config realize --input-format yaml -o out.yaml` could be simplified to `cat some.yaml | uw config realize -o out.yaml`.
- `cat some.yaml | uw config realize --input-format yaml --output-format yaml` could be simplified to `cat some.yaml | uw config realize`.

**Current Version**

Incubating version 2.10.0 and all previous releases.

maddenp-cu:
**Synopsis**

Fixes #790 .

Previous behavior example:
```
$ echo 'n: !int "{{ 1 + 2 }}"' | uw config realize --input-format yaml --output-format yaml
n: 3
```

New behavior example:
```
$ echo 'n: !int "{{ 1 + 2 }}"' | uw config realize
n: 3
```

You can still be explicit as needed, of course:

```
$ echo "&nl n = 3 /" | uw config realize --input-format nml
nl:
  n: 3
```

And deducing config format from filename extensions still works as expected, e.g.

```
$ echo "&nl n = 3 /" | uw config realize --input-format nml -o a.nml
$ cat a.nml
&nl
    n = 3
/
```

**Type**

- [x] Documentation

**Impact**

**Checklist**

### DIFF
--- a/notebooks/fs.ipynb
+++ b/notebooks/fs.ipynb
@@ -31,7 +31,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 13,
    "id": "77864d80-e6f4-48c2-a5d5-88fc512106a9",
    "metadata": {},
    "outputs": [],
@@ -57,7 +57,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 14,
    "id": "834654da-dfa9-4997-bcc5-846420381b18",
    "metadata": {},
    "outputs": [
@@ -88,7 +88,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "0585971b-47c6-48aa-9f1f-d5890cbb2061",
+   "id": "a939b7ac-8064-4469-a29e-819afb5128f6",
    "metadata": {},
    "source": [
     "Files to be copied are specified by a mapping from keys destination-pathname keys to source-pathname values, either in a YAML file or a a Python ``dict``."
@@ -96,7 +96,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 15,
    "id": "a959522f-d769-48c6-918d-d42776b3600a",
    "metadata": {},
    "outputs": [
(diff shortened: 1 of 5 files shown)


notebooks/fs.ipynb
```diff
@@ -773,15 +796,8 @@
...
-      "[2025-05-17T15:29:22]     INFO Schema validation succeeded for fs config\n",
-      "[2025-05-17T15:29:22]  WARNING Target fixtures/fs/missing-file.nml: Not ready [external asset]\n",
-      "[2025-05-17T15:29:22]  WARNING Link tmp/link-target/missing-link.nml -> fixtures/fs/missing-file.nml: Not ready\n",
-      "[2025-05-17T15:29:22]  WARNING Link tmp/link-target/missing-link.nml -> fixtures/fs/missing-file.nml: Requires:\n",
-      "[2025-05-17T15:29:22]  WARNING Link tmp/link-target/missing-link.nml -> fixtures/fs/missing-file.nml: ✖ Target fixtures/fs/missing-file.nml\n",
-      "[2025-05-17T15:29:22]  WARNING File links: Not ready\n",
-      "[2025-05-17T15:29:22]  WARNING File links: Requires:\n",
-      "[2025-05-17T15:29:22]  WARNING File links: ✖ Link tmp/link-target/missing-link.nml -> fixtures/fs/missing-file.nml\n"
+      "[2025-09-10T19:06:16]     INFO Validating config against internal schema: files-to-stage\n",
+      "[2025-09-10T19:06:16]     INFO Schema validation succeeded for fs config\n"
```
christinaholtNOAA, line 800: It's odd that this one used to be missing things and now seems to work....

christinaholtNOAA (approved):
This is a really nice change, especially when it's typically what _**I**_ want. Thanks!

notebooks/fs.ipynb
maddenp-cu: I just checked, and this change predates this PR -- the behavior change is already on `main`, we just haven't updated the notebooks in a while.

I think I see what's happening: The `Copier` class is `yield`ing a bunch of calls to `uwtools.utils.tasks.filecopy()`, but `filecopy()` is not a _task_, it's just a regular function. So it's working, because `filecopy()` does actually copy the file, but it doesn't become part of the `iotaa` task graph or do any of the expected logging. I'll fix this in a separate PR. Nice catch!

maddenp-cu: No, that's not quite true: `filecopy()` just delegated to other task functions, which _are_ `iotaa` `@task` functions. And the logging works fine on the command line:
```
$ echo "/tmp/copy: /tmp/no-such-file" | uw fs copy
[2025-09-10T23:20:02]     INFO Validating config against internal schema: files-to-stage
[2025-09-10T23:20:02]     INFO Schema validation succeeded for fs config
[2025-09-10T23:20:02]  WARNING File /tmp/no-such-file: Not ready [external asset]
[2025-09-10T23:20:02]  WARNING Local /tmp/no-such-file -> /tmp/copy: Not ready
[2025-09-10T23:20:02]  WARNING Local /tmp/no-such-file -> /tmp/copy: Requires:
[2025-09-10T23:20:02]  WARNING Local /tmp/no-such-file -> /tmp/copy: ✖ File /tmp/no-such-file
[2025-09-10T23:20:02]  WARNING File copies: Not ready
[2025-09-10T23:20:02]  WARNING File copies: Requires:
[2025-09-10T23:20:02]  WARNING File copies: ✖ Local /tmp/no-such-file -> /tmp/copy
```
and in the REPL:
```
>>> fs.copy(config={"/tmp/copy": "/tmp/no-such-file"})
[2025-09-10T23:23:20]     INFO Validating config against internal schema: files-to-stage
[2025-09-10T23:23:20]     INFO Schema validation succeeded for fs config
[2025-09-10T23:23:20]  WARNING File /tmp/no-such-file: Not ready [external asset]
[2025-09-10T23:23:20]  WARNING Local /tmp/no-such-file -> /tmp/copy: Not ready
[2025-09-10T23:23:20]  WARNING Local /tmp/no-such-file -> /tmp/copy: Requires:
[2025-09-10T23:23:20]  WARNING Local /tmp/no-such-file -> /tmp/copy: ✖ File /tmp/no-such-file
[2025-09-10T23:23:20]  WARNING File copies: Not ready
[2025-09-10T23:23:20]  WARNING File copies: Requires:
[2025-09-10T23:23:20]  WARNING File copies: ✖ Local /tmp/no-such-file -> /tmp/copy
{'ready': [], 'not-ready': ['/tmp/copy']}
```
So something else is going on. Not sure what yet.

PR #805, opened 2025-09-17
Fail rocoto realize if XML contains unrendered expressions.

christinaholtNOAA:
**Synopsis**

If any unrendered jinja expressions remain in the Rocoto XML, report which values are needed and raise an error.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/rocoto.py
+++ b/src/uwtools/rocoto.py
@@ -14,14 +14,16 @@ from pathlib import Path
 from time import sleep
 from typing import TYPE_CHECKING, Any
 
+from jinja2 import Environment, StrictUndefined, meta
 from lxml import etree
 from lxml.builder import E  # type: ignore[import-not-found]
 from lxml.etree import Element, SubElement, _Element
 
 from uwtools.config.formats.yaml import YAMLConfig
+from uwtools.config.jinja2 import unrendered
 from uwtools.config.validator import validate_external as validate_yaml
-from uwtools.exceptions import UWConfigError, UWError
-from uwtools.logging import log
+from uwtools.exceptions import UWConfigError, UWConfigRealizeError, UWError
+from uwtools.logging import INDENT, log
 from uwtools.utils.file import readable, resource_path, writable
 from uwtools.utils.processing import run_shell_cmd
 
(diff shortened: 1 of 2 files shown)


src/uwtools/rocoto.py
```diff
@@ -58,6 +60,15 @@ def realize(config: YAMLConfig | Path | None, output_file: Path | None = None) -
     """
     rxml = _RocotoXML(config)
     xml = str(rxml).strip()
+    if unrendered(xml):
+        log.error(xml)
+        log.error("Value(s) needed to render this XML are:")
+        for var in meta.find_undeclared_variables(
+            Environment(undefined=StrictUndefined).parse(xml)
+        ):
+            log.error("%s%s", INDENT, var)
```
christinaholtNOAA, line 69: These lines exist in the jinja module, but are spread among several private methods/functions. I felt like it might be simple enough just to repeat it here. Thoughts?

I also chose not to break it out into a helper function because it was pretty short, and clearly shows what to do if there's unrendered expressions in the XML.

christinaholtNOAA: Oh, also, none of the private Jinja methods are set up to take a template from memory.

maddenp-cu: I'm happy with this low-key repetition: If we ever wanted slightly different behavior, messages, etc. for this use case, we wouldn't have to coordinate it with other use cases.

src/uwtools/rocoto.py
```diff
@@ -58,6 +60,15 @@ def realize(config: YAMLConfig | Path | None, output_file: Path | None = None) -
     """
     rxml = _RocotoXML(config)
     xml = str(rxml).strip()
+    if unrendered(xml):
+        log.error(xml)
```
maddenp-cu, line 64: Is `xml` potentially a big block of XML? If so, we might prefer
```
for line in xml.split("\n"):
    log.error(line)
```
to make it log nicely over multiple lines, vs one huge multiline string. That's what we generally do elsewhere when we're expecting big multiline output. Your call.

maddenp-cu (approved):

src/uwtools/rocoto.py
christinaholtNOAA: It is indeed a single, large block of XML. I kind of appreciated that the XML showed up without log message prefixes on every line in this particular case. It looks like an XML would look if it were written to a file.

PR #815, opened 2025-10-31
Tooling updates

maddenp-cu:
**Synopsis**

Updates to runtime and test dependency packages, plus trivial code updates addressing issues flagged by the newer `ruff`:

- Use `|` for union types instead of the `Union` constructor.
- Use `Path.symlink()` instead of `os.symlink`.
- Use the `strict=` keyword argument with `zip()` to be explicit about whether or not the two sequences are expected to be the same length.
- Import `Callable` from `collections.abc` instead of `typing`.

No behavioral changes.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/notebooks/tests/conftest.py
+++ b/notebooks/tests/conftest.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
 from pathlib import Path
-from collections.abs import Callable
+from typing import TYPE_CHECKING
 
 from pytest import fixture
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 
 @fixture


src/uwtools/utils/time.py
```diff
@@ -25,5 +25,5 @@ def to_timedelta(value: int | str | timedelta) -> timedelta:
     if isinstance(value, int):
         return timedelta(hours=value)
     keys = ["hours", "minutes", "seconds"]
-    args = dict(zip(keys, map(int, value.split(":"))))
+    args = dict(zip(keys, map(int, value.split(":")), strict=False))
```
christinaholtNOAA: This one makes sense that we might not have equal lengths for both zipped items.

maddenp-cu: Exactly. Thank you for the close look on these.

PR #816, opened 2025-11-03
Enable coverage of test code

maddenp-cu:
**Synopsis**

Let `coverage` report coverage of unit-test code, too. Make changes to get coverage back to 100%.

**Type**

**Impact**

**Checklist**


src/pyproject.toml
```diff
@@ -3,7 +3,6 @@ exclude_also = [
   "if TYPE_CHECKING:"
 ]
 fail_under = 100
-omit = ["*/tests/*"]
```
maddenp-cu: This is the main change. Everything that follows deals with the side effects.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -40,30 +40,25 @@ class ConcreteConfig(Config):
     """
 
     @classmethod
-    def _dict_to_str(cls, cfg):
-        pass
+    def _dict_to_str(cls, cfg): ...
```
maddenp-cu: The use of `[Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis)` instead of `pass` continues to make it unnecessary to define a real function/method body, but does not produce an additional line of code that would need to be covered.

src/uwtools/tests/config/formats/test_base.py
```diff
@@ -96,12 +91,12 @@ def test_config_base__depth(config):
...
 
 def test_config_base__load_paths(config, tmp_path):
-    paths = (tmp_path / fn for fn in ("f1", "f2"))
+    paths = [tmp_path / fn for fn in ("f1", "f2")]
     for path in paths:
         path.write_text(yaml.dump({path.name: "defined"}))
     cfg = config._load_paths(config_files=paths)
     for path in paths:
-        assert cfg[path.name] == "present"
+        assert cfg[path.name] == "defined"
```
maddenp-cu: It turned out that `paths` was empty and that the `for` loop's body was never executed, because `paths` was previously a generator, and was exhausted by the previous `for` loop. By making it a list instead, both `for` loops have something to iterate over. Once the second loop was actually executing, the test failed since the code had been updated to behave differently.

src/uwtools/tests/config/test_validator.py
```diff
@@ -43,41 +42,6 @@ def config(tmp_path) -> dict[str, Any]:
...
-                                    "task": "hello",
-                                },
-                            },
-                        },
-                    },
-                },
-            },
-        }
-    }
-    return kwargs, config
```
maddenp-cu: Never called, dead code.

src/uwtools/tests/test_cli.py
```diff
@@ -23,9 +23,8 @@
 
 def actions(parser: Parser) -> list[str]:
     # Return actions (named subparsers) belonging to the given parser.
-    if actions := [x for x in parser._actions if isinstance(x, _SubParsersAction)]:
-        return list(actions[0].choices.keys())
-    return []
```
maddenp-cu: Line 28 was never executed in the old version of the test.

src/uwtools/tests/test_schemas.py
```diff
@@ -359,13 +359,13 @@ def ww3_prop():
 
 
 def non_empty_dict(errors: list[str]) -> bool:
-    for msg in [
-        "{} does not have enough properties",  # jsonschema [4.18.0,4.20.*]
-        "{} should be non-empty",  # jsonschema [4.21.0,?]
-    ]:
-        if msg in errors:
-            return True
-    return False
```
maddenp-cu: One of these (can't remember which) was never executed by the tests. The new version returns the appropriate `bool` value with no uncovered code.

PR #824, opened 2025-11-18
Update to iotaa 1.4.x

maddenp-cu:
**Synopsis**

Update conda recipe to use `iotaa` `v1.4.x` and make corresponding code changes.

There are a lot of individual changes, but they are almost all copies of a small set of changes:

- Import and call the `iotaa` `Asset` constructor to create an asset. (The former asset-creating `asset()` function now replaces the old `assets()` function, which can be passed an `iotaa` `Node` object to retrieve the asset(s) readied by that  task node.)
- Import and use the `@collection` decorator instead of the old `@tasks` decorator.
- `Node` objects now provide a `.req` property instead of `.requirements`. As with `.ref` and `.asset`, it's common for tasks to have only a single requirement, making the plural nonsensical. And expressions like `req[0]` to refer to the 0th requirement in a `list` of requirements, or `req["gribfile"]` to refer to a required task responsible for readying a GRIB file, read well.

Also:

- Fix a documentation link.
- Re-render notebooks to get some changes made in a previous PR.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/api/execute.py
```diff
@@ -58,7 +58,7 @@ def execute(
     :param graph_file: Write Graphviz DOT output here.
     :param key_path: Path of keys to config block to use.
     :param stdin_ok: OK to read from stdin?
-    :return: The assets yielded by the task, if it completes without raising an exception.
+    :return: The task-graph Node yielded by the task, if it completes without raising an exception.
```
maddenp-cu: This is more accurate.

src/uwtools/cli.py
```diff
@@ -338,7 +338,7 @@ def _dispatch_execute(args: Args) -> bool:
 
     :param args: Parsed command-line args.
     """
-    assets = uwtools.api.execute.execute(
+    node = uwtools.api.execute.execute(
```
maddenp-cu: This is more accurate.

PR #834, opened 2025-12-17
Update tooling

maddenp-cu:
**Synopsis**

Update dev packages and install required type packages via conda instead of having `mypy` install them separately via pip.

**Type**

**Impact**

**Checklist**


recipe/meta.yaml
```diff
@@ -29,12 +29,18 @@ test:
...
+    - pytest 9.0.*
     - pytest-cov 7.0.*
     - pytest-xdist 3.8.*
     - python {{ pymin }}
     - ruff 0.14.*
+    - types-PyYAML
+    - types-jsonschema
+    - types-python-dateutil
+    - types-requests
+    - types-setuptools
```
maddenp-cu: Installing these packages (plus `lxml-stubs`) allows us to simplify the `mypy` invocation in `recipe/run_script.sh`, and prevents `mypy` from installing typing packages via pip, leading to a pure-conda dev environment.

In the future, when updating the versions of dev packages, it would be a good idea to temporarily comment these type packages out and make sure they're still required: Any package could choose, in a new version, to include typing information in its core packages, eliminating the need for a separate type package. `mypy` will complain if types are missing and will suggest a type package to install. In most cases, the conda-forge package name matches the `mypy`-suggested pip package name.

PR #837, opened 2026-01-08
Serialize timedelta objects in parseable form

maddenp-cu:
**Synopsis**

This issue came up in practice. Given

`a.yaml`
```
start: !datetime 2025-01-01T00
stop: !datetime 2025-01-03T03
raw: !timedelta '{{ ((stop - start) * 0.8).total_seconds() / 3600 }}'
```
Current behavior:
```
$ uw config realize -i a.yaml | tee b.yaml
start: 2025-01-01T00:00:00
stop: 2025-01-03T03:00:00
raw: !timedelta '1 day, 16:48:00'
```
```
$ python
>>> from uwtools.api.config import get_yaml_config
>>> c = get_yaml_config("b.yaml")
>>> c["raw"]
  File "/home/maddenp/git/uwtools/src/uwtools/config/support.py", line 204, in __repr__
    return "%s %s" % (self.tag, self.converted)
  File "/home/maddenp/git/uwtools/src/uwtools/config/support.py", line 226, in converted
    return dict(zip(UWYAMLConvert.TAGS, converters, strict=True))[self.tag](self.value)
  File "/home/maddenp/git/uwtools/src/uwtools/utils/time.py", line 28, in to_timedelta
    args = dict(zip(keys, map(float, value.split(":")), strict=False))
ValueError: could not convert string to float: '1 day, 16'
```
The problem is that the value `'1 day, 16:48:00'` that appears in `b.yaml`, which is the standard representation of the `timedelta` object given by calling `str()` on it, is not parseable by `uwtools` as a `timedelta` object.

New behavior:
```
$ uw config realize -i a.yaml | tee b.yaml
start: 2025-01-01T00:00:00
stop: 2025-01-03T03:00:00
raw: !timedelta '40:48:00'
```
```
$ python
>>> from uwtools.api.config import get_yaml_config
>>> c = get_yaml_config("b.yaml")
>>> c["raw"]
!timedelta 1 day, 16:48:00
```
It's fine for the `timedelta` object to be _internally_ represented as `1 day, 16:48:00`, but when it is serialized (dumped) to YAML, it needs to be in the format that `uwtools` supports so that it can be read in again later.

After this merges, I'd like to do a `v2.12.2` bugfix release of this, too.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/support.py
+++ b/src/uwtools/config/support.py
@@ -51,6 +51,16 @@ def add_yaml_representers() -> None:
     """
     Add representers to the YAML dumper for custom types.
     """
+
+    def timedelta2str(dumper: yaml.dumper.Dumper, data: timedelta) -> yaml.ScalarNode:
+        seconds = int(data.total_seconds())
+        hours = seconds // 3600
+        seconds %= 3600
+        minutes = seconds // 60
+        seconds %= 60
+        s = f"{hours}:{minutes:02d}:{seconds:02d}"
+        return dumper.represent_scalar("!timedelta", s)
+
     yaml.add_representer(Namelist, _represent_namelist)
     yaml.add_representer(OrderedDict, _represent_ordereddict)
     yaml.add_representer(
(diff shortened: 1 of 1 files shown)


src/uwtools/tests/config/test_support.py
```diff
@@ -182,6 +199,17 @@ def test_UWYAMLConvert_list_ok(self, loader):
...
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!timedelta", value="null"))
+        with raises(ValueError, match="could not convert string to float"):
+            assert ts.converted
+
+    @mark.parametrize("value", ["49", "49:00", "49:00:00"])
+    def test_UWYAMLConvert__timedelta_ok(self, loader, value):
+        ts = support.UWYAMLConvert(loader, yaml.ScalarNode(tag="!timedelta", value=value))
+        assert ts.converted == timedelta(hours=49)
+        self.comp(ts, f"!timedelta '{value}'")
+
```
maddenp-cu: Tangentially related tests that seemed to have been missing.

PR #845, opened 2026-03-19
Addition of an ecFlow tool.

christinaholtNOAA:
**Synopsis**

This is the basis for an ecFlow tool that translates YAML to an ecFlow definition. 

Fixes Issue #847 

Tests/docs/JSONSchema are all meant to be contributed separately.

In my working example, I'm testing with the following YAML:

```
ecflow:
  scheduler: "{{ platform.scheduler }}"
  suites_{{ ec.type }}:
    expand:
      type: [prod]
    family_primary:
      families_{{ ec.synop }}:
        expand:
          synop: !list '[{% for h in range(0, 23, 23) %}"{{ "%02d" % h }}",{% endfor %} ]'
        family_aigfs:
          vars:
            PROJ: AIGFS
            PACKAGEHOME: /scratch3/BMC/wrfruc/cholt/aigfs/aigfs/ecf
            EVAL: "YES"
          family_v1.0:
            family_forecast:
              task_jaigfs_prep:
                script: &script
                  account: myacct
                  pre_includes:
                    - head.h
                    - envir-1.h
                  post_includes:
                    - tail.h
                  manual: "Prepares AIGFS ICS from GFS."
                  execution: !dict "{{ prep.aigfs_ics.execution}}"
                  rundir: "{{ prep.aigfs_ics.rundir }}"
              task_jaigfs_forecast:
                trigger: jaigfs_prep == complete
                events: !list '[{% for h in range(0, 13, 6) %}"{{ "release_f%03d" % h }}",{% endfor %} ]'
                script:
                  <<: *script
                  manual: "Runs AIGFS forecast."
                  execution: !dict "{{ forecast.graphcast_model.execution}}"
                  rundir: "{{ forecast.graphcast_model.rundir }}"
              family_post:
                tasks_{{ ec.synop }}Z_post_f{{ ec.fhr }}:
                  expand:
                    fhr: !list '[{% for h in range(0, 13, 6) %}"{{ "%03d" % h }}",{% endfor %} ]'
                  trigger: "../jaigfs_forecast:release_f{{ ec.fhr }}"
                  script:
                    <<: *script
                    manual: "Runs AIGFS post for {{ ec.fhr }}."
                    execution: !dict "{{ post.aigfs_post.execution}}"
                    rundir: "{{ post.aigfs_post.rundir }}"
                  vars:
                    FHR: "{{ ec.fhr }}"
```

And the result is:

```
#5.15.2
suite prod
  family primary
    family 00
      family aigfs
        edit PROJ 'AIGFS'
        edit PACKAGEHOME '/scratch3/BMC/wrfruc/cholt/aigfs/aigfs/ecf'
        edit EVAL 'YES'
        family v1.0
          family forecast
            task jaigfs_prep
            task jaigfs_forecast
              trigger jaigfs_prep == complete
              event release_f000
              event release_f006
              event release_f012
            family post
              task 00Z_post_f000
                trigger ../jaigfs_forecast:release_f000
                edit FHR '000'
              task 00Z_post_f006
                trigger ../jaigfs_forecast:release_f006
                edit FHR '006'
              task 00Z_post_f012
                trigger ../jaigfs_forecast:release_f012
                edit FHR '012'
            endfamily
          endfamily
        endfamily
      endfamily
    endfamily
  endfamily
endsuite
# enddef
```

One key aspect that I haven't seen a in any prior solutions is the `expand` support. When included, as in the `family_post` set of `tasks_*`, multiple tasks will be created with the variables included in the `expand` section realized. This adds a looping mechanism to reduce duplication in the YAML definition.

The tool also writes all the task scripts, as defined by the `task: script:` blocks.

For example, `prod/primary/00/aigfs/v1.0/forecast/forecast.ecf` contains:

```
#SBATCH --account=myacct
#SBATCH --chdir={{ user.experiment_dir }}/{{ cycle.strftime("%Y%m%d%H") }}/forecast
#SBATCH --mem=150GB
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=128
#SBATCH --output={{ user.experiment_dir }}/{{ cycle.strftime("%Y%m%d%H") }}/forecast.out
#SBATCH --time=01:30:00
#SBATCH --verbose

model=%MODEL%

%include <head.h>
%include <envir-1.h>

source /scratch3/BMC/wrfruc/cholt/aigfs/etc/profile.d/conda.sh
conda activate aigfs

run graphcast
if [[ $? -ne 0 ]]; then
   ecflow_client --msg="***JOB $ECF_NAME ERROR RUNNING J-SCRIPT ***"
   ecflow_client --abort
   exit 1
fi

%include <tail.h>

%manual
Runs AIGFS forecast.
%end
```

I've been testing with a super simple script:

```
from pathlib import Path
from uwtools.ecflowtool import _ecFlowDef

config = Path("aigfs/ush/wflow.yaml")
e = _ecFlowDef(config)
print(e)
e.write_ecf_scripts(Path(".", "ecf"))
e.write_suite_definition(Path(".", "sd"))
```

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -26,6 +26,7 @@ from ecflow import (  # type: ignore[import-untyped]
 
 from uwtools.config.formats.base import Config
 from uwtools.config.formats.yaml import YAMLConfig
+from uwtools.exceptions import UWConfigError
 from uwtools.logging import log
 from uwtools.scheduler import JobScheduler
 from uwtools.strings import STR
@@ -40,16 +41,16 @@ class _ECFlowDef:
     """
 
     def __init__(self, config: dict | Config | Path | None = None) -> None:
-        self.scripts: dict[Path, str] = {}
+        self._scripts: dict[Path, str] = {}
         cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
         cfgobj = cfgobj.dereference()
-        self._config = cfgobj.data.get(STR.ecflow, cfgobj.data)
-        self.scheduler = self._config.get("scheduler")
-        self.d = Defs()
+        self._config = cfgobj.data[STR.ecflow]
+        self._scheduler = self._config.get("scheduler")
+        self._d = Defs()
         self._add_workflow_components()
 
     def __str__(self):
-        return self.d.__str__()
+        return self._d.__str__()
 
     def write_ecf_scripts(self, path: Path | str) -> None:
         """
@@ -57,7 +58,12 @@ class _ECFlowDef:
 
         :param path: Where to write the ecFlow scripts.
         """
-        for subpath, content in self.scripts.items():
+
+        if not self._scripts:
+            log.warning("No scripts are configured for this workflow.")
+            return
+
+        for subpath, content in self._scripts.items():
             outpath = path / subpath
             outpath.parent.mkdir(parents=True, exist_ok=True)
             outpath.write_text(content)
@@ -68,9 +74,10 @@ class _ECFlowDef:
 
         :param path: Where to write the suite definition.
         """
-        path = Path(path, "suite.def")
-        path.parent.mkdir(parents=True, exist_ok=True)
-        path.write_text(self.d.__str__())
+        path = Path(path)
+        path.mkdir(parents=True, exist_ok=True)
+        suite = path / "suite.def"
+        suite.write_text(self._d.__str__())
 
     def _add_workflow_components(self) -> None:
         """
@@ -81,13 +88,13 @@ class _ECFlowDef:
             match tag:
                 case "extern":
                     for ext in subconfig:
-                        self.d.add_extern(ext)
+                        self._d.add_extern(ext)
                 case "vars":
-                    self.d.add_variable(subconfig)
+                    self._d.add_variable(subconfig)
                 case "suite":
-                    self._add_node(subconfig, Suite(name), self.d)
+                    self._add_node(subconfig, Suite(name), self._d)
                 case "suites":
-                    self._expand_block(subconfig, name, Suite, self.d)
+                    self._expand_block(subconfig, name, Suite, self._d)
 
     def _expand_block(
         self,
@@ -112,11 +119,10 @@ class _ECFlowDef:
 
         # Check to make sure all lists are the same length.
         try:
-            for _i in zip(*expand.values(), strict=True):
-                pass
-        except ValueError:
-            log.error("All expand variables under %s must be the same length" % (parent.name()))
-            raise
+            assert list(zip(*expand.values(), strict=True))
+        except ValueError as e:
+            msg = "All expand variables under %s must be the same length" % (parent.name())
+            raise UWConfigError(msg) from e
 
         # Build up the new blocks in the suite definition.
         primary_variable = list(expand.keys())[0]
@@ -153,6 +159,8 @@ class _ECFlowDef:
         for key, subconfig in config.items():
             tag, name = self._tag_name(key)
             match tag:
+                # Tree buiding cases
+
                 case "family":
                     self._add_node(subconfig, Family(name), node, refs)
                 case "families":
@@ -161,6 +169,9 @@ class _ECFlowDef:
                     self._add_node(subconfig, Task(name), node, refs)
                 case "tasks":
                     self._expand_block(subconfig, name, Task, node, refs)
+
+                # Node attribute cases
+
                 case "defstatus":
                     node.add_defstatus(subconfig)
                 case "events":
@@ -193,38 +204,18 @@ class _ECFlowDef:
         :param name: The name of the repeat.
         :param node: The node to add the repeat to.
         """
-        """
-        YAML:
-
-        repeat:
-          variable: myvar
-
-          WITH
-
-          start:
-          end:
-          step: (optional)
-
-          OR
-
-          step:
-
-          OR
-
-          list:
-        """
         match name:
             case "date":
                 config["delta"] = config.pop("step", None)
                 repeat = RepeatDate
-            case "datetime":
-                repeat = RepeatDateTime
-            case "int":
-                repeat = RepeatInteger
-            case "day":
-                repeat = RepeatDay
             case "datelist" | "enumerated" | "string":
                 repeat = RepeatEnumerated
+            case "datetime":
+                repeat = RepeatDateTime
+            case "day":
+                repeat = RepeatDay
+            case "int":
+                repeat = RepeatInteger
         node.add_repeat(repeat(**config))
 
     def _create_ecf_script(self, config: dict, task: Task) -> None:
@@ -235,12 +226,12 @@ class _ECFlowDef:
         :param task: The task node.
         """
         scheduler = (
-            self._scheduler(
+            self._jobscheduler(
                 account=config.get("account", ""),
                 execution=config.get("execution", ""),
                 rundir=config.get("rundir", ""),
             )
-            if self.scheduler
+            if self._scheduler
             else None
         )
         execution = config[STR.execution]
@@ -254,8 +245,11 @@ class _ECFlowDef:
             scheduler=scheduler,
         )
 
-        path = Path(task.get_abs_node_path().lstrip("/")).parent /  f"{task.name().split('_', 1)[-1]}.ecf"
-        self.scripts[path] = es
+        path = (
+            Path(task.get_abs_node_path().lstrip("/")).parent
+            / f"{task.name().split('_', 1)[-1]}.ecf"
+        )
+        self._scripts[path] = es
 
     def _ecflowscript(
         self,
@@ -318,7 +312,7 @@ class _ECFlowDef:
         )
         return re.sub(r"\n\n\n+", "\n\n", rs.strip())
 
-    def _scheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
+    def _jobscheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
         """
         Use the execution block to build a JobScheduler object.
 
@@ -330,7 +324,7 @@ class _ECFlowDef:
         resources = {
             STR.account: account,
             STR.rundir: rundir,
-            STR.scheduler: self.scheduler,
+            STR.scheduler: self._scheduler,
             STR.stdout: "%s.out" % Path(rundir),
             **({STR.threads: threads} if threads else {}),
             **execution.get(STR.batchargs, {}),


src/uwtools/ecflow.py
```diff
@@ -0,0 +1,353 @@
...
+                repeat = RepeatDate
+            case "datetime":
+                repeat = RepeatDateTime
+            case "int":
+                repeat = RepeatInteger
+            case "day":
+                repeat = RepeatDay
+            case "datelist|enumerated|string":
+                repeat = RepeatEnumerated
+        node.add_repeate(repeat(**config))
```
AnnaSmoot-NOAA: Possible typo. Should be add_repeat()?

src/uwtools/ecflow.py
AnnaSmoot-NOAA, line 1: My understanding is that `"datelist|enumerated|string"` is a literal string match. I think we may want `"datelist" | "enumerated" | "string"`

AnnaSmoot-NOAA, line 1: This may be a down the road issue. Do we want some kind of `raise ValueError` if the correct information isn't provided for `case "datelist" | "enumerated" | "string"` ? Or for any of the cases in the code? Or should that be handled elsewhere?

christinaholtNOAA: We might consider that. We often lean heavily on the schema checking to ensure that nothing gets to this point that doesn't already meet the code requirements. Let's see how strict the schema ends up, and then of course, we can revisit.

src/uwtools/ecflow.py
christinaholtNOAA: Absolutely!

src/uwtools/ecflow.py
AnnaSmoot-NOAA: That makes sense.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+class _ECFlowDef:
+    """
+    Generate an ecFlow definition file from a YAML config.
+    """
+
+    def __init__(self, config: dict | Config | Path | None = None) -> None:
+        self.scripts: dict[Path, str] = {}
+        cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
+        cfgobj = cfgobj.dereference()
+        self._config = cfgobj.data.get(STR.ecflow, cfgobj.data)
```
maddenp-cu: It looks like we're saying that we don't care whether or not the config we're given has a top-level `ecflow` key. I also don't immediately see where we're validating the config -- maybe that's not in place yet; fine if so. I don't yet understand why we'd accept configs with or without the `ecflow` key, and then how that'll impact validation. Maybe we'd only validate the block under the `ecflow` key. I might be getting ahead of myself, but am curious about the use case for this flexibility.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        self.scheduler = self._config.get("scheduler")
```
maddenp-cu: Is there a use case for the `scheduler` attribute being public? Same question about `scripts`, probably. We don't have API support implemented yet, but maybe you already have something in mind that uses these.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        self.d = Defs()
```
maddenp-cu: If this attribute is going to be public, it probably deserves a more meaningful name, eventually at least. I wouldn't object to an obscure name if it's going to be private and only accessed via public methods.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def write_ecf_scripts(self, path: Path | str) -> None:
+        """
+        The ecf scripts for this workflow.
+
+        :param path: Where to write the ecFlow scripts.
+        """
+        for subpath, content in self.scripts.items():
+            outpath = path / subpath
+            outpath.parent.mkdir(parents=True, exist_ok=True)
```
maddenp-cu, line 68: Would this be equivalent to
```
path.mkdir(parents=True, exist_ok=True)
```
if moved up and out of the loop?

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+            outpath.write_text(content)
+
+    def write_suite_definition(self, path: Path | str) -> None:
+        """
+        The suite definition artifact.
+
+        :param path: Where to write the suite definition.
+        """
+        path = Path(path, "suite.def")
+        path.parent.mkdir(parents=True, exist_ok=True)
```
maddenp-cu: I think you can omit the `.parent` call if you move this up one line, before the redefinition of `path`.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+    def __str__(self):
+        return self.d.__str__()
+
+    def write_ecf_scripts(self, path: Path | str) -> None:
+        """
+        The ecf scripts for this workflow.
+
+        :param path: Where to write the ecFlow scripts.
+        """
+        for subpath, content in self.scripts.items():
```
maddenp-cu: Should we log a warning if this call is going to be a no-op because `self.scripts` is empty?

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        refs = refs if refs is not None else {}
+        expand = config["expand"]
+
+        # Check to make sure all lists are the same length.
+        try:
+            for _i in zip(*expand.values(), strict=True):
+                pass
+        except ValueError:
+            log.error("All expand variables under %s must be the same length" % (parent.name()))
+            raise
```
maddenp-cu: `uwtools.exceptions` has a set of `UWError`-derived exceptions, which are caught by the CLI and handled IIRC gracefully. We should probably raise one of those, or define a new one there if none of the existing ones makes sense.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        :param refs: Variable/value pairs used in higher-level expand blocks.
+        """
+
+        refs = refs if refs is not None else {}
+        expand = config["expand"]
+
+        # Check to make sure all lists are the same length.
+        try:
+            for _i in zip(*expand.values(), strict=True):
+                pass
```
maddenp-cu: Interesting construct. Might
```
assert list(zip(*expand_values(), strict=True))
```
do as well? It seems to do what I'd hope in the REPL, and maybe avoids the illusion that anything is _happening_ here.

Or, maybe even more explicitly expressing the idea of list-length testing, one of
```
assert len(set(map(len, expand_values()))) == 1
assert len({len(x) for x in expand_values()}) == 1
```
i.e. assert that the set of lengths of all the lists has only one value, i.e. all lists have the same length?

And then, in either case, catch `AssertionError` and re-raise some kind of `UWError` in its place.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        except ValueError:
+            log.error("All expand variables under %s must be the same length" % (parent.name()))
+            raise
+
+        # Build up the new blocks in the suite definition
```
```suggestion
        # Build up the new blocks in the suite definition.
```

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+            new_ref = deepcopy(refs)
+            new_ref.update({k: v[i] for k, v in expand.items()})
+            new_block = YAMLConfig({name: config}).dereference(context={"ec": new_ref})
+            new_name = list(new_block.keys())[0]
+            args = {
+                "config": new_block[new_name],
+                "node": nodetype(new_name),
+                "parent": parent,
+                "refs": new_ref,
+            }
```
```suggestion
            new_refs = deepcopy(refs)
            new_refs.update({k: v[i] for k, v in expand.items()})
            new_block = YAMLConfig({name: config}).dereference(context={"ec": new_refs})
            new_name = list(new_block.keys())[0]
            args = {
                "config": new_block[new_name],
                "node": nodetype(new_name),
                "parent": parent,
                "refs": new_refs,
            }
```
I feel like `new_refs` should be plural because `refs` is.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        :param config: Configuration data for these components.
+        :param node: The node to add to the parent.
+        :param parent: The parent object to add this node to.
+        :param refs: Optional references from expanded nodes from higher in the tree.
+        """
+        parent.add(node)
+        add_items = lambda n, cfg: (n(*args) for args in cfg)
+        for key, subconfig in config.items():
+            tag, name = self._tag_name(key)
+            match tag:
```
maddenp-cu, line 161: 1. There's no default handler on this `match` block -- I think we're counting on schema validation to eventually ensure that we don't get an unexpected `tag` value, yes? Fine by me if so.
2. Is there a rationale for the order of the `case` clauses?

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+
+          OR
+
+          step:
+
+          OR
+
+          list:
+        """
+        match name:
```
maddenp-cu, line 207: Same question as above -- is there a rationale for the `case`-clause order?

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        Write the ecf script for the task to disk.
+
+        :param config: The configuration for the script.
+        :param task: The task node.
+        """
+        scheduler = (
+            self._scheduler(
+                account=config.get("account", ""),
+                execution=config.get("execution", ""),
+                rundir=config.get("rundir", ""),
```
maddenp-cu, line 232: Are these items expected to be optional ultimately?

Same question applies elsewhere wherever `.get()` is used in place of `[]`.

Maybe so! Just checking.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+            envcmds=execution.get("envcmds", []),
+            pre_includes=config.get("pre_includes", []),
+            post_includes=config.get("post_includes", []),
+            scheduler=scheduler,
+        )
+
+        path = Path(
+            Path(task.get_abs_node_path().lstrip("/")).parent,
+            f"{task.name().split('_', 1)[-1]}.ecf",
+        )
```
```suggestion
        path = Path(task.get_abs_node_path().lstrip("/")).parent /  f"{task.name().split('_', 1)[-1]}.ecf"
```
But maybe that line is too long -- maybe that's why you have what you have.

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+          step: (optional)
+
+          OR
+
+          step:
+
+          OR
+
+          list:
+        """
```
maddenp-cu: This is just documentation, right? Are there pros or cons for it being part of the docstring instead of a separate string?

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+        fi
+
+        {post_includes}
+
+        %manual
+        {manual}
+        %end
+        """
+        pre_includes = [] if pre_includes is None else pre_includes
+        post_includes = [] if post_includes is None else post_includes
```
```suggestion
        pre_includes = pre_includes or []
        post_includes = post_includes or []
```
I think...

src/uwtools/ecflow.py
```diff
@@ -0,0 +1,354 @@
...
+            ECF_NAME="ECF_NAME",
+        )
+        return re.sub(r"\n\n\n+", "\n\n", rs.strip())
+
+    def _scheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
+        """
+        Use the execution block to build a JobScheduler object.
+
+        :param account: The user account for the batch system.
+        :param execution: The standard UWYAML execution block.
```
```suggestion
        :param execution: The standard UW YAML execution block.
```
Just how it's referred to elsewhere in docs, etc., I think.

maddenp-cu (approved):
A few questions and ideas for your consideration...

src/uwtools/ecflow.py
christinaholtNOAA: This is a great point. I think at some point in my manual testing, I was just passing in the `ecflow:` block but not the top-level key. I should definitely have this always require that `ecflow:` block.

We do not yet have validation in place. It's coming in my next ticket. I've started working on the JSON Schema, but having gotten to the hooking it in part yet.

src/uwtools/ecflow.py
christinaholtNOAA: No, I think they can both be private.

src/uwtools/ecflow.py
christinaholtNOAA: I _think_ this one can also be private.

src/uwtools/ecflow.py
christinaholtNOAA: I like that idea.

src/uwtools/ecflow.py
christinaholtNOAA: I don't think so. `subpath` can be quite deep depending on the number of families in your workflow.

src/uwtools/ecflow.py
christinaholtNOAA: I tested `assert zip(*expand.values(), strict=True)` in the REPL and didn't get what I wanted out of it, hence the iterator. I like the list conversion, it's much clearer. 

I do think that the `list(zip(` approach is a bit more understandable than the set options.

src/uwtools/ecflow.py
christinaholtNOAA: 1. That's right. If I don't end up with a strict enough schema for whatever reason, I will certainly add a default here.
2. It's probably not a strong enough reason, but I felt that the tags that continue to grow the tree would be nice to highlight up front, while the others (below `tasks`) are attributes added to the current node. Perhaps it is strong enough and could just use a comment since it's defying convention.

src/uwtools/ecflow.py
christinaholtNOAA: I think I will likely take it out now that the schema is mostly capturing it.

src/uwtools/ecflow.py
christinaholtNOAA: No good rationale here. I forgot to alphabetize after following along with the ecFlow documentation ordering. I'll reorder it.

src/uwtools/ecflow.py
christinaholtNOAA: Yes, I think they should all be optional to keep with the ultimate flexibility with ecFlow working without a scheduler. I don't think any/many of our workflows would leave them out, but I am trying to mirror the tool capability more closely.

src/uwtools/ecflow.py
christinaholtNOAA: Yeah, it was a pretty line. I'll double-check, though.

src/uwtools/ecflow.py
christinaholtNOAA: Seems like it should work.

PR #862, opened 2026-04-10
Add pytest coverage for ecflow.py

AnnaSmoot-NOAA:
**Synopsis**

**Summary**
This PR adds a complete test suite for `src/uwtools/ecflow.py` targeting 100% code coverage.

**Changes**
Tests Added (src/uwtools/tests/test_ecflow.py)
Fixtures: Created reusable fixtures (`instance`, `instance_with_scheduler`, `minimal_config`) for test setup
`_tag_name` tests: Parametrized tests for key parsing into tag/name tuples
`_ecflowscript` tests: Coverage for script generation with various options (envcmds, envvars, includes, scheduler)
`_jobscheduler` tests: Mocked scheduler configuration with/without threads
`_add_repeat` tests: All repeat types (date, int, day, enumerated, datelist, string, datetime)
`__init__` tests: Initialization from dict, Path, Config object, with/without scheduler
`_add_workflow_components` tests: extern, vars, suite, and suites expansion
`_add_node` tests: All node attribute cases (family, task, vars, trigger, events, defstatus, inlimits, labels, late, limits, meters, repeat, script, expand)
`_expand_block` tests: Basic expansion and mismatched length validation
`_create_ecf_script` tests: Script creation with/without scheduler
`write_ecf_scripts` tests: Writing scripts, handling no scripts case, string paths
`write_suite_definition` tests: Directory creation, string paths
Integration tests: Full workflow creation and expand block parameterization
Source Changes (`src/uwtools/ecflow.py`)
Added handling for expand tag (silently skipped as already processed by _expand_block)
Added catch-all case with `AssertionError` for unrecognized tags
Minor comment punctuation fixes

**Testing**
All tests pass and provide full branch coverage for ecflow.py.

**Type**

- [ ] Documentation
- [X] Tooling (CI, code-quality, packaging, revision-control, etc.)

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -195,8 +195,10 @@ class _ECFlowDef:
                     node.add_trigger(subconfig)
                 case "vars":  # add_variable accepts a dict.
                     node.add_variable(subconfig)
+                case "expand":  # Already processed by _expand_block.
+                    pass
                 case _:
-                    pass  # Ignore unrecognized tags.
+                    assert False, f"Unrecognized tag: {tag}"
 
     def _add_repeat(self, config: dict, name: str, node: Node) -> None:
         """
@@ -314,9 +316,7 @@ class _ECFlowDef:
         )
         return re.sub(r"\n\n\n+", "\n\n", rs.strip())
 
-    def _jobscheduler(
-        self, account: str, execution: dict, rundir: Path | str
-    ) -> JobScheduler:
+    def _jobscheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
         """
         Use the execution block to build a JobScheduler object.
 
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -40,7 +40,7 @@ class _ECFlowDef:
     Generate an ecFlow definition file from a YAML config.
     """
 
-    def __init__(self, config: dict | Config | Path | None = None) -> None:  # pragma: no cover
+    def __init__(self, config: dict | Config | Path | None = None) -> None:
         self._scripts: dict[Path, str] = {}
         cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
         cfgobj = cfgobj.dereference()
@@ -49,10 +49,10 @@ class _ECFlowDef:
         self._d = Defs()
         self._add_workflow_components()
 
-    def __str__(self):  # pragma: no cover
+    def __str__(self):
         return self._d.__str__()
 
-    def write_ecf_scripts(self, path: Path | str) -> None:  # pragma: no cover
+    def write_ecf_scripts(self, path: Path | str) -> None:
         """
         The ecf scripts for this workflow.
 
@@ -68,7 +68,7 @@ class _ECFlowDef:
             outpath.parent.mkdir(parents=True, exist_ok=True)
             outpath.write_text(content)
 
-    def write_suite_definition(self, path: Path | str) -> None:  # pragma: no cover
+    def write_suite_definition(self, path: Path | str) -> None:
         """
         The suite definition artifact.
 
@@ -79,7 +79,7 @@ class _ECFlowDef:
         suite = path / "suite.def"
         suite.write_text(self._d.__str__())
 
-    def _add_workflow_components(self) -> None:  # pragma: no cover
+    def _add_workflow_components(self) -> None:
         """
         Add suite(s) and other attributes to the suite definition.
         """
@@ -96,7 +96,7 @@ class _ECFlowDef:
                 case "suites":
                     self._expand_block(subconfig, name, Suite, self._d)
 
-    def _expand_block(  # pragma: no cover
+    def _expand_block(
         self,
         config: dict,
         name: str,
@@ -139,7 +139,7 @@ class _ECFlowDef:
             }
             self._add_node(**args)
 
-    def _add_node(  # noqa: C901,PLR0912  # pragma: no cover
+    def _add_node(  # noqa: C901,PLR0912
         self,
         config: dict,
         node: Node,
@@ -187,16 +187,18 @@ class _ECFlowDef:
                     add_items(node.add_limit, subconfig)
                 case "meters":
                     add_items(node.add_meter, subconfig)
-                case "repeat":  # Only one repeat is allowed per node
+                case "repeat":  # Only one repeat is allowed per node.
                     self._add_repeat(subconfig, name, node)
-                case "trigger":  # Only one trigger is allowed per node
-                    node.add_trigger(subconfig)
-                case "vars":  # add_variable accepts a dict
-                    node.add_variable(subconfig)
                 case "script":
                     self._create_ecf_script(subconfig, node)
+                case "trigger":  # Only one trigger is allowed per node.
+                    node.add_trigger(subconfig)
+                case "vars":  # add_variable accepts a dict.
+                    node.add_variable(subconfig)
+                case _:
+                    pass  # Ignore unrecognized tags.
 
-    def _add_repeat(self, config: dict, name: str, node: Node) -> None:  # pragma: no cover
+    def _add_repeat(self, config: dict, name: str, node: Node) -> None:
         """
         Adds a repeat to a node.
 
@@ -218,7 +220,7 @@ class _ECFlowDef:
                 repeat = RepeatInteger
         node.add_repeat(repeat(**config))
 
-    def _create_ecf_script(self, config: dict, task: Task) -> None:  # pragma: no cover
+    def _create_ecf_script(self, config: dict, task: Task) -> None:
         """
         Write the ecf script for the task to disk.
 
@@ -251,7 +253,7 @@ class _ECFlowDef:
         )
         self._scripts[path] = es
 
-    def _ecflowscript(  # pragma: no cover
+    def _ecflowscript(
         self,
         execution: list[str],
         manual: str,
@@ -314,7 +316,7 @@ class _ECFlowDef:
 
     def _jobscheduler(
         self, account: str, execution: dict, rundir: Path | str
-    ) -> JobScheduler:  # pragma: no cover
+    ) -> JobScheduler:
         """
         Use the execution block to build a JobScheduler object.
 
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -198,7 +198,7 @@ class _ECFlowDef:
                 case "expand":  # Already processed by _expand_block.
                     pass
                 case _:
-                    assert False, f"Unrecognized tag: {tag}"
+                    raise AssertionError(f"Unrecognized tag: {tag}")
 
     def _add_repeat(self, config: dict, name: str, node: Node) -> None:
         """
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -40,7 +40,7 @@ class _ECFlowDef:
     Generate an ecFlow definition file from a YAML config.
     """
 
-    def __init__(self, config: dict | Config | Path | None = None) -> None:
+    def __init__(self, config: dict | Config | Path | None = None) -> None:  # pragma: no cover
         self._scripts: dict[Path, str] = {}
         cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
         cfgobj = cfgobj.dereference()
@@ -49,10 +49,10 @@ class _ECFlowDef:
         self._d = Defs()
         self._add_workflow_components()
 
-    def __str__(self):
+    def __str__(self):  # pragma: no cover
         return self._d.__str__()
 
-    def write_ecf_scripts(self, path: Path | str) -> None:
+    def write_ecf_scripts(self, path: Path | str) -> None:  # pragma: no cover
         """
         The ecf scripts for this workflow.
 
@@ -68,7 +68,7 @@ class _ECFlowDef:
             outpath.parent.mkdir(parents=True, exist_ok=True)
             outpath.write_text(content)
 
-    def write_suite_definition(self, path: Path | str) -> None:
+    def write_suite_definition(self, path: Path | str) -> None:  # pragma: no cover
         """
         The suite definition artifact.
 
@@ -79,7 +79,7 @@ class _ECFlowDef:
         suite = path / "suite.def"
         suite.write_text(self._d.__str__())
 
-    def _add_workflow_components(self) -> None:
+    def _add_workflow_components(self) -> None:  # pragma: no cover
         """
         Add suite(s) and other attributes to the suite definition.
         """
@@ -96,7 +96,7 @@ class _ECFlowDef:
                 case "suites":
                     self._expand_block(subconfig, name, Suite, self._d)
 
-    def _expand_block(
+    def _expand_block(  # pragma: no cover
         self,
         config: dict,
         name: str,
@@ -139,7 +139,7 @@ class _ECFlowDef:
             }
             self._add_node(**args)
 
-    def _add_node(  # noqa: C901,PLR0912
+    def _add_node(  # noqa: C901,PLR0912  # pragma: no cover
         self,
         config: dict,
         node: Node,
@@ -196,7 +196,7 @@ class _ECFlowDef:
                 case "script":
                     self._create_ecf_script(subconfig, node)
 
-    def _add_repeat(self, config: dict, name: str, node: Node) -> None:
+    def _add_repeat(self, config: dict, name: str, node: Node) -> None:  # pragma: no cover
         """
         Adds a repeat to a node.
 
@@ -218,7 +218,7 @@ class _ECFlowDef:
                 repeat = RepeatInteger
         node.add_repeat(repeat(**config))
 
-    def _create_ecf_script(self, config: dict, task: Task) -> None:
+    def _create_ecf_script(self, config: dict, task: Task) -> None:  # pragma: no cover
         """
         Write the ecf script for the task to disk.
 
@@ -251,7 +251,7 @@ class _ECFlowDef:
         )
         self._scripts[path] = es
 
-    def _ecflowscript(
+    def _ecflowscript(  # pragma: no cover
         self,
         execution: list[str],
         manual: str,
@@ -312,7 +312,9 @@ class _ECFlowDef:
         )
         return re.sub(r"\n\n\n+", "\n\n", rs.strip())
 
-    def _jobscheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
+    def _jobscheduler(
+        self, account: str, execution: dict, rundir: Path | str
+    ) -> JobScheduler:  # pragma: no cover
         """
         Use the execution block to build a JobScheduler object.
 
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING
 
 from ecflow import (  # type: ignore[import-untyped]
     Defs,
+    DState,
     Family,
     Late,
     Node,
@@ -32,6 +33,8 @@ from uwtools.scheduler import JobScheduler
 from uwtools.strings import STR
 
 if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable, Sequence
+
     from ecflow import NodeContainer
 
 
@@ -155,7 +158,13 @@ class _ECFlowDef:
         :param refs: Optional references from expanded nodes from higher in the tree.
         """
         parent.add(node)
-        add_items = lambda n, cfg: (n(*args) for args in cfg)
+
+        def add_items(
+            method: Callable[..., object], cfg: Iterable[Sequence[object]]
+        ) -> None:
+            for args in cfg:
+                method(*args)
+
         for key, subconfig in config.items():
             tag, name = self._tag_name(key)
             match tag:
@@ -173,7 +182,7 @@ class _ECFlowDef:
                 # Node attribute cases
 
                 case "defstatus":
-                    node.add_defstatus(subconfig)
+                    node.add_defstatus(getattr(DState, subconfig))
                 case "events":
                     for event in subconfig:
                         node.add_event(event)
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -159,9 +159,7 @@ class _ECFlowDef:
         """
         parent.add(node)
 
-        def add_items(
-            method: Callable[..., object], cfg: Iterable[Sequence[object]]
-        ) -> None:
+        def add_items(method: Callable[..., object], cfg: Iterable[Sequence[object]]) -> None:
             for args in cfg:
                 method(*args)
 
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -198,7 +198,8 @@ class _ECFlowDef:
                 case "expand":  # Already processed by _expand_block.
                     pass
                 case _:
-                    raise AssertionError(f"Unrecognized tag: {tag}")
+                    msg = f"Unrecognized tag: {tag}"
+                    raise AssertionError(msg)
 
     def _add_repeat(self, config: dict, name: str, node: Node) -> None:
         """
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -5,7 +5,7 @@ Tests for uwtools.ecflow module.
 from pathlib import Path
 from unittest.mock import Mock, patch
 
-from ecflow import Defs, Family, Suite, Task
+from ecflow import Defs, Suite, Task
 from pytest import fixture, mark, raises
 
 from uwtools import ecflow
@@ -369,8 +369,8 @@ class TestECFlowDef:
 
     # write_ecf_scripts tests
 
-    def test_write_ecf_scripts__no_scripts(self, instance, logged):
-        instance.write_ecf_scripts(Path("/tmp"))
+    def test_write_ecf_scripts__no_scripts(self, instance, logged, tmp_path):
+        instance.write_ecf_scripts(tmp_path)
         assert logged("No scripts are configured for this workflow")
 
     def test_write_ecf_scripts__with_scripts(self, instance, tmp_path):
@@ -452,9 +452,8 @@ class TestECFlowDef:
         suite = Suite("test")
         task = Task("t1")
         config = {"late": {"submitted": "+00:15", "active": "+01:00"}}
-        with patch.object(ecflow, "Late") as mock_late:
-            with patch.object(task, "add_late") as mock_add_late:
-                instance._add_node(config, task, suite)
+        with patch.object(ecflow, "Late") as mock_late, patch.object(task, "add_late"):
+            instance._add_node(config, task, suite)
         mock_late.assert_called_once()
 
     def test__add_node__with_limits(self, instance):
@@ -587,7 +586,7 @@ class TestECFlowDef:
         ecf.write_ecf_scripts(tmp_path)
         assert len(list(tmp_path.rglob("*.ecf"))) == 1
 
-    def test_integration__with_expand(self, tmp_path):
+    def test_integration__with_expand(self):
         """
         Test workflow with expand blocks for parameterized tasks.
         """
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -5,7 +5,7 @@ Tests for uwtools.ecflow module.
 from pathlib import Path
 from unittest.mock import Mock, patch
 
-from ecflow import Defs, Suite, Task
+from ecflow import Defs, Suite, Task  # type: ignore[import-untyped]
 from pytest import fixture, mark, raises
 
 from uwtools import ecflow
@@ -146,7 +146,7 @@ class TestECFlowDef:
         assert call_args["queue"] == "batch"
 
     def test__jobscheduler__no_threads(self, instance_with_scheduler):
-        execution = {}
+        execution: dict = {}
         with patch.object(ecflow.JobScheduler, "get_scheduler") as mock_get:
             instance_with_scheduler._jobscheduler(
                 account="myaccount",
@@ -235,7 +235,7 @@ class TestECFlowDef:
         assert isinstance(ecf._d, Defs)
 
     def test__init__missing_ecflow_key(self):
-        config = {"not_ecflow": {}}
+        config: dict = {"not_ecflow": {}}
         with raises(KeyError):
             _ECFlowDef(config=config)
 
@@ -264,13 +264,13 @@ class TestECFlowDef:
 
     def test__add_node__basic(self, instance):
         suite = Suite("test")
-        config = {}
+        config: dict = {}
         instance._add_node(config, suite, instance._d)
         assert "test" in str(instance._d)
 
     def test__add_node__with_task(self, instance):
         suite = Suite("test")
-        config = {"task_hello": {}}
+        config: dict = {"task_hello": {}}
         with patch.object(instance, "_add_node", wraps=instance._add_node) as mock:
             instance._add_node(config, suite, instance._d)
         # Called twice: once for suite, once for task
@@ -278,7 +278,7 @@ class TestECFlowDef:
 
     def test__add_node__with_family(self, instance):
         suite = Suite("test")
-        config = {"family_myfam": {"task_t1": {}}}
+        config: dict = {"family_myfam": {"task_t1": {}}}
         with patch.object(instance, "_add_node", wraps=instance._add_node) as mock:
             instance._add_node(config, suite, instance._d)
         # Called for suite, family, and task
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -13,13 +13,14 @@ from uwtools.config.formats.yaml import YAMLConfig
 from uwtools.ecflow import _ECFlowDef
 from uwtools.exceptions import UWConfigError
 
-
 # Fixtures
 
 
 @fixture
 def instance():
-    """Create an _ECFlowDef instance without calling __init__."""
+    """
+    Create an _ECFlowDef instance without calling __init__.
+    """
     obj = object.__new__(_ECFlowDef)
     obj._scripts = {}
     obj._scheduler = None
@@ -30,7 +31,9 @@ def instance():
 
 @fixture
 def instance_with_scheduler():
-    """Create an _ECFlowDef instance with a scheduler configured."""
+    """
+    Create an _ECFlowDef instance with a scheduler configured.
+    """
     obj = object.__new__(_ECFlowDef)
     obj._scripts = {}
     obj._scheduler = "slurm"
@@ -41,27 +44,19 @@ def instance_with_scheduler():
 
 @fixture
 def minimal_config():
-    """Minimal config for instantiation."""
+    """
+    Minimal config for instantiation.
+    """
     return {"ecflow": {}}
-
-
-@fixture
-def suite_config():
-    """Config with a simple suite."""
-    return {
-        "ecflow": {
-            "suite_test": {
-                "task_hello": {}
-            }
-        }
-    }
 
 
 # Tests
 
 
 class TestECFlowDef:
-    """Tests for class uwtools.ecflow._ECFlowDef."""
+    """
+    Tests for class uwtools.ecflow._ECFlowDef.
+    """
 
     # _tag_name tests
 
@@ -197,7 +192,9 @@ class TestECFlowDef:
 
     @mark.parametrize("repeat_type", ["datelist", "string"])
     def test__add_repeat__enumerated_variants(self, instance, repeat_type):
-        """Test that datelist and string repeat types also use RepeatEnumerated."""
+        """
+        Test that datelist and string repeat types also use RepeatEnumerated.
+        """
         node = Mock()
         config = {"name": "VAR", "values": ["a", "b"]}
         with patch.object(ecflow, "RepeatEnumerated") as mock_repeat:
@@ -320,10 +317,7 @@ class TestECFlowDef:
     # _expand_block tests
 
     def test__expand_block__basic(self, instance):
-        config = {
-            "expand": {"MEMBER": ["m01", "m02"]},
-            "task_run": {}
-        }
+        config = {"expand": {"MEMBER": ["m01", "m02"]}, "task_run": {}}
         suite = Suite("test")
         instance._d.add(suite)
         with patch.object(instance, "_add_node") as mock_add_node:
@@ -380,18 +374,14 @@ class TestECFlowDef:
         assert logged("No scripts are configured for this workflow")
 
     def test_write_ecf_scripts__with_scripts(self, instance, tmp_path):
-        instance._scripts = {
-            Path("test/hello.ecf"): "#!/bin/bash\necho hello"
-        }
+        instance._scripts = {Path("test/hello.ecf"): "#!/bin/bash\necho hello"}
         instance.write_ecf_scripts(tmp_path)
         outfile = tmp_path / "test" / "hello.ecf"
         assert outfile.exists()
         assert "echo hello" in outfile.read_text()
 
     def test_write_ecf_scripts__with_string_path(self, instance, tmp_path):
-        instance._scripts = {
-            Path("suite/task.ecf"): "#!/bin/bash\necho test"
-        }
+        instance._scripts = {Path("suite/task.ecf"): "#!/bin/bash\necho test"}
         instance.write_ecf_scripts(str(tmp_path))
         outfile = tmp_path / "suite" / "task.ecf"
         assert outfile.exists()
@@ -492,7 +482,9 @@ class TestECFlowDef:
         mock_repeat.assert_called_once()
 
     def test__add_node__with_script_then_continue(self, instance):
-        """Test that loop continues after processing script case (covers branch 196->159)."""
+        """
+        Test that loop continues after processing script case (covers branch 196->159).
+        """
         suite = Suite("test")
         task = Task("t1")
         # Place script FIRST so loop must continue to process trigger afterward.
@@ -504,7 +496,9 @@ class TestECFlowDef:
         assert task.get_trigger() is not None
 
     def test__add_node__with_script_last(self, instance):
-        """Test that loop exits after processing script as the last item."""
+        """
+        Test that loop exits after processing script as the last item.
+        """
         suite = Suite("test")
         task = Task("t1")
         # Place script LAST so loop exits after processing it.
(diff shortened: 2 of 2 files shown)


src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+def instance_with_scheduler():
+    """
+    Create an _ECFlowDef instance with a scheduler configured.
+    """
+    obj = object.__new__(_ECFlowDef)
+    obj._scripts = {}
+    obj._scheduler = "slurm"
+    obj._d = Defs()
+    obj._config = {}
+    return obj
```
maddenp-cu:
```suggestion
@fixture
def instance_with_scheduler(instance):
    """
    Create an _ECFlowDef instance with a scheduler configured.
    """
    instance._scheduler = "slurm"
    return instance
```
If this works the same, it's a little more DRY.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+            _ECFlowDef(config=config)
+
+    # _add_workflow_components tests
+
+    def test__add_workflow_components__extern(self, instance):
+        instance._config = {"extern": ["/path/to/ext1", "/path/to/ext2"]}
+        instance._add_workflow_components()
+        # Externs are added to the Defs object
+        def_str = str(instance._d)
+        assert "extern" in def_str or instance._d is not None
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Fixed.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+        }
+        with patch.object(ecflow, "RepeatDateTime") as mock_repeat:
+            instance._add_repeat(config.copy(), "datetime", node)
+        mock_repeat.assert_called_once()
+        node.add_repeat.assert_called_once()
+
+    def test__add_repeat__unknown_type(self, instance):
+        node = Mock()
+        config = {"name": "X"}
+        with raises(UnboundLocalError):
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: I'm not looking to change ecflow.py

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+        assert "VAR1" in str(instance._d)
+
+    def test__add_node__with_trigger(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        suite.add(task)
+        config = {"trigger": "t1 == complete"}
+        task2 = Task("t2")
+        instance._add_node(config, task2, suite)
+        assert "trigger" in str(instance._d).lower() or task2 is not None
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+        assert len(instance._scripts) == 1
+
+    def test__add_node__unrecognized_tag(self, instance):
+        """
+        Test that unrecognized tags raise AssertionError.
+        """
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"unknown_tag": "value"}
+        with raises(AssertionError, match="Unrecognized tag: unknown"):
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: I'm not looking to change ecflow.py so skipping.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+    def test__ecflowscript__minimal(self, instance):
+        result = instance._ecflowscript(
+            execution=["echo hello"],
+            manual="Test script",
+        )
+        assert "echo hello" in result
+        assert "Test script" in result
+        assert "%manual" in result
+        assert "%end" in result
+        assert "model=%MODEL%" in result
```
maddenp-cu: It might strengthen these tests to assert that the strings appear in the result as _complete lines_ -- it looks like that's the intention. For example, we could add something like
```
def assert_line_in(result: str, line: str) -> None:
    assert line in result.split("\n")
```
and call it like
```
assert_line_int(result, "echo hello")
```
instead of the existing `assert` statements.

This might break down if we also need to account for whitespace, especially leading whitespace, but then either we could pass in the expected leading whitespace as part of `line`, or have a regex-based check that disregarded leading whitespace. Please let me know if that's needed and I could suggest something, but I think it would look something like
```
assert any([re.match(fr"^ *{line}$", x) for x in result.split("\n")])
```
i.e. `any(...)` will be `True` if a regex that is "the start of the line, then any number of leading spaces ,then the line, then the end of the line" matches any line `x` in `result`

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+    def test__ecflowscript__with_includes(self, instance):
+        result = instance._ecflowscript(
+            execution=["echo hello"],
+            manual="Test script",
+            pre_includes=["head.h", "setup.h"],
+            post_includes=["tail.h"],
+        )
+        assert "%include <head.h>" in result
+        assert "%include <setup.h>" in result
+        assert "%include <tail.h>" in result
```
maddenp-cu: In addition to the risk of accidental partial-line matches mentioned earlier, these assertions would pass if these lines appeared in _any order_ in the result. We could strengthen this with something like
```
expected = ["%include <head.h>", "%include <setup.h>", "%include <tail.h>"]
actual = [line for line in result.split("\n") if any([re.match(fr"^ *{x}$", line) for x in expected])]
assert actual == expected
```
I think... Check my logic, but I think that `actual` will contain the lines from `expected` if they are found in `result`, but also in the _order_ they are found in `result` -- and that order had better match. Is that right? If you tried this out, you should be able to change the order of something in the code-under-test to make the test break, then put it back and see that the test now passes.

If this works, it could be factored out as a helper function and used in these tests wherever the order of lines we're looking for in `result` matters.

maddenp-cu (commented):
I started to review but see that you've added some comments, so I'll send these first few thoughts, read yours, then circle back.

src/uwtools/tests/test_ecflow.py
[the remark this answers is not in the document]
maddenp-cu: It took me a minute but I agree with you here: We'd discussed the fact that we want to _assert_ something here rather than raise a user-facing exception, because we believe that a failure to match a key here is a failure in our schema checking, and we want users to report it to us so we can fix it. I was momentarily confused by the `raise` in `ecflow.py` instead of an `assert`, but I see that raising an `AssertionError` is the right thing to do: We _could_ have an `assert` statement before match block that said something like
```
assert key in ["family", "families", "task", ...
```
but that would have the ugly side effect of forcing us to repeat all those key names, which already appear in the `case` statements. I'm just thinking it through in public -- does it sound right?

AnnaSmoot-NOAA:
@maddenp-cu 
I'll start looking at your comments. I just finished looking at copilots comments and decided to incorporate a couple. 

Otherwise, I'm just waiting for these CI test to pass: 
CI failed due to external link checker finding 503 errors on two GitHub Pages sites (christopherwharrop.github.io/rocoto/ and mpas-dev.github.io/atmosphere/atmosphere_download). These are temporary outages unrelated to my changes. Will re-run CI when the sites are back up.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+                execution=execution,
+                rundir="/path/to/run",
+            )
+        mock_get.assert_called_once()
+        call_args = mock_get.call_args[0][0]
+        assert call_args["account"] == "myaccount"
+        assert call_args["rundir"] == "/path/to/run"
+        assert call_args["scheduler"] == "slurm"
+        assert call_args["threads"] == 4
+        assert call_args["queue"] == "batch"
```
maddenp-cu:
```suggestion
        mock_get.assert_called_once_with(
            account="myaccount",
            rundir=/path/to/run",
            scheduler="slurm",
            threads=4,
            queue="batch",
        )
```
If I'm not mistaken. And, if not, this simpler pattern can probably be used elsewhere, too. There's also `<mock>.assert_called_with()`, which makes the same assertion about the _latest_ call to a mock, if it was called more than once by the code-under-test. The only time I find myself using `call_args` is when I need to make assertions about more than one call to a mock. (And now that I'm saying it, I wonder if there's not even a better way to to that.)

You can also have positional arguments (not just keyword arguments) , those assertions.

If you have an argument whose value you don't know or just don't care about, you can import `ANY` from `unittest.mock` and supply that as the value and the assertion will hold no matter what value was passed as that particular argument. This might be useful for the `stdout` argument in this case, which is not mentioned in the list of individual `assert` calls commented on here, but would need to be present (I think...) in the `.assert_called_once_with()` alternative.

I guess I can see pros and cons, but the `.assert_called_once_with()` method exists for a reason and it seems a bit less noisy to me. It also seems more rigorous to me, in that you have to account for _all_ the arguments: If the signature of `JobScheduler.get_scheduler()` changed to, for example, add new arguments, and the code-under-test was updated to match, this test, with the individual `assert` statements, might continue to pass, and we wouldn't necessarily know that we needed to update it to keep it robust.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+    def test__jobscheduler__no_threads(self, instance_with_scheduler):
+        execution: dict = {}
+        with patch.object(ecflow.JobScheduler, "get_scheduler") as mock_get:
+            instance_with_scheduler._jobscheduler(
+                account="myaccount",
+                execution=execution,
+                rundir="/path/to/run",
+            )
+        call_args = mock_get.call_args[0][0]
+        assert "threads" not in call_args
```
maddenp-cu: This seems like a good use of `call_args` -- a negative assertion. Otherwise, I guess we'd have to do a full `.assert_called_once_with()` with every argument _but_ the one we want to assert doesn't appear.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+
+    # _add_repeat tests
+
+    def test__add_repeat__date(self, instance):
+        node = Mock()
+        config = {"name": "YMD", "start": 20240101, "end": 20240131, "step": 1}
+        with patch.object(ecflow, "RepeatDate") as mock_repeat:
+            instance._add_repeat(config.copy(), "date", node)
+        mock_repeat.assert_called_once()
+        node.add_repeat.assert_called_once()
```
maddenp-cu: I think this test doesn't capture some of the nuance of what happens in the code-under-test, namely 1. that `config` is modified (whether or not that's the best idea), and 2. that `RepeatDate` gets called in a certain way. Consider this alternative:
```
    def test__add_repeat__date(self, instance):
        node = Mock()
        common = {"name": "YMD", "start": 20240101, "end": 20240131}
        initial = {**common, "step": 1}
        final = {**common, "delta": 1}
        with patch.object(ecflow, "RepeatDate") as mock_repeat:
            instance._add_repeat(initial, "date", node)
        assert initial == final
        mock_repeat.assert_called_once_with(**final)
        node.add_repeat.assert_called_once()
```
I think this captures those additional facts.

This formulation also suggests to me that maybe `_add_repeat()` should _not_ modify config, but construct its own copy. That would simplify this test.

It looks like `"date"` is the only case that results in config modification, so the other tests for this function can probably be simpler, but probably should also make assertions about the config passed in and what is passed on to the ecFlow API object initializers.

src/uwtools/ecflow.py
```diff
@@ -187,16 +187,21 @@ def _add_node(  # noqa: C901,PLR0912  # pragma: no cover
...
-                case "repeat":  # Only one repeat is allowed per node
+                case "repeat":  # Only one repeat is allowed per node.
                     self._add_repeat(subconfig, name, node)
-                case "trigger":  # Only one trigger is allowed per node
-                    node.add_trigger(subconfig)
-                case "vars":  # add_variable accepts a dict
-                    node.add_variable(subconfig)
                 case "script":
                     self._create_ecf_script(subconfig, node)
+                case "trigger":  # Only one trigger is allowed per node.
```
maddenp-cu, line 201: I feel like it might have been discussed elsewhere, maybe in @christinaholtNOAA's previous PR, but we probably should soon add a class to `uwtools.strings` to capture all these ecFlow magic strings so that they can be verified by `pylint` et al. Doing that could also be a good time to see if we could borrow from [`wxvx`](https://github.com/NOAA-GSL/wxvx/blob/main/src/wxvx/strings.py) (which initially borrowed from `uwtools`), whose `strings` module avoids the need to repeat the same word as both a key and a value. It allows for having the value match they key by default, but also for overriding with a special value.

maddenp-cu (commented):
I'm about halfway through but need to work on something else for a while this afternoon. I think some of the suggestions here, if adopted, might touch some of the tests I haven't read yet, so I'll send these now and check back later and/or in the morning.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        call_args = mock_get.call_args[0][0]
+        assert call_args["account"] == "myaccount"
+        assert call_args["rundir"] == "/path/to/run"
+        assert call_args["scheduler"] == "slurm"
+        assert call_args["threads"] == 4
+        assert call_args["queue"] == "batch"
+
+    def test__jobscheduler__no_threads(self, instance_with_scheduler):
+        execution: dict = {}
+        with patch.object(ecflow.JobScheduler, "get_scheduler") as mock_get:
```
maddenp-cu: I think the `uwtools` test convention is usually to name mocks directly after the object they're mocking, so e.g. `as get_scheduler` here instead of `mock_get`. Truncating at just `get` isn't too bad here, but in the `_add_repeat` tests below, a variety of different classes -- `RepeatDate`, `RepeatEnumerated`, etc. -- are all called `mock_repeat` in different tests. I think that  might be confusing to someone who's reading a bunch of different tests; it seems like it'd be easy to forget which ecFlow class you're dealing with at the moment. For a similar number of characters, I think I'd glean more from e.g. `repeast_date` than from `mock_repeat`.

Sometimes we do see mocked objects with very long names that are tedious to reproduce, so I think we do also have cases like `with mock.object(obj, "foo_bar_baz_quz_asdf_etc") as fbbqae` where the acronym seems less awful that the full name.

src/uwtools/ecflow.py
christinaholtNOAA: I can add this as a ticket.

src/uwtools/ecflow.py
```diff
@@ -187,16 +187,21 @@ def _add_node(  # noqa: C901,PLR0912  # pragma: no cover
...
+                    node.add_trigger(subconfig)
+                case "vars":  # add_variable accepts a dict.
+                    node.add_variable(subconfig)
+                case "expand":  # Already processed by _expand_block.
+                    pass
+                case _:
+                    msg = f"Unrecognized tag: {tag}"
+                    raise AssertionError(msg)
```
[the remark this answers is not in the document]
christinaholtNOAA: Well, this is a nice suggestion. :)

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+def instance():
+    """
+    Create an _ECFlowDef instance without calling __init__.
+    """
+    obj = object.__new__(_ECFlowDef)
+    obj._scripts = {}
+    obj._scheduler = None
+    obj._d = Defs()
+    obj._config = {}
+    return obj
```
christinaholtNOAA: Would problems arise from just having something like:

```
return _ECFlowDef({"ecflow": {}})
```

This might help with maintainability, especially if we miss adding an instance variable here that we add in the code. 

EDIT: And now I'm seeing below that you already have this config as a fixture. Perhaps:

```
def instance(minimal_config):
    return return _ECFlowDef(minimal_config)
```

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: I'm not sure that having the code make its own copy of `config` so it isn't modified would actually change the test.

The user interface requires arguments `name`, `start`, `end`, `step` so it is consistent with all the other repeat classes, but the ecflow API requires arguments `name`, `start`, `end`, `delta`.  In the test `_add_repeat` would still need `initial`, while the `add_repeat` API call still needs `delta`.

Perhaps I'm missing some nuance here?

I am in no way suggesting that `config` _should_ be modified in this code. I'm happy to make sure it's protected appropriately with a `deepcopy`.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        task2 = Task("t2")
+        instance._add_node(config, task2, suite)
+        assert str(task2.get_trigger()) == "t1 == complete"
+
+    def test__add_node__with_events(self, instance):
+        suite = Suite("test")
+        config = {"events": ["event1", "event2"]}
+        task = Task("t1")
+        instance._add_node(config, task, suite)
+        # Events are added to the task
```
christinaholtNOAA: This one needs a test.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+
+    def test__add_node__with_defstatus(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"defstatus": "complete"}
+        with patch.object(task, "add_defstatus") as mock_defstatus:
+            instance._add_node(config, task, suite)
+        mock_defstatus.assert_called_once_with("complete")
```
christinaholtNOAA: For tests like these (defstatus, events), would it be stronger to use the `Node.get_*` or `Node.find_*` methods from ecflow to ensure that the code is actually adding the attribute to the Node?

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+
+    # _expand_block tests
+
+    def test__expand_block__basic(self, instance):
+        config = {"expand": {"MEMBER": ["m01", "m02"]}, "task_run": {}}
+        suite = Suite("test")
+        instance._d.add(suite)
+        with patch.object(instance, "_add_node") as mock_add_node:
+            instance._expand_block(config, "suite", Suite, instance._d)
+        assert mock_add_node.call_count == 2
```
christinaholtNOAA, line 344: It might be nice to ensure that these `_add_node` calls get called with `m01` and `m01` names.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        with patch.object(instance, "_expand_block") as mock_expand:
+            instance._add_node(config, suite, instance._d)
+        mock_expand.assert_called_once()
+
+    def test__add_node__with_inlimits(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"inlimits": [["limit1", "/path"], ["limit2", "/path2"]]}
+        instance._add_node(config, task, suite)
+        # Note: add_items returns unconsumed generator in ecflow.py
```
christinaholtNOAA: This one needs a test.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+
+    def test__add_node__with_labels(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"labels": [["label1", "value1"], ["label2", "value2"]]}
+        instance._add_node(config, task, suite)
+        # Note: add_items returns unconsumed generator in ecflow.py
```
christinaholtNOAA: This one needs a test.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        with patch.object(ecflow, "Late") as mock_late, patch.object(task, "add_late"):
+            instance._add_node(config, task, suite)
+        mock_late.assert_called_once()
+
+    def test__add_node__with_limits(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"limits": [["limit1", 5], ["limit2", 10]]}
+        instance._add_node(config, task, suite)
+        # Note: add_items returns unconsumed generator in ecflow.py
```
christinaholtNOAA: This and the next need tests.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
     )
-    def test__tag_name(self, key, expected):
-        # _tag_name is called on self, so we need a minimal instance.
-        ecf = _ECFlowDef(config={"ecflow": {}})
-        assert ecf._tag_name(key) == expected
+    def test__tag_name(self, instance, key, expected):
+        assert instance._tag_name(key) == expected
+
+    # _ecflowscript tests
+
```
elcarpenterNOAA: We generally do not include comments like these in test files. The test names are already explicit, and we also keep tests alphabetized. Here, the tests appear to be sorted within each function group, but the function groups themselves should be sorted as well.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,606 @@
...
+        task = Task("t1")
+        config = {"repeat_int": {"name": "STEP", "start": 0, "end": 10}}
+        with patch.object(instance, "_add_repeat") as mock_repeat:
+            instance._add_node(config, task, suite)
+        mock_repeat.assert_called_once()
+
+    def test__add_node__with_script_then_continue(self, instance):
+        """
+        Test that loop continues after processing script case (covers branch 196->159).
+        """
```
elcarpenterNOAA: We generally do not include docstrings in our unit tests, and at a minimum, we should avoid referencing specific line numbers, since those become outdated so quickly.

src/uwtools/tests/test_ecflow.py
maddenp-cu: @christinaholtNOAA 

> I'm not sure that having the code make its own copy of `config` so it isn't modified would actually change the test.

Maybe only in that `assert initial == final` would no longer be necessary or interesting, and some variables could then be inlined because they wouldn't be needed more than once:

```
    def test__add_repeat__date(self, instance):
        node = Mock()
        common = {"name": "YMD", "start": 20240101, "end": 20240131}
        with patch.object(ecflow, "RepeatDate") as mock_repeat:
            instance._add_repeat({**common, "step": 1}, "date", node)
        mock_repeat.assert_called_once_with(**{**common, "delta": 1})
        node.add_repeat.assert_called_once()
```
I think we _should_ avoid changing config data that the user might have passed to us via the API. Where, in the initializer, we have

```
cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
```

the `YAMLConfig` initializer does a `deepcopy` for us, but if we take `config` as-is, we should probably avoid modifying it. Even if we take care not to modify it in `add_repeat()` or similar, we might want to `deeocopy` it anyway: It's not that expensive, and then doubly safe.

Also on that line -- sorry, this is going way off-topic for this PR -- do we really want to accept _any_ `Config`? I think it will work because our JSON Schema logic should operate on the `dict` form of whatever it is, but do we really want to support e.g. `NMLConfig`? Probably not. `NMLConfig` might not even be deep enough (it's limited to `{"namelist": {"key": "val"}}` complexity) for a minimum viable ecFlow def. We could consider making the interface here `dict | YAMLConfig`.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: Not sure if the :) denotes sarcasm, but I think copilot is wrong here: If we raised a `UWConfigError` we'd be telling the user that they need to fix something; but if we let them get this far without a schema-validation error, *we* need to fix something -- and a raised assertion would tell us that. An assertion means "We don't think this can ever happen", whereas an exception means "You're doing it wrong".

It _is_ nice, though. It's a subtle point, and I don't blame copilot for missing it.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        assert "suite test" in suite_def
+        assert "family prep" in suite_def
+        assert "task setup" in suite_def
+        assert "task run" in suite_def
+        # Write suite definition.
+        ecf.write_suite_definition(tmp_path)
+        assert (tmp_path / "suite.def").exists()
+        # Write ecf scripts.
+        ecf.write_ecf_scripts(tmp_path)
+        assert len(list(tmp_path.rglob("*.ecf"))) == 1
```
maddenp-cu: Can we predict, and if so do we care, what the script's filename is? If not, no worries, but if so should we be more specific, like
```
assert (tmp_path / "expected.ecf").is_file()
```
?

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        # Verify suite definition was created.
+        suite_def = str(ecf)
+        assert "suite test" in suite_def
+        assert "family prep" in suite_def
+        assert "task setup" in suite_def
+        assert "task run" in suite_def
+        # Write suite definition.
+        ecf.write_suite_definition(tmp_path)
+        assert (tmp_path / "suite.def").exists()
+        # Write ecf scripts.
```
maddenp-cu:
```suggestion
        # Write ecf script.
```
In this case.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
christinaholtNOAA: No sarcasm. But you make an excellent point. I was suggesting that it was nice that CoPilot gave suggestions based on consistency with the rest of the code base. 

If somehow the user _does_ get here, though...will this be the appropriate way to handle it?

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: I can also open a ticket to take care of the deepcopy on config and ensure that the original user config isn't modified anywhere. I will want to check that I haven't inadvertently done this elsewhere, and it will be nice to have tests before making a change like that.

We probably do want a YAMLConfig here, and I can also make that change with the same new ticket. I think my original thought was that if a user had somehow created a "plain" Config from a dict, that would be acceptable, but no, that's probably not right.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: I think so: They'll see an `AssertionError` (same as they'd see from an `assert` statement) and -- hopefully -- open an issue with us.

src/uwtools/tests/test_ecflow.py
maddenp-cu: Plain `Config` is abstract, so they should not be able to instantiate one.

src/uwtools/tests/test_ecflow.py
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Yes. That makes sense and what I was thinking as well.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: I agree, this is cleaner and more DRY. (I hade to look up DRY.)
I switched `instance_with_scheduler` to build from `instance` by `setting _scheduler = "slurm"`.
Current tests don’t request both fixtures together, so behavior is unchanged.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Thank you. That's great suggestion. I also applied it to the related ecflowscript tests for consistency.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Great point! I've added an `assert_lines_in_order` helper and switched the include assertions to use it, so the test now validates both line identity and order.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+
+    def test_write_ecf_scripts__no_scripts(self, instance, logged, tmp_path):
+        instance.write_ecf_scripts(tmp_path)
+        assert logged("No scripts are configured for this workflow")
+
+    def test_write_ecf_scripts__with_scripts(self, instance, tmp_path):
+        instance._scripts = {Path("test/hello.ecf"): "#!/bin/bash\necho hello"}
+        instance.write_ecf_scripts(tmp_path)
+        outfile = tmp_path / "test" / "hello.ecf"
+        assert outfile.exists()
```
maddenp-cu: I think that, here and below, assertions using `.is_file()` or `.is_dir()`, rather than `.exists()`, would make for strong tests, as they are more specific. `.exists()` returns true whether the object is a file or a directory so, in addition to being a stronger test of the code, the more-specific methods are helpful documentation for users, who can see at a glance whether the filesystem entity we're talking about is a file or a directory.

src/uwtools/tests/test_ecflow.py
```diff
@@ -2,22 +2,602 @@
...
+        config = {"limits": [["limit1", 5], ["limit2", 10]]}
+        instance._add_node(config, task, suite)
+        # Note: add_items returns unconsumed generator in ecflow.py
+
+    def test__add_node__with_meters(self, instance):
+        suite = Suite("test")
+        task = Task("t1")
+        config = {"meters": [["meter1", 0, 100, 50]]}
+        instance._add_node(config, task, suite)
+        # Note: add_items returns unconsumed generator in ecflow.py
```
maddenp-cu: Maybe this test is missing an assertion?

maddenp-cu (commented):
A couple more small suggestions. I'll do another pass after the next round of updates gets pushed. This is looking good, though.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Yes, I agree that the test should verify the full mock call directly instead of pulling pieces out of `call args`. I have updated it to use `assert_called_once_with`, but with one adjustment: in this code path `JobScheduler.get_scheduler()` receives a single mapping argument, not keyword arguments, so the assertion compares the full dictionary that was passed.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: @maddenp-cu That makes sense sense. I ended up going with a full `assert_called_once_with{{...}}` here as well. It still covers the negative case since the assertion only passes if the call exactly matches the expected dictionary and therefore doesn't include `threads`. So I agree `call_args` would have been reasonable here, but the full-call assertion keeps the style consistent and is a bit stricter. What do you think?

 ```
def test__jobscheduler(self, instance_with_scheduler):
        execution = {"threads": 4, "batchargs": {"queue": "batch"}}
        with patch.object(ecflow.JobScheduler, "get_scheduler") as mock_get:
            instance_with_scheduler._jobscheduler(
                account="myaccount",
                execution=execution,
                rundir="/path/to/run",
            )
        mock_get.assert_called_once_with(
            {
                "account": "myaccount",
                "rundir": "/path/to/run",
                "scheduler": "slurm",
                "stdout": "/path/to/run.out",
                "threads": 4,
                "queue": "batch",
            }
        )

    def test__jobscheduler__no_threads(self, instance_with_scheduler):
        execution: dict = {}
        with patch.object(ecflow.JobScheduler, "get_scheduler") as mock_get:
            instance_with_scheduler._jobscheduler(
                account="myaccount",
                execution=execution,
                rundir="/path/to/run",
            )
        mock_get.assert_called_once_with(
            {
                "account": "myaccount",
                "rundir": "/path/to/run",
                "scheduler": "slurm",
                "stdout": "/path/to/run.out",
            }
        )

```

File: src/uwtools/tests/test_ecflow.py (hunk 10 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: I strengthened these tests to assert the actual ecFlow API calls rather than only checking that the mocks were called. For the `date` case, the test now verifies the `step -> delta` translation and that the constructed repeat object is what gets passed to `node.add_repeat`. I did not make the test assert in-place mutation of `config`, since that seems more like an implementation detail we may want to remove in follow-up work rather than behavior to lock in here.

```

 def test__add_repeat__date(self, instance):
        node = Mock()
        common = {"name": "YMD", "start": 20240101, "end": 20240131}
        with patch.object(ecflow, "RepeatDate") as mock_repeat:
            instance._add_repeat({**common, "step": 1}, "date", node)
        mock_repeat.assert_called_once_with(**{**common, "delta": 1})
        node.add_repeat.assert_called_once_with(mock_repeat.return_value)

    def test__add_repeat__int(self, instance):
        node = Mock()
        config = {"name": "STEP", "start": 0, "end": 10, "step": 1}
        with patch.object(ecflow, "RepeatInteger") as mock_repeat:
            instance._add_repeat(config.copy(), "int", node)
        mock_repeat.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(mock_repeat.return_value)

    def test__add_repeat__day(self, instance):
        node = Mock()
        config = {"step": 1}
        with patch.object(ecflow, "RepeatDay") as mock_repeat:
            instance._add_repeat(config.copy(), "day", node)
        mock_repeat.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(mock_repeat.return_value)

    def test__add_repeat__enumerated(self, instance):
        node = Mock()
        config = {"name": "MEMBER", "values": ["m01", "m02", "m03"]}
        with patch.object(ecflow, "RepeatEnumerated") as mock_repeat:
            instance._add_repeat(config.copy(), "enumerated", node)
        mock_repeat.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(mock_repeat.return_value)
```

File: src/uwtools/tests/test_ecflow.py (hunk 9 above)
Comment (maddenp-cu) [reply src/uwtools/tests/test_ecflow.py]: I like it. I think being explicit helps us understand the intent later when this is no longer fresh in our minds.

File: src/uwtools/ecflow.py (hunk 11 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/ecflow.py:201]: I agree that a ticket is the right follow-up.

File: src/uwtools/tests/test_ecflow.py (hunk 12 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Yes. That's a very good point on the naming. Here's the update: 

```
 def test__add_repeat__date(self, instance):
        node = Mock()
        common = {"name": "YMD", "start": 20240101, "end": 20240131}
        with patch.object(ecflow, "RepeatDate") as repeat_date:
            instance._add_repeat({**common, "step": 1}, "date", node)
        repeat_date.assert_called_once_with(**{**common, "delta": 1})
        node.add_repeat.assert_called_once_with(repeat_date.return_value)

    def test__add_repeat__int(self, instance):
        node = Mock()
        config = {"name": "STEP", "start": 0, "end": 10, "step": 1}
        with patch.object(ecflow, "RepeatInteger") as repeat_integer:
            instance._add_repeat(config.copy(), "int", node)
        repeat_integer.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(repeat_integer.return_value)

    def test__add_repeat__day(self, instance):
        node = Mock()
        config = {"step": 1}
        with patch.object(ecflow, "RepeatDay") as repeat_day:
            instance._add_repeat(config.copy(), "day", node)
        repeat_day.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(repeat_day.return_value)

    def test__add_repeat__enumerated(self, instance):
        node = Mock()
        config = {"name": "MEMBER", "values": ["m01", "m02", "m03"]}
        with patch.object(ecflow, "RepeatEnumerated") as repeat_enumerated:
            instance._add_repeat(config.copy(), "enumerated", node)
        repeat_enumerated.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(repeat_enumerated.return_value)

    @mark.parametrize("repeat_type", ["datelist", "string"])
    def test__add_repeat__enumerated_variants(self, instance, repeat_type):
        """
        Test that datelist and string repeat types also use RepeatEnumerated.
        """
        node = Mock()
        config = {"name": "VAR", "values": ["a", "b"]}
        with patch.object(ecflow, "RepeatEnumerated") as repeat_enumerated:
            instance._add_repeat(config.copy(), repeat_type, node)
        repeat_enumerated.assert_called_once_with(**config)
        node.add_repeat.assert_called_once_with(repeat_enumerated.return_value)
```

File: src/uwtools/tests/test_ecflow.py (hunk 14 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Yes, that's a good point. Thank you. I switched the  `instance` fixture to use `return _ECFlowDef(minimal_config)`. With the minimal config fixture, `__init__` still yields the same minimal test object. This should be more maintainable because the fixture now follows the class’s real initialization path instead of manually reproducing its internal state.

```
@fixture
def instance(minimal_config):
    """
    Create a minimal _ECFlowDef instance.
    """
    return _ECFlowDef(minimal_config)

```

File: src/uwtools/tests/test_ecflow.py (hunk 15 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Agreed. I replaced the placeholder with a real test that patches `task.add_event` and asserts the expected event calls in order. 

```
  config = {"events": ["event1", "event2"]}
        task = Task("t1")
        with patch.object(task, "add_event") as add_event:
            instance._add_node(config, task, suite)
        add_event.assert_has_calls([call("event1"), call("event2")])
```

For your comment about lines 161-167, Paul brought that up as well and I addressed it in a comment to him.

File: src/uwtools/tests/test_ecflow.py (hunk 16 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: I spent a decent amount of time on this. At first I didn't think this suggestion was necessarily better, then I did some digging into the Ecflow API and it does support stronger direct-state assertions. 

I updated the tests to use ecFlow’s own node accessors instead of patched method calls. The `events` test now checks `find_event()` and the `events` list on the `Task`, and the `defstatus` test now asserts `task.get_defstatus() == DState.complete`, so these now verify the actual node state rather than only that the corresponding add-methods were invoked.

```
        config = {"events": ["event1", "event2"]}
        task = Task("t1")
        instance._add_node(config, task, suite)
        assert task.find_event("event1") is not None
        assert task.find_event("event2") is not None
        assert [event.name() for event in task.events] == ["event1", "event2"]
```

File: src/uwtools/tests/test_ecflow.py (hunk 17 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py:344]: Good point. I updated so that the test asserts the expanded node names directly. It now uses a templated task name and checks that the `_add_node` calls receive nodes named `m01` and `m02`, instead of only verifying that `_add_node` was called twice.

`def test__expand_block__basic(self, instance):
        config = {"expand": {"MEMBER": ["m01", "m02"]}, "task_{{ ec.MEMBER }}": {}}
        suite = Suite("test")
        instance._d.add(suite)
        with patch.object(instance, "_add_node") as mock_add_node:
            instance._expand_block(config, "{{ ec.MEMBER }}", Task, suite)
        assert mock_add_node.call_count == 2
        assert [x.kwargs["node"].name() for x in mock_add_node.call_args_list] == ["m01", "m02"]`

File: src/uwtools/tests/test_ecflow.py (hunk 18 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Great catch! Thank you!
I added a real test for `inlimits` and it exposed an implementation issue as well. `_add_node()` was routing `inlimits` through a helper that returned an unconsumed generator, so those `add_inlimit()` calls never executed. The test now asserts both configured inlimits are applied, and the helper in `ecflow.py` now eagerly invokes the node methods. Here is what I did: 

In `ecflow.py`:
```
parent.add(node)
def add_items(method, cfg) -> None:
    for args in cfg:
        method(*args)

for key, subconfig in config.items():
    tag, name = self._tag_name(key)
    match tag:
        ...
        case "inlimits":
            add_items(node.add_inlimit, subconfig)
        case "labels":
            add_items(node.add_label, subconfig)
```
         
In `test_pyecflow.py`: 
```
def test__add_node__with_inlimits(self, instance):
    suite = Suite("test")
    task = Task("t1")
    config = {"inlimits": [["limit1", "/path"], ["limit2", "/path2"]]}
    with patch.object(task, "add_inlimit") as add_inlimit:
        instance._add_node(config, task, suite)
    assert add_inlimit.call_args_list == [
        call("limit1", "/path"),
        call("limit2", "/path2"),
    ]
```

File: src/uwtools/tests/test_ecflow.py (hunk 19 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: I replaced the placeholder with a real `labels` test in `test_ecflow.py`.

```
    def test__add_node__with_labels(self, instance):
        suite = Suite("test")
        task = Task("t1")
        config = {"labels": [["label1", "value1"], ["label2", "value2"]]}
        with patch.object(task, "add_label") as add_label:
            instance._add_node(config, task, suite)
        assert add_label.call_args_list == [
            call("label1", "value1"),
            call("label2", "value2"),
        ]
```

File: src/uwtools/tests/test_ecflow.py (hunk 20 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: I replaced both placeholders with real tests. They now assert that `_add_node()` forwards each configured limits entry to `task.add_limit()` and each configured meters entry to `task.add_meter()`. 


```
def test__add_node__with_limits(self, instance):
    suite = Suite("test")
    task = Task("t1")
    config = {"limits": [["limit1", 5], ["limit2", 10]]}
    with patch.object(task, "add_limit") as add_limit:
        instance._add_node(config, task, suite)
    assert add_limit.call_args_list == [
        call("limit1", 5),
        call("limit2", 10),
    ]

def test__add_node__with_meters(self, instance):
    suite = Suite("test")
    task = Task("t1")
    config = {"meters": [["meter1", 0, 100, 50]]}
    with patch.object(task, "add_meter") as add_meter:
        instance._add_node(config, task, suite)
    assert add_meter.call_args_list == [
        call("meter1", 0, 100, 50),
    ]
```

File: src/uwtools/tests/test_ecflow.py (hunk 23 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Good catch! Yes, the filename is predictable here, so I tightened the integration test to assert the exact output path. 

Updated code: 
```
# Write ecf scripts.
ecf.write_ecf_scripts(tmp_path)
assert (tmp_path / "test" / "run.ecf").is_file()
```

File: src/uwtools/tests/test_ecflow.py (hunk 24 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Done.

File: src/uwtools/tests/test_ecflow.py (hunk 25 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: I updated those filesystem assertions to use `.is_file()` instead of `.exists()` throughout the relevant tests. These checks are all asserting generated file artifacts, so `.is_file()` is more specific and better documents the expected output type.

File: src/uwtools/tests/test_ecflow.py (hunk 26 above)
Comment (AnnaSmoot-NOAA) [reply src/uwtools/tests/test_ecflow.py]: Thanks for catching that. I updated it so it now asserts `_add_node()` forwards the configured meter entry to `task.add_meter()` rather than only executing the branch. 

```
def test__add_node__with_meters(self, instance):
    suite = Suite("test")
    task = Task("t1")
    config = {"meters": [["meter1", 0, 100, 50]]}
    with patch.object(task, "add_meter") as add_meter:
        instance._add_node(config, task, suite)
    assert add_meter.call_args_list == [
        call("meter1", 0, 100, 50),
    ]
```

AnnaSmoot-NOAA:
I think everything is resolved at this point. Let me know if there is any other feedback. 

Also, thank you both @maddenp-cu and @christinaholtNOAA for giving such clear and thoughtful feedback. I know it takes a lot of time on your end and it's been so helpful for me.

maddenp-cu (approved):
Thanks for incorporating so much feedback and iterating on this with us. This represents a lot of new tests, enough that we've probably collectively missed _something_, but it looks great overall and we have a solid foundation here to build on if we later find that we've missed something.

christinaholtNOAA (approved):
Thanks for all the great work and quick turnarounds. LGTM.

PR #863, opened 2026-04-15
Add an ecFlow JSON schema

### LINKED ISSUE #849: Write a JSON schema for the ecFlow YAML language.
**Description**

The new YAML language for the ecFlow tool needs to be schema checked. Write the JSON schema necessary to ensure appropriate validation and test the schema via unittests.

**Acceptance Criteria**

- [ ] Write a JSON schema for the ecFlow YAML
- [ ] Unit test it as thoroughly as possible at the time the ticket is worked. Future additions to the language will be the responsibility of the developer working on the new feature.

christinaholtNOAA:
**Synopsis**

Adds an ecFlow schema and tests for it.

Fixes #849.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


christinaholtNOAA:
Most of what I have here comes from the API Docs given the signatures of the classes under ecflow. (https://ecflow.readthedocs.io/en/latest/python_api/python_api.html#api).

When more than one property is allowed for a node (like events, variables, etc.), I have allowed for an array of those in UW YAML. Otherwise, it follows the object signature exactly(e.g., trigger).

I learned that when composing multiple refs together under a higher-level section (like in `familycontainer`), you must have the higher level handle the "additionalProperties" requirement, and that tag doesn't work as expected (so says ChatGPT), so "unevaluatedProperties" should be used instead to get the expected behavior.

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+                    "expand": {
+                        "myvar": ["foo", "bar"],
+                    },
+                    "family_three": {},
+                },
+            },
+        },
+    }
+    assert not errors(config)
+    # At least one suite is required
```
maddenp-cu:
```suggestion
    # At least one suite is required:
```
For consistency with existing comments in this module.

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+                    "families_a{{ ec.var }}": {"family_b": {}, "expand": {"myvar": [1, 2]}}
+                },
+                "expand": {
+                    "avar": ["foo", "bar"],
+                },
+            }
+        }
+    }
+    assert not errors(config)
+    # Ensure that the requried "expand" and node entries are present
```
maddenp-cu:
```suggestion
    # Ensure that the requried "expand" and node entries are present:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+
+
+def test_schema_ecflow_refs_addons_events():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Basic spec:
+    assert not errors({"events": [[1, "foo"]]})
+    assert not errors({"events": [1, "foo"]})
+    assert not errors({"events": [1, 2, 3]})
+    assert not errors({"events": ["foo", "bar", "baz"]})
+    # Must be one of the specified patterns
```
maddenp-cu:
```suggestion
    # Must be one of the specified patterns:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert "is not valid under any of the given schemas" in errors({"events": [["foo", 2]]})
+
+
+def test_schema_ecflow_refs_addons_inlimits():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Supports one of four categories
```
maddenp-cu:
```suggestion
    # Supports one of four categories:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert not errors({"inlimits": [["foo"]]})
+    assert not errors({"inlimits": [["foo", "bar"]]})
+    assert not errors({"inlimits": [["foo", "bar", 2], ["baz", "qux"]]})
+    assert not errors({"inlimits": [["foo", "bar", 2, True]]})
+    # Exactly one is required
```
maddenp-cu:
```suggestion
    # Exactly one is required:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert "['foo', 2] is not valid under any of the given schemas" in errors(
+        {"inlimits": [["foo", 2]]}
+    )
+    assert "{} is not of type 'array'" in errors({"inlimits": {}})
+    assert "[] should be non-empty" in errors({"inlimits": []})
+
+
+def test_schema_ecflow_refs_addons_labels():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Allows for a list of 2-lists
```
maddenp-cu:
```suggestion
    # Allows for a list of 2-lists:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert not errors({"labels": [["foo", "bar"], ["baz", "qux"]]})
+    # List must contain 2-item lists
```
maddenp-cu:
```suggestion
    # List must contain 2-item lists:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Basic spec:
+    assert not errors({"late": {"active": "00:15", "complete": "+02:00", "submitted": "20:00"}})
+
+
+def test_schema_ecflow_refs_addons_limits():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Basic spec:
+    assert not errors({"limits": [["foo", 1], ["bar", 2]]})
+    # It's a list of lists
```
maddenp-cu:
```suggestion
    # It's a list of lists:
```

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert "'foo' is not of type 'array'" in errors({"limits": ["foo"]})
+    assert "[] should be non-empty" in errors({"limits": []})
+
+
+def test_schema_ecflow_refs_addons_meters():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Basic spec:
+    assert not errors({"meters": [["foo", 1, 2], ["bar", 2, 8]]})
+    # It's a list of lists
```
maddenp-cu:
```suggestion
    # It's a list of lists:
```

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
...
+        },
+        "events": {
+          "items": {
+            "oneOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "integer"
+              },
```
maddenp-cu, line 25: Per the docs, it looks like this could also be `{"type": ["integer", "string"]}` if I'm understanding the intent correctly. Ever so slightly more concise...

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
...
+              {
+                "prefixItems": [
```
maddenp-cu, line 27: TIL `prefixItems` -- very nice!

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
...
+                  }
+                ],
+                "type": "array"
+              }
+            ]
+          },
+          "minItems": 1,
+          "type": "array"
+        },
+        "inlimits": {
```
maddenp-cu, line 42: I really appreciate the rigor of this schema.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
...
+              {
+                "maxItems": 1,
+                "minItems": 1,
+                "prefixItems": [
+                  {
+                    "type": "string"
+                  }
+                ],
+                "type": "array"
+              },
```
maddenp-cu, line 53: Per [this section](https://json-schema.org/understanding-json-schema/reference/array#additionalitems) in the docs, it sounds like a simpler alternative to `maxItems` / `minItems` might be to just say `"items": false` to indicate that nothing beyond what is enumerated by `prefixItems` is permitted.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
 {
+  "$defs": {
+    "addons": {
+      "properties": {
+        "defstatus": {
+          "type": "string"
+        },
+        "events": {
+          "items": {
+            "oneOf": [
```
maddenp-cu: Just a piece of trivia I picked up somewhere: Apparently, `oneOf` is more expensive to compute than `anyOf` because JSON Schema has to ensure that exactly one sub-schema matches, rather than just OKing the data at the first match. It has to invalidate all the others. I seriously doubt it matters at all in `uwtools`, and I'm not asking for a change, but I found it interesting and lean toward `anyOf` by default and haven't seen a problem yet. I'm sure there are cases where it matters, though.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,517 @@
...
+            "type": "string"
+          },
+          "type": "array"
+        },
+        "scheduler": {
+          "enum": [
+            "lsf",
+            "pbs",
+            "slurm"
+          ],
```
maddenp-cu, line 501: Not now, but we should probably factor this out, as it also occurs in `platform.jsonschema` and, if I'm not mistaken, the two need to be kept in-sync.

maddenp-cu (approved):
Looks good -- a few comments/suggestions as usual.

src/uwtools/resources/jsonschema/ecflow.jsonschema
christinaholtNOAA: Now that I have tests...let's see!

src/uwtools/resources/jsonschema/ecflow.jsonschema
christinaholtNOAA: I like this! It does require the other, nicer switch to `anyOf`, as well.

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+    assert "[] should be non-empty" in errors({"limits": []})
+
+
+def test_schema_ecflow_refs_addons_meters():
+    errors = schema_validator("ecflow", "$defs", "addons")
+    # Basic spec:
+    assert not errors({"meters": [["foo", 1, 2], ["bar", 2, 8]]})
+    # It's a list of lists:
+    assert "'foo' is not of type 'array'" in errors({"limits": ["foo"]})
+    assert "[] should be non-empty" in errors({"limits": []})
```
elcarpenterNOAA: Should these say `"meters"` instead of `"limits"`?

src/uwtools/tests/test_schemas.py
```diff
@@ -670,6 +670,246 @@ def test_schema_chgres_cube_namelist_update_values(chgres_cube_config, chgres_cu
...
+                    "families_a{{ ec.var }}": {"family_b": {}, "expand": {"myvar": [1, 2]}}
+                },
+                "expand": {
+                    "avar": ["foo", "bar"],
+                },
+            }
+        }
+    }
+    assert not errors(config)
+    # Ensure that the requried "expand" and node entries are present:
```
elcarpenterNOAA:
```suggestion
    # Ensure that the required "expand" and node entries are present:
```

elcarpenterNOAA (approved):

src/uwtools/tests/test_schemas.py
christinaholtNOAA: Yes, I do believe they should. Nice catch!

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,511 @@
 {
+  "$defs": {
+    "addons": {
+      "properties": {
+        "defstatus": {
+          "type": "string"
```
AnnaSmoot-NOAA, line 15: For ecFlow, the valid `defstatus` values are typically `complete`, `suspended`, `queued`, and `aborted`.  Since those are predefined values, enum might be a good addition here:, like how you handled `scheduler` on lines 497-502. 

```
"defstatus": {
  "enum": ["complete", "suspended", "aborted", "queued", "submitted", "active", "unknown"],
  "type": "string"
}
```

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,511 @@
...
+            },
+            "variable": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "variable",
+            "list"
+          ]
+        },
```
AnnaSmoot-NOAA, line 236: I'm still new to JSON schema so this might not be important or correct, but I think you're missing `"type": "object"` here for consistency with other repeat definitions like `repeat_date`, `repeat_datetime`, and `repeat_string`

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,511 @@
...
+            "variable": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "variable",
+            "start",
+            "end"
+          ]
+        },
```
AnnaSmoot-NOAA, line 316: Same comment about possibly missing `"type": "object"` here for consistency with other repeat definitions like `repeat_date`, `repeat_datetime`, and `repeat_string`

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -1,6 +1,511 @@
...
+        {
+          "propertyNames": {
+            "pattern": "^suite_.+$"
+          }
+        },
+        {
+          "propertyNames": {
+            "pattern": "^suites_.+$"
+          }
+        }
```
AnnaSmoot-NOAA: Once again, this is my first time really looking at JSCON schema so please correct my understanding if wrong. 
The `propertyNames` pattern requires all properties to match `^suite_.+$` or `^suites_.+$`, but `extern`, `scheduler`, and `vars` are also valid properties. Is the `anyOf` contraint too strict here?

Do you think a test case to confirm behavior might be helpful? If a test like the one below fails, maybe the `propertyNames` should be tweaked for allowing `properties` (`extern`, `vars`, `scheduler`)?

```
def test_schema_ecflow_suite_with_scheduler():
    errors = schema_validator("ecflow")
    config = {
        "ecflow": {
            "scheduler": "slurm",
            "suite_one": {
                "task_two": {"script": {"execution": {"executable": "echo hi"}}},
            },
        },
    }
    assert not errors(config)
```

AnnaSmoot-NOAA (commented):
Just a couple of quick questions. 
1. `defstatus` and possibly using `enum`
2.  possibly missing `"type": "object"`
3. question about `propertyNames` pattern constraint.

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu: I think you're right. I'm surprised that JSON Schema is ok with there not being a `'type"` specified here (the `repeat_int` block below is also missing it). Apparently, per [this post](https://github.com/json-schema/json-schema/issues/172#issuecomment-114076650), anything can be accepted if `"type"` is missing and the other constraints in the schema are only applied to types that they match -- like `properties` would match for `object`s, `minItems` and `maxItems` would match for `array`s, etc. It sounds like we would normally want to specify `"type"`.

This is what the schema unit tests are for: To verify that the schema we've written actually does what we think it does. We usually try passing values of types that don't match what's expected and assert that JSON Schema complains. @christinaholtNOAA this could be a good opportunity to see if maybe we're in general not testing something that we should -- if, that is, what I wrote above is true. Would, stay, a `string` value pass validation as a `repeat_datetime` value if we don't have `"type": "object"` here?

christinaholtNOAA: Oh, this is scary! I feel like we do a pretty good job of testing the properties that go into objects, but rarely throw non-objects at these blocks!

src/uwtools/resources/jsonschema/ecflow.jsonschema
christinaholtNOAA: If ChatGPT is to be believed, there isn't a clean way to make this scenario super strict, i.e., require at least one patternProperty and allow any others.

What do folks think? Do we need that strictness? I don't have requirements at other levels that a suite must have a family or task, or that a family must have a task. I honestly don't know if the API will allow the suite definition to stop at the suite or family level and not include a task.

Does it make sense to put a check in the code to warn, "you didn't really build a tree, you'll need a suite at the top level to get that started"? I suppose we might have a similar warning at the suite and family levels that lets folks know that their workflow tree has ended at a level that won't continue on to host a task.

src/uwtools/resources/jsonschema/ecflow.jsonschema
christinaholtNOAA: I think I might like to do this sort of test for the whole set of schemas. Would it be okay to separate that one out into its own PR a little later?

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu: I'm not familiar with the required syntax. Could you say in English here what the rule is that we'd like to enforce?

christinaholtNOAA: Yes. I'd like to require that at least one `suite_` or `suites_` entry is present. All other properties are optional.

maddenp-cu: Quick initial thought and then I'll see if I can suggest something in answer to this question: I think we can summarize `"pattern": "^suite_.+$"` and `"pattern": "^suites_.+$"` as `"pattern": "^suite[s]?_.+$"`, i.e. the `s` after `suite` is optional. This should be close to the regex form accepted by JSON Schema if not exactly right.

maddenp-cu: That previous regex suggestion might not be as useful as I thought, as I now see that those regexes appear again below, and there they can't be combined because they have different subschemas. I'm still trying to understand why we need both `propertyNames` and `patternProperties`.

I was thinking about something like the following pseudocode, to be used... somewhere:
```
"allOf": [
  {
    "patternProperties": {
      "^suite_.+$": {...},
      "^suites_.+$": {...},
    },
    "minProperties": 1
  },
  {
    "patternProperties": {
      "^suite_.+$": {...},
      "^suites_.+$": {...},
    },
    "properties": {
      "extern": {},
      "scheduler": {},
      "vars": {}
    }
  }
]
```
like: It must be the case that at least 1 `suite_` or `suites_` property exists, and then it must also be the case that certain other properties are allowed along with a `suite_` or `suites_` property, but none of the other properties are required. Maybe the values of the `suite_` and `suites_` pattern properties could be DRYed out to `$defs`. (I guess they already contain `$defs` references so maybe they're not too bad to repeat, but maybe an extra level of redirection would make it even more concise.)

Is that useful at all? I'm still unclear on the use of `propertyNames` and `patternProperties` together, referring to the same things, so maybe I'm missing something.

maddenp-cu: Or, looking at the docs about `unevaluatedProperties`, maybe that could be something like
```
{
  "allOf": [
    {
      "patternProperties": {
        "^suite_.+$": {...},
        "^suites_.+$": {...},
      },
      "minProperties": 1
    }
  ],
  "properties": {
    "extern": {...},
    "scheduler": {...},
    "vars": {...}
  },
  "type": "object",
  "unevaluatedProperties": false
}
```
as in, `extern` / `scheduler` / `vars` are allowed, plus there _must_ be at least one `suite_` or `suites_` property, and if the subschema under `allOf` validates any `suite_` / `suites_` properties, then they are considered allowed by the enclosing schema?

AnnaSmoot-NOAA (approved):

src/uwtools/resources/jsonschema/ecflow.jsonschema
christinaholtNOAA: I've worked with several iterations of this and am still not seeing the desired outcome -- that it throws an error when there isn't at least one `suite_` or `suites_` entry. If I can figure this out, I'd like to incorporate it at each level where that pattern applies. I think I will leave it in this more permissive state for now.

PR #869, opened 2026-04-18
Strings container for ecFlow strings

### LINKED ISSUE #865: Add a string class to ecflow.py for all the ecFlow-specific terminology we expect from the schema.
**Description**

Adding a dataclass to the ecflow.py module to describe all the ecFlow-specific strings. Having this connection between the code and the schema will help us identify and debug typos in the development process more quickly, as the linter will catch them easily for us.

**Acceptance Criteria**

- [ ] Identify all the strings used as keys in the config for the ecflow interface to YAML and include them in a dataclass of strings.
- [ ] Replace the identified strings with the dataclass instances instead.
- [ ] We don't have tests for other similar data classes in the strings or rocoto modules, so none should be required for this ticket either.

maddenp-cu:
**Synopsis**

Resolves #865.

- Create a `uwtools.strings.EC` container for ecFlow-related strings.
- Borrow `wxvx` code that eliminates the need to repeat the string both as a key and a value in string containers.
- Update some string keys to include underscores where their values include them, reducing the number of special cases where keys and values do not match.

On that last point, it seems to me now that the value of short keys is lower than the value of consistency. Also, many of the updated keys were short only where they omitted underscores that, when included, probably make them more readable. There are still a few unavoidable special cases, but far fewer.

Almost all of the diffs in this PR are due to string-container keys being updated to match their values, so you should be able to ignore most of them: `mypy` checks that member references on these string containers are valid.

**Type**

**Impact**

**Checklist**


src/uwtools/ecflow.py
```diff
@@ -30,7 +30,7 @@
 from uwtools.exceptions import UWConfigError
 from uwtools.logging import log
 from uwtools.scheduler import JobScheduler
-from uwtools.strings import STR
+from uwtools.strings import EC, STR
```
maddenp-cu, line 33: The use of the new `EC` string container in this module addresses #869.

elcarpenterNOAA (approved):

src/uwtools/ecflow.py
christinaholtNOAA: I believe the rocoto module defines its own `STR` data class. I'm curious if we might want to similarly move it to the strings module. Probably not in this PR.

src/uwtools/strings.py
```diff
@@ -2,38 +2,90 @@
...
+# the risk of typos in string literals, and exposes references to the type checker for validation.
+
 from dataclasses import dataclass, fields
 
+# Private
+
+_ = ""  # default value, to be replaced by key
+
+
+class _ValsMatchKeys:
```
christinaholtNOAA, line 17: I like this approach!

christinaholtNOAA (approved):
This is quite a nice upgrade to the strings module and to the code's readability.

maddenp-cu:
Sorry, I just realized that a lot of the diffs here are due to the `validate-xml` changes from `main` that are here now but not yet on `ecflow_tool`. #871 should fix that and make a lot of the diffs here disappear.

PR #872, opened 2026-04-20
Adding API interface for ecFlow.

christinaholtNOAA:
**Synopsis**

Adds the ecflow realize and validate (placeholder) to the API. Includes unit tests.

Fixes #851. 

**Type**

- [ ] Documentation

**Impact**

**Checklist**


src/uwtools/api/ecflow.py
```diff
@@ -0,0 +1,58 @@
...
+    config: _YAMLConfig | Path | str | None,
+    output_path: Path | str | None = None,
+    scripts_path: Path | str | None = None,
+    stdin_ok: bool = False,
+) -> bool:
+    """
+    Realize the ecFlow workflow defined in the given YAML as a Suite Definition and ecf scripts.
+
+    If no input file is specified, ``stdin`` is read. A ``YAMLConfig`` object may also be provided
+    as input. If no output file is specified, the Suite Definition is written to``stdout``. The ecf
```
elcarpenterNOAA:
```suggestion
    as input. If no output file is specified, the Suite Definition is written to ``stdout``. The ecf
```

```suggestion
    as input. If no output file is specified, the Suite Definition is written to ``stdout``. The ecf
```

src/uwtools/api/ecflow.py
```diff
@@ -0,0 +1,58 @@
...
+    scripts are not produced if scripts_path is not provided.
```
```suggestion
    scripts are not produced if ``scripts_path`` is not provided.
```

src/uwtools/api/ecflow.py
```diff
@@ -0,0 +1,58 @@
...
+    _realize(
+        config=_ensure_data_source(_str2path(config), stdin_ok),
+        output_path=_str2path(output_path),
+        scripts_path=_str2path(scripts_path),
+    )
+    return True
+
+
+def validate(
+    yaml_file: Path | str | None = None,
```
maddenp-cu, line 48: Not for this PR, but I wonder if we should support validating a `dict` or `YAMLConfig` here, too. I think the drivers have this flexibility, which could be handy for programmatic use cases. (As opposed to the `rocoto validate-xml` case, where it's unlikely that one would have an XML document without also having a file.)

src/uwtools/tests/api/test_ecflow.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_api_ecflow_realize():
+    path1, path2 = Path("foo"), Path("bar")
+    with patch.object(ecflow, "_realize") as _realize:
+        ecflow.realize(config=path1, output_path=path2)
+    _realize.assert_called_once_with(config=path1, output_path=path2, scripts_path=None)
+
+
+def test_api_ecflow_realize_with_scripts():
```
maddenp-cu:
```suggestion
def test_api_ecflow_realize__with_scripts():
```

src/uwtools/tests/test_ecflow.py
```diff
@@ -432,12 +442,6 @@ def test_write_suite_definition__creates_directory(self, instance, tmp_path):
         instance.write_suite_definition(nested_path)
         assert (nested_path / "suite.def").is_file()
 
-    def test_write_suite_definition__with_string_path(self, instance, tmp_path):
-        suite = Suite("test")
-        instance._d.add(suite)
-        instance.write_suite_definition(str(tmp_path))
-        assert (tmp_path / "suite.def").is_file()
-
```
AnnaSmoot-NOAA, line 440: I'm just wondering why this test was removed?  Is this because `write_suite_definition` now accepts `Path | None` instead of `Path | str`?

src/uwtools/tests/test_ecflow.py
```diff
@@ -644,3 +648,46 @@ def test_integration__with_expand(self):
...
+    assert capsys.readouterr().out == expected
+
+
+def test_ecflow_validate__file(assets):
+    cfgfile, _, _ = assets
+    assert ecflow.validate_file(cfgfile)
+
+
+def test_ecflow_validate__stdin():
+    assert ecflow.validate_file()
```
AnnaSmoot-NOAA, line 693: This only asserts `validate_file()` returns `True`. Once I have my validate work ready, will this be something I should test with mocking?

maddenp-cu, line 693: This is as it is because `validate_file()` is just a stub right now, right?

src/uwtools/ecflow.py
```diff
@@ -357,3 +359,28 @@ def _tag_name(self, key: str) -> tuple[str, str]:
...
+
+
+def realize(
+    config: YAMLConfig | Path | None,
+    output_path: Path | None = None,
+    scripts_path: Path | None = None,
+) -> str:
+    """
+    Realize the ecFlow suite defined in a given YAML as a Suite Definition and corresponding ecf
+    scripts (if scripts_path is provided).
```
```suggestion
    scripts (if ``scripts_path`` is provided).
```

src/uwtools/ecflow.py
```diff
@@ -357,3 +359,28 @@ def _tag_name(self, key: str) -> tuple[str, str]:
...
+
+    :param config: Path to YAML input file (None => read stdin), or YAMLConfig object.
```
```suggestion
    :param config: Path to YAML input file (None => read ``stdin``), or YAMLConfig object.
```

src/uwtools/ecflow.py
```diff
@@ -357,3 +359,28 @@ def _tag_name(self, key: str) -> tuple[str, str]:
...
+    :param output_path: Path to write the rendered Suite Definition file (None => write to stdout).
```
```suggestion
    :param output_path: Path to write the rendered Suite Definition file (None => write to ``stdout``).
```

elcarpenterNOAA (approved):

src/uwtools/api/ecflow.py
```diff
@@ -0,0 +1,58 @@
...
+    :param stdin_ok: OK to read from ``stdin``?
+    :return: ``True``.
+    """
+    _realize(
+        config=_ensure_data_source(_str2path(config), stdin_ok),
+        output_path=_str2path(output_path),
+        scripts_path=_str2path(scripts_path),
+    )
+    return True
+
```
AnnaSmoot-NOAA, line 45: I'm looking at these lines and the ones listed below in `src/uwtools/api/ecflow.py`. 

```
def realize(
    config: YAMLConfig | Path | None,
    output_path: Path | None = None,
    scripts_path: Path | None = None,
) -> str:
    """
    Realize the ecFlow suite defined in a given YAML as a Suite Definition and corresponding ecf
    scripts (if scripts_path is provided).

    :param config: Path to YAML input file (None => read stdin), or YAMLConfig object.
    :param output_path: Path to write the rendered Suite Definition file (None => write to stdout).
    :param scripts_path: Path to write the rendered ecf scripts (None => do not write scripts).
    :return: Suite Definition as a string.
    """
    suite = _ECFlowDef(config)
    suite.write_suite_definition(output_path)
    if scripts_path:
        suite.write_ecf_scripts(scripts_path)
    return str(suite)
```

`uwtools.ecflow.realize(...)` returns a string and `uwtools.api.ecflow.realize(...)` returns a boolean. 
Is it okay for these to be returning different types or could that make problems?

AnnaSmoot-NOAA (changes requested):

src/uwtools/ecflow.py
```diff
@@ -357,3 +359,28 @@ def _tag_name(self, key: str) -> tuple[str, str]:
...
+    :return: Suite Definition as a string.
+    """
+    suite = _ECFlowDef(config)
+    suite.write_suite_definition(output_path)
+    if scripts_path:
+        suite.write_ecf_scripts(scripts_path)
+    return str(suite)
+
+
+def validate_file(yaml_file: Path | None = None) -> bool:  # noqa: ARG001
```
maddenp-cu, line 386: Am I right in thinking that we can ignore `stdin_ok` at this level because it will already have been dealt with in the API, i.e. if we get here an `yaml_file` is `None` then the user has said `stdin_ok=True`?

maddenp-cu (approved):
A couple small suggestions and questions.

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: That's right. I'm not sure that we necessarily need to accept a string here if this isn't exposed to end users.

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: Yes, mocking will likely be sufficient for this test once it's no longer a placeholder.

src/uwtools/api/ecflow.py
christinaholtNOAA: This is typically how we deal with the API vs the implementation of the code in the module. The API in this case is going to have its output go to file or STDOUT, so the return value is more about returning that it did what it was asked (similar to a zero exit status in bash). The method in the ecflow module gives us a public method within uwtools to get to the string representation of the workflow in memory. Giving the end-user access to the in-memory string version seems like a poor choice given that users could start to mangle that in any number of ways that may no longer work with ecflow.

We haven't seen a problem from this pattern with our other API implementations, and have found the different use of the two to be helpful in various scenarios.

src/uwtools/api/ecflow.py
christinaholtNOAA: I think this is great feedback for #866. Here, it's just a placeholder.

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: Yep.

src/uwtools/ecflow.py
christinaholtNOAA: That's right. Do we still think that's sufficient, given how we've done it similarly elsewhere?

maddenp-cu: I think so, and it agrees with what you mentioned in a response to @AnnaSmoot-NOAA about `Path | str` not being necessary if in the API we handle the `str` -> `Path` conversion. Simpler types, simpler interfaces beyond the API make for easier maintenance I think.

AnnaSmoot-NOAA (approved):

PR #873, opened 2026-04-21
Add EcFlow Config Validation

### LINKED ISSUE #866: Add a validate method in the ecflow module.
**Description**

Add a validation method to the ecflow module. This will likely be very similar to the implementation in the base driver. The ecflow module should somehow know how to find its own JSON schema (uses something like "resources" importing).

**Acceptance Criteria**

- [ ] Add validation to the ecflow module.
- [ ] Include unit tests
- [ ] ecflow module knows where its own JSON schema is

AnnaSmoot-NOAA:
**Synopsis**

Adds programmatic and file-based validation for ecFlow configs in the core
`uwtools.ecflow` module. Replaces the placeholder `validate_file()` introduced
in #872 with a real implementation, and broadens the input contract to accept
a `dict`, `YAMLConfig`, `Path`, or `None` (read `stdin`).

Fixes #866 & NOAA-EMC/global-workflow/#4797

**Type**

- [ ] Documentation

**Impact**

**Checklist**

**Self-Review Comment**

This PR builds on #872 and addresses the follow-up question raised by
@maddenp-cu in that review: *"I wonder if we should support validating a `dict`
or `YAMLConfig` here, too."* The answer here is yes — the new `validate()`
accepts all of `dict`, `YAMLConfig`, `Path`, and `None`.

#872 has now merged into `ecflow_tool`. This branch has been merged with the
new `ecflow_tool`, and `src/uwtools/api/ecflow.py` (plus its test) has been
updated to import and call `validate(config=...)` instead of the placeholder
`validate_file(yaml_file=...)`.

## Follow-ups discovered while implementing

While testing fail-fast validation (ultimately not adopted in this PR), two
pre-existing issues in the ecFlow module surfaced and should be filed
separately:

1. **Schema / runtime mismatch on `execution`.** `execution-serial.jsonschema` requires `executable`, but `_ecflowscript_for_task` in `ecflow.py` reads `jobcmd`.
2. **Schema rejects single-key `expand` blocks.** `test_integration__with_expand` uses an `expand`-only block, which violates the schema's `minProperties: 2` rule on `expander`. Either the test or the schema should be updated.

### DIFF
--- a/src/uwtools/api/ecflow.py
+++ b/src/uwtools/api/ecflow.py
@@ -45,15 +45,16 @@ def realize(
 
 
 def validate(
-    yaml_file: Path | str | None = None,
+    config: _YAMLConfig | dict | Path | str | None = None,
     stdin_ok: bool = False,
 ) -> bool:
     """
-    Validate the input YAML against its schema.
+    Validate an ecFlow config against its schema.
 
-    :param yaml_file: Path to YAML file (``None`` or unspecified => read ``stdin``).
+    :param config: An ecFlow config as a ``dict``, a ``YAMLConfig``, a path to a YAML file
+        (``Path`` or ``str``), or ``None`` (read ``stdin``).
     :param stdin_ok: OK to read from ``stdin``?
-    :return: ``True`` if the YAML conforms to the schema.
+    :return: ``True`` if the config conforms to the schema.
     :raises: ``UWConfigError`` if validation fails.
     """
-    return _validate(config=_ensure_data_source(_str2path(yaml_file), stdin_ok))
+    return _validate(config=_ensure_data_source(_str2path(config), stdin_ok))
--- a/src/uwtools/api/ecflow.py
+++ b/src/uwtools/api/ecflow.py
@@ -54,6 +54,6 @@ def validate(
     :param yaml_file: Path to YAML file (``None`` or unspecified => read ``stdin``).
     :param stdin_ok: OK to read from ``stdin``?
     :return: ``True`` if the YAML conforms to the schema.
-    :raises: UWConfigError if validation fails.
+    :raises: ``UWConfigError`` if validation fails.
     """
     return _validate(config=_ensure_data_source(_str2path(yaml_file), stdin_ok))
--- a/src/uwtools/api/ecflow.py
+++ b/src/uwtools/api/ecflow.py
@@ -53,6 +53,7 @@ def validate(
 
     :param yaml_file: Path to YAML file (``None`` or unspecified => read ``stdin``).
     :param stdin_ok: OK to read from ``stdin``?
-    :return: ``True`` if the YAML conforms to the schema, ``False`` otherwise.
+    :return: ``True`` if the YAML conforms to the schema.
+    :raises: UWConfigError if validation fails.
     """
     return _validate(config=_ensure_data_source(_str2path(yaml_file), stdin_ok))
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -4,7 +4,6 @@ Support for creating ecFlow suite definitions and ecf scripts.
 
 from __future__ import annotations
 
-import json
 import re
 from copy import deepcopy
 from pathlib import Path
@@ -28,7 +27,7 @@ from ecflow import (  # type: ignore[import-untyped]
 
 from uwtools.config.formats.base import Config
 from uwtools.config.formats.yaml import YAMLConfig
-from uwtools.config.validator import bundle, internal_schema_file, validate_internal
+from uwtools.config.validator import validate_internal
 from uwtools.exceptions import UWConfigError
 from uwtools.logging import log
 from uwtools.scheduler import JobScheduler
@@ -385,14 +384,6 @@ def realize(
     return str(suite)
 
 
-def schema() -> dict:
-    """
-    Return the ecFlow module's internal schema.
-    """
-    path = internal_schema_file(schema_name=EC.ecflow)
-    return bundle(json.loads(path.read_text()))
-
-
 def validate(config: dict | YAMLConfig | Path | None = None) -> bool:
     """
     Validate an ecFlow config against the internal ecFlow schema.
@@ -400,9 +391,9 @@ def validate(config: dict | YAMLConfig | Path | None = None) -> bool:
     :param config: A ``dict``, a ``YAMLConfig``, a path to a YAML file, or ``None``
         (``None`` => read ``stdin``).
     :return: ``True`` if the config conforms to the schema.
-    :raises: UWConfigError if validation fails.
+    :raises: ``UWConfigError`` if validation fails.
     """
-    kwargs: dict = {"schema_name": EC.ecflow, "desc": "ecflow config"}
+    kwargs: dict = {"schema_name": EC.ecflow, "desc": "ecFlow config"}
     if isinstance(config, (dict, YAMLConfig)):
         kwargs["config_data"] = config
     else:
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -57,20 +57,6 @@ class _ECFlowDef:
     def __str__(self):
         return self._d.__str__()
 
-    def validate(self) -> bool:
-        """
-        Validate the ecFlow config against the internal ecFlow schema.
-
-        :return: ``True`` if the config conforms to the schema.
-        :raises: UWConfigError if validation fails.
-        """
-        validate_internal(
-            schema_name=EC.ecflow,
-            desc="ecflow config",
-            config_data={EC.ecflow: self._config},
-        )
-        return True
-
     def write_ecf_scripts(self, path: Path | str) -> None:
         """
         The ecf scripts for this workflow.
@@ -383,13 +369,19 @@ def schema() -> dict:
     return bundle(json.loads(path.read_text()))
 
 
-def validate_file(yaml_file: Path | None = None) -> bool:
+def validate(config: dict | YAMLConfig | Path | None = None) -> bool:
     """
-    Validate an ecFlow YAML config against the internal ecFlow schema.
+    Validate an ecFlow config against the internal ecFlow schema.
 
-    :param yaml_file: Path to YAML file (``None`` => read ``stdin``).
+    :param config: A ``dict``, a ``YAMLConfig``, a path to a YAML file, or ``None``
+        (``None`` => read ``stdin``).
     :return: ``True`` if the config conforms to the schema.
     :raises: UWConfigError if validation fails.
     """
-    validate_internal(schema_name=EC.ecflow, desc="ecflow config", config_path=yaml_file)
+    kwargs: dict = {"schema_name": EC.ecflow, "desc": "ecflow config"}
+    if isinstance(config, (dict, YAMLConfig)):
+        kwargs["config_data"] = config
+    else:
+        kwargs["config_path"] = config
+    validate_internal(**kwargs)
     return True
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -4,6 +4,7 @@ Support for creating ecFlow suite definitions and ecf scripts.
 
 from __future__ import annotations
 
+import json
 import re
 from copy import deepcopy
 from pathlib import Path
@@ -27,6 +28,7 @@ from ecflow import (  # type: ignore[import-untyped]
 
 from uwtools.config.formats.base import Config
 from uwtools.config.formats.yaml import YAMLConfig
+from uwtools.config.validator import bundle, internal_schema_file, validate_internal
 from uwtools.exceptions import UWConfigError
 from uwtools.logging import log
 from uwtools.scheduler import JobScheduler
@@ -54,6 +56,20 @@ class _ECFlowDef:
 
     def __str__(self):
         return self._d.__str__()
+
+    def validate(self) -> bool:
+        """
+        Validate the ecFlow config against the internal ecFlow schema.
+
+        :return: ``True`` if the config conforms to the schema.
+        :raises: UWConfigError if validation fails.
+        """
+        validate_internal(
+            schema_name=EC.ecflow,
+            desc="ecflow config",
+            config_data={EC.ecflow: self._config},
+        )
+        return True
 
     def write_ecf_scripts(self, path: Path | str) -> None:
         """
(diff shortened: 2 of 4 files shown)


src/uwtools/api/ecflow.py
```diff
@@ -55,4 +55,4 @@ def validate(
     :param stdin_ok: OK to read from ``stdin``?
     :return: ``True`` if the YAML conforms to the schema, ``False`` otherwise.
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: The misleading `"False otherwise"` wording was inherited from #872 when `validate_file()` was a no-op stub; now that the wrapper routes through `validate_internal`, schema failures raise `UWConfigError` and there's no `False` return path.

Applied a minor variation of the suggestion to match the `:raises:` style used elsewhere in `uwtools.ecflow` (e.g. the module-level `validate())`:

```
:return: ``True`` if the YAML conforms to the schema.
:raises: UWConfigError if validation fails.
```

src/uwtools/ecflow.py
```diff
@@ -383,5 +385,27 @@ def realize(
...
+def validate(config: dict | YAMLConfig | Path | None = None) -> bool:
+    """
+    Validate an ecFlow config against the internal ecFlow schema.
+
+    :param config: A ``dict``, a ``YAMLConfig``, a path to a YAML file, or ``None``
+        (``None`` => read ``stdin``).
+    :return: ``True`` if the config conforms to the schema.
+    :raises: UWConfigError if validation fails.
+    """
+    kwargs: dict = {"schema_name": EC.ecflow, "desc": "ecflow config"}
```
```suggestion
    kwargs: dict = {"schema_name": EC.ecflow, "desc": "ecFlow config"}
```

src/uwtools/api/ecflow.py
```diff
@@ -53,6 +53,7 @@ def validate(
 
     :param yaml_file: Path to YAML file (``None`` or unspecified => read ``stdin``).
     :param stdin_ok: OK to read from ``stdin``?
-    :return: ``True`` if the YAML conforms to the schema, ``False`` otherwise.
+    :return: ``True`` if the YAML conforms to the schema.
+    :raises: UWConfigError if validation fails.
```
```suggestion
    :raises: ``UWConfigError`` if validation fails.
```
For nicer rendering in ReadTheDocs.

src/uwtools/ecflow.py
```diff
@@ -383,5 +385,27 @@ def realize(
     return str(suite)
 
 
-def validate_file(yaml_file: Path | None = None) -> bool:  # noqa: ARG001
+def schema() -> dict:
+    """
+    Return the ecFlow module's internal schema.
+    """
+    path = internal_schema_file(schema_name=EC.ecflow)
+    return bundle(json.loads(path.read_text()))
```
maddenp-cu: I might be missing something, but I'm not sure we need this. I'm guessing that this flowed from this statement in the issue:

> This will likely be very similar to the implementation in the base driver.

For the drivers, we expect that some users will be doing development on the underlying components themselves, and so might be changing them in ways that require changes to their UW YAML config and, therefore, changes to their config schemas. So we've given them a way to extract the driver's built-in schema (a `--show-schema` flag at the driver-mode level, e.g. `uw  mpas --show-schema`) so that they can modify their copy of it, and then a `--schema-file <path>` option (e.g. `uw mpas run --schema-file <path> ...`) to override use of the built-in schema to use their modified version.

This scenario seems to me less likely with ecFlow, so I'm not sure users will need to extract its schema -- and I don't see that we're giving them an option to override the built-in schema anyway.

So, unless I'm overlooking a requirement, this function and its corresponding test might not be necessary.

(This might raise an interesting question as to whether users should be able to extract _any_ schema from the package without visiting the source code, but this seems out-of-scope at the moment.)

src/uwtools/tests/test_ecflow.py
```diff
@@ -684,10 +685,58 @@ def test_ecflow_realize__write_scripts(capsys, assets):
...
+    ):
+        actual = ecflow.schema()
+    assert actual == expected
+    bundle.assert_called_once_with(expected)
+
+
+def test_validate__path(tmp_path):
+    yaml_file = tmp_path / "ecflow.yaml"
+    yaml_file.write_text("ecflow: {}\n")
+    assert ecflow.validate(yaml_file)
```
maddenp-cu: Does this work because we didn't really resolve [this question](https://github.com/ufs-community/uwtools/pull/863/changes/BASE..a94b360193cd7909a19fcd8baf2d164f0d5157bb#r3093609231)? I think we're using the real schema here, and we're successfully validating a no-op workflow. Is that what we want and/or do we need to open a new issue to improve on this soon? It's good that this test is telling the truth about the current behavior if this is how it is, and that it will break -- leading us to improve it -- if we improve the schema. Just want to make sure we're all on the same page about this.

maddenp-cu: In the tests here, since we have `minimal_config` and `tmp_path` fixtures available, it might be even easier to dump to a file and really validate it, without any mocking, something like

```
path = tmp_path / "config.yaml"
YAMLConfig(minimal_config).dump(path)
assert ecflow.validate(path)
```

for this one. And I think we need to check the failure case where `UWConfigError` is raised on a bad config, via `with raises(UWConfigError) as e:` after `from pytest import raises`, and then making an assertion about the string content of `e`.

Similarly for the `YAMLConfig` and `dict` tests: I think we can easily just really do the validation. We don't need to test the failure case for every combination -- just one bad config with one of the path/config/dict inputs.

I think the mocking at the API level is appropriate, but the fact that we can mock there depends on us doing thorough testing here. We might be able to convince ourselves that the mocking here is ok, too, as the even-deeper library code is tested, but it seems easy enough to not mock here and not have to use inductive reasoning to convince ourselves of the tests' validity. That's how my brain works, anyway.

EDIT: Sorry, never mind about the failure-mode test, I see it below!

src/uwtools/tests/test_ecflow.py
```diff
@@ -684,10 +685,58 @@ def test_ecflow_realize__write_scripts(capsys, assets):
...
+
+
+def test_validate__stdin():
+    with patch.object(ecflow, "validate_internal") as validate_internal:
+        assert ecflow.validate()
+    validate_internal.assert_called_once_with(
+        schema_name="ecflow",
+        desc="ecflow config",
+        config_path=None,
+    )
```
maddenp-cu: We can also do a "real" test here without mocking `validate_external` but by mocking `stdin`. I forget the exact mechanism, but if you search the codebase for `StringIO`, IIRC that's used by other tests that read from `stdin`. Please let me know if you don't find it.

maddenp-cu (changes requested):

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Good catch.

src/uwtools/ecflow.py

src/uwtools/ecflow.py
AnnaSmoot-NOAA: I was a little on the fence about this when I did and I agree with you. You're right, ecFlow doesn't expose  `--show-schema` CLI flag or `--schema-file` override, so this function isn't reachable from outside the package and serves no current use case. Removed `schema()`, its test, and the now-unused `json` / `bundle` / `internal_schema_file` imports.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Yes, you're right.  `{"ecflow": {}}` passes the real schema today because the schema doesn't require any suites at the ecflow root level. The test is intentionally exercising the real schema with the minimum input it currently accepts, so if we later tighten the schema, like requiring a minimum of one suite, the test will fail and need updating. 

I think improving the schema would be a separate PR. Do you want me to add it as a follow-up issue to handle after this PR is merged along with the other two follow-ups I have noted?

maddenp-cu: I agree that improving the schema is out-of-scope here. Due to my ignorance about ecFlow, I think you and @christinaholtNOAA would have to decide whether a schema with no suites is something we'd ever want to allow. If so, maybe the schema is ok. If not, I think we should open an Issue so we don't forget.

AnnaSmoot-NOAA: That's a good idea. I like it! I've dropped the mocking from `test_validate__path`, `test_validate__dict`, and `test_validate__yamlconfig` so they exercise the real `validate_internal` path against the real schema.

I also switched `test_validate__path` to use `YAMLConfig(minimal_config).dump(path)` per your suggestion — cleaner than hand-writing YAML text. 

I also restructured `test_validate_invalid` to the explicit `with raises(UWConfigError) as e` with an assertion against `str(e.value)`.

src/uwtools/ecflow.py
```diff
@@ -383,5 +385,27 @@ def realize(
...
+
+
+def validate(config: dict | YAMLConfig | Path | None = None) -> bool:
```
christinaholtNOAA, line 387: In PR #872, [Paul brought up](https://github.com/ufs-community/uwtools/pull/872#discussion_r3113227230) adding support for validating `dict` and `YAMLConfig` at the API level. We [also discussed](https://github.com/ufs-community/uwtools/pull/872#discussion_r3114029236) keeping the interface at this level simpler because it isn't exposed to the end user.

What are your thoughts on keeping this as `validate_file` so it's consistent with that goal and updating the API with that support?

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: I see. I used `StringIO` along with `patch.object(sys, "stdin", new=sio)` pattern (with `_stdinproxy.cache_clear()` like the YAML tests do). `test_validate__stdin` now feeds a real, serialized `minimal_config` through `sys.stdin` and exercises `validate_internal` against the real schema. No mocks are left in the validate suite.

src/uwtools/ecflow.py
maddenp-cu: If we accept a `YAMLConfig` or a `dict` via the API, don't we need to accept them here too? I agree that we don't need to accept `str` here because the API can do the `str` -> `Path` conversion for us, but what could the API do with a `dict` or `YAMLConfig` except pass them through? I guess it could write them to a file and then pass that file's `Path` here, but that seems convoluted to me, and we'd have to take care to clean up the temporary file. Maybe I got confused somewhere along the way, but the current implementation here looks about like what I'd expect.

christinaholtNOAA (commented):
I'm good to go with my one concern. I don't want to hold up any progress in my absence.

src/uwtools/api/ecflow.py
maddenp-cu, line 48: This PR adds support for `YAMLConfig` and `dict` ecFlow config objects at the library level, but users will not be able to make use of that unless we also accept those types here in the API. So I think the signature here should maybe be
```
config: Path | str | YAMLConfig | dict | None
```
Then maybe this?
```
return _validate(
    _ensure_data_source(_str2path(yaml_file), stdin_ok)
    if isinstance(config, (Path, str, NoneType))
    else config
)
```

maddenp-cu (changes requested):

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Ok. I'm going to open a few issues as follow-ups to this PR. If we decide we're ok with an empty suite, we can just close the issue.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: Ok. I updated the API signature: `config: YAMLConfig | dict | Path | str | None` (renamed `yaml_file` → `config` to match the core), and updated `test_api_ecflow_validate` to match. Skipped the explicit `isinstance` branch since `_str2path` is a no-op on non-str types and `_ensure_data_source` only enforces the `None` + `stdin_ok=False` case, so the single line return `_validate(config=_ensure_data_source(_str2path(config), stdin_ok))` handles all input types cleanly.

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Done and resolved in above comment: 

I updated the API signature: config: YAMLConfig | dict | Path | str | None (renamed yaml_file → config to match the core), and updated `test_api_ecflow_validate` to pass `config=`. the original one-liner handles all input types correctly:
`return _validate(config=_ensure_data_source(_str2path(config), stdin_ok))`

src/uwtools/api/ecflow.py
```diff
@@ -45,14 +45,16 @@ def realize(
...
-    :param yaml_file: Path to YAML file (``None`` or unspecified => read ``stdin``).
+    :param config: An ecFlow config as a ``dict``, a ``YAMLConfig``, a path to a YAML file
+        (``Path`` or ``str``), or ``None`` (read ``stdin``).
     :param stdin_ok: OK to read from ``stdin``?
-    :return: ``True`` if the YAML conforms to the schema, ``False`` otherwise.
+    :return: ``True`` if the config conforms to the schema.
+    :raises: ``UWConfigError`` if validation fails.
     """
-    return _validate(yaml_file=_ensure_data_source(_str2path(yaml_file), stdin_ok))
+    return _validate(config=_ensure_data_source(_str2path(config), stdin_ok))
```
maddenp-cu, line 60: Does this properly handle the case where `config` is a `YAMLConfig` or a `dict`? And I'm not seeing that we're testing either of those cases yet. Maybe I'm looking too soon and there's something that hasn't been pushed yet?

AnnaSmoot-NOAA: Thanks for catching that!

Coverage now (split out from the original single test):

API: `test_api_ecflow_validate__path / __str / __dict / __yamlconfig / __stdin / __no_stdin_no_config`
Core: `test_validate__path / __dict / __yamlconfig / __stdin / __invalid`
So `uwtools.api.ecflow.validate(config=my_dict)` and `validate(config=my_yamlconfig)` are both end-to-end tested.

src/uwtools/api/ecflow.py
maddenp-cu: Sorry, I asked about this in a separate comment but see your statement here that all the input types are handled correctly -- which kind of surprises me, but I'm glad that it's true! If we haven't yet, I do think we should demonstrate all accepted types in the API unit test. I need to look at the latest updates more carefully after I get out of meetings and stop multitasking.

AnnaSmoot-NOAA: Understood and appreciated.

maddenp-cu (approved):
All looks good, thank you!

elcarpenterNOAA (approved):

PR #880, opened 2026-04-24
Improvements to values-needed behavior

maddenp-cu:
**Synopsis**

- Update `uw config realize --values-needed` (and its API equivalent) behavior to report only on incomplete content (i.e. stop reporting values that are completely rendered), and to do so for both keys and values containing unrendered Jinja2 content.
- Update `uwtools.api.config.realize()` so that, when `values_needed=True`, the call's return value is a `dict` with `keys` and `values` items, each of which is a list of keypaths leading to incomplete keys and values, respectively. The caller can use these keypaths as they like, for example by following the chain of `dict` keys / `list` indexes to incomplete content and reporting on it.
- Update docs, notebooks, docstrings, and tests as needed.

See the update docs, notebooks, and unit tests for examples of the new behavior.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/config/formats/base.py
```diff
@@ -23,6 +23,8 @@
 from uwtools.logging import INDENT, MSGWIDTH, log
 from uwtools.utils.file import str2path
 
+NIL = object()
```
maddenp-cu, line 20: Config values might intentionally be `None` values, so checking against `None` to see if an argument is uninitialized in `incomplete()`, below, does not work. This value is guaranteed to be unique during the lifetime of one invocation of the Python interpreter, so use it instead of `None` to signal an uninitialized argument.

src/uwtools/tests/conftest.py
```diff
@@ -49,3 +49,10 @@ def utc(*args, **kwargs) -> datetime:
         return dt.replace(tzinfo=None)
 
     return utc
+
+
+@fixture
+def uwcaplog(caplog):
+    caplog.handler.setFormatter(logging.Formatter("%(message)s"))
+    caplog.set_level(logging.DEBUG)
+    return caplog
```
maddenp-cu, line 58: This fixture makes it easier to work with `caplog.text` in tests by ensuring the log messages have no `INFO` et al. or timestamp prefixes, instead consisting of just the logged message; and by setting `caplog`'s capture level to `DEBUG` so that all messages are captured.

notebooks/config.ipynb
```diff
@@ -740,18 +746,17 @@
...
+      "[2000-01-01T00:00:00]     INFO Values with unrendered content:\n",
       "[2000-01-01T00:00:00]     INFO   memo.sender_id: {{ id }}\n",
       "[2000-01-01T00:00:00]     INFO   memo.message: {{ greeting }}, {{ recipient }}!\n"
      ]
     },
     {
      "data": {
       "text/plain": [
-       "{}"
+       "{'keys': [], 'vals': [['memo', 'sender_id'], ['memo', 'message']]}"
```
maddenp-cu, line 759: Here's an example of the new return value from `realize(..., values_needed=True)`. Currently the keypaths appear in the order they were encountered in the input config. We could also opt to sort them. Both choices have benefits, but the user can sort if they like, but they can't unsort.

notebooks/fs.ipynb
```diff
@@ -546,7 +546,7 @@
       "\u001b[01;34mtmp/copier-target\u001b[0m\n",
       "├── \u001b[01;34mdata\u001b[0m\n",
       "│   ├── \u001b[00mfile2-copy.txt\u001b[0m\n",
-      "│   └── \u001b[00mfile3-copy.csv\u001b[0m\n",
+      "���   └── \u001b[00mfile3-copy.csv\u001b[0m\n",
```
elcarpenterNOAA: What's happening here with the ���? Is this just a GitHub display issue? There's another one on line 1481.

src/uwtools/config/formats/base.py
```diff
@@ -54,48 +56,37 @@ def __repr__(self) -> str:
...
-
-    def _characterize_values(self, values: dict, parent: str) -> tuple[list[str], list[str]]:
-        """
-        Characterize values as either complete or as incomplete template placeholders.
-
-        :param values: The dictionary to examine.
-        :param parent: Parent key.
-        :return: Lists of complete and incomplete-template-placeholder values.
-        """
+    def incomplete(
```
maddenp-cu: The new (renamed) version of this function just collects lists of keypaths to incomplete keys and values and stays out of formatting/presentation. That's now handled by the updated `_realize_values_needed()`, but since the lists are also passed back to API users now, they can do any other reporting or response they like.

elcarpenterNOAA (approved):

notebooks/fs.ipynb
maddenp-cu: Huh, I thought that was maybe Jupyter's decision to encode those characters a certain way, but I see that they actually look this way in the rendered notebook. Not good. Thanks for pointing this out, I'll find a fix.

notebooks/fs.ipynb
```diff
@@ -37,13 +37,14 @@
...
     "import os\n",
+    "import yaml\n",
     "from pathlib import Path\n",
     "from shutil import rmtree\n",
     "from freezegun import freeze_time\n",
     "from uwtools.api import fs\n",
     "from uwtools.api.logging import use_uwtools_logger\n",
-    "import yaml\n",
     "\n",
+    "os.environ[\"TREE_CHARSET\"] = \"ascii\"\n",
```
maddenp-cu, line 47: The only way I have found so to to eliminate those � characters from the `tree` output has been to use regular ASCII characters for the trees. They don't look as nice, but I think they'll have to do for now.

src/uwtools/config/support.py
```diff
@@ -208,10 +208,16 @@ class UWYAMLConvert(UWYAMLTaggedStr):
     TaggedValT = bool | datetime | dict | float | int | list | timedelta
 
     def __repr__(self) -> str:
-        return "%s %s" % (self.tag, self.converted)
+        try:
+            return "%s %s" % (self.tag, self.converted)
+        except yaml.constructor.ConstructorError:
+            return self.tagged_string
```
maddenp-cu, line 214: This change, and the parallel one to `__str__()` below, allow `UWYAMLConvert` objects to be printed/displayed even when they cannot be rendered. For example, the YAML text `!int '{{ x }}'`, parsed to a `UWYAMLConvert` object, previously would raise an exception if `x` was undefined. Now the text `!int '{{ x }}'` will be reproduced verbatim as the representation of the object. This simplifies some other things, including the code in the updated `incomplete()` function.

src/uwtools/config/formats/base.py
maddenp-cu: I relocated this method in the class and simplified it.

Issue #876, opened 2026-04-22
ecFlow schema rejects single-key `expand` blocks (`expander.minProperties: 2`)


AnnaSmoot-NOAA:
## Description

`src/uwtools/resources/jsonschema/ecflow.jsonschema` defines `expander` with `"minProperties": 2` (line 346), which causes any block that specifies only `expand` (with no sibling key) to fail validation. This blocks the `test_integration__with_expand` style of input, which uses an `expand`-only block. This is a mismatch between the ecFlow YAML pattern uwtools intends to support and the JSON Schema rules it currently enforces.

## What the test is trying to do and what tests imply is valid. 

```yaml
ecflow:
  some_suite:
    expand:
      var: [1, 2, 3]
```

This is rejected by `uwtools.ecflow.validate()` even though it is the intended pattern exercised by `test_integration__with_expand.`

What the test is trying to do above means generate multiple variants of this node by expanding over `var`. Currently, In `src/uwtools/resources/jsonschema/ecflow.jsonschema` the definition `"$defs"."expander"` is:

- `"required": ["expand"]` → so expand must be present (fine)
- but also `"minProperties": 2` → the object must have at least two keys (this is the problem)
- and it has `patternProperties` like `^family_.+$`, `^task_.+$`, etc. which represent “sibling keys” (children) that can exist alongside expand.

So an object like: 
```
expand:
  var: [1, 2, 3]
```
has only one property (`expand`) → it violates `minProperties: 2` → schema validation rejects it, which then makes `uwtools.ecflow.validate()` reject it too. As shown here: 
```
"expander": {
  "minProperties": 2,
  ...
  "properties": { "expand": ... },
  "required": ["expand"],
  "type": "object"
}
```
`minProperties: 2` encodes an assumption: an expander block should have `expand` plus at least one “real content” sibling key (like a `family_*`, `task_*`, `families_*`, `tasks_*`, etc.), so that expanding actually expands something. `test_integration__with_expand` seems to use (or expects support for) an “expand-only” block, meaning: allow `expand` even if there’s no sibling key in that same mapping. 

## Acceptance criteria

Decide and implement one of:
1) Drop/lower `minProperties` on `expander` so an `expand`-only block validates. This relaxes the schema  matches the test pattern. 
2) Keep strictness of  `minProperties: 2` and update the test/docs to require a sibling key alongside `expand`.
Either way, schema and integration tests must agree.

## References
PR #873 (follow-up #2 in PR description)

[...]

PR #886, opened 2026-04-27
Validate config before building the workflow.

### LINKED ISSUE #884: ecFlow config should be validated upon instantiation
**Description**

The validation of a config for ecFlow should be done in the `__init__` method of `_ECFlowDef` to ensure the config is valid before the workflow building begins.

**Acceptance Criteria**

- [ ] Validation is included as part of the `__init__` method.
- [ ] All existing tests pass, which may mean updating test configs throughout the ecflow tests.

christinaholtNOAA:
**Synopsis**

Fixes #884 and #876

Adds call to `validate` in the `_ECFlowDef.__init__` method. Updates the failing configs in the tests.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -25,7 +25,6 @@ from ecflow import (  # type: ignore[import-untyped]
     Task,
 )
 
-from uwtools.config.formats.base import Config
 from uwtools.config.formats.yaml import YAMLConfig
 from uwtools.config.validator import validate_internal
 from uwtools.exceptions import UWConfigError
@@ -45,10 +44,11 @@ class _ECFlowDef:
     Generate an ecFlow definition file from a YAML config.
     """
 
-    def __init__(self, config: dict | Config | Path | None = None) -> None:
+    def __init__(self, config: dict | YAMLConfig | Path | None = None) -> None:
         self._scripts: dict[Path, str] = {}
-        cfgobj = config if isinstance(config, Config) else YAMLConfig(config)
-        cfgobj = cfgobj.dereference()
+        cfgobj = config if isinstance(config, YAMLConfig) else YAMLConfig(config)
+        cfgobj.dereference()
+        validate(config)
         self._config = cfgobj.data[EC.ecflow]
         self._scheduler = self._config.get(STR.scheduler)
         self._d = Defs()
@@ -252,7 +252,7 @@ class _ECFlowDef:
             else None
         )
         execution = config[STR.execution]
-        cmd = execution.get(EC.jobcmd)
+        cmd = execution.get(EC.jobcmd, "")
         es = self._ecflowscript(
             execution=[cmd],
             manual=config.get(EC.manual, f"Script to run {task.name()}"),
@@ -316,7 +316,7 @@ class _ECFlowDef:
         pre_includes = pre_includes or []
         post_includes = post_includes or []
         directives = scheduler.directives if scheduler else ""
-        initcmds = scheduler.initcmds if scheduler else []
+        initcmds = scheduler.initcmds if scheduler else [""]
         rs = dedent(template).format(
             directives="\n".join(directives),
             envcmds="\n".join(envcmds or []),
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -252,8 +252,11 @@ class _ECFlowDef:
             else None
         )
         execution = config[STR.execution]
-        assert "jobcmd" in execution, f"execution block for {task.name()} must include 'jobcmd'"
-        cmd = execution[EC.jobcmd]
+        try:
+            cmd = execution[EC.jobcmd]
+        except KeyError as e:
+            msg = "The execution block for %s must include 'jobcmd'" % task.name()
+            raise UWConfigError(msg) from e
         es = self._ecflowscript(
             execution=[cmd],
             manual=config.get(EC.manual, f"Script to run {task.name()}"),
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -48,7 +48,7 @@ class _ECFlowDef:
         self._scripts: dict[Path, str] = {}
         cfgobj = config if isinstance(config, YAMLConfig) else YAMLConfig(config)
         cfgobj.dereference()
-        validate(config)
+        validate(cfgobj)
         self._config = cfgobj.data[EC.ecflow]
         self._scheduler = self._config.get(STR.scheduler)
         self._d = Defs()
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -252,7 +252,8 @@ class _ECFlowDef:
             else None
         )
         execution = config[STR.execution]
-        cmd = execution.get(EC.jobcmd, "")
+        assert "jobcmd" in execution, f"execution block for {task.name()} must include 'jobcmd'"
+        cmd = execution[EC.jobcmd]
         es = self._ecflowscript(
             execution=[cmd],
             manual=config.get(EC.manual, f"Script to run {task.name()}"),
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -267,7 +267,7 @@ class TestECFlowDef:
 
     def test__init__missing_ecflow_key(self):
         config: dict = {"not_ecflow": {}}
-        with raises(KeyError):
+        with raises(UWConfigError):
             _ECFlowDef(config=config)
 
     # _add_workflow_components tests
@@ -608,12 +608,15 @@ class TestECFlowDef:
                     "family_prep": {
                         "task_setup": {
                             "trigger": "1==1",
-                        }
+                            "script": {
+                                "execution": {"executable": "prep.exe", "jobcmd": "echo running"},
+                            },
+                        },
                     },
                     "task_run": {
                         "trigger": "/test/prep/setup == complete",
                         "script": {
-                            "execution": {"jobcmd": "echo running"},
+                            "execution": {"jobcmd": "echo running", "executable": "run.exe"},
                         },
                     },
                 }
@@ -642,6 +645,7 @@ class TestECFlowDef:
                 "suite_ensemble": {
                     "tasks_member_{{ ec.MEM }}": {
                         "expand": {"MEM": ["01", "02", "03"]},
+                        "script": {"execution": {"executable": "run me", "jobcmd": "hello"}},
                     }
                 }
             }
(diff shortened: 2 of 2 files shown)


elcarpenterNOAA (approved):

src/uwtools/ecflow.py
```diff
@@ -252,7 +252,7 @@ def _create_ecf_script(self, config: dict, task: Task) -> None:
             else None
         )
         execution = config[STR.execution]
-        cmd = execution.get(EC.jobcmd)
+        cmd = execution.get(EC.jobcmd, "")
```
maddenp-cu: Is there a use case where it makes sense to have an empty `jobcmd`? Is it when we're not creating scripts at all? It looks like this becomes `{execution}` in the script in `_ecflowscript()`, and I'm not sure that script makes sense with an empty `{execution}`. Does it? If we think this doesn't make sense but that we've somehow previously prevented the nonsensical thing from happening, would an `assert` be appropriate here to document/enforce that understanding?

AnnaSmoot-NOAA: Yes, I was thinking it might be good to check explicitly for or asserting a non-empty `jobcmd` unless we have a reason for it to be empty. If we want to allow and empty case we should add a test to cover that.

christinaholtNOAA: Yes, I do think we do want to enforce the existence of a `jobcmd`. I like the idea of an `assert` to do it.

AnnaSmoot-NOAA (commented):
I think this is good to go once there is an `assert` test for an empty  `jobcmd`.

src/uwtools/ecflow.py
```diff
@@ -252,7 +252,8 @@ def _create_ecf_script(self, config: dict, task: Task) -> None:
             else None
         )
         execution = config[STR.execution]
-        cmd = execution.get(EC.jobcmd)
+        assert "jobcmd" in execution, f"execution block for {task.name()} must include 'jobcmd'"
```
maddenp-cu: I'm sorry, I think I failed to follow along with some of the changes here as I should have.

I realize now the the updated `execution-serial` and `execution-parallel` schemas now have an optional `jobcmd` property. But does that property make sense in any context other than the `ecflow` tool? I'm guessing not, and if not, I think maybe we should roll back that change and move that property to new schemas (maybe defined in `ecflow.jsonschema`) and enforce (maybe via an `allOf` block) both the appropriate underlying `execution-*` schema _and_ a required `jobcmd` proeprty. Like inheritance, but for schemas, e.g. the `ecflow` `execution-serial` schema would enforce the base `execution-serial` schema plus the `jobcmd` requirement.

Then I think the assert here would make sense: It should not be possible for users to trigger an `AssertionError`, so we should have prevented them from getting this far, probably via a schema check. But then `jobcmd` has to be _required_ in the schema and of course it can't be if we add it to the existing `execution-serrial` (likewise `-parallel`) schema, because it's not appropriate for other use cases.

Another way of looking at it is that every driver that uses e.g. `execution-serial` would now accept, in its config, a `jobcmd` property that would be ignored. That would be confusing. And since driver users can dump a driver's schema for modification, they'd be mystified by this undocumented `jobcmd` property.

Sorry that I'm only now realizing this -- unless I'm also misunderstanding, which would be less work to fix.

maddenp-cu (changes requested):

src/uwtools/ecflow.py
maddenp-cu: Also, I'd be fine with approving this and then circling back to improve the schemas in a follow-up PR if we don't want to complicate this one. This PR does achieve the goal of doing the schema check.

christinaholtNOAA: I think it could be useful for Rocoto workflows, too. Right now, all of the Rocoto workflow definition files define their run-time resources by Jinja2-reference to the appropriate `execution` block. The one task-specific piece of information that's left in the Rocoto task definition block is the `command` -- how to run the task. To change that, a user actually has to go to that separate workflow YAML to change the run command. With this addition, the app can similarly point to the `execution.jobcmd` block and let users have full control from the top-level defaults file.

Here's an example from MPAS App:

```
    task_ungrib_ics:
      command:
        cyclestr:
          value: 'source &MPAS_APP;/load_wflow_modules.sh &PLATFORM; && &MPAS_APP;/scripts/ungrib.py
            -c &EXPERIMENT_CONFIG;
            --cycle @Y-@m-@dT@H:@M:@S
            --key-path prepare_grib_ics'
      account: "{{ platform.account }}"
      exclusive: "true"
      join:
        cyclestr:
          value: '&LOGDIR;/{{ jobname }}_@Y@m@d@H.log'
      walltime: "{{ prepare_grib_ics.ungrib.execution.batchargs.walltime }}"
      nodes: "{{ prepare_grib_ics.ungrib.execution.batchargs.nodes }}:ppn={{ prepare_grib_ics.ungrib.execution.batchargs.tasks_per_node }}"
      dependency:
        taskdep:
          attrs:
            task: get_ics_data
```

I think it could be nice to have all of the information about how to run a given task together in the execution block. If we keep this enabled for any driver via the execution block, we can reference it from either Rocoto or ecFlow YAML definitions, getting us one step closer to interchangeability.

maddenp-cu: I'm not sure I understand quiet yet. In that example, what would the value of `jobcmd` in the `execution` block be?

christinaholtNOAA: Great question. It looks a bit messier than I want it to be, but:

```
runcmd: 'source &MPAS_APP;/load_wflow_modules.sh &PLATFORM; && &MPAS_APP;/scripts/ungrib.py
            -c &EXPERIMENT_CONFIG;
            --cycle @Y-@m-@dT@H:@M:@S
            --key-path prepare_grib_ics'
```

All of these XML entity references _could_ be reformulated as Jinja2 expressions or environment variables for apps that want to support interoperability with both Rocoto and ecFlow. The `--cycle` argument value can be factored out to an environment variable, as well, removing the Rocoto-specific syntax.

maddenp-cu: This strikes me as maybe still too Rocoto-specific (e.g. the XML entity references, even if the cycle string were factored out) to belong in a driver's config block. What would we put there if we wanted to support both Rocoto and ecFlow workflows in the same app, which seems like something we want to be able to do in general? Maybe it can be parameterized further and be really generic?

Am I right in starting to interpret the proposed `jobcmd` as containing instructions for running `uw` itself? That feels self-referential in a way that's making me uneasy, but I'm not sure I can explain/justify that feeling yet. I might be misunderstanding something.

Maybe I should sleep on it, it's been a long day.

AnnaSmoot-NOAA: I'm trying to make sure I understand this. `jobcmd` should be for ecFlow only schemas and workflows, right? So it should be in an ecFlow schema block and use the `allOf` to extend the base execution schema for ecFlow? 

And either now or later, we should have a runtime check (`assert`) too. 

To make it more ecFlow friendly, can we just use ecFlow syntax like: 
1) Shell-like option and setting environment variables: 
```
jobcmd: |
  source $MPAS_APP/load_wflow_modules.sh $PLATFORM &&
  $MPAS_APP/scripts/ungrib.py \
    -c $EXPERIMENT_CONFIG \
    --cycle $CYCLE_TIME \
    --key-path prepare_grib_ics
```
or 2) Jinja2-style: 
```
jobcmd: |
  source {{ MPAS_APP }}/load_wflow_modules.sh {{ PLATFORM }} &&
  {{ MPAS_APP }}/scripts/ungrib.py \
    -c {{ EXPERIMENT_CONFIG }} \
    --cycle {{ CYCLE_TIME }} \
    --key-path prepare_grib_ics
```
or 3) ecFlow env vars style: 
```
jobcmd: |
  source %MPAS_APP%/load_wflow_modules.sh %PLATFORM% &&
  %MPAS_APP%/scripts/ungrib.py \
    -c %EXPERIMENT_CONFIG% \
    --cycle %ECF_DATE% \
    --key-path prepare_grib_ics
```

Based on my limited knowledge of uw-tools, I would lean towards option 3 or option 2.

christinaholtNOAA: @maddenp-cu I see what you mean about being self-referential. This is basically a "how do you run _**me**_ with this app" feature. That's a slight departure from "how do you run **_my executable_** with the resources configured for this app" that exists in the rest of the execution block. In the case of the `jobcmd`, it could be a call to `uw`, or it could be several layers of scripts removed from `uw`, depending on the app.

@AnnaSmoot-NOAA These are all valid options for the `jobcmd`, and the version that is included as a default will be totally up to the app developers. The `uwtools` implementation should not be opinionated at all about what's included there. Also, it's not _just_ for ecFlow. It can be used for defining the Rocoto command, too.

In Option 2, everything with the Jinja2 expressions would be fine except for `CYCLE_TIME`, which won't be available until run time, so would need to be an environment variable (as in your Option 1), a Rocoto cyclestring (as in my copied example above), or an ecFlow variable (as in your Option 3).

maddenp-cu: I think both of your questions are what we're trying to figure out. I think @christinaholtNOAA is suggesting that `jobcmd` may be useful to both ecFlow _and_ Rocoto and that it would then be beneficial to have it specified in the `execution:` block, and accepted by the existing schemas for that block (serial and parallel variants).

If you wanted a `jobcmd` value that could be used either with ecFlow or Rocoto workflows, I think your option 3 would be out because it wouldn't support Rocoto. Option 1 could work if your `envcmds` added the necessary `export` command to the runscript.

One could argue that which workflow manager to use if up to an application, and if they want to support only ecFlow they could provide values like option 3, and if they wanted to be agnostic they could go with options 1 or 2. So, I guess workflow-manager flexibility is not necessarily an argument against `jobcmd` being part of a driver config via the `execution:` block.

I'm still uneasy about this because it doesn't seem like the value of `jobcmd` has anything to do with the driver specifically. But, then, I guess neither does the Slurm account name the driver's job is to be run under, or many/any of the other things set under `execution:`. So, again, not really an argument against it.

Maybe I'm coming around to the idea. Maybe a different, more-descriptive name would help. What exactly _is_ this? I think the fact that we're doing the documentation last has contributed to my confusion: There's no cogent description of this value yet, and the examples in the tests don't show a realistic, enlightening use case -- which is fine! Could we propose here a description that would appear in the docs and then see if this is the best setting name we can think of?

BTW, if this value is optional in the schema (seems like it will have to be) and only enforced at the ecFlow or Rocoto level where it's used, we will after all want to raise an exception. Sorry about the confusion on that point.

maddenp-cu (approved):

AnnaSmoot-NOAA (approved):

PR #891, opened 2026-05-01
Ecflow tool cli

AnnaSmoot-NOAA:
-->

**Synopsis**

This PR adds a new `uw ecflow` CLI tool for working with ecFlow suite definitions, enabling users to realize (generate) and validate ecFlow configurations from UW YAML files.

**Key Features:**
- **New CLI commands:**
  - `uw ecflow realize` - Converts UW YAML config to ecFlow suite definition and ecf scripts
  - `uw ecflow validate` - Validates ecFlow YAML configuration against JSON schema

- **Flexible execution configuration:** The ecFlow execution blocks now accept either `executable` (standard UW format) or `incantation` (ecFlow-specific format), with `incantation` taking precedence when both are provided. This allows users to:
  - Use the standard UW `executable` field for consistency across drivers
  - Use ecFlow's native `incantation` field when preferred
  - Migrate existing ecFlow configurations
 
 **Example Usage:**
```
# Validate an ecFlow config
uw ecflow validate --config-file workflow.yaml

# Generate suite definition and scripts
uw ecflow realize --config-file workflow.yaml --output-dir /path/to/output
```
 
 
**Type**

- [X] Documentation

**Impact**

**Checklist**

- [X] I have reviewed the documentation and have made any updates necessitated by this change., backup, or binary files.
- Please remove commented-out code.
- Please ensure code, config files, etc., contain no hardcoded paths.
- Please format code snippets in PR description/comments with ```code block``` or `inline code`.
- Please consider adding your own review comments to guide other reviewers.

-->


src/uwtools/cli.py
```diff
@@ -21,17 +19,34 @@
 import uwtools.api
 import uwtools.api.config
 import uwtools.api.driver
+import uwtools.api.ecflow
 import uwtools.api.execute
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Ok. This change seems reasonable to make the CLI more usable, allows optional dependency, and clearer error messaging.

src/uwtools/cli.py
```diff
@@ -21,17 +19,34 @@
...
+# --- workaround for logging.py shadowing stdlib logging ---
+# Remove "uwtools.logging" from sys.modules if present
+if "uwtools.logging" in sys.modules:  # pragma: no cover
+    del sys.modules["uwtools.logging"]
+
+# Force import of stdlib logging before anything else
+builtins.std_logging = logging  # type: ignore[attr-defined]
+
+# Dynamically import log from uwtools.logging
+log = __import__("uwtools.logging", fromlist=["log"]).log
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
```diff
@@ -1352,6 +1461,8 @@ def _parse_args(raw_args: list[str]) -> tuple[Args, Checks]:
         STR.fs: partial(_add_subparser_fs, subparsers),
         STR.rocoto: partial(_add_subparser_rocoto, subparsers),
         STR.template: partial(_add_subparser_template, subparsers),
+        STR.ecflow: partial(_add_subparser_ecflow, subparsers),
+        STR.ecflow: partial(_add_subparser_ecflow, subparsers),
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/setup.py
```diff
@@ -42,7 +42,7 @@
     kwargs["install_requires"] = [
         pkg.replace(" ==", "==")
         for pkg in meta["packages"]["run"]
-        if not re.match(r"^python .*$", pkg)
+        if not re.match(r"^python .*$", pkg) and not pkg.strip().startswith("ecflow")
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: @maddenp-cu @christinaholtNOAA What do you guys think of this? 

Currently, `setup.py` excludes `ecflow` from `install_requires`, so it's an optional dependency. `cli.py` has lazy import with error handling so the CLI doesn't crash if ecflow is missing. 

Copilot is suggesting to remove the filter in `setup.py` to make `ecflow` a required dependency and have it auto-installed with uw-tools. 

Which approach do we think would be best?

src/uwtools/ecflow.py
```diff
@@ -252,11 +263,13 @@ def _create_ecf_script(self, config: dict, task: Task) -> None:
...
-        try:
-            cmd = execution[EC.incantation]
-        except KeyError as e:
-            msg = "The execution block for %s must include 'incantation'" % task.name()
-            raise UWConfigError(msg) from e
+        cmd = execution.get(EC.incantation) or execution.get(STR.executable)
+        if not cmd:
+            msg = (
+                "The execution block for %s must include 'incantation' or 'executable'"
+                % task.name()
```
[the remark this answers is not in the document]

src/uwtools/cli.py
```diff
@@ -68,35 +64,33 @@ def main() -> None:
...
+            STR.mpassit: partial(_dispatch_to_driver, STR.mpassit),
+            STR.orog: partial(_dispatch_to_driver, STR.orog),
+            STR.orog_gsl: partial(_dispatch_to_driver, STR.orog_gsl),
+            STR.schism: partial(_dispatch_to_driver, STR.schism),
+            STR.sfc_climo_gen: partial(_dispatch_to_driver, STR.sfc_climo_gen),
+            STR.shave: partial(_dispatch_to_driver, STR.shave),
+            STR.ungrib: partial(_dispatch_to_driver, STR.ungrib),
+            STR.upp: partial(_dispatch_to_driver, STR.upp),
+            STR.upp_assets: partial(_dispatch_to_driver, STR.upp_assets),
+            STR.ww3: partial(_dispatch_to_driver, STR.ww3),
```
elcarpenterNOAA: What was the reason for this change? I personally think the previous version was cleaner and easier to read.

src/uwtools/cli.py
```diff
@@ -308,6 +373,21 @@ def _dispatch_config_validate(args: Args) -> bool:
...
+    """
+    actions = {
+        STR.compare: _dispatch_config_compare,
+        STR.compose: _dispatch_config_compose,
+        STR.realize: _dispatch_config_realize,
+        STR.validate: _dispatch_config_validate,
+    }
+    return actions[args[STR.action]](args)
+
+
```
elcarpenterNOAA: This should be moved back to its original location. This file is organized by `# Mode ...` sections, with the appropriate parser and dispatch functions grouped together under each mode. The ecflow tool should have its own `# Mode ecflow` section, with all of the ecflow-specific functions grouped alphabetically underneath it.

src/uwtools/cli.py
```diff
@@ -32,6 +28,11 @@
 from uwtools.strings import FORMAT, STR
 from uwtools.utils.file import get_config_format, resource_path
 
+"""
+Modal CLI.
+"""
+
+
```
elcarpenterNOAA: Can we move this back to where it was? Down here, it's no longer treated as a module docstring.

src/uwtools/cli.py
```diff
@@ -46,11 +47,6 @@ def main() -> None:
     """
     Main entry point.
     """
-
-    # Silence logging initially, then process the command-line arguments by parsing them. Run all
-    # defined checks for the appropriate [sub]mode. Reconfigure logging after quiet/verbose choices
-    # are known, then dispatch to the [sub]mode handler.
-
```
elcarpenterNOAA, line 53: Why was this removed? I think it still adds value.

docs/sections/user_guide/cli/tools/ecflow/Makefile
```diff
@@ -0,0 +1,26 @@
+.PHONY: all clean
```
christinaholtNOAA: This file should be a copy of the generic Makefile included with the other tools at the same level.

src/uwtools/resources/jsonschema/execution-parallel.jsonschema
```diff
@@ -1,5 +1,17 @@
...
+      "required": [
+        "executable"
+      ]
+    },
+    {
+      "required": [
+        "incantation"
+      ]
+    }
+  ],
```
christinaholtNOAA: This is not the intended behavior.`executable` is always required. `incantation` is optional. This should remain the case for all the drivers, which is the primary use for the execution block. The code should also reflect that this behavior should not change.

src/setup.py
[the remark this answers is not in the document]
christinaholtNOAA: For now, I'd prefer to see ecflow installed as a required dependency. It's easier, and we don't have pushback on it at the moment. If we do have some opposition to the extra install in the future, we can assess the feasibility of the optional install then. 

One issue with _not_ installing it when users think they don't need it on RDHPCS platforms is that the version installed in the module is quite old and doesn't support the features we use in uwtools.

src/uwtools/cli.py
```diff
@@ -106,6 +100,72 @@ def main() -> None:
...
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.realize, "Realize an ecFlow suite definition")
+    optional = _basic_setup(parser)
+    _add_arg_config_file(optional)
+    _add_arg_output_dir(optional)
+    return _add_args_verbosity(optional)
+
+
+def _add_subparser_ecflow(subparsers: Subparsers) -> dict[str, ActionChecks]:
```
christinaholtNOAA, line 114: This one should move above `_add_subparser_ecflow_realize` to maintain appropriate alphabetical order.

src/uwtools/cli.py
```diff
@@ -106,6 +100,72 @@ def main() -> None:
...
+
+    :param subparsers: Parent parser's subparsers, to add this subparser to.
+    """
+    parser = _add_subparser(subparsers, STR.ecflow, "Handle ecflow suite definitions")
+    _basic_setup(parser)
+    ecflow_subparsers = _add_subparsers(parser, STR.action, STR.action.upper())
+    return {
+        STR.realize: _add_subparser_ecflow_realize(ecflow_subparsers),
+        STR.validate: _add_subparser_ecflow_validate(ecflow_subparsers),
+    }
```
christinaholtNOAA, line 126: We probably want to use the more generic `subparsers` here, as is done in all other functions in this module.

src/uwtools/cli.py
```diff
@@ -548,13 +633,13 @@ def _add_subparser_rocoto(subparsers: Subparsers) -> ModeChecks:
...
     _basic_setup(parser)
-    subparsers = _add_subparsers(parser, STR.action, STR.action.upper())
+    rocoto_subparsers = _add_subparsers(parser, STR.action, STR.action.upper())
     return {
-        STR.iterate: _add_subparser_rocoto_iterate(subparsers),
-        STR.realize: _add_subparser_rocoto_realize(subparsers),
-        STR.validate_xml: _add_subparser_rocoto_validate_xml(subparsers),
+        STR.iterate: _add_subparser_rocoto_iterate(rocoto_subparsers),
+        STR.realize: _add_subparser_rocoto_realize(rocoto_subparsers),
+        STR.validate_xml: _add_subparser_rocoto_validate_xml(rocoto_subparsers),
```
christinaholtNOAA: Was there a reason behind these changes? We have used `subparsers` in the same context in every other `_add_subparser` function in this module. I'm not sure this one should be singled out.

src/uwtools/cli.py
```diff
@@ -566,7 +651,7 @@ def _add_subparser_rocoto_iterate(subparsers: Subparsers) -> ActionChecks:
     """
     parser = _add_subparser(subparsers, STR.iterate, "Iterate a Rocoto workflow")
     required = parser.add_argument_group(TITLE_REQ_ARG)
-    _add_arg_cycle(required)
+    _add_arg_cycle(required, required=True)
```
christinaholtNOAA: Why this change?

src/uwtools/cli.py
```diff
@@ -941,6 +1026,23 @@ def _add_arg_output_format(group: Group, choices: list[str], required: bool = Fa
...
+def _add_arg_output_dir(group: Group, required: bool = False) -> None:
+    """
+    Add --output-dir argument.
+
+    :param group: The argparse group to add the argument to.
+    :param required: Whether the argument is required.
+    """
+    group.add_argument(
+        _switch(STR.output_dir),
+        "-o",
```
christinaholtNOAA: I'm not sure that we want to use `-o` here as it typically denotes the specific file that comes out. This will be slightly different in that you always get a `suite.def` in this directory. Perhaps we just stick with the long name?

src/uwtools/cli.py
```diff
@@ -1352,6 +1454,7 @@ def _parse_args(raw_args: list[str]) -> tuple[Args, Checks]:
         STR.fs: partial(_add_subparser_fs, subparsers),
         STR.rocoto: partial(_add_subparser_rocoto, subparsers),
         STR.template: partial(_add_subparser_template, subparsers),
+        STR.ecflow: partial(_add_subparser_ecflow, subparsers),
```
christinaholtNOAA: This should move up to its appropriate alphabetical spot.

src/uwtools/cli.py
```diff
@@ -1459,3 +1562,7 @@ def _version() -> str:
     """
     info = json.loads(resource_path("info.json").read_text())
     return "version %s build %s" % (info["version"], info["buildnum"])
+
+
+if __name__ == "__main__":  # pragma: no cover
+    main()
```
christinaholtNOAA: I don't think we need this addition. This file should never be called directly, but its `main` function is the entry point defined by the package as what you get when you type `uw` on the command line.

src/uwtools/ecflow.py
```diff
@@ -80,8 +87,12 @@ def write_suite_definition(self, path: Path | None) -> None:
         :param path: Where to write the suite definition.
         """
         if path:
+            log.debug("Creating output directory for suite definition: %s", path)
             path.mkdir(parents=True, exist_ok=True)
             path = path / "suite.def"
+            log.debug("Writing suite definition to: %s", path)
+        else:
+            log.debug("No output path provided, writing suite definition to stdout.")
```
christinaholtNOAA:
```suggestion
            log.debug("No output path provided, writing the suite definition to stdout.")
```

test_ecflow.yaml
```diff
@@ -0,0 +1,20 @@
+ecflow:
```
christinaholtNOAA: This file should not be in the repo.

christinaholtNOAA (changes requested):
In general, I don't think the behavior for using either `incantation` or `executable` should be changed. Users really do need to provide an `incantation` separate from the `executable`. They are not interchangeable.

docs/sections/user_guide/cli/tools/ecflow/Makefile
maddenp-cu: I think it should be a symlink, right? I think that's what the others at the same level are. Then, if anything needs to change, it changes for all of the subdirectories.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
maddenp-cu, line 1: I assume that this file was generated by a `.cmd` file, such that the fact that it doesn't end with a a newline, as the GitHub diff complains about with its ⛔ icon, is accurate. But probably it should -- maybe we need to add a newline where the code creates this file. Could be worth fixing in this PR if it's an easy change.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/suite.def
```diff
@@ -0,0 +1,14 @@
...
+    task fetch
+      trigger 1==1
+    task process
+      trigger /workflow/data_prep/fetch == complete
+  endfamily
+  task run_model
+    trigger /workflow/data_prep == complete
+endsuite
+# enddef
+
```
maddenp-cu: lol, and this one has an extra trailing newline. It probably should not. Also probably worth a fix in this PR if it's easy. It's no surprise we didn't see this before, but nice that the generated documentation is pointing us to it.

docs/sections/user_guide/cli/tools/ecflow/Makefile
christinaholtNOAA: Yes! I missed that when I was poking around in my clone.

src/uwtools/tests/test_cli.py
```diff
@@ -475,6 +491,95 @@ def test_cli__dispatch_fs_report_yes(capsys):
...
+
+
+def test_cli__dispatch_ecflow_missing_ecflow():
+    """
+    Test that _dispatch_ecflow raises UWError when ecflow is not installed.
+    """
+    args = {STR.action: STR.realize}
+    with (
+        patch("builtins.__import__", side_effect=ImportError("No module named 'ecflow'")),
+        raises(UWError, match=r"ecflow is not installed\. Install it with: pip install ecflow"),
```
maddenp-cu: `uwtools` is a conda package, so we do not want to advise our users to instal packages with `pip`. I'll read further to see how this might come about, but wanted to at least make note of this here so I don't forget. AFAIK, we'll always have `ecflow` in a `uwtools` conda environment.

src/uwtools/tests/test_cli.py
```diff
@@ -475,6 +491,95 @@ def test_cli__dispatch_fs_report_yes(capsys):
...
+def test_cli__dispatch_ecflow_other_import_error():
+    """
+    Test that _dispatch_ecflow re-raises non-ecflow ImportErrors.
+    """
+    args = {STR.action: STR.realize}
+    with (
+        patch("builtins.__import__", side_effect=ImportError("No module named 'something_else'")),
+        raises(ImportError, match="No module named 'something_else'"),
+    ):
+        cli._dispatch_ecflow(args)
```
maddenp-cu: I suspect this isn't necessary: `uwtools` specifies its required packages when its conda environment is created, so there shouldn't be a case where a required module cannot be imported. If there is, it's probably a bug we've introduced and we probably can't improve on the Python's built-in error reporting.

src/uwtools/cli.py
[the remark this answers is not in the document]
maddenp-cu: I don't think we want this behavior. We don't have anything in place for (and I'm honestly not sure what conda's support is for) optionally-installed requirement packages. I think we can assume that all required packages are available in a conda environment and avoid conditional imports and the complications associated with them.

src/uwtools/cli.py
```diff
@@ -264,6 +309,26 @@ def _dispatch_config_compose(args: Args) -> bool:
...
+
+    :param args: Parsed command-line args.
+    """
+    try:
+        import uwtools.api.ecflow  # noqa: F401, PLC0415
+    except ImportError as e:
+        if "ecflow" in str(e):
+            msg = "ecflow is not installed. Install it with: pip install ecflow"
+            raise UWError(msg) from e
+        raise
```
maddenp-cu: I think we can assume that `ecflow` is installed and omit this `try` / `except` block.

Doing conditional imports based on user-selected features is an interesting idea, but I'm not sure if/how conda supports it, and I think we'd want to consider this approach holistically and see what other packages might be user-selectable instead of always-installed.

src/uwtools/ecflow.py
```diff
@@ -38,6 +37,9 @@
 
     from ecflow import NodeContainer
 
+# Import log after ensuring stdlib logging is properly set up
+from uwtools.logging import log
```
maddenp-cu: I don't see a change to `uwtools.logging` that would explain this comment or why this import statement needed to move. I think the comment might be inaccurate, too: Logging is not actually set up until `uwtools.logging.setup_logging()` gets called. Please let me know if I'm missing something, but I think maybe this change should be reverted.

src/uwtools/ecflow.py
```diff
@@ -46,13 +48,16 @@ class _ECFlowDef:
 
     def __init__(self, config: dict | YAMLConfig | Path | None = None) -> None:
         self._scripts: dict[Path, str] = {}
+        log.debug("Initializing _ECFlowDef with config: %s", config)
```
```suggestion
        log.debug("Creating ecFlow definition from %s", f"config: {config} if config else "stdin")
```
I think it might be nicer to see (even though this is a debug-level log message) a description of the object than its (private) class name, and if `config` is `None`, we'll be reading the config from `stdin`, so it might be more helpful to report that than `None`.

src/setup.py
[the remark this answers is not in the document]
maddenp-cu: I agree. And this bit of code only applies to users who make the uncommon choice to install `uwtools` via pip instead of via conda. `ecflow` will certainly be available in conda environments due to the contents of `recipe/meta.yaml` at the repo root.

maddenp-cu (changes requested):
I'd like to do another pass after the first round of comments are deal with.

src/uwtools/cli.py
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Excellent. I appreciate the clarification.

src/setup.py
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Thanks. That makes sense. I removed the filter.

src/uwtools/cli.py
AnnaSmoot-NOAA: I think it got changed as part of my debugging struggles. I agree with you both. It should be changed back to how it was.

src/uwtools/cli.py
AnnaSmoot-NOAA: That makes sense. Thanks for pointing that out! `#Mode config` is now restored to it's original position and `_dispatch_ecflow` is no longer misplaced and the `#Mode ecflow` functions are now in alphabetical order.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
AnnaSmoot-NOAA: Thanks for catching this. There was a debugging spiral I ran into that revolved around logging.py and it shouldn't have. This is a remnement of that that I didn't revert back. I just did diff now to double check anything related to logging.

docs/sections/user_guide/cli/tools/ecflow/Makefile
AnnaSmoot-NOAA: Done. `Makefile` is now a symlink to `../..Makefile.outputs`, matching other tool subdirectories.

src/uwtools/resources/jsonschema/execution-parallel.jsonschema
AnnaSmoot-NOAA: Got it. Thanks!

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
AnnaSmoot-NOAA: I've fixed it back. I think this and the previous change you mentioned came from when I was debugging the CLI and it wasn't giving me output errors so I was trying to enforce required arguments, but I've changed it back.

For this particular PR I struggled with that debugging without output for a bit and had two rounds of merge conflicts. I've learned that I need to realize analyze the diff before pushing and/or opening the PR.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done. Thanks for catching that.

src/uwtools/cli.py
AnnaSmoot-NOAA: Agreed. I've removed it. I kept debating about those two lines when I was debugging.

test_ecflow.yaml
AnnaSmoot-NOAA: Thanks for catching that. I should've caught that. My apologies.

AnnaSmoot-NOAA:
@christinaholtNOAA Thank you so much for all of your thoughtful feedback! I see some careless mistakes on my end I'll do better on those in the future. 

I think I've addressed all of your comments. In summary:

`ecflow.py`: `_create_ecf_script` requires `incantation` strictly via try/except — `executable` cannot substitute for it.

`execution-parallel.jsonschema `/ `execution-serial.jsonschema`: `required: ["executable"]` unchanged from base; `incantation` remains a separate optional field (renamed from `jobcmd`).

`test_ecflow.py`: Removed tests that covered the executable-as-fallback behavior; `test__create_ecf_script__without_incantation` raises correctly.

`test_schemas.py`: Reverted 3 functions back to simple `executable`-only forms.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
AnnaSmoot-NOAA: Yes. Thanks for catching that. I made 2 changes: 

1. Added `+ "\n"` to the return value so generated `.ecf` files always end with a newline.
2. For `workflow_output/*.ecf files` : Added the trailing newline to all three committed `.ecf` example files (`fetch.ecf`, `process.ecf`, `model.ecf`).

docs/sections/user_guide/cli/tools/ecflow/workflow_output/suite.def
AnnaSmoot-NOAA: That is funny after your previous comment. I fixed both newline issues: 
1. In `.ecf scripts`:` _ecflowscript()` now appends `\n` after `strip()`, so generated scripts end with exactly one newline. Updated the three committed `.ecf` example files accordingly.
2. In `suite.def`: `write_suite_definition()` now strips the trailing newline that `Defs.__str__()` already includes before passing to `print()`, so the output ends with exactly one newline. Updated `suite.def` and `realize-stdout.out`.

src/uwtools/tests/test_cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_cli.py
AnnaSmoot-NOAA: Agreed and done.

src/uwtools/cli.py
AnnaSmoot-NOAA: Fixed while I was going through Christina's comments.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: Fixed while I was going through Christina's comments.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: Done. Changed to: 

`log.debug("Creating ecFlow definition from %s", config if config else "stdin")`

AnnaSmoot-NOAA: Actually, that failed linting, so now it's: 
  `      log.debug("Creating ecFlow definition from %s", config or "stdin")`

AnnaSmoot-NOAA:
@maddenp-cu I've made all of the changes based on feedback from you, Christina, and Emily. This should be ready for you to look at a second time.

Once again, thank you to you and @christinaholtNOAA and @elcarpenterNOAA for the thoughtful feedback. I think I can help minimize the smaller problems in my code in the future by carefully going over the diff in draft.

maddenp-cu (approved):
Nice work, thank you!

docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
```diff
@@ -0,0 +1,12 @@
+model=%MODEL%
```
christinaholtNOAA, line 1: I'm realizing that these examples are included here in the workflow output, but they don't seem to have been incorporated into the documentation.

src/uwtools/tests/test_ecflow.py
```diff
@@ -642,7 +642,7 @@ def test_integration__with_expand(self):
                 "suite_ensemble": {
                     "tasks_member_{{ ec.MEM }}": {
                         "expand": {"MEM": ["01", "02", "03"]},
-                        "script": {"execution": {"executable": "run me", "incantation": "hello"}},
+                        "script": {"execution": {"incantation": "hello"}},
```
christinaholtNOAA, line 645: It's surprising to me that this would pass schema checking.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
AnnaSmoot-NOAA: Thanks for catching that!
1. I added a new `ecflow-workflow.yaml` example file that produces the `workflow_output/` artifacts, so readers can see the input that generates those scripts.
2. Updated the `realize --scripts-path` example to use `ecflow-workflow.yaml` and point output to `workflow_output/`.
3. Added `literalinclude` directives in `ecflow.rst` to display the contents of all three generated `.ecf` scripts (`fetch.ecf`, `process.ecf`, and `model.ecf`) directly in the docs.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: I've been digging into this this morning. Here's what I have: 

Python silently deduplicates dict literal keys at parse time — the interpreter keeps only the last value. So if two `"script"` keys appear in the same dict literal, the schema never sees a duplicate; it just receives whichever value Python kept last. Schema validation can't help here because by the time the dict reaches the validator, the duplication is already gone.

The fix was to remove the extra `"script"` key. To catch this class of bug automatically in the future, a linter rule like `pylint`'s `W0109` (duplicate-key) or `flake8-bugbear`'s `B006` variant could flag duplicate dict keys at development time.

AnnaSmoot-NOAA:
Looking into Christina's comment on the schema, I decided to dig deeper and found 2 more potential issues. I would like to open them as issues outside of this PR and get this PR merged. Does that sound ok @maddenp-cu @christinaholtNOAA? 

Here is what I think needs further investigation: 

1.  issue #893 
`test_integration__with_expand` passes with an execution block containing only `incantation` (no `executable`) because expand blocks are validated against the `expander` schema, which doesn't have `additionalProperties: false` and doesn't recurse into the task-level `script` content — so that execution block is never schema-checked.

2. issue #894 
Separately, `execution-serial.jsonschema` and `execution-parallel.jsonschema` both declare `"required": ["executable"]`, but `_create_ecf_script` only ever reads `incantation` from the execution block. So there's a mismatch between the schema and the implementation — the schema requires `executable` but the code never uses it.

src/uwtools/tests/test_ecflow.py
maddenp-cu: At the point in `_ECFlowDef.__init__()` where the validation is done, the config looks like (from a session with the debugger) this:
```
ecflow:
  suite_ensemble:
    tasks_member_{{ ec.MEM }}:
      expand:
        MEM:
        - '01'
        - '02'
        - '03'
      script:
        execution:
          incantation: hello
```
Looking at the schema, it looks like `suite_ensemble` would match `^suite_.+$` under `ecflow`. Its value would match the `addons` part of the `allOf` trivially (because `addons` doesn't _require_ anything),  then the `nodecontainer` part of the `allOf` would match `tasks_member_{{ ec.MEM }}` to its `^tasks_.+$` option. That then leads to `expander`, which requires an `expand` child (ok), but is not expecting `script`, so it seems like validation would fail here.

Does `expand` need `"additionalProperties": false`? If I set that, then this test fails, even on `ecflow_tool`.

maddenp-cu:
>    `test_integration__with_expand` passes with an execution block containing only `incantation` (no `executable`) because expand blocks are validated against the `expander` schema, which doesn't have `additionalProperties: false` and doesn't recurse into the task-level `script` content — so that execution block is never schema-checked.

Yeah, sorry, I wrote basically the same thing before reading this. I concur.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: No, `expand` should not have `additionalProperties: false`. The siblings of `expand` (like `script`, `trigger`, `vars`) are intentional — they get passed through to `_add_node` for each expanded node, so they act as shared attributes applied to every node in the expansion. The schema is correctly permissive here. `test_integration__with_expand` covers this case and confirms the behavior is working as designed.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
christinaholtNOAA: There are many additional changes that I expected to show up since so many of the `.out` files were generated by AI. Perhaps this is best handled in a follow-on ticket to update the docs. I'll go ahead and approve here and add a new ticket to the board.

christinaholtNOAA (approved):
The docs need a follow-on, but the addition meets the needs of the ticket.

src/uwtools/tests/test_ecflow.py
maddenp-cu: Sorry, I mistyped: I meant `expander`, not `expand`. Is `script` really supposed to be present at this position in the config? It's not explicitly mentioned in the `expander` schema, but that schema also does not set `additionalProperties` to `False`, so _anything_ would validate successfully here -- I just tested and changing `script` to `foo` and that validates.

elcarpenterNOAA (approved):

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: You're right, Paul. Thanks!  My previous response was imprecise. 

The behavior is intentional — siblings of `expand` are applied to each expanded node, but I think the schema is too permissive because `expander` has no `additionalProperties: false` and no reference to `addons`, so arbitrary keys like `foo` pass validation unchecked.  I think a  fix would be to have `expander` also `$ref addons` and, for task expanders specifically, allow `script`, but that's non-trivial since `expander` is shared between `tasks_*` and `families_*` contexts. I'll open a follow-up issue on tightening the `expander` schema.

Opened: https://github.com/ufs-community/uwtools/issues/897

PR #892, opened 2026-05-04
Don't modify input in methods.

### LINKED ISSUE #864: Don't change input config in ecflow.py.
**Description**

In at least one case in ecflow.py, an input config is changed (in the `_add_repeat` method). In general, we should use deep copies of input config dicts before modifying them so we do not inadvertently change users' configs.

This is the one example identified in a recent code review, but the whole module should be checked for any other instances of this type of config change.

**Acceptance Criteria**

- [ ] Fix any cases where the input config may be modified and update it to modify only a local copy.
- [ ] Update unit tests as needed.

christinaholtNOAA:
**Synopsis**

Fixes #864.

Completely removes the modification of config in the method in question. No other instances were found. Learned that ecFlow doesn't accept kwargs.

Tests that add properties to nodes have been updated to remove mocking and add the items specifically to ensure the calls are correct for the ecFlow API.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -34,8 +34,6 @@ from uwtools.strings import EC, STR
 from uwtools.utils.file import writable
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable, Sequence
-
     from ecflow import NodeContainer
 
 
@@ -161,10 +159,7 @@ class _ECFlowDef:
         """
         parent.add(node)
 
-        def add_items(method: Callable[..., object], cfg: Iterable[Sequence[object]]) -> None:
-            for args in cfg:
-                method(*args)
-
+        add_items = lambda method, cfg: [method(*args) for args in cfg]
         for key, subconfig in config.items():
             tag, name = self._tag_name(key)
             match tag:
@@ -223,7 +218,6 @@ class _ECFlowDef:
         """
         match name:
             case EC.date:
-                config["delta"] = config.pop("step", None)
                 repeat = RepeatDate
             case EC.datelist | EC.enumerated | EC.string:
                 repeat = RepeatEnumerated
@@ -233,7 +227,12 @@ class _ECFlowDef:
                 repeat = RepeatDay
             case EC.int:
                 repeat = RepeatInteger
-        node.add_repeat(repeat(**config))
+        args = [
+            config.get(k)
+            for k in ("variable", "start", "end", "step", "values")
+            if config.get(k) is not None
+        ]
+        node.add_repeat(repeat(*args))
 
     def _create_ecf_script(self, config: dict, task: Task) -> None:
         """
@@ -257,7 +256,7 @@ class _ECFlowDef:
         except KeyError as e:
             msg = "The execution block for %s must include 'incantation'" % task.name()
             raise UWConfigError(msg) from e
-        es = self._ecflowscript(
+        script_contents = self._ecflowscript(
             execution=[cmd],
             manual=config.get(EC.manual, f"Script to run {task.name()}"),
             envcmds=execution.get(EC.envcmds, []),
@@ -270,7 +269,7 @@ class _ECFlowDef:
             Path(task.get_abs_node_path().lstrip("/")).parent
             / f"{task.name().split('_', 1)[-1]}.ecf"
         )
-        self._scripts[path] = es
+        self._scripts[path] = script_contents
 
     def _ecflowscript(
         self,
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -227,9 +227,13 @@ class _ECFlowDef:
                 repeat = RepeatDay
             case EC.int:
                 repeat = RepeatInteger
+
+        # The schema-checked config blocks for each of these will be the variable and either the
+        # start/end/[step] keys or the list key. Since they must be passed as positional arguments,
+        # ensure their order is either (argument, start, end, [step]) or (argument, list).
         args = [
             config.get(k)
-            for k in ("variable", "start", "end", "step", "values")
+            for k in (EC.variable, EC.start, EC.end, EC.step, EC.list)
             if config.get(k) is not None
         ]
         node.add_repeat(repeat(*args))
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -218,7 +218,7 @@ class TestECFlowDef:
 
     def test__add_repeat__enumerated(self, instance):
         node = Task("t1")
-        config = {"variable": "MEMBER", "values": ["m01", "m02", "m03"]}
+        config = {"variable": "MEMBER", "list": ["m01", "m02", "m03"]}
         instance._add_repeat(config.copy(), "enumerated", node)
         assert 'repeat enumerated MEMBER "m01" "m02" "m03"' in node.get_repeat().__str__()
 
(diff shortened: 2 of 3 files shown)


src/uwtools/tests/test_ecflow.py
```diff
@@ -190,48 +191,46 @@ def test__jobscheduler__no_threads(self, instance_with_scheduler):
...
-        repeat_date.assert_called_once_with(**{**common, "delta": 1})
-        node.add_repeat.assert_called_once_with(repeat_date.return_value)
+        node = Task("t1")
+        config = {"variable": "YMD", "step": 4, "start": 20240101, "end": 20240131}
+        start_config = deepcopy(config)
+        instance._add_repeat(config, "date", node)
+        assert config == start_config
+        expected = ["YMD", 20240101, 20240131, 4]
+        repeat = node.get_repeat()
+        actual = [repeat.__getattribute__(a)() for a in ("name", "start", "end", "step")]
```
christinaholtNOAA, line 201: This test also ensures that the arguments end up in the correct order.

src/uwtools/tests/test_ecflow.py
```diff
@@ -190,48 +191,46 @@ def test__jobscheduler__no_threads(self, instance_with_scheduler):
...
 
     def test__add_repeat__date(self, instance):
-        node = Mock()
-        common = {"name": "YMD", "start": 20240101, "end": 20240131}
-        with patch.object(ecflow, "RepeatDate") as repeat_date:
-            instance._add_repeat({**common, "step": 1}, "date", node)
-        repeat_date.assert_called_once_with(**{**common, "delta": 1})
-        node.add_repeat.assert_called_once_with(repeat_date.return_value)
+        node = Task("t1")
+        config = {"variable": "YMD", "step": 4, "start": 20240101, "end": 20240131}
```
christinaholtNOAA, line 195: I used a non-default value for step here just to make sure it was actually getting set properly.

src/uwtools/tests/test_ecflow.py
```diff
@@ -190,48 +191,46 @@ def test__jobscheduler__no_threads(self, instance_with_scheduler):
...
-        node = Mock()
-        config = {"name": "MEMBER", "values": ["m01", "m02", "m03"]}
-        with patch.object(ecflow, "RepeatEnumerated") as repeat_enumerated:
-            instance._add_repeat(config.copy(), "enumerated", node)
-        repeat_enumerated.assert_called_once_with(**config)
-        node.add_repeat.assert_called_once_with(repeat_enumerated.return_value)
+        node = Task("t1")
+        config = {"variable": "MEMBER", "values": ["m01", "m02", "m03"]}
+        instance._add_repeat(config.copy(), "enumerated", node)
+        assert 'repeat enumerated MEMBER "m01" "m02" "m03"' in node.get_repeat().__str__()
```
christinaholtNOAA, line 223: The values didn't seem to show up in any of the public attributes, oddly enough.

src/uwtools/tests/test_ecflow.py
```diff
@@ -590,17 +573,15 @@ def test__add_node__with_expand_tag(self, instance):
         assert task.get_trigger() is not None
 
     def test__add_repeat__datetime(self, instance):
-        node = Mock()
+        task = Task("t1")
         config = {
-            "name": "DT",
+            "variable": "DT",
```
christinaholtNOAA, line 578: Here and elsewhere, the schema-required key is `variable`.

src/uwtools/ecflow.py
```diff
@@ -223,7 +218,6 @@ def _add_repeat(self, config: dict, name: str, node: Node) -> None:
         """
         match name:
             case EC.date:
-                config["delta"] = config.pop("step", None)
```
christinaholtNOAA, line 226: Don't need this since the ecFlow API doesn't seem to allow keyword arguments.

src/uwtools/ecflow.py
```diff
@@ -233,7 +227,12 @@ def _add_repeat(self, config: dict, name: str, node: Node) -> None:
                 repeat = RepeatDay
             case EC.int:
                 repeat = RepeatInteger
-        node.add_repeat(repeat(**config))
+        args = [
+            config.get(k)
+            for k in ("variable", "start", "end", "step", "values")
+            if config.get(k) is not None
+        ]
```
christinaholtNOAA, line 238: This should put the arguments in the right order since we can't rely on keyword args.

src/uwtools/ecflow.py
```diff
@@ -257,7 +256,7 @@ def _create_ecf_script(self, config: dict, task: Task) -> None:
         except KeyError as e:
             msg = "The execution block for %s must include 'incantation'" % task.name()
             raise UWConfigError(msg) from e
-        es = self._ecflowscript(
+        script_contents = self._ecflowscript(
```
christinaholtNOAA, line 263: The name just annoyed me in my review.

elcarpenterNOAA (approved):
Very nice.

src/uwtools/ecflow.py
```diff
@@ -161,10 +159,7 @@ def _add_node(  # noqa: PLR0912
         """
         parent.add(node)
 
-        def add_items(method: Callable[..., object], cfg: Iterable[Sequence[object]]) -> None:
-            for args in cfg:
-                method(*args)
-
+        add_items = lambda method, cfg: [method(*args) for args in cfg]
```
maddenp-cu, line 162: I like this, but am a little (pleasantly) surprised the linter doesn't complain about the fact that we don't use the `list` return value.

src/uwtools/ecflow.py
maddenp-cu, line 238: Am I right in thinking that this is safe -- e.g. there's no risk that `values` might be present but `step` might not, such that `values` would end up as the 4th element in `args` instead of `step` -- because `step` and `values` (and maybe some of the other items, except `variable`) are mutually exclusive, so that `step` and `values` would never be present together, due to the schema? It might be worth a small comment here to explain that to nervous readers like me. Maybe referring them to the schema would be useful.

Is `"values"` what's called `"list"` in the schema, which we can't use here because it's a keyword?

I guess we're lucky that this works so elegantly, and that an uglier but more explicit version might look something like
```
match set(config.keys()):
    case {"step"}:
        args = ...
    case {"variable", "list"}:
        args = ....
    case {"variable", "start", "end"} | {"variable", "start", "end", "step"}:
        args = ...
    case _:
        raise ...
```
I think maybe that covers the possibilities I see in the schema. Each of those `args =` might call a lambda or a defined function, but there'd still be a lot of duplication with all those argument names.

On that note: Would it be worth factoring out those strings to the `strings.EC` container?

maddenp-cu (approved):
Just a couple questions to make sure I understand, and small suggestion for a comment.

src/uwtools/ecflow.py
christinaholtNOAA: You're absolutely right! The key should be `list`. Nice catch. 

I believe there is a decent enough distinction between the way the "start, end, step" options work and the way the "enumerate" ones do that we can leave it. Are you okay with the less explicit version?

I will add a comment and factor these into `strings.EC` either way.

maddenp-cu: Yes, I like the less explicit version and am happy that it works.

AnnaSmoot-NOAA:
The looks great, especially after the feedback from Paul and Emily!

AnnaSmoot-NOAA (approved):

PR #895, opened 2026-05-06
Require at least one suite in ecFlow config schema (fixes #877)

### LINKED ISSUE #877: Decide whether ecFlow schema should permit a config with no suites
## Description

Today `{"ecflow": {}}` validates successfully against
`src/uwtools/resources/jsonschema/ecflow.jsonschema` because the schema
does not require any suites at the `ecflow` root. Surfaced during review
of PR #873 by @maddenp-cu:

> I think you and @christinaholtNOAA would have to decide whether a
> schema with no suites is something we'd ever want to allow. If so,
> maybe the schema is ok. If not, I think we should open an Issue so we
> don't forget.

## Decision needed

Is a no-suite ecFlow workflow ever a legitimate input?

- **If yes:** close this issue; current schema is correct.
- **If no:** tighten the schema to require at least one suite under
  `ecflow` (e.g., `minProperties: 1` on the `ecflow` object, or an
  explicit "at least one suite" constraint).

## Follow-on work if tightened

- Update `test_validate__dict`, `test_validate__yamlconfig`,
  `test_validate__path`, and `test_validate__stdin` in
  `src/uwtools/tests/test_ecflow.py` to use a `minimal_config` that
  contains a real (minimal) suite instead of `{}`.
- Audit any other fixtures that rely on an empty `ecflow` body.

## References

- PR #873 review thread (maddenp / AnnaSmoot exchange)

AnnaSmoot-NOAA:
**Synopsis**

Closes #877

## Problem ##

During review of issue #877, the team discussed whether `uw ecflow` validate should reject configs with no suites. After testing, the team agreed on a convention: accept what ecFlow accepts. Since `Defs.check()` accepts empty and suite-free definitions without complaint, the schema should not be stricter than ecFlow itself.

The gap identified in #877 is better addressed by running ecFlow's own correctness checker — analogous to how `uwtools` validates Rocoto XML against Rocoto's official schema — rather than adding JSON Schema constraints that may reject valid configs.

## Fix ##

Added a call to `self._d.check()` in `_ECFlowDef.__init__` after building the suite definition. If ecFlow reports any errors (e.g. unresolvable trigger expressions, limit violations), a `UWConfigError` is raised. This gives ecFlow's own client-side validation layer on top of the JSON Schema check.

No schema constraint was added — the `ecflow` block permits empty and suite-free configs consistent with what ecFlow itself accepts.

## Changes ##

- `ecflow.py`: Call `self._d.check()` after `_add_workflow_components()`; raise `UWConfigError` on non-empty output

- `test_ecflow.py`:
    - Reverted `minimal_config` to `{"ecflow": {}}`
    - Reverted `assets` fixture and `realize()` tests to use `{"ecflow": {}}` minimal config; `realize()` tests assert only `"# enddef" in output`
    - Removed `test_validate__empty_ecflow_fails` and `test_validate__properties_only_fails` - these behaviors are now permitted
    - Added `test__init__defs_check_bad_trigger`: verifies that a trigger referencing a nonexistent task raises `UWConfigError` via `Defs.check()`
- `test_schemas.py`:
    - Added `test_schema_ecflow_scheduler`: valid scheduler values pass; unknown values fail
    - Added `test_schema_ecflow_extern`: string list passes; non-string items and scalar values fail
    - Added `test_schema_ecflow_refs_taskcontainer_script_incantation`: `incantation` + `executable` valid; `incantation` alone fails, schema requires `executable`

**Type**

- [X] Bug fix (corrects a known issue)
- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -54,9 +54,8 @@ class _ECFlowDef:
         log.debug("Adding workflow components to suite definition.")
         self._add_workflow_components()
         log.debug("Workflow components added. Scripts: %s", list(self._scripts.keys()))
-        if check_errors := self._d.check():
-            msg = f"ecFlow definition check failed:\n{check_errors}"
-            raise UWConfigError(msg)
+        check_errors = self._d.check()
+        assert not check_errors, check_errors
 
     def __str__(self):
         return self._d.__str__()
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -54,6 +54,9 @@ class _ECFlowDef:
         log.debug("Adding workflow components to suite definition.")
         self._add_workflow_components()
         log.debug("Workflow components added. Scripts: %s", list(self._scripts.keys()))
+        if check_errors := self._d.check():
+            msg = f"ecFlow definition check failed:\n{check_errors}"
+            raise UWConfigError(msg)
 
     def __str__(self):
         return self._d.__str__()
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -470,15 +470,6 @@
   "properties": {
     "ecflow": {
       "additionalProperties": false,
-      "not": {
-        "propertyNames": {
-          "enum": [
-            "extern",
-            "scheduler",
-            "vars"
-          ]
-        }
-      },
       "patternProperties": {
         "^suite_.+$": {
           "allOf": [
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -470,6 +470,15 @@
   "properties": {
     "ecflow": {
       "additionalProperties": false,
+      "not": {
+        "propertyNames": {
+          "enum": [
+            "extern",
+            "scheduler",
+            "vars"
+          ]
+        }
+      },
       "patternProperties": {
         "^suite_.+$": {
           "allOf": [
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -284,9 +284,8 @@ class TestECFlowDef:
                 }
             }
         }
-        with raises(UWConfigError) as e:
+        with raises(AssertionError):
             _ECFlowDef(config=config)
-        assert "ecFlow definition check failed" in str(e.value)
 
     # _add_workflow_components tests
 
@@ -678,28 +677,28 @@ def test_ecflow_realize__cfg_to_file(tmp_path, assets):
     cfgfile, _ = assets
     ecflow.realize(config=YAMLConfig(cfgfile), output_path=tmp_path)
     output = (tmp_path / "suite.def").read_text()
-    assert "# enddef" in output
+    assert output == str(Defs())
 
 
 def test_ecflow_realize__cfg_to_stdout(capsys, assets):
     cfgfile, _ = assets
     ecflow.realize(config=YAMLConfig(cfgfile))
     output = capsys.readouterr().out
-    assert "# enddef" in output
+    assert output == str(Defs())
 
 
 def test_ecflow_realize__file_to_file(tmp_path, assets):
     cfgfile, _ = assets
     ecflow.realize(config=cfgfile, output_path=tmp_path)
     output = (tmp_path / "suite.def").read_text()
-    assert "# enddef" in output
+    assert output == str(Defs())
 
 
 def test_ecflow_realize__file_to_stdout(capsys, assets):
     cfgfile, _ = assets
     ecflow.realize(config=cfgfile)
     output = capsys.readouterr().out
-    assert "# enddef" in output
+    assert output == str(Defs())
 
 
 def test_ecflow_realize__write_scripts(capsys, assets):
@@ -708,7 +707,7 @@ def test_ecflow_realize__write_scripts(capsys, assets):
         ecflow.realize(config=cfgfile, scripts_path=script_path)
         write_scripts.assert_called_once_with(script_path)
     output = capsys.readouterr().out
-    assert "# enddef" in output
+    assert output == str(Defs())
 
 
 def test_validate__path(tmp_path, minimal_config):
--- a/src/uwtools/tests/test_ecflow.py
+++ b/src/uwtools/tests/test_ecflow.py
@@ -52,7 +52,7 @@ def minimal_config():
     """
     Minimal config for instantiation.
     """
-    return {"ecflow": {"suite_minimal": {}}}
+    return {"ecflow": {}}
 
 
 def assert_line_in(result: str, line: str) -> None:
@@ -241,18 +241,18 @@ class TestECFlowDef:
 
     def test__init__with_dict(self, minimal_config):
         ecf = _ECFlowDef(config=minimal_config)
-        assert ecf._config == {"suite_minimal": {}}
+        assert ecf._config == {}
         assert ecf._scheduler is None
         assert isinstance(ecf._d, Defs)
 
     def test__init__with_scheduler(self):
-        config = {"ecflow": {"scheduler": "slurm", "suite_test": {}}}
+        config = {"ecflow": {"scheduler": "slurm"}}
         ecf = _ECFlowDef(config=config)
         assert ecf._scheduler == "slurm"
 
     def test__init__with_path(self, tmp_path):
         config_file = tmp_path / "config.yaml"
-        config_file.write_text("ecflow:\n  scheduler: pbs\n  suite_test: {}\n")
+        config_file.write_text("ecflow:\n  scheduler: pbs\n")
         ecf = _ECFlowDef(config=config_file)
         assert ecf._scheduler == "pbs"
         assert isinstance(ecf._d, Defs)
@@ -260,13 +260,33 @@ class TestECFlowDef:
     def test__init__with_config_object(self, minimal_config):
         cfg = YAMLConfig(minimal_config)
         ecf = _ECFlowDef(config=cfg)
-        assert ecf._config == {"suite_minimal": {}}
+        assert ecf._config == {}
         assert isinstance(ecf._d, Defs)
 
     def test__init__missing_ecflow_key(self):
         config: dict = {"not_ecflow": {}}
         with raises(UWConfigError):
             _ECFlowDef(config=config)
+
+    def test__init__defs_check_bad_trigger(self):
+        config = {
+            "ecflow": {
+                "suite_test": {
+                    "task_a": {
+                        "script": {
+                            "execution": {
+                                "executable": "run.exe",
+                                "incantation": "echo hello",
+                            }
+                        },
+                        "trigger": "nonexistent_task == complete",
+                    }
+                }
+            }
+        }
+        with raises(UWConfigError) as e:
+            _ECFlowDef(config=config)
+        assert "ecFlow definition check failed" in str(e.value)
 
     # _add_workflow_components tests
 
@@ -658,7 +678,6 @@ def test_ecflow_realize__cfg_to_file(tmp_path, assets):
     cfgfile, _ = assets
     ecflow.realize(config=YAMLConfig(cfgfile), output_path=tmp_path)
     output = (tmp_path / "suite.def").read_text()
-    assert "suite minimal" in output
     assert "# enddef" in output
 
 
@@ -666,7 +685,6 @@ def test_ecflow_realize__cfg_to_stdout(capsys, assets):
     cfgfile, _ = assets
     ecflow.realize(config=YAMLConfig(cfgfile))
     output = capsys.readouterr().out
-    assert "suite minimal" in output
     assert "# enddef" in output
 
 
@@ -674,7 +692,6 @@ def test_ecflow_realize__file_to_file(tmp_path, assets):
     cfgfile, _ = assets
     ecflow.realize(config=cfgfile, output_path=tmp_path)
     output = (tmp_path / "suite.def").read_text()
-    assert "suite minimal" in output
     assert "# enddef" in output
 
 
@@ -682,7 +699,6 @@ def test_ecflow_realize__file_to_stdout(capsys, assets):
     cfgfile, _ = assets
     ecflow.realize(config=cfgfile)
     output = capsys.readouterr().out
-    assert "suite minimal" in output
     assert "# enddef" in output
 
 
@@ -692,7 +708,6 @@ def test_ecflow_realize__write_scripts(capsys, assets):
         ecflow.realize(config=cfgfile, scripts_path=script_path)
         write_scripts.assert_called_once_with(script_path)
     output = capsys.readouterr().out
-    assert "suite minimal" in output
     assert "# enddef" in output
 
 
@@ -744,21 +759,3 @@ def test_validate__suite_with_properties():
         }
     }
     assert ecflow.validate(config)
-
-
-def test_validate__empty_ecflow_fails():
-    """
-    An empty ecflow block should not pass validation.
-    """
-    config: dict = {"ecflow": {}}
-    with raises(UWConfigError):
-        ecflow.validate(config)
-
-
-def test_validate__properties_only_fails():
-    """
-    An ecflow block with only optional properties (no suite) should not pass validation.
-    """
-    config = {"ecflow": {"scheduler": "slurm", "vars": {"FOO": "bar"}}}
-    with raises(UWConfigError):
-        ecflow.validate(config)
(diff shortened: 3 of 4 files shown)


AnnaSmoot-NOAA:
Addressed the Copilot review comment on `assets` / the `realize()` tests.

**Problem:** The `assets` fixture was building `expected` by calling `str(_ECFlowDef(minimal_config))`, which made the five `realize()` tests circular — since `realize()` also uses `_ECFlowDef` internally, the tests would pass even if `__str__` produced garbage, as long as it was *consistently* garbage.

**Fix:** Removed `expected` from the `assets` fixture entirely and replaced all equality assertions (`assert output == expected`) with two structural invariant checks:
- `assert "suite minimal" in output` — verifies the suite name is rendered (note: `suite_minimal` → tag=`suite`, name=`minimal` after `_tag_name` parsing)
- `assert "# enddef" in output` — verifies the suite definition is properly closed

This avoids hardcoding the ecFlow version string (the original concern) while still genuinely testing that `realize()` produces correct suite-definition content. Regressions in `_ECFlowDef.__str__` will now be caught.

src/uwtools/tests/test_ecflow.py
```diff
@@ -27,7 +27,7 @@ def assets(tmp_path, minimal_config):
     YAMLConfig(minimal_config).dump(yaml_file)
     script_path = tmp_path / "scripts"
     script_path.mkdir(exist_ok=True, parents=True)
-    expected = "#5.15.2\n# enddef\n\n"
+    expected = str(_ECFlowDef(minimal_config)) + "\n"
     return yaml_file, script_path, expected
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: 1. I removed `expected = str(_ECFlowDef(minimal_config)) + "\n"` from the `assets` fixture — it was self-referential since `realize()` uses the same `_ECFlowDef` internally
2. I replaced the five `assert output == expected` checks in the `realize()` tests with:
-`assert "suite minimal" in output` (the suite name, after `_tag_name` strips the `suite_` prefix)
-`assert "# enddef" in output` (confirms the definition is properly closed)

src/uwtools/tests/test_ecflow.py
```diff
@@ -715,3 +722,43 @@ def test_validate__invalid(tmp_path):
     with raises(UWConfigError) as e:
         ecflow.validate(yaml_file)
     assert "YAML validation errors" in str(e.value)
+
+
+def test_validate__suite_only():
+    """
+    A config with only a suite should pass validation.
+    """
+    config: dict = {"ecflow": {"suite_mysuite": {}}}
```
maddenp-cu, line 747: Not necessarily for this PR, but now of course I have to wonder: Is an empty `suite_*` something we want? Maybe: It occurs to me that we might not want to drift far from what ecFlow itself allows, to reject configs it would accept, or vice versa. I'm not sure any of us knows exactly what that is, though.

Did I hear in discussion at some point that ecFlow might have a correctness checker we could run on the final `.def` file? We do that with Rocoto, too: We check the final XML that `uwtools` produces against Rocoto's official schema.

src/uwtools/tests/test_ecflow.py
maddenp-cu, line 1: I think it would be good to add some tests/assertions to `test_schemas.py`, too, to do some targeted tests of the schema independent of the library code.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Great questions. Very helpful.

On the empty suite: ecFlow itself accepts an empty suite without complaint, so allowing `suite_mysuite: {}` is consistent with  not rejecting configs ecFlow would accept. I guess it's useless in practice but not invalid.

On the checker: yes! 
I think I mentioned `Defs.check()` in one of our meetings before. It's in the ecFlow Python API validates trigger/complete expressions and limits (client-side), returning a non-empty string on errors. After some digging, I see there's also `Defs.check_job_creation()` for verifying that `.ecf` script files are locatable. We could call `self._d.check()` in `_ECFlowDef` after building the suite definition — analogous to how Rocoto XML is validated against Rocoto's official schema. That would give us ecFlow's own validation layer on top of the JSON schema check. Should that be a follow up issue or add it into this one?

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -470,6 +470,15 @@
...
       "additionalProperties": false,
+      "not": {
+        "propertyNames": {
+          "enum": [
+            "extern",
+            "scheduler",
+            "vars"
+          ]
+        }
+      },
```
christinaholtNOAA: This was a surprising addition to get the answer we wanted. I'm concerned about how useful the error we get for the schema check. For example, when adding this test to the `test_schemas.py` suite:

```
def test_schema_ecflow_no_suite():
    errors = schema_validator("ecflow")
    config = {
        "ecflow": {
            "scheduler": "pbs",
            }
        }
    assert "foo" in errors(config)
```

JSON Schema reports:
```
"{'scheduler': 'pbs'} should not be valid under {'propertyNames': {'enum': ['extern', 'scheduler', 'vars']}}
```

This doesn't feel like it's guiding users to do the right thing.

AnnaSmoot-NOAA:
Addressed two review comments in follow-up commits:

---

**1. Copilot review — self-referential `expected` in `realize()` tests** ([b80d5bf](https://github.com/AnnaSmoot-NOAA/uwtools/commit/b80d5bf))

The `assets` fixture was building `expected` via `str(_ECFlowDef(minimal_config))`, which made the five `realize()` tests circular — since `realize()` also uses `_ECFlowDef` internally, the tests would pass even if `__str__` produced garbage.

Removed `expected` from `assets` and replaced all equality assertions with structural invariant checks:
- `assert "suite minimal" in output` — the suite name is rendered (note: `suite_minimal` → tag=`suite`, name=`minimal` after `_tag_name` parsing)
- `assert "# enddef" in output` — the suite definition is properly closed

These are version-agnostic (no hardcoded ecFlow version string) and will catch regressions in rendering logic.

---

**2. Review feedback — add targeted schema tests to `test_schemas.py`** ([2936c4d](https://github.com/AnnaSmoot-NOAA/uwtools/commit/2936c4d))

Added four new schema-level tests independent of the library code:

- `test_schema_ecflow_suite_required` — confirms the `not/propertyNames` constraint: empty `ecflow` block fails; optional-properties-only fails; an empty suite (`suite_minimal: {}`) passes (ecFlow itself accepts this); suite + optional properties passes.
- `test_schema_ecflow_scheduler` — valid values (`lsf`, `pbs`, `slurm`) pass; unknown values (e.g. `torque`) fail.
- `test_schema_ecflow_extern` — string list passes; non-string items and scalar values fail.
- `test_schema_ecflow_refs_taskcontainer_script_incantation` — `incantation` alongside `executable` is valid; `incantation` alone is invalid per schema (`executable` is required by `execution-serial`/`execution-parallel`; note that `incantation` being required is a separate code-level enforcement in `_create_ecf_script`).

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu: FWIW we have other cases already where the failure advice isn't very helpful. A common one is something like `is not valid under any of the given schemas`, which happens with `anyOf` (maybe also `allOf`?) keywords. (Search for `is not valid` in `test_schemas.py` to see many cases where these appear in practice.

In `uwtools.config.validator.validate()`, we added [some logic](https://github.com/ufs-community/uwtools/blob/main/src/uwtools/config/validator.py#L99-L105) to detect this case and try to construct some more helpful output, showing the candidate subschemas that _could_ have matched but didn't -- better IMO than just `is not valid under any of the given schemas`, in that it at least answers the question "_Which_ schemas?"

Maybe there's another way to formulate this schema the leads to better out-of-the-box advice to users. That's worth looking for. Alternatively, we could see if we could detect this case in `validate()`, intercept it, and work with the error object at that point to construct a better report for the user. I think it'd be worth trying to find another schema with the same semantics but better user feedback, but there may not be one. Some things are just tricky/messy.

I'll vote for correctness first every time, though -- even if the feedback is confusing. Maybe we can improve the feedback. Maybe we have to resort to documentation.

One other thing, that I mentioned in another message: Are we forbidding ecFlow configurations that ecFlow itself actually accepts? If it accepts nonsensical configs, maybe we should, too. I don't know enough about ecFlow. Should we implement a call to their correctness checker, if such a thing exists, as soon as possible to keep us aligned with their idea of "correct", like the way we validate the Rocoto XML we generate against Rocoto's own XML schema?

src/uwtools/tests/test_ecflow.py
maddenp-cu: What do @christinaholtNOAA @elcarpenter think about maybe adopting the convention that "we accept what ecFlow accepts"? We also have the option to try to force users to do better with stricter checks, but I'm wary of assuming that we know best when there might be legit reasons to do something that seems pointless at first glance. Adopting such a convention would guide our schema checking.

elcarpenterNOAA: I'm in favor of that convention. I'm not sure we have a enough collective experience with ecFlow to start dictating what users can and can't do at this point, but I'd be open to revisiting the idea of stricter checks in the future.

christinaholtNOAA: If we go with that convention, it seems that `Defs.check()` accepts a totally empty suite definition file, so we wouldn't need to require a `suite` entry at the top level.

With `test.def`:

```
#5.15.2
# enddef
```

```
>>> defs = Defs("test.def")
>>> print(defs.check())

>>>
```

Same for these:

```
#5.15.2
extern /foo/bar
# enddef
```

and

```
#5.15.2
extern /foo/bar
suite baz
  edit VAR 12
# enddef
```

I did find it odd that if I put a variable at the top level, it didn't show up when loaded this way. As in this example:

test.def:
```
#5.15.2
extern /foo/bar
edit VAR 12
# enddef
```
Results in:
```
>>> defs = Defs("test.def")
>>> print(defs)
#5.15.2
extern /foo/bar
# enddef

>>> 
```

maddenp-cu: I feel like we should follow what `Defs.check()` allows (and validate by actually calling it on the suite def we produce), and if that makes some of the schema-checking complexity disappear, I'm not going to complain.

christinaholtNOAA: I'm okay with that, too.

AnnaSmoot-NOAA: Based on this discussion and the team's agreement on "accept what ecFlow accepts":
 
1. I reverted the `not/propertyNames` constraint from `ecflow.jsonschema`; `{"ecflow": {}}` and optional-properties-only configs now pass schema validation, consistent with what ecFlow itself accepts.

2. I added `Defs.check()` in `_ECFlowDef.__init__`. After building the suite definition, we call `self._d.check()` and raise `UWConfigError` if it returns errors. This gives ecFlow's own validation layer (trigger/complete expression and limit checking) analogous to how we validate Rocoto XML against Rocoto's schema.

So Issue #877 is resolved by convention + `Defs.check()` rather than a JSON Schema restriction.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: Both concerns addressed by the changes in the latest commits: 

Confusing error message — moot. The `not/propertyNames` constraint has been reverted entirely, so that error message no longer appears.

"Are we forbidding configs ecFlow accepts?" — The schema no longer rejects empty or suite-free configs.

ecFlow correctness checker — implemented.

AnnaSmoot-NOAA:
Updated based on review discussion:  the PR has pivoted significantly and is ready for another look.

What changed:

Based on the team's agreement in this thread, thanks @maddenp-cu, @christinaholtNOAA, @elcarpenterNOAA, the approach shifted from "enforce a suite requirement via JSON Schema" to "accept what ecFlow accepts." Here's what that looks like in code:

1. Reverted the `not/propertyNames` schema constraint: empty and suite-free `ecflow` configs now pass schema validation, consistent with what `Defs.check()` accepts.

2. Added `Defs.check()` in `_ECFlowDef.__init__`: after building the suite definition, we call `self._d.check()` and raise `UWConfigError` if ecFlow reports errors. This is the direct analogue of how we validate Rocoto XML against Rocoto's own schema.

3. Reverted `minimal_config` back to `{"ecflow": {}}` and updated all dependent tests accordingly. Removed `test_validate__empty_ecflow_fails` and `test_validate__properties_only_fails`. Added `test__init__defs_check_bad_trigger` to cover the new `Defs.check()` path.

Issue #877 is resolved by convention + Defs.check() rather than a schema restriction.

src/uwtools/tests/test_ecflow.py
```diff
@@ -656,37 +675,40 @@ def test_integration__with_expand(self):
 
 
 def test_ecflow_realize__cfg_to_file(tmp_path, assets):
-    cfgfile, _, expected = assets
+    cfgfile, _ = assets
     ecflow.realize(config=YAMLConfig(cfgfile), output_path=tmp_path)
     output = (tmp_path / "suite.def").read_text()
-    assert output == expected
+    assert "# enddef" in output
```
maddenp-cu: Why this change, and the elimination of `expected`? It looks like the output is still
```
(Pdb) p output
'#5.15.2\n# enddef\n'
```
An assertion that one line is present is weaker than an assertion about the entire contents of the output. The single-line check would pass even if all kinds of other unexpected nonsense was present in the output, while a full-contents check excludes that possibility.

src/uwtools/ecflow.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, config: dict | YAMLConfig | Path | None = None) -> None:
         log.debug("Adding workflow components to suite definition.")
         self._add_workflow_components()
         log.debug("Workflow components added. Scripts: %s", list(self._scripts.keys()))
+        if check_errors := self._d.check():
+            msg = f"ecFlow definition check failed:\n{check_errors}"
+            raise UWConfigError(msg)
```
```suggestion
        check_errors = self._d.check()
        assert not check_errors, check_errors
```
I see that we're not currently doing this correctly in `rocoto.py` (and should fix that), but following another `uwtools` convention, we should raise exceptions when the user has made a mistake that they might be able to correct, but we should fail assertions when we, the developers, have made a mistake that users cannot correct. Here, if we've taken a validated YAML config and produced an invalid ecFlow `Def`, that's on us -- there's nothing the user can do differently to correct the problem; they need to report the bug to us.

maddenp-cu (changes requested):

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: I see. I wasn't treating it as installed from the ecflow package when it obviously is. 

I restored full-contents assertions using `str(Defs())` as the expected value. This is derived dynamically from the installed ecflow library, so it remains version-agnostic while still asserting the complete output.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: Ah I see. That's a good point. 

I changed to assert not `check_errors, check_errors` per the convention. Like you said, this error would indicate a developer bug (bad Def produced from a valid config) rather than a user-correctable mistake.

src/uwtools/tests/test_ecflow.py
```diff
@@ -27,8 +27,7 @@ def assets(tmp_path, minimal_config):
     YAMLConfig(minimal_config).dump(yaml_file)
     script_path = tmp_path / "scripts"
     script_path.mkdir(exist_ok=True, parents=True)
-    expected = "#5.15.2\n# enddef\n"
-    return yaml_file, script_path, expected
+    return yaml_file, script_path
```
christinaholtNOAA: I personally prefer the original formulation that ended up simpler in each of the tests and had an independent "expected" outcome from the code.

maddenp-cu: I agree. The current `assert output == str(Defs())` feels too close to a tautology. It just says "The `Defs` object will appear as a string on `stdout`". A test is a good place to be explicit about how we believe this should look.

AnnaSmoot-NOAA: Done! 
I restored the explicit `expected` string in the `assets` fixture using `f"#{ecflowlib.__version__}\n# enddef\n"`. This keeps the assertion independent of the code under test (not a tautology - which I had to look up!) while remaining version-agnostic so it won't break on ecflow upgrades.

maddenp-cu (approved):

christinaholtNOAA (approved):

elcarpenterNOAA (approved):

PR #900, opened 2026-05-12
Add YAML !extend support for sequences

maddenp-cu:
**Synopsis**

Add support for a custom YAML tag, `!extend`, that can be used to append elements to a YAML sequence.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/config/support.py
```diff
@@ -21,6 +21,10 @@
 INCLUDE_TAG = "!include"
 YAMLKey = bool | float | int | str
 
+
+class _UWYAMLLoader(yaml.SafeLoader): ...
```
maddenp-cu: Previously, `uw_yaml_loader()` was modifying `yaml.SafeLoader` directly. That could be unsafe when `uwtools` is mixed in with other code that wants to use that loader. Create a subclass to modify for `uwtools` purposes.

src/uwtools/config/support.py
```diff
@@ -124,12 +128,13 @@ def log_and_error(msg: str) -> Exception:
     return UWConfigError(msg)
 
 
+@cache
```
maddenp-cu, line 132: There's no need to do this more than once.

elcarpenterNOAA (approved):

christinaholtNOAA (approved):

Issue #893, opened 2026-05-06
`ecflow`: `expander` schema skips validation of task `script`/`execution` blocks


AnnaSmoot-NOAA:
**Current Behavior**

When tasks are defined inside an `expand` block (e.g., `tasks_member_{{ ec.MEM }}`), the `script`/`execution` block is never validated against the schema. This means invalid or incomplete execution configs can pass `uw ecflow validate` without error.

Cause: The `expander` schema (`#/$defs/expander` in `ecflow.jsonschema`) uses `patternProperties` to match child keys but does **not** set `additionalProperties: false`, and it does not recurse into the `taskcontainer` schema for inline task content within the expand template. As a result, the `execution` block inside an expand template is invisible to the validator.

**Expected Behavior**

Validation should recurse into the task `script/execution` block within expand templates and enforce the same constraints as for non-expanded tasks.

**To Reproduce**

```yaml
ecflow:
  suite_test:
    tasks_member_{{ ec.MEM }}:
      expand:
        MEM: ["01", "02"]
      script:
        execution:
          incantation: echo hello   # missing required 'executable' — no error raised

`uw ecflow validate` passes without complaint.

**Additional Context**

Discovered during review of PR #891.

AnnaSmoot-NOAA:
This issue is somewhat based on the outcomes from #894, so needs to be done after #894.

AnnaSmoot-NOAA:
closed with PR #905

PR #901, opened 2026-05-13
ecflow: fix expander schema composition and enforce unevaluatedProperties

### LINKED ISSUE #897: `ecflow: expander` schema accepts arbitrary sibling properties alongside `expand`
**Current Behavior**

The `expander` `$def` in `ecflow.jsonschema` does not restrict which properties may appear as siblings of the `expand` key. Because `additionalProperties` is not set to `false` and no `addons` reference is included, any arbitrary key validates successfully. For example, the following passes `uw ecflow validate` without error:

```
ecflow:
  suite_test:
    tasks_member_{{ ec.MEM }}:
      expand:
        MEM: ["01", "02"]
      foo: bar   # nonsense key — should fail validation
```

**Expected Behavior**

Only properties that are intentionally valid alongside `expand` should validate. For a `tasks_*` expander, valid siblings are node attributes from `addons` (e.g., `trigger`, `vars`, `defstatus`) plus `script`. For a `families_*` expander, `script` is not valid. Currently the schema cannot distinguish these and accepts anything.

**Additional Context**

Sibling properties of `expand` are intentional at runtime: `_expand_block()` passes the full config dict (including all siblings) to `_add_node()` for each expanded node, where `expand` is skipped and other keys, like `script`, `trigger`, are processed normally. The schema just fails to reflect this intent.

Discovered during review of PR #891 (comment by @maddenp-cu).

A possible complete fix would likely require splitting `expander` into `task_expander` and `family_expander` schemas so that `script` is only permitted in the task variant. This is related to but distinct from #893, which covers the failure to validate the content of the `script`/`execution` block inside expand templates.

AnnaSmoot-NOAA:
**Synopsis**

Fixes #897. The single `expander` `$def` in `ecflow.jsonschema` did not restrict which properties could appear as siblings of the `expand` key, allowing arbitrary keys to pass `uw ecflow validate` without error. It also could not distinguish between family expanders (where `script` is invalid) and task expanders (where script is a required sibling).

This PR addresses both issues by:

-Keeping a single `expander` def (which validates only the expand key and its value)
-Using `allOf` composition at each `patternProperties` usage site to pair `expander` with the appropriate container type, `familycontainer` for `families_*` keys, `taskcontainer` for `tasks_*` keys, so that `script` is required for task expanders and rejected for family expanders
-Adding `unevaluatedProperties: false` at each usage site (all four patterns in `nodecontainer` and the two top-level suite patterns) so that unknown properties are rejected across all node levels
-Simplifying the `expand` inner schema to use `additionalProperties: {"type": "array"}` instead of `patternProperties`
-Removing a pre-existing bug where `"unevaluatedProperties": false` appeared as a literal key inside `patternProperties` at the ecflow level, where it was treated as a regex pattern rather than a schema keyword

Tests are updated to cover valid and invalid sibling properties at each node level.

Note: `script` content within task expanders is not validated by this change, which is tracked separately in #893.

**Type**

- [X] Bug fix (corrects a known issue)
- [ ] Documentation

**Impact**

**Checklist**

- [X] Where helpful, I have written comments in this PR's _Files changed_ view to assist reviewers.

### DIFF
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -342,22 +342,7 @@
       },
       "type": "object"
     },
-    "expander": {
-      "minProperties": 2,
-      "patternProperties": {
-        "^families_.+$": {
-          "$ref": "#/$defs/expander"
-        },
-        "^family_.+$": {
-          "$ref": "#/$defs/familycontainer"
-        },
-        "^task_.+$": {
-          "$ref": "#/$defs/taskcontainer"
-        },
-        "^tasks_.+$": {
-          "$ref": "#/$defs/expander"
-        }
-      },
+    "expand_block": {
       "properties": {
         "expand": {
           "additionalProperties": false,
@@ -372,12 +357,30 @@
       },
       "required": [
         "expand"
-      ],
-      "type": "object"
+      ]
     },
     "family": {
       "$ref": "#/$defs/familycontainer",
       "unevaluatedProperties": false
+    },
+    "family_expander": {
+      "$ref": "#/$defs/family_expander_base",
+      "unevaluatedProperties": false
+    },
+    "family_expander_base": {
+      "allOf": [
+        {
+          "$ref": "#/$defs/addons"
+        },
+        {
+          "$ref": "#/$defs/nodecontainer"
+        },
+        {
+          "$ref": "#/$defs/expand_block"
+        }
+      ],
+      "minProperties": 2,
+      "type": "object"
     },
     "familycontainer": {
       "allOf": [
@@ -393,7 +396,7 @@
     "nodecontainer": {
       "patternProperties": {
         "^families_.+$": {
-          "$ref": "#/$defs/expander"
+          "$ref": "#/$defs/family_expander"
         },
         "^family_.+$": {
           "$ref": "#/$defs/familycontainer"
@@ -402,7 +405,7 @@
           "$ref": "#/$defs/taskcontainer"
         },
         "^tasks_.+$": {
-          "$ref": "#/$defs/expander"
+          "$ref": "#/$defs/task_expander"
         }
       },
       "type": "object"
@@ -410,6 +413,45 @@
     "task": {
       "$ref": "#/$defs/taskcontainer",
       "unevaluatedProperties": false
+    },
+    "task_expand_block": {
+      "properties": {
+        "expand": {
+          "additionalProperties": false,
+          "minProperties": 1,
+          "patternProperties": {
+            "^.+$": {
+              "type": "array"
+            }
+          },
+          "type": "object"
+        },
+        "script": {
+          "type": "object"
+        }
+      },
+      "required": [
+        "expand"
+      ]
+    },
+    "task_expander": {
+      "$ref": "#/$defs/task_expander_base",
+      "unevaluatedProperties": false
+    },
+    "task_expander_base": {
+      "allOf": [
+        {
+          "$ref": "#/$defs/addons"
+        },
+        {
+          "$ref": "#/$defs/nodecontainer"
+        },
+        {
+          "$ref": "#/$defs/task_expand_block"
+        }
+      ],
+      "minProperties": 2,
+      "type": "object"
     },
     "taskcontainer": {
       "allOf": [
@@ -497,7 +539,7 @@
           ]
         },
         "^suites_.+$": {
-          "$ref": "#/$defs/expander"
+          "$ref": "#/$defs/family_expander"
         },
         "unevaluatedProperties": false
       },
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -342,7 +342,7 @@
       },
       "type": "object"
     },
-    "expand_block": {
+    "expander": {
       "properties": {
         "expand": {
           "additionalProperties": false,
@@ -364,22 +364,6 @@
       "$ref": "#/$defs/familycontainer",
       "unevaluatedProperties": false
     },
-    "family_expander": {
-      "allOf": [
-        {
-          "$ref": "#/$defs/addons"
-        },
-        {
-          "$ref": "#/$defs/nodecontainer"
-        },
-        {
-          "$ref": "#/$defs/expand_block"
-        }
-      ],
-      "minProperties": 2,
-      "type": "object",
-      "unevaluatedProperties": false
-    },
     "familycontainer": {
       "allOf": [
         {
@@ -394,7 +378,14 @@
     "nodecontainer": {
       "patternProperties": {
         "^families_.+$": {
-          "$ref": "#/$defs/family_expander"
+          "allOf": [
+            {
+              "$ref": "#/$defs/expander"
+            },
+            {
+              "$ref": "#/$defs/familycontainer"
+            }
+          ]
         },
         "^family_.+$": {
           "$ref": "#/$defs/familycontainer"
@@ -405,18 +396,12 @@
         "^tasks_.+$": {
           "allOf": [
             {
-              "$ref": "#/$defs/addons"
+              "$ref": "#/$defs/expander"
             },
             {
-              "$ref": "#/$defs/nodecontainer"
-            },
-            {
-              "$ref": "#/$defs/task_expand_block"
+              "$ref": "#/$defs/taskcontainer"
             }
-          ],
-          "minProperties": 2,
-          "type": "object",
-          "unevaluatedProperties": false
+          ]
         }
       },
       "type": "object"
@@ -424,27 +409,6 @@
     "task": {
       "$ref": "#/$defs/taskcontainer",
       "unevaluatedProperties": false
-    },
-    "task_expand_block": {
-      "properties": {
-        "expand": {
-          "additionalProperties": false,
-          "minProperties": 1,
-          "patternProperties": {
-            "^.+$": {
-              "type": "array"
-            }
-          },
-          "type": "object"
-        },
-        "script": {
-          "type": "object"
-        }
-      },
-      "required": [
-        "expand"
-      ],
-      "type": "object"
     },
     "taskcontainer": {
       "allOf": [
@@ -532,7 +496,14 @@
           ]
         },
         "^suites_.+$": {
-          "$ref": "#/$defs/family_expander"
+          "allOf": [
+            {
+              "$ref": "#/$defs/expander"
+            },
+            {
+              "$ref": "#/$defs/familycontainer"
+            }
+          ]
         },
         "unevaluatedProperties": false
       },
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -365,10 +365,6 @@
       "unevaluatedProperties": false
     },
     "family_expander": {
-      "$ref": "#/$defs/family_expander_base",
-      "unevaluatedProperties": false
-    },
-    "family_expander_base": {
       "allOf": [
         {
           "$ref": "#/$defs/addons"
@@ -381,7 +377,8 @@
         }
       ],
       "minProperties": 2,
-      "type": "object"
+      "type": "object",
+      "unevaluatedProperties": false
     },
     "familycontainer": {
       "allOf": [
@@ -437,10 +434,6 @@
       "type": "object"
     },
     "task_expander": {
-      "$ref": "#/$defs/task_expander_base",
-      "unevaluatedProperties": false
-    },
-    "task_expander_base": {
       "allOf": [
         {
           "$ref": "#/$defs/addons"
@@ -453,7 +446,8 @@
         }
       ],
       "minProperties": 2,
-      "type": "object"
+      "type": "object",
+      "unevaluatedProperties": false
     },
     "taskcontainer": {
       "allOf": [
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -359,7 +359,16 @@
     },
     "family": {
       "$ref": "#/$defs/familycontainer",
-      "unevaluatedProperties": false
+      "propertyNames": {
+        "anyOf": [
+          {
+            "$ref": "#/$defs/addon_property_names"
+          },
+          {
+            "pattern": "^(families|family|task|tasks)_.+$"
+          }
+        ]
+      }
     },
     "familycontainer": {
       "allOf": [
@@ -383,15 +392,45 @@
               "$ref": "#/$defs/familycontainer"
             }
           ],
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "const": "expand"
+              },
+              {
+                "$ref": "#/$defs/addon_property_names"
+              },
+              {
+                "pattern": "^(families|family|task|tasks)_.+$"
+              }
+            ]
+          }
         },
         "^family_.+$": {
           "$ref": "#/$defs/familycontainer",
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "$ref": "#/$defs/addon_property_names"
+              },
+              {
+                "pattern": "^(families|family|task|tasks)_.+$"
+              }
+            ]
+          }
         },
         "^task_.+$": {
           "$ref": "#/$defs/taskcontainer",
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "const": "script"
+              },
+              {
+                "$ref": "#/$defs/addon_property_names"
+              }
+            ]
+          }
         },
         "^tasks_.+$": {
           "allOf": [
@@ -402,14 +441,35 @@
               "$ref": "#/$defs/taskcontainer"
             }
           ],
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "const": "expand"
+              },
+              {
+                "const": "script"
+              },
+              {
+                "$ref": "#/$defs/addon_property_names"
+              }
+            ]
+          }
         }
       },
       "type": "object"
     },
     "task": {
       "$ref": "#/$defs/taskcontainer",
-      "unevaluatedProperties": false
+      "propertyNames": {
+        "anyOf": [
+          {
+            "const": "script"
+          },
+          {
+            "$ref": "#/$defs/addon_property_names"
+          }
+        ]
+      }
     },
     "taskcontainer": {
       "allOf": [
@@ -480,6 +540,26 @@
           "type": "object"
         }
       ]
+    },
+    "addon_property_names": {
+      "enum": [
+        "defstatus",
+        "events",
+        "inlimits",
+        "labels",
+        "late",
+        "limits",
+        "meters",
+        "repeat_date",
+        "repeat_datelist",
+        "repeat_datetime",
+        "repeat_day",
+        "repeat_enumerated",
+        "repeat_int",
+        "repeat_string",
+        "trigger",
+        "vars"
+      ]
     }
   },
   "properties": {
@@ -495,7 +575,16 @@
               "$ref": "#/$defs/addons"
             }
           ],
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "$ref": "#/$defs/addon_property_names"
+              },
+              {
+                "pattern": "^(families|family|task|tasks)_.+$"
+              }
+            ]
+          }
         },
         "^suites_.+$": {
           "allOf": [
@@ -506,7 +595,19 @@
               "$ref": "#/$defs/familycontainer"
             }
           ],
-          "unevaluatedProperties": false
+          "propertyNames": {
+            "anyOf": [
+              {
+                "const": "expand"
+              },
+              {
+                "$ref": "#/$defs/addon_property_names"
+              },
+              {
+                "pattern": "^(families|family|task|tasks)_.+$"
+              }
+            ]
+          }
         }
       },
       "properties": {
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -1,5 +1,25 @@
 {
   "$defs": {
+    "addon_property_names": {
+      "enum": [
+        "defstatus",
+        "events",
+        "inlimits",
+        "labels",
+        "late",
+        "limits",
+        "meters",
+        "repeat_date",
+        "repeat_datelist",
+        "repeat_datetime",
+        "repeat_day",
+        "repeat_enumerated",
+        "repeat_int",
+        "repeat_string",
+        "trigger",
+        "vars"
+      ]
+    },
     "addons": {
       "properties": {
         "defstatus": {
@@ -539,26 +559,6 @@
           ],
           "type": "object"
         }
-      ]
-    },
-    "addon_property_names": {
-      "enum": [
-        "defstatus",
-        "events",
-        "inlimits",
-        "labels",
-        "late",
-        "limits",
-        "meters",
-        "repeat_date",
-        "repeat_datelist",
-        "repeat_datetime",
-        "repeat_day",
-        "repeat_enumerated",
-        "repeat_int",
-        "repeat_string",
-        "trigger",
-        "vars"
       ]
     }
   },
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -403,7 +403,20 @@
           "$ref": "#/$defs/taskcontainer"
         },
         "^tasks_.+$": {
-          "$ref": "#/$defs/task_expander"
+          "allOf": [
+            {
+              "$ref": "#/$defs/addons"
+            },
+            {
+              "$ref": "#/$defs/nodecontainer"
+            },
+            {
+              "$ref": "#/$defs/task_expand_block"
+            }
+          ],
+          "minProperties": 2,
+          "type": "object",
+          "unevaluatedProperties": false
         }
       },
       "type": "object"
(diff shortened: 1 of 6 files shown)


src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -372,13 +357,31 @@
       },
       "required": [
         "expand"
-      ],
-      "type": "object"
+      ]
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done. Added `"type": "object"` to both `expand_block` and `task_expand_block` in `ecflow.jsonschema`, making them self-contained.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -411,6 +415,46 @@
...
+        {
+          "$ref": "#/$defs/nodecontainer"
+        },
+        {
+          "$ref": "#/$defs/task_expand_block"
+        }
+      ],
+      "minProperties": 2,
+      "type": "object"
+    },
```
maddenp-cu: Could `task_expander` and `task_expander_base` be combined, since the latter is only, as far as I can tell, referenced by the former? Do we anticipate another use case such that it makes sense for `task_expander_base` to continue to exist separately? Or am I missing a reason that they should be separate?

Same question for `family_expander` and `family_expander_base`.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -393,7 +397,7 @@
     "nodecontainer": {
       "patternProperties": {
         "^families_.+$": {
-          "$ref": "#/$defs/expander"
+          "$ref": "#/$defs/family_expander"
```
christinaholtNOAA: I have a vague recollection that `"unevaluatedProperties": false` can't be used in blocks that will be used as a reference to build up more complicated sections. Does this one need `family_expand_base`?

Here is an example that concerns me. I am using `expand.yaml` with contents that pass:

```
ecflow:
  suite_mysuite:
    families_members{{ ec.var }}:
      expand:
        var: ["01", "02"]
      task_forecast:
        script:
          execution:
            incantation: run me
            executable: forecast.exe
      tasks_post{{ ec.fff }}:
        expand:
          fff: ["012", "024"]
        script:
          execution:
            incantation: run me
            executable: post.exe
```

When I comment out the `script:` subsection of `tasks_post:`, the error seems wrong.

```
$ uw ecflow validate -c expand.yaml 
[2026-05-13T10:43:06]     INFO Validating config against internal schema: ecflow
[2026-05-13T10:43:06]    ERROR 2 schema-validation errors found in ecFlow config
[2026-05-13T10:43:06]    ERROR Error at ecflow.suite_mysuite.families_members{{ ec.var }}.tasks_post{{ ec.fff }}:
[2026-05-13T10:43:06]    ERROR   {'expand': {'fff': ['012', '024']}} does not have enough properties
[2026-05-13T10:43:06]    ERROR Error at ecflow.suite_mysuite.families_members{{ ec.var }}:
[2026-05-13T10:43:06]    ERROR   Unevaluated properties are not allowed ('task_forecast', 'tasks_post{{ ec.fff }}' were unexpected)
[2026-05-13T10:43:06]    ERROR YAML validation errors
```

The existing schema reports a simpler error that seems less confusing:

```
$ uw ecflow validate -c expand.yaml 
[2026-05-13T10:44:45]     INFO Validating config against internal schema: ecflow
[2026-05-13T10:44:45]    ERROR 1 schema-validation error found in ecFlow config
[2026-05-13T10:44:45]    ERROR Error at ecflow.suite_mysuite.families_members{{ ec.var }}.tasks_post{{ ec.fff }}:
[2026-05-13T10:44:45]    ERROR   {'expand': {'fff': ['012', '024']}} does not have enough properties
[2026-05-13T10:44:45]    ERROR YAML validation errors
```

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -402,7 +406,7 @@
           "$ref": "#/$defs/taskcontainer"
         },
         "^tasks_.+$": {
-          "$ref": "#/$defs/expander"
+          "$ref": "#/$defs/task_expander"
```
maddenp-cu: Can `task_expander` just be inlined here? I think I only see this single reference to it, such that it looks like unnecessary indirection.

I see that `family_expander` is referenced in two places so it makes sense for that one to be a separate `def` to keep it DRY.

Maybe there's an argument to be made for parallel structure between these two sets of schemas, but I think I prefer as little indirection via `$ref`s as possible, as it makes the schema easier to read and reason about, IMO.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -379,6 +364,25 @@
...
+    "family_expander_base": {
+      "allOf": [
+        {
+          "$ref": "#/$defs/addons"
+        },
+        {
+          "$ref": "#/$defs/nodecontainer"
+        },
+        {
+          "$ref": "#/$defs/expand_block"
```
maddenp-cu: Seems like this schema could also be inlined here.

But I'm curious to hear what other reviewers think about this: Is the schema easier to read / reason about with more or fewer `$ref`s? In this specific case maybe it's nice that we just say "These three `$ref`s all have to validate" and we don't have to look at what they entail _right here_. And only `expand_block` could be inlined here, as `addons` and `nodecontainer` are referenced multiple times and so should remain factored out.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -379,6 +364,25 @@
...
+        }
+      ],
+      "minProperties": 2,
```
maddenp-cu: Does this say that _any two_ properties from any of the `addons`, `nodecontainer`, or `expand_block` sections satisfy this schema, and Is this something that we potentially still need to address (maybe it's already captured in an issue, I haven't checked)? For example, are two properties defined in `addons` sufficient in a config? This seems familiar...

maddenp-cu (commented):
Some questions -- I'll re-review after we see what we want to do about these, especially since the tests might get updated if we make changes.

christinaholtNOAA (commented):
I'm concerned about how complicated this has gotten, and about the new reporting that we see in the example I've provided below.

After playing around a bit, I can solve the issue of a missing script for the tasks with only the following mods:

```
@@ -346,7 +346,10 @@
       "minProperties": 2,
       "patternProperties": {
         "^families_.+$": {
-          "$ref": "#/$defs/expander"
+          "allOf": [
+            {"$ref": "#/$defs/expander"},
+            {"$ref": "#/$defs/famil"}
+          ]
         },
         "^family_.+$": {
           "$ref": "#/$defs/familycontainer"
@@ -355,7 +358,10 @@
           "$ref": "#/$defs/taskcontainer"
         },
         "^tasks_.+$": {
-          "$ref": "#/$defs/expander"
+          "allOf": [
+            {"$ref": "#/$defs/expander"},
+            {"$ref": "#/$defs/taskcontainer"}
+          ]
         }
       },
```

This schema would still have the problem of allowing arbitrary keys under expander blocks. I'm not actually sure that we can get around this constraint easily with the necessary level of code reuse that we have and the unpredictable behavior of `unevaluatedProperties` when used deep within references when they get combined with `anyOf`, `allOf`, etc.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: Ok. I had to dig into this a little bit. You're right, neither of the  `_base` def is referenced anywhere else. It looks like the DEV-uw-tools verion of JSON scheme is  4.25.1 and supports `allOf + unevaluatedProperties: false` in the same object. I collapsed `family_expander_base` into `family_expander` and `task_expander_base` into `task_expander` directly.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: That's a good point and more DRY. 
I've inlined the `task_expander` reference into `nodecontainer's ^tasks_.+$` pattern. I've kept `family_expander` as a named def since it's referenced in two places (`^families_.+$` in `nodecontainer` and also in `familycontainer`'s `allOf`). I also updated the corresponding test to navigate the schema path directly, not looking up `$defs/task_expander`.

christinaholtNOAA:
Here's an example of a super simple way that using `unevaluatedProperties` in composed refs can cause headaches. 

The schema:
```
{
  "$defs": {
    "a": {
      "type": "object",
      "properties": {
        "foo": {"type": "integer"}
      },
      "required": ["foo"],
      "unevaluatedProperties": false
    },
    "b": {
      "type": "object",
      "properties": {
        "bar": {"type": "string"}
      }
    }
  },
  "allOf": [
      {"$ref": "#/$defs/a"},
      {"$ref": "#/$defs/b"}
  ],
  "unevaluatedProperties": false,
  "type": "object"
}
```

When one branch fails, things go off the rails. It tells us that even though `foo` is required, it is still unexpected:
```
$ echo "{foo: 1, bar: 1}" | uw config validate --schema-file oneof.jsonschema 
[2026-05-13T12:10:18]    ERROR 3 schema-validation errors found in config
[2026-05-13T12:10:18]    ERROR Error at top level:
[2026-05-13T12:10:18]    ERROR   Unevaluated properties are not allowed ('bar' was unexpected)
[2026-05-13T12:10:18]    ERROR Error at bar:
[2026-05-13T12:10:18]    ERROR   1 is not of type 'string'
[2026-05-13T12:10:18]    ERROR Error at top level:
[2026-05-13T12:10:18]    ERROR   Additional properties are not allowed ('bar', 'foo' were unexpected)
```

These two are incompatible, even when they are correct:
```
$ echo "{foo: 1, bar: a}" | uw config validate --schema-file oneof.jsonschema 
[2026-05-13T12:13:08]    ERROR 2 schema-validation errors found in config
[2026-05-13T12:13:08]    ERROR Error at top level:
[2026-05-13T12:13:08]    ERROR   Unevaluated properties are not allowed ('bar' was unexpected)
[2026-05-13T12:13:08]    ERROR Error at top level:
[2026-05-13T12:13:08]    ERROR   Unevaluated properties are not allowed ('foo' was unexpected)
```

If we take out `unevaluatedProperties` at the top-level, the one in `a` still causes us problems:
```
$ echo "{foo: 1, bar: a}" | uw config validate --schema-file oneof.jsonschema 
[2026-05-13T12:14:19]    ERROR 1 schema-validation error found in config
[2026-05-13T12:14:19]    ERROR Error at top level:
[2026-05-13T12:14:19]    ERROR   Unevaluated properties are not allowed ('bar' was unexpected)
```

It seems that only when we take the `unevaluatedProperties` at the deeper level that things actually start to work the way we expect.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: You're right that both `expand_block` and `task_expand_block` are each referenced exactly once, so both could be inlined.
I share your thought that the `allOf` with three `$refs` reads fairly cleanly as "these three constraints all apply". I'm happy to inline both if the consensus is to minimize indirection. 
What do other reviewers think? @elcarpenter @christinaholtNOAA

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: You're right to flag this. 

`expand` is required by `expand_block`, so `minProperties: 2` basically says "at least one more property beyond expand", but that second property can be anything. A config with only `expand` + `vars` and no actual task/family children passes validation. `nodecontainer` uses `patternProperties` to describe what child keys may look like, but doesn't require any of them. So yes, a `family_expander` (or the inlined task expander) with no expandable children would be silently accepted. 

Do we think this should be a separate issue?

christinaholtNOAA: I thought we had already looked at the problem and discovered that we can't require `patternProperties`.

maddenp-cu: The latest I recall about this is that we could not (yet?) find an obvious way to specify that we need _at least_ one of the `patternProperties`-spec'ed properties, plus zero or more of the others. I feel like maybe @AnnaSmoot-NOAA presented a solution that seemed to work, but we were concerned about the confusing user feedback seen when validation fails -- I remember mentioning that we already had some custom code to improve on the default JSON Schema reporting for `anyOf` / `allOf` failures, and suggesting that we might be able to combine some custom reporting code plus Anna's solution to get something that works.

Or, it may be that this requirement is too tricky to get right in JSON Schema and that we'll need to handle it in Python instead. I don't think we should give up on JSON Schema just yet, since it's better that users get feedback as early as possible, and config validation is the earliest we can get, but we might admit defeat at some point.

But, in any case, I'm not sure this `minProperties` is doing anything useful, and might just be giving false hope, in which case I think we should consider removing it. (I think there might be a couple instances of this, too.)

christinaholtNOAA: I'm in favor of removing the `minProperties` where it doesn't give us what we need.

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu: Quick question about this: Is `script` *required* under `^tasks_.+$`, e.g. `tasks_post{{ ec.fff }}`? I think so, and maybe have some ideas to share, but want to make sure about this first.

AnnaSmoot-NOAA: @christinaholtNOAA: 
Thank you, Christina! Yes, I thought code as getting a bit cumbersome. This is a much simpler approach! Using `allOf` to compose `expander` with the appropriate container type gives us the key fix (task expanders now require `script` via `taskcontainer`) without the `unevaluatedProperties` complexity that was causing the confusing error messages. I've updated the PR to implement your suggestion. The `family_expander` and `task_expand_block defs` are gone, and `nodecontainer/^suites_.+$` now use `allOf: [expander, familycontainer/taskcontainer]`. The known limitation (arbitrary sibling keys are still accepted) is acknowledged and, as you noted, likely not worth fighting with JSON Schema alone.

@maddenp-cu:
Yes, `script` is required under `^tasks_.+$`. It's enforced by `taskcontainer`, which is now (after I've added Christina's fix) composed directly via` allOf: [expander, taskcontainer]`. With this approach, commenting out script from `tasks_post{{ ec.fff }}` now produces exactly one clean error:

```
Error at ecflow.suite_mysuite.families_members{{ ec.var }}.tasks_post{{ ec.fff }}:
  'script' is a required property
```

This is now no more `unevaluatedProperties` noise at the parent level. I'm happy to consider other ideas if there's a case this approach doesn't handle well.

AnnaSmoot-NOAA:
Thank you @christinaholtNOAA for showing how my previous appraoch created confusing errors. Much appreciated. 

I've used yours fix and will push shortly. It's working much better.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: Based on the changes Christina proposed further down on this PR which I've implemented, `minProperties: 2` has been removed along with `family_expander` in the latest revision.

On the question of requiring at least one `patternProperties`-matched child: I think it's worth opening a separate issue rather than blocking this PR. I wonder it might be handled in Python rather than JSON. I can look into it and open another issue: #903

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu, line 374:
```suggestion
        "expand": {
          "additionalProperties": {
            "type": "array"
          },
          "minProperties": 1,
          "type": "object"
        }
      },
```

maddenp-cu:
Is this schema valid?
```
ecflow:
  suite_mysuite:
    families_members{{ ec.var }}:
      foo: bar
      expand:
        var: ["01", "02"]
      task_forecast:
        script:
          execution:
            incantation: run me
            executable: forecast.exe
        foo: bar
      tasks_post{{ ec.fff }}:
        expand:
          fff: ["012", "024"]
        script:
          execution:
            incantation: run me
            executable: post.exe
        foo: bar
```
I'm guessing not. I guess all those `foo: bar` items are wrong. But, given the current PR code:
```
$ uw ecflow validate -c config.yaml 
[2026-05-14T14:02:50]     INFO Validating config against internal schema: ecflow
[2026-05-14T14:02:50]     INFO Schema validation succeeded for ecFlow config
```
Whereas if I make this diff:
```
$ git diff -U0
diff --git a/src/uwtools/resources/jsonschema/ecflow.jsonschema b/src/uwtools/resources/jsonschema/ecflow.jsonschema
index 39931a3..a1a8e73 100644
--- a/src/uwtools/resources/jsonschema/ecflow.jsonschema
+++ b/src/uwtools/resources/jsonschema/ecflow.jsonschema
@@ -404 +404,2 @@
-          ]
+          ],
+          "unevaluatedProperties": false
```
I then get
```
$ uw ecflow validate -c config.yaml 
[2026-05-14T14:04:04]     INFO Validating config against internal schema: ecflow
[2026-05-14T14:04:04]    ERROR 1 schema-validation error found in ecFlow config
[2026-05-14T14:04:04]    ERROR Error at ecflow.suite_mysuite.families_members{{ ec.var }}.tasks_post{{ ec.fff }}:
[2026-05-14T14:04:04]    ERROR   Unevaluated properties are not allowed ('foo' was unexpected)
[2026-05-14T14:04:04]    ERROR YAML validation errors
```
Is that not what we want? All the current PR tests still pass, so I think that either 1. We are not validating rigorously enough, or 2. The tests are not pointing out why this use of `unevaluatedProperties` isn't acceptable. I haven't tested disallowing the other two `foo: bar` items shown above, but there might be a similar fix.

In addition to being very useful to users, this schema records our own understanding of what this mini-language allows, so I'd rather not skimp on rigor if we can possibly help it.

src/uwtools/resources/jsonschema/ecflow.jsonschema
maddenp-cu: Thanks for #903. I agree that we should work that separately.

AnnaSmoot-NOAA: I ended up closing #903 because ecflow does in fact accept a family (or expander block) with no child nodes.

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: Done, thanks!  `additionalProperties: {"type": "array"}` is cleaner. Updated.

AnnaSmoot-NOAA:
@maddenp-cu 

First on a surface level, I thought I had gotten rid of those foo and bar names and had made them more meaningful. I've gone through the code and renamed those. 

However, referencing `foo: bar` as it was before - you're right. The schema was too permissive. `foo: bar`  should've failed everywhere. Thank you for catching that. 

The cause: 
I removed `unevaluatedProperties: false` from inside the `$defs`, but I should have added it back at the usage sites in `patternProperties` instead, which is exactly what your diff demonstrates.

My fix: 
I added `unevaluatedProperties: false` to all six usage sites: all four patterns in `nodecontainer` and the two top-level patterns (`^suite_.+$`, `^suites_.+$`). The `familycontainer`/`taskcontainer`/`nodecontainer` defs themselves remain open, since they're building blocks composed at those usage sites.

I also removed a pre-existing bug: `"unevaluatedProperties": false` was sitting inside `patternProperties` as a literal key. That made it a regex pattern rather than a schema keyword.

I also updated the tests to verify that unknown properties are rejected at every node level while valid ecflow properties, while addon keys like `defstatus` and `trigger` continue to pass.

AnnaSmoot-NOAA:
I ran some CL validation as we discussed in out meeting and I'm not getting the same error as Christina, but deeper. 

The cascading `unevaluated properties are not allowed` errors on the parent nodes (the family expander and suite levels) are still present and are now one deeper. These are artifacts of `unevaluatedProperties: false` propagating up through the composed `allOf` + `$ref` chain when a child fails.

I'm going to dig into this so it might be tomorrow (Friday) that I have something. 

The positive though it The primary/first error Christina got is clearer: `'script' is a required property` is precise and actionable, vs the old `does not have enough properties`.

AnnaSmoot-NOAA:
I think I have some options on this error issue we're seeing. I'd love input on this: @maddenp-cu @christinaholtNOAA @elcarpenter. I'd lean towards either of the first two options.  

**Option 1: Python-level Cascade Suppression**

In `_validation_errors` (or the `validate` function in `validator.py`), post-process the error list: drop any `"Unevaluated properties are not allowed"` error at path `X` if a deeper error already exists at path `X.<property>`.

The logic: if `tasks_post{{ ec.fff }}` already carries a real error (`'script' is a required property`), the parent-level "unevaluated" complaints about that same key are redundant noise.

What's good about this is that the schema stays entirely unchanged and it works generically for any future pattern. The actionable error stays prominent.
The downside is it requires parsing the property name out of the `unevaluatedProperties` error message string and could accidentally suppress a legitimate cascade in a corner case.


**Option 2: Replace `unevaluatedProperties: false` with `propertyNames`**

`propertyNames` validates each key's name against a schema without relying on annotation propagation at all. It fires purely on the name, independent of whether value-schemas succeed or fail, so it never cascades.

What's good here is that it's a pure schema fix. No cascading by design. Easy to reason about. However, the error message change. Instead of `"Unevaluated properties are not allowed ('foo'…)"` users see `"'foo' does not match any of the given schemas"`. This requires explicitly enumerating every valid addon property name and tests need updating for the new message.

For `^families_.+$`:
```
"propertyNames": {
  "anyOf": [
    {"const": "expand"},
    {"enum": ["defstatus", "events", "inlimits", ...]},
    {"pattern": "^(families|family|task|tasks)_.+$"}
  ]
}
```

For `^tasks_.+$`:
```
"propertyNames": {
  "anyOf": [
    {"const": "expand"},
    {"const": "script"},
    {"enum": ["defstatus", "events", "inlimits", ...]}
  ]
}
```


**Option 3: Declare `expand` and `script` directly at the pattern level**

The cascade happens because `unevaluatedProperties: false` only considers a property "evaluated" if the subschema that matched it succeeded. If we declare `expand` and `script` for `tasks_*` in `properties` at the same schema level as the `unevaluatedProperties: false`, those keys are always marked evaluated regardless of whether refs succeed:

```
"^tasks_.+$": {
  "properties": {
    "expand": { "$ref inline..." },
    "script": {}
  },
  "required": ["expand"],
  "allOf": [
    {"$ref": "#/$defs/expander"},
    {"$ref": "#/$defs/taskcontainer"}
  ],
  "unevaluatedProperties": false
}
```

This reduces but doesn't fully eliminate the cascade. The child nodes inside a `families_*` expander (`task_forecast`, `tasks_post*`) are still evaluated via `familycontainer` --> `nodecontainer` patternProperties, which can still fail to annotate.

What's good is that this is a contained schema change and it keeps `unevaluatedProperties` semantics. However, It only partially solves the problem. The cascading for child nodes inside family expanders would still occur.


**Option 4: Custom `unevaluatedProperties` keyword handler**

Use `jsonschema`'s `validators.extend`, which is already in use in `_validation_errors`, to override the `unevaluatedProperties` keyword with a version that marks properties as "evaluated" based on whether a `patternProperties` pattern matches the key name, rather than whether the pattern's value-schema succeeded.

What's good is that it fixes the root cause cleanly for all schema patterns and schema changes are needed! However, this would be complex to implement and test. It couples the validator code tightly to jsonschema internals.

maddenp-cu:
@AnnaSmoot-NOAA I think option 2 looks promising and worth a try. We already have [some code](https://github.com/ufs-community/uwtools/blob/main/src/uwtools/config/validator.py#L92-L105) that intercepts `anyOf` / `allOf` errors and tries to give more useful feedback to the user, and I'd be interested to know if it handles the errors that would be emitted after this change. If not, that code block might be a place to add a handler for the new errors if it looks like we can improve on the default JSON Schema reporting.

In re: option 1, it could still be an option, though I'd suggest trying -- again, probably near or in the code block referenced above -- to intercept and improve on the error rather than just suppressing it. Or do suppress it, but only if we can verify that it's really useless as in the chaining case where there's a well-reported root cause that makes the chaining errors superflous.

In re: option 3, does this imply redeclaring property names in multiple schemas? The [JSON Schema docs](https://json-schema.org/understanding-json-schema/reference/object#unevaluatedproperties) discuss this downside as a rationale for `undeclaredProperties`. It's not the end of the world if this turned out to be the best option for both user and developer experience; there's just a small maintenance cost for having to keep property names in-sync across schemas.

AnnaSmoot-NOAA:
@maddenp-cu Implemented Option 2 — here's what I found:

**On the existing `anyOf`/`oneOf` handler:** `propertyNames` errors have `validator == "propertyNames"`, so they fall through to the default path in `validate()` rather than hitting the existing handler at lines 92–105. That said, the default message is already clean: `'bad_key' is not valid under any of the given schemas`. I'm not getting a cascade, and the error names the offending key directly.

**On the Option 3 maintenance concern (property names in multiple schemas):** Addressed — all 8 `propertyNames` schemas share a single `$defs/addon_property_names` enum via `$ref`. Adding a new addon property now requires updating exactly 2 places: the `addons` def (the value schema) and `addon_property_names` (the name allowlist).

`foo: bar` test case now produces exactly 1 error — `'foo' is not valid under any of the given schemas` — at the correct path. No parent-level cascade happens.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -393,23 +404,92 @@
...
+              "$ref": "#/$defs/expander"
+            },
+            {
+              "$ref": "#/$defs/familycontainer"
+            }
+          ],
+          "propertyNames": {
+            "anyOf": [
+              {
+                "const": "expand"
```
maddenp-cu, line 418: I think we should stick with this solution, but just wanted to point out that we are, in fact, redeclaring the `expand` property name, among others, using this technique: We have to declare `expand` here, while it's also declared in `expander`, under `$defs` above. These values will always have to be kept in-sync, which is what the `undeclaredProperties` mechanism is meant to help avoid. But I think that clean error messages and rigorous validation is worth the cost.

maddenp-cu (approved):
Thanks for your patience getting this through -- we had to learn multiple new things implementing it. If, when working with this new tool, we get ideas for improvements, we can circle back with fresh eyes and maybe new ideas.

AnnaSmoot-NOAA:
@maddenp-cu That's a fair point — `expand` (and `script`) do have to be kept in sync between the `expander`/`taskcontainer` defs and the `propertyNames` allowlists. The `addon_property_names` `$def` at least centralizes the addon list to a single sync point, but the expander-specific keys (`expand`, `script`) are still duplicated by necessity. Agreed that clean error messages and rigorous validation are worth that cost.

christinaholtNOAA (approved):
I like how this shaped up!

elcarpenterNOAA (approved):

PR #902, opened 2026-05-13
ecflow: add ecflow-specific execution schema requiring incantation

### LINKED ISSUE #894: ecflow: schema requires `executable` in execution block but implementation only uses `incantation`
**Current Behavior**
`execution-serial.jsonschema` and `execution-parallel.jsonschema` both declare `"required": ["executable"]`, but `_create_ecf_script` in `ecflow.py` never reads `executable` from the execution block — it only reads `incantation`. This is a mismatch between the schema contract and the implementation.

**Expected Behavior**
Users are required by the schema to supply `executable` even though it has no effect on the generated `.ecf` script. Either:

The schema should be updated to not require `executable` (and perhaps require `incantation` instead), or
The implementation should be updated to use `executable` as intended.

**To Reproduce**
- Schema (`execution-serial.jsonschema`, `execution-parallel.jsonschema`): `"required": ["executable"]`
- Implementation (`ecflow.py`,` _create_ecf_script`):
- `cmd = execution[EC.incantation]`
- `executable` is extracted but never used to construct the `.ecf` script.

**Additional Context**

Discovered during review of PR #891.

AnnaSmoot-NOAA:
**Synopsis**

Closes #894

The shared `execution-serial` and `execution-parallel` schemas require `executable`, but `ecflow.py` never reads `executable` — it only reads `incantation`. This mismatch forces users to supply a required key that has no effect on the generated `.ecf` script.

This PR implements Option A from the issue discussion: give ecFlow its own execution schema that reflects what ecFlow actually uses and requires. A small amount of duplication (`batchargs`, `envcmds`, `threads`) is preferable to requiring users to supply a meaningless key.

**Changes:**

-`ecflow.jsonschema`: `taskcontainer.script.execution` now uses an inline ecflow-specific schema instead of `anyOf` with `execution-parallel`/`execution-serial`. The new schema requires `incantation`; allows optional `batchargs`, `envcmds`, `threads`; omits `executable`, `mpiargs`, `mpicmd` (driver-only keys that `ecflow.py` never reads); and enforces `additionalProperties: false`.
-`test_schemas.py`: replaced `executable` with `incantation` in ecflow fixtures; replaced `test_schema_ecflow_refs_taskcontainer_script_incantation` (which asserted incorrect behavior) with `test_schema_ecflow_execution` that tests the new schema directly.
-`test_ecflow.py`: removed `executable` from execution block fixtures.

**Type**

**Impact**

Note: I marked Impact as "breaking change" since existing ecflow configs that include `executable` will now fail schema validation. 

**Checklist**


src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -425,14 +425,7 @@
...
                 "execution": {
-                  "anyOf": [
-                    {
-                      "$ref": "urn:uwtools:execution-parallel"
-                    },
-                    {
-                      "$ref": "urn:uwtools:execution-serial"
-                    }
-                  ]
+                  "$ref": "urn:uwtools:ecflow-execution"
```
maddenp-cu: Any reason to not just inline that subschema here and have one less file / place to have to jump to to understand the overall schema? I don't see a benefit to factoring it out if it's only used in one place.

AnnaSmoot-NOAA: That's true! Thank you and done. The test for it is updated too.

maddenp-cu (approved):

elcarpenterNOAA (approved):

PR #908, opened 2026-05-27
Fix issue with datetime re-parsing during realize

maddenp-cu:
**Synopsis**

Given `a.yaml`
```
a:
  t: !datetime 2022-02-01T00
c: !dict '{{ b }}'
```
and `b.yaml`
```
b:
  x: !datetime '{{ a.t }}'
```
Old behavior:
```
$ uw config compose --realize a.yaml b.yaml
a:
  t: 2022-02-01T00:00:00
c:
  x: datetime.datetime(2022
  2: null
  1: null
  0: null
  0): null
b:
  x: 2022-02-01T00:00:00
```
New behavior:
```
$ uw config compose --realize a.yaml b.yaml
a:
  t: 2022-02-01T00:00:00
c:
  x: 2022-02-01T00:00:00
b:
  x: 2022-02-01T00:00:00
```
I spent a lot of time looking for a better solution, but couldn't find one. This solution requires a caveat, which I've added to the docs. I think the potential for a problem with the documented behavior is low.

**Type**

- [x] Documentation

**Impact**

Technically a breaking change, though I highly doubt any users have `datetime` `repr` strings in their configs.

**Checklist**

### DIFF
--- a/src/uwtools/config/formats/ini.py
+++ b/src/uwtools/config/formats/ini.py
@@ -36,7 +36,7 @@ class INIConfig(Config):
         # INI configs have one level for the [section], and one level for each key, so are exactly
         # depth 2.
 
-        return depth == 2  # noqa: PLR2004
+        return depth == 2
 
     @classmethod
     def _dict_to_str(cls, cfg: dict) -> str:
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -138,6 +138,7 @@ def dereference(
     :param keys: The dict keys leading to this value.
     :return: The input value, with Jinja2 syntax rendered.
     """
+    report = lambda x: deref_debug("Rendering", x)
     rendered: _ConfigVal
     if isinstance(val, dict):
         keys = keys or []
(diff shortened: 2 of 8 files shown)


src/uwtools/config/formats/ini.py
```diff
@@ -36,7 +36,7 @@ def _depth_ok(depth: int) -> bool:
         # INI configs have one level for the [section], and one level for each key, so are exactly
         # depth 2.
 
-        return depth == 2  # noqa: PLR2004
```
maddenp-cu, line 39: This change is now mostly unrelated, though I initially made it to support a change that I later rolled back. Still: I think we can trust our team to use numeric constants appropriately and to assign ones with vague or important meaning to constants.

src/uwtools/config/jinja2.py
```diff
@@ -301,9 +304,10 @@ def _deref_render(val: str, context: dict, local: dict | None = None) -> str:
         rendered = val
         deref_debug("Rendering failed", val)
         for line in str(e).split("\n"):
-            deref_debug(line)
+            deref_debug(f"  Exception text: {line}")
```
maddenp-cu, line 307: During debugging I was confused about the source of these log messages so added this prefix to clarify.

elcarpenterNOAA (approved):

christinaholtNOAA (approved):

PR #911, opened 2026-05-28
Sync with main

maddenp-cu:
**Synopsis**

Bring branch `ecflow_tool` up to date with `main`.

**Type**

**Impact**

**Checklist**


src/uwtools/config/jinja2.py
```diff
@@ -401,6 +426,42 @@ def _supplement_values(
...
+            except Exception:  # noqa: BLE001
+                return nil
+        if isinstance(v, dict):
+            return {k: Pending() if (r := resolve(x)) is nil else r for k, x in v.items()}
+        if isinstance(v, list):
+            return [resolve(x) for x in v]
+        return v
+
+    kvpairs = {**(local or {}), **context}.items()
+    return {k: r for k, v in kvpairs if (r := resolve(v)) is not nil}
```
AnnaSmoot-NOAA, line 462: This seems to drop any top-level key whose value is a `UWYAMLConvert` that fails to convert. This differs from the `dict` case (line 456), where unresolvable values inside a dict become `Pending()` sentinels that trigger a graceful `UndefinedError`. Was it intentional to drop top-level unresolvable tagged values rather than also using a `Pending()` sentinel? Could this cause a template that references that key to fail silently or with a misleading error?

I'm wondering if for the ecflow tool this could cause a top-level key
of a config that can't be converted. It might disappear from the context silently rather than being held for a later iteration. Could this cause ecflow task configs to render with missing values, or silently produce incorrect output, in cases where tagged values depend on other values in the same config?

src/uwtools/config/support.py
```diff
@@ -124,12 +130,13 @@ def log_and_error(msg: str) -> Exception:
...
     """
     A loader with basic UW constructors added.
     """
-    loader = yaml.SafeLoader
-    for tag_class in (UWYAMLConvert, UWYAMLGlob, UWYAMLRemove):
+    loader = _UWYAMLLoader
+    for tag_class in (UWYAMLConvert, UWYAMLExtend, UWYAMLGlob, UWYAMLRemove):
         for tag in tag_class.TAGS:
             loader.add_constructor(tag, tag_class)
     return loader
```
AnnaSmoot-NOAA, line 142: `uw_yaml_loader()` is now cached and returns the same `_UWYAMLLoader` class instance every call. The constructors are only registered once, right? 
If the ecflow tool adds its own custom YAML constructors or tags anywhere in its code path or if future ecflow-specific tags are added to `UWYAMLConvert.TAGS` after the cache has already been populated those would be silently ignored. Is there any plan to add ecflow-specific tags or loaders, and if so, does the `@cache` need to be accounted for? I'm not sure, I'm just curious.

src/uwtools/tests/config/test_tools.py
```diff
@@ -666,6 +781,63 @@ def test_config_tools_realize__output_file_format(tmp_path):
...
+def test_config_tools_realize__reference_tagged_val_1(capsys, tmp_path):
+    yaml1 = """
+    a: !int '{{ 1 + 1 }}'
+    b: '{{ a }} is 2'
+    """
+    yaml2 = """
+    a: 2
+    b: 2 is 2
+    """
+    path = tmp_path / "config,yaml"
```
AnnaSmoot-NOAA, line 793: Should this be "config.yaml" not "config,yaml"? Repeated on lines 813 and 834.

docs/sections/user_guide/cli/tools/config.rst
```diff
@@ -198,7 +198,7 @@ Examples
 ``realize``
 -----------
 
-In ``uw`` terminology, to realize a configuration file is to transform it from its raw form into its final, usable state. The ``realize`` action can build a complete config file from two or more separate files.
+In ``uw`` terminology, to realize a configuration file is to transform it from its raw form into its final, usable state. Specificallly, the ``realize`` action replaces `YAML aliases <https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/>`_ with their anchored content, and renders :jinja2:`Jinja2 expressions<templates>`. It can build a complete config file from two or more separate files.
```
AnnaSmoot-NOAA, line 201: Typo: Specificallly --> Specifically

maddenp-cu: This PR is just to sync recent changes from `main` onto `ecflow_tool` so that, hopefully next week, we can have a smooth merge of this branch back to `main`. So I won't make any changes here, but will take notes and make any necessary fixes on `main` directly.

AnnaSmoot-NOAA (approved):

docs/sections/user_guide/cli/tools/config.rst
AnnaSmoot-NOAA: Sounds good. Based on that, I approve.

src/uwtools/config/jinja2.py
maddenp-cu: This function creates an updated _context_ dict, from which values can be drawn to render expressions in the main config dict. If a value is not present here, but would be needed to render an expression in the main config, that expression will remain in place, potentially to be rendered later when a better context dict is available. The context dict is repeatedly re-created, so items dropped in one iteration may will be present again later. If we can find a failing use case, we can write a failing test and then make a fix.

src/uwtools/config/support.py
maddenp-cu: I don't anticipate any new YAML tags to support ecFlow per se.

src/uwtools/tests/config/test_tools.py
maddenp-cu: WIll fix on `main`.

elcarpenterNOAA (approved):

PR #913, opened 2026-05-29
Fix to allow drivers to support both hardlink and symlink actions.

christinaholtNOAA:
**Synopsis**

Fixes a bug in which the `Linker.go()` iotaa task isn't unique when harlinks and symlinks are both requested. It adds an optional name field to the `go()` task and has all the `stager` methods that rely on the task pass a unique name. The same was done for the `Copier.go()` task for consistency and to break the pattern.

**Type**

- [ ] Documentation

**Impact**

**Checklist**


maddenp-cu:
I feel like the information I need is in the PR and I'm failing to put it together, but: How does this bug get triggered? My intuition is that e.g. a `Copier` object would be instantiated, its `go` task method would be called, that that `go` task would be the only task with the name `File copies` in the task tree. How do duplicate / ambiguous task names appear? If there are calls to `go` on _other_ `Copier` (or `Linker`) objects, those should not interfere, as there should be a separate namespace for that task tree. If this isn't the case, there might be a more subtle bug, maybe in `iotaa`.

christinaholtNOAA:
I found the bug when running the MPAS driver with both the `files_to_link` and `files_to_hardlink` sections in the config. When both were included, the `provisioned_rundir` task would only run the first (`files_to_hardlink` in this case). I assumed this is because the `.go()` task, despite being associated with a different object, was still registered as a task with the same name in the `iotaa` tree, so I parameterized it so I could give it a different name.

I was able to duplicate it with the test. I initially had a copy/paste error in this PR where I'd named both the hardlink and symlink `.go()` tasks identically. When I did that, the new test failed with no linked files present.

christinaholtNOAA:
Per group discussion, this will become a bugfix release on the current release. The underlying issue may be fixed in `iotaa` soon; otherwise, this change may be cherry-picked into `main` later.

src/uwtools/resources/info.json
```diff
@@ -1,4 +1,4 @@
 {
   "buildnum": "0",
-  "version": "2.16.0"
+  "version": "2.17.0"
```
maddenp-cu: I think you'll want to set this to `2.16.1` and then run `make meta` in your `base` conda environment (with the `condev` package installed) to bring `recipe/meta.json` up to date, too.

elcarpenterNOAA (approved):

src/uwtools/drivers/enkf.py
```diff
@@ -47,7 +47,7 @@ def background_files(self):
             Linker(
                 config=files.as_dict(),
                 target_dir=self.rundir,
-            ).go()
+            ).go("background_files")
```
maddenp-cu: FWIW, I think this will just appear in a log message, and it might look cleaner without the underscore. The fact that this method is called `background_files` will probably not be familiar to most people seeing the log; maybe an English prose description would serve them best (which might just be what you have without the underscore).

Same comment applies to the other `.go()` calls below...

maddenp-cu (commented):
A couple small suggetion

maddenp-cu (approved):

PR #921, opened 2026-06-04
Improve !extend handling and associated documentation

maddenp-cu:
**Synopsis**

- Detect when a user is trying to `!extend` a tagged, unrealized Jinja2 expression literal sequence and provide a more specific error message.
- Be clearer in the documentation about the limitations of `!extend`.
- Add a tip about addling and slicing lists in Jinja2 expressions to do things `!extend` cannot.

**Type**

- [x] Documentation

**Impact**

**Checklist**


recipe/meta.yaml
```diff
@@ -28,7 +28,7 @@ requirements:
     - requests >=2.32,<2.35
 test:
   requires:
-    - docformatter 1.7.7 # 1.7.8 seems buggy
+    - docformatter 1.7.7.* # 1.7.8 seems buggy
```
maddenp-cu, line 31: This fixes a complaint from `conda create`.

recipe/meta.yaml
```diff
@@ -16,7 +16,7 @@ requirements:
     - python {{ pymin }}
     - setuptools
   run:
-    - ecflow 5.15.*
+    - ecflow 5.16.*
```
maddenp-cu, line 19: `ecflow 5.17` is also available, but has a conflict with something else in the `uwtools` environment, which I didn't look into. We can address that later if/when we want to upgrade `ecflow` further. The bump from `5.15` to `5.16` was low-hanging fruit.

christinaholtNOAA (approved):

Issue #868, opened 2026-04-16
Tool action to provision an ecFlow server.


christinaholtNOAA:
**Description**

**Alternatives Considered**

**Acceptance Criteria**

**Additional Context**

**Current Version**

AnnaSmoot-NOAA:
I had two questions after reading this issue and did some digging. Here is what I've found: 

**1. Does the ecflow Python API expose server-start functionality?** 
No. The ecFlow Python API is designed exclusively for interacting with an already-running server. There is no `ecflow.Client.start_server()` or equivalent, so `ecflow_server` must be launched via `subprocess`. There is no Python API alternative.

**2. In considering a Python API for `openssl` over `subprocess`: is using `cryptography` Python package preferable to `openssl`?** 
This is all quite new to me, so here is my understanding:  

_`cryptography` pros:_
- has no shell exposure or private keys, no dependency on `openssl` binary in PATH, has native Python error handling, is easy to mock, has full control over file permissions (write to file only after permission is set like we want), and can generate ecflow-specific files (`server.key`, `server.crt`). 

_`cryptography` cons:_
- `dh2048.pem` generation is not directly supported by the `cryptography`  package and will likely still require `subprocess` to call `openssl dhparam`. 


_`openssl` subprocess pros:_ 
-has full control over file permissions (write to file only after permission is set like we want), can generate ecflow-specific files (`server.key`, `server.crt`, and `dh2048.pem` ). 

_`openssl` subprocess cons:_ 
-keys/passphrases can be exposed to shell env, requires correct `openssl` version in PATH, must parse CLI output, harder to unit test


**Possible implementation approach:** 
1. Start the server: use `subprocess` to call `ecflow_server --port=<port>`
2. SSL certificate generation: Use the `cryptography` package for `server.key` and `server.crt`  as this is safer, more Pythonic, easier to test, but use `subprocess` to `openssl dhparam` for `dh2048.pem` since `cryptography` doesn't support DH parameter generation. OR just use `openssl`?
3. File permissions: Create the empty file with `600` permissions first, then write to file.  
4. Reuse exsisting certs: Check `~/.ecflowrc/ssl` first; reuse if all three files are present, generate if not.

Please correct me if I'm wrong on any of these. I'm also not sure about my 2. question, which is also 2. in the implementation approach.

christinaholtNOAA:
Notes from the mob session:
[...]

PR #920, opened 2026-06-04
Add initial ecflow tool with suite-definition support

maddenp-cu:
**Synopsis**

Merge initial `ecflow` tool work, which provides a UW YAML language for describing ecFlow suite definitions and a `uw ecflow realize` tool/action pair for creating files containing such definitions, as well as supporting ecf scripts. Support for configuring and running ecFlow servers will follow soon.

Note that everything in this PR was already reviewed via PRs to the `ecflow_tool` branch. This PR just merges those vetted changes to `main`.

**Type**

- [x] Documentation
- [x] Enhancement (adds new functionality)

**Impact**

- [x] This is a non-breaking change (existing functionality continues to work as expected)

**Checklist**

- [x] I have added myself and any co-authors to the PR's _Assignees_ list.
- [x] I have reviewed the documentation and have made any updates necessitated by this change.
- [x] Where helpful, I have written comments in this PR's _Files changed_ view to assist reviewers.

PR #923, opened 2026-06-09
Add uw ecflow server tool

AnnaSmoot-NOAA:
**Synopsis**

Adds a new `uw ecflow server` action (CLI + API) that provisions and runs and ecflow server. Closes issue #868. 

The action:

- Starts an `ecflow_server` on an available TCP port, picking a random port in the registered range (`1024-49151`) and retrying until a free one is found, or using an explicit `--port`.
- Enables SSL security by default: reuses existing certificates in `$HOME/.ecflowrc/ssl`, or generates them (`server.key`, `server.crt`, `dh2048.pem`) via `openssl`, creating each file with `600` permissions before writing. A `--insecure` switch disables SSL.
- Runs in the foreground and shuts down gracefully on `CTRL-C` (`SIGINT`), terminating the server via `ecflow_client --terminate` rather than emitting a stack trace.
- Follows the uwtools convention: log/status messages go to `stderr`; with `--report`, a JSON object of server connection variables (`ECF_HOST`, `ECF_PORT`, and `ECF_SSL` when secure) is written to `stdout` for downstream consumers (e.g. piped to jq).
- Is driven by a required UW YAML config validated against a new internal schema (`ecflow-server.jsonschema`). `ECF_HOME` is required and any of the documented `ECF_*` server settings (enumerated in the schema) are optional and exported into the server environment. `ECF_PORT` and `ECF_SSL` are intentionally excluded from config. The tool manages the port (CLI `--port`) and SSL state (`--insecure`) itself.

CLI: 
```
uw ecflow server [-h] [--version] [--config-file PATH] [--insecure]
                 [--port PORT] [--report] [--quiet] [--verbose]
```

The feature was verified against a live `ecflow_server` (ecFlow 5.15.2), exercising both secure and insecure paths, port auto-selection, the `--report` JSON, and clean `CTRL-C` shutdown. 

Note for Reviewers: 
This branch was rebuilt cleanly on top of `main` after the `ecflow_tool` branch was squash-merged (#920); it contains server-tool changes only. The arbitrary `ECF_SSL`-cert-path option is intentionally deferred (MVP uses the default `$HOME/.ecflowrc/ssl`) and will be a follow-up.

**Type**

- [X] Documentation

**Impact**

**Checklist**

- [X] Where helpful, I have written comments in this PR's _Files changed_ view to assist reviewers.


src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+    Generate a 2048-bit RSA private key (no password) at ``path``.
+
+    :param path: Destination for the private key file.
+    :raises UWError: If ``openssl`` reports failure.
+    """
+    log.info("Generating SSL private key: %s", path)
+    success, _ = run_shell_cmd(f"openssl genrsa -out {path} 2048")
+    if not success:
+        msg = f"Failed to generate SSL private key at {path}"
+        raise UWError(msg)
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+    :param key_path: Path to the existing RSA private key.
+    :raises UWError: If ``openssl`` reports failure.
+    """
+    hostname = socket.gethostname()
+    log.info("Generating SSL certificate: %s", path)
+    cmd = f"openssl req -x509 -key {key_path} -new -out {path} -days 3650 -subj '/CN={hostname}'"
+    success, _ = run_shell_cmd(cmd)
+    if not success:
+        msg = f"Failed to generate SSL certificate at {path}"
+        raise UWError(msg)
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+) -> None:
+    """
+    Start an ecFlow server on an available TCP port with SSL security enabled.
+
+    The server runs in the foreground until interrupted (e.g. via CTRL-C), at which point it is
+    shut down gracefully via ``ecflow_client --terminate``.
+
+    :param config: A ``dict``, a ``YAMLConfig``, or a path to a YAML file providing server settings.
+    :param port: TCP port to use (``None`` => pick a random available port from 1024-49151).
+    :param insecure: Start the server without SSL security.
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+                    thread.terminal.set()
+                    thread.error = f"Requested port {candidate} is unavailable"
+                    return
+                log.debug("Port %s already in use", candidate)
+                continue
+            thread.terminal.set()
+            thread.error = f"ecflow_server failed on port {candidate}: {e.stdout}"
+            for line in (e.stdout or "").split("\n"):
+                log.error(line)
+            return
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Good catch. Done.

src/uwtools/api/ecflow.py
```diff
@@ -45,6 +46,39 @@ def realize(
...
 
+def server(
+    config: _YAMLConfig | dict | Path | str | None = None,
+    port: int | None = None,
+    insecure: bool = False,
+    report: bool = False,
+    stdin_ok: bool = False,
+) -> bool:
+    """
+    Start an ecFlow server on an available TCP port with SSL security enabled.
```
```suggestion
    Start an ecFlow server on an available TCP port, optionally with SSL security enabled.
```

src/uwtools/api/ecflow.py
```diff
@@ -45,6 +46,39 @@ def realize(
...
+
+    If no port is specified, a random port in the registered range (1024-49151) is chosen,
+    retrying if the port is unavailable. SSL certificates are read from or written to
+    ``$HOME/.ecflowrc/ssl``. Use ``insecure`` to skip SSL. Use ``report`` to emit a JSON report
+    of the server details (hostname, port) to ``stdout``.
```
```suggestion
    of the server details (e.g. hostname, port) to ``stdout``.
```
Because it might not contain `SSL` if the documentation above is accurate, and might include other items if we add them later.

src/uwtools/api/ecflow.py
```diff
@@ -45,6 +46,39 @@ def realize(
...
+
+    :param config: A ``dict``, a ``YAMLConfig``, or a path to a YAML config file (``None`` => read
+        ``stdin``).
+    :param port: TCP port to use; overrides config value (``None`` => pick a random available port
+        from 1024-49151).
+    :param insecure: Start the server without SSL security.
+    :param report: Output server details (hostname, port) as JSON to ``stdout``.
```
```suggestion
    :param report: Output server details (e.g. hostname, port) as JSON to ``stdout``.
```

src/uwtools/resources/jsonschema/ecflow-server.jsonschema
```diff
@@ -0,0 +1,91 @@
...
+    "ECF_URL_CMD": {
+      "description": "Command executed by the client to open a browser with the job status URL.",
+      "type": "string"
+    }
+  },
+  "required": [
+    "ECF_HOME"
+  ],
+  "type": "object"
+}
```
maddenp-cu: I think this schema should be incorporated into the existing `ecflow.jsonschema`. It was recently updated on `main` with `server` and `suitedef` sub-schemas, and the former is currently a stub that accepts any object (dict, mapping). That stub can be fleshed out with this content. The PR branch may need to merge `main` in to see these changes if that hasn't happened yet.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+    :param port: A specific port to use, or ``None`` to pick a random port (1024-49151).
+    :param insecure: Start the server without SSL security.
+    """
+    thread = cast(_ServerThread, current_thread())
+    fixed = port is not None
+    cmd = ["ecflow_server", *([] if insecure else ["--ssl"])]
+    while not thread.terminal.is_set():
+        # ecFlow accepts ports only in the registered range 1024-49151.
+        candidate = port if fixed else random.randint(1024, 49151)  # noqa: S311
+        log.debug("Trying to start server on port %s", candidate)
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: This is based on ecflow's own error messages: 
```
ERROR Server port 54321 not set correctly.
ERROR The port numbers are divided into three ranges.
ERROR    o the Well Known Ports, (require root permission)      0 -1023
ERROR    o the Registered Ports,                             1024 -49151
ERROR    o Dynamic and/or Private Ports.                    49151 -65535
ERROR
ERROR Please set in the range 1024-49151 via argument or
ERROR Server/src/environment.cfg.h or set the environment variable ECF_PORT.
```

src/uwtools/cli.py
```diff
@@ -1047,6 +1087,15 @@ def _add_arg_output_dir(group: Group, required: bool = False) -> None:
     )
 
 
+def _add_arg_port(group: Group) -> None:
+    group.add_argument(
+        _switch(STR.port),
+        help="TCP port for the ecFlow server (default: random from 1024-49151)",
```
maddenp-cu: This `1024-49151` range recurs so frequently that I think we should try to DRY it out, perhaps importing it from the `uwtools.ecflow` module, maybe as module-level variables like `ECFLOW_PORT_MIN` and `ECFLOW_PORT_MAX`.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,257 @@ def realize(
...
+    """
+    while not thread.terminal.is_set():
+        if thread.port:
+            cmd = f"ecflow_client {ssl_opt}--port {thread.port} --ping"
+            success, _ = run_shell_cmd(cmd=cmd, quiet=True)
+            if success:
+                log.info("Server started on port %s", thread.port)
+                if report_vars is not None:
+                    _server_report(port=thread.port, report_vars=report_vars)
+                break
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Yes, he loop could busy-spin a core and hammer `ecflow_client` while the port is unknown or ping is failing.
Fix: 
-In `ecflow.py` I added `sleep(0.2) at the end of each `_server_wait` iteration. This is placed after the success `break`. Alos added top-level `from time import sleep` 
-In `test_ecflow.py` the two looping tests now patch `sleep` and assert it's called once with `0.2`

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,269 @@ def realize(
...
+def _provision_ssl(ssl_dir: Path = _SSL_DIR) -> None:
+    """
+    Ensure SSL certificates exist in ``$HOME/.ecflowrc/ssl``.
+
+    If all required files exist, logs that they will be reused. If the directory exists but is
+    missing one or more required files, logs an error and raises ``UWError``. If the directory does
+    not exist, creates it and generates the required SSL files using ``openssl``.
+
+    :param ssl_dir: Directory where SSL certificates are stored. Defaults to
+        ``$HOME/.ecflowrc/ssl``.
```
maddenp-cu: On the one hand, this seems maybe premature, as if a non-default directory is used, I don't think we have a way for the user to specify it for use later, like via the CLI. On the other hand, we discussed the fact that ecFlow apparently supports specification of a non-standard location via (IIRC) the `ECF_SSL` variable, so we might make use of this parameter later. On the balance, I think I'd prefer to remove this -- unless I'm missing a reason why it needs to exist for current functionality -- and add it back later if and when we expose that flexibility to users. Thoughts?

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,269 @@ def realize(
...
+    Create an empty file with owner-only (600) permissions.
+
+    The file is created with restricted permissions before any content is written, so that sensitive
+    key material is never exposed with open permissions.
+
+    :param path: Path of the file to create.
+    """
+    fd = os.open(str(path), os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+    os.close(fd)
+    path.chmod(0o600)
```
maddenp-cu: I think in manual testing we determined that `openssl` creates files with the correct permissions. Do we still need this function or the calls to it? We could simplify if not, but it'd be worth a manual test.

src/uwtools/resources/jsonschema/ecflow-server.jsonschema
```diff
@@ -0,0 +1,91 @@
+{
+  "additionalProperties": false,
+  "properties": {
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: I'm going to fix the PR description. The schema is intentional.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: But is this true? I think we tested it and saw that `openssl` uses the correct permissions. If you verified that this is factual, we should keep the `_ssl_touch()` calls, but it would be worth checking.

src/uwtools/ecflow.py
```diff
@@ -402,6 +412,269 @@ def realize(
...
+    if not success:
+        msg = f"Failed to generate SSL certificate at {path}"
+        raise UWError(msg)
+
+
+def _ssl_generate_dhparam(path: Path) -> None:
+    """
+    Generate 2048-bit Diffie-Hellman parameters at ``path``.
+
+    This step can take several minutes.
```
maddenp-cu: Is this true? Can we time it and see? That seems inordinately long.

maddenp-cu (commented):
First pass comments... I want to read more, especially the tests, and to test manually from this branch, but can provide this initial feedback until then.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: A good test might be to run `umask 0000`, which sets an as-permissive-as-possible mask, then remove the ecFlow SSL directory, then run `uw ecflow server` once and check the permission on the security files it creates. (Then exit that shell so that you don't accidentally create more files/directories with too-open permissions!)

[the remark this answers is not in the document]
maddenp-cu: I just tried this locally and got
```
-rw-rw-rw- 1 maddenp maddenp  428 Jun  9 16:05 dh2048.pem
-rw-rw-rw- 1 maddenp maddenp 1099 Jun  9 16:05 server.crt
-rw------- 1 maddenp maddenp 1704 Jun  9 16:05 server.key
```
IIRC `server.crt` is the file that's actually used to encrypt communications, so I think these permissions are too permissive and that we should keep the "touch" logic that we have.

maddenp-cu:
I did a test with `ECF_HOME` in my config set to a directory that doesn't exist yet and get
```
[2026-06-09T16:10:56]    ERROR Failed to launch ecflow_server: [Errno 2] No such file or directory: PosixPath('/home/maddenp/git/uwtools/server')
[2026-06-09T16:10:57]    ERROR Failed to launch ecflow_server: [Errno 2] No such file or directory: PosixPath('/home/maddenp/git/uwtools/server')
```
I think we should create the directory the user requests on their behalf and not fail. We usually just execute something like `path.mkdir(parents=True, exist_ok=True)` to ensure the a directory exists before proceeding.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,570 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+        patch.object(ecflow, "_ssl_generate_key") as mock_key,
+        patch.object(ecflow, "_ssl_generate_cert") as mock_cert,
+        patch.object(ecflow, "_ssl_generate_dhparam") as mock_dh,
+    ):
+        ecflow._provision_ssl(ssl_dir)
+    assert ssl_dir.is_dir()
+    mock_key.assert_called_once_with(ssl_dir / "server.key")
+    mock_cert.assert_called_once_with(ssl_dir / "server.crt", ssl_dir / "server.key")
+    mock_dh.assert_called_once_with(ssl_dir / "dh2048.pem")
+    assert logged("Creating SSL directory")
```
christinaholtNOAA, line 738: This test would be stronger without mocking.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,570 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+
+
+def test_ecflow__ssl_generate_key__success(tmp_path):
+    path = tmp_path / "server.key"
+    with patch.object(ecflow, "run_shell_cmd", return_value=(True, "")) as mock_cmd:
+        ecflow._ssl_generate_key(path)
+    mock_cmd.assert_called_once()
+    assert f"-out {path}" in mock_cmd.call_args[0][0]
+    # The key file is pre-created with owner-only permissions before openssl writes to it.
+    assert oct(path.stat().st_mode)[-3:] == "600"
```
christinaholtNOAA, line 747: This one might also benefit from running the code.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,570 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+
+def test_ecflow__ssl_generate_cert__success(tmp_path):
+    cert_path = tmp_path / "server.crt"
+    key_path = tmp_path / "server.key"
+    with patch.object(ecflow, "run_shell_cmd", return_value=(True, "")) as mock_cmd:
+        ecflow._ssl_generate_cert(cert_path, key_path)
+    mock_cmd.assert_called_once()
+    assert f"-out {cert_path}" in mock_cmd.call_args[0][0]
+    # The cert file is pre-created with owner-only permissions before openssl writes to it.
+    assert oct(cert_path.stat().st_mode)[-3:] == "600"
```
christinaholtNOAA, line 767: Same here.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,570 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+def test_ecflow__ssl_generate_dhparam__success(tmp_path):
+    path = tmp_path / "dh2048.pem"
+    with (
+        patch.object(ecflow, "run_shell_cmd", return_value=(True, "")) as mock_cmd,
+        patch.object(ecflow, "_ssl_touch") as mock_touch,
+    ):
+        ecflow._ssl_generate_dhparam(path)
+    mock_touch.assert_called_once_with(path)
+    mock_cmd.assert_called_once()
+    assert f"-out {path}" in mock_cmd.call_args[0][0]
```
christinaholtNOAA: Also here.

src/uwtools/ecflow.py
```diff
@@ -402,6 +413,272 @@ def realize(
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.port: int | None = None
+        self.terminal = Event()
+        self.error: str | None = None
+
+
+def server(
+    config: dict | YAMLConfig | Path | None = None,
```
christinaholtNOAA: I was thinking that we'd decided that the config should be required for telling the server where to spin up. I understand that in the CLI we might not have that requirement since we accept configs from STDIN, but here, it seems like we would want that. Maybe I missed a shift in that requirement along the way, though.

christinaholtNOAA (changes requested):

src/uwtools/ecflow.py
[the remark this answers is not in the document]
AnnaSmoot-NOAA: I see. I thought things were fine, then copilot made be second guess myself, but this is more than that too. Touch alone isn't enough. It seems like whether or not `0600` survives depends on how openssl builds writes the file. 

I kept `_ssl_touch` and added `chmod(0o600)` after the openssl call.

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Good catch. I updated the API docstring, and CLI `--report` help and `server-help.out`.

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/resources/jsonschema/ecflow-server.jsonschema
AnnaSmoot-NOAA: Yes, I agree with you and that took a bit of shuffling. The `server` block now lives alongside `suitedef` under the top-level `ecflow:` key, with `ECF_HOME` required and `additionalProperties: false`.

src/uwtools/ecflow.py
maddenp-cu: I think we determined that, in fact, `openssl` relies on the user's umask and that we _do_ need to do something here, so I think my previous comment is OBE.

However, could we potentially use [this](https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch) instead? It appears to support setting the mode right in the call. If that works, we could eliminate this function in just call this `pathlib` method instead.

EDIT: But in another comment I suggested that we make a `umask` call  in the shell that `openssl` runs in, which might be the simplest option of all. Check that first.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: Hmm. If the permissions set by our "touch" are replaced by `openssl` and then fixed by the `chmod` just below, then there's a span of time during which the files are unprotected, so we need to think of something else.

An approach might be to prefix the `openssl` calls with `umask 0077`, which masks out all "group" and "everyone" bits. So, instead of a command like
```
f"openssl genrsa -out {path} 2048
```
we might have
```
f"umask 0077 && openssl genrsa -out {path} 2048
```
I think this will ensure that the files `openssl` creates have restricted permissions, and that no `touch` or `chmod` will be necessary. I'll test and follow up here.

src/uwtools/ecflow.py
maddenp-cu: In my tests it took like a few seconds. 10 maybe? I don't think it takes long enough to mention, to be honest.

src/uwtools/ecflow.py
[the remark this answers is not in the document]
maddenp-cu: Yes, prefixing the `openssl` commands with `umask 0077 && ` appears to work as hoped, without the need for any touch/chmod logic.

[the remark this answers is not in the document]
AnnaSmoot-NOAA: Oh yes, that's a good idea. Thanks, Paul! Done: I switched to prefixing the `openssl` commands with `unmask 0077 &&` and dropped `_ssl_touch/ chmod`. The generated `server.key`, `server.crt`, and `dh2048.pem` now come out as `0600` and shouldn't have a window of open permissions.

src/uwtools/cli.py
AnnaSmoot-NOAA: Thanks. I had noticed that those port numbers were repetitive and wasn't sure if it was too bad or not. I've applied your proposed fix with  `ECFLOW_PORT_MIN` and `ECFLOW_PORT_MAX`.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: Yes, I think that's reasonable to remove it. It was working as more than dead code in some tests, but I rewrote that; done.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: OBE. Resolved above with `unmask` fix.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: You're right. It only took about 1 second when I just now ran it. Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: @christinaholtNOAA  True! Thanks! I've reworked the test to use uw-tools own `_ssl_generate_*`, asserts the actually certificate files. I also democked the `_server_start` test to use the real `tmp_path` run dirs. 

@maddenp-cu Done. `_server_start` now does `path.mkdir(parents=True, exist_ok=True)` before launching the server so `ECF_HOME` gets created instead of failing.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: True and that was insightful. Thank you.  `test_ecflow__ssl_generate_key__success` and `test_ecflow__ssl_generate_cert__success` now invoke real `openssl` and assert the generated files are `0600`, so they genuinely verify the `umask 0077 &&` permission guarantee rather than checking a mock-created file.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: @christinaholtNOAA & @maddenp-cu 
I made the generate key and cert tests real. I hesitate to do that for the dhparam. The `umask 0077 &&` permission mechanism is identical across all three helpers, so the key and cert tests (which now run real `openssl`) already verify the real `0600` outcome of that shared code path. 
Running dhparam for real adds essentially no extra permission coverage, but DH-parameter generation time seems to be slow and variable. That makes a flake/timeout risk in our CI. 

For perspective, the key and cert tests run in about ~0.05-0.07s, but the dhparam has taken ~0.8-48.7s. What do you both think? Should I leave it mocked or make it real?

src/uwtools/ecflow.py
AnnaSmoot-NOAA: You're right and done. The config should be required. That was a change we made. I removed `=None` default so `config` is not a required argument on `server()` functions, matching realize.

src/uwtools/api/ecflow.py
```diff
@@ -45,6 +46,40 @@ def realize(
...
+
+    :param config: A ``dict``, a ``YAMLConfig``, or a path to a YAML config file (``None`` => read
+        ``stdin``).
+    :param port: TCP port to use; overrides config value (``None`` => pick a random available port
+        from ``ECFLOW_PORT_MIN``-``ECFLOW_PORT_MAX``).
+    :param insecure: Start the server without SSL security.
+    :param report: Output server details (e.g. hostname, port) as JSON to ``stdout``.
+    :param stdin_ok: OK to read from ``stdin``?
+    :return: ``True``.
+    """
```
maddenp-cu, line 74:
```suggestion
    f"""
    Start an ecFlow server on an available TCP port, optionally with SSL security enabled.

    If no port is specified, a random port in the registered range
    (``{ECFLOW_PORT_MIN}``-``{ECFLOW_PORT_MAX}``) is chosen,
    retrying if the port is unavailable. SSL certificates are read from or written to
    ``$HOME/.ecflowrc/ssl``. Use ``insecure`` to skip SSL. Use ``report`` to emit a JSON report
    of the server details (e.g. hostname, port) to ``stdout``.

    :param config: A ``dict``, a ``YAMLConfig``, or a path to a YAML config file (``None`` => read
        ``stdin``).
    :param port: TCP port to use; overrides config value (``None`` => pick a random available port
        from ``{ECFLOW_PORT_MIN}``-``{ECFLOW_PORT_MAX}``).
    :param insecure: Start the server without SSL security.
    :param report: Output server details (e.g. hostname, port) as JSON to ``stdout``.
    :param stdin_ok: OK to read from ``stdin``?
    :return: ``True``.
    """
```
Does it work to make this docstring an f-string and interpolate in the actual values of `ECFLOW_PORT_MIN` and `ECFLOW_PORT_MAX`? They'll have to be `import`ed, of coruse. Since this docstring will be used in the generated API docs in ReadTheDocs, it would be nice if the true min/max port values were reported.

src/uwtools/strings.py
```diff
@@ -156,10 +156,12 @@ class _STR(_ValsMatchKeys):
     gsi: str = _
     hardlink: str = _
     help: str = _
+    hostname: str = _
     hsi: str = _
     htar: str = _
     input_file: str = _
     input_format: str = _
+    insecure: str = _
```
maddenp-cu, line 164: Would this string, and possibly all 4 of the new additions here, better reside in the `_EC` container, rather than this generic `_STR` container, given that they all pertain to ecFlow?

src/uwtools/tests/test_schemas.py
```diff
@@ -715,6 +715,21 @@ def test_schema_ecflow_suitedef(ecflow_config):
...
+    errors = schema_validator("ecflow", "properties", "ecflow", "properties", "server")
+    # Basic spec:
+    assert not errors({"ECF_HOME": "/path/to/ecf"})
+    # ECF_HOME is required:
+    assert "'ECF_HOME' is a required property" in errors({})
+    # Additional properties are not allowed:
+    assert "Additional properties are not allowed" in errors(
+        {"ECF_HOME": "/path/to/ecf", "ECF_BOGUS": "x"}
+    )
+    # ECF_CHECKMODE is constrained to an enum:
```
maddenp-cu, line 754: The schema places useful constraints on many more of the supported `ECF_` values. I think we should test those constraints here, too. For values that simply must be e.g. strings, a [loop like this](https://github.com/ufs-community/uwtools/blob/main/src/uwtools/tests/test_schemas.py#L488-L489) is a concise way to check all those. It looks like a handful need to be integers > 0, and `ECF_MICRO` has special requirements. These are all worth locking down here.

src/uwtools/cli.py
```diff
@@ -28,6 +28,7 @@
 import uwtools.api.template
 import uwtools.config.jinja2
 import uwtools.rocoto
+from uwtools.ecflow import ECFLOW_PORT_MAX, ECFLOW_PORT_MIN
```
maddenp-cu: There's a suggestion to import these into `uwtools.api.ecflow` to use in rendering the docstring there. If that's done, then this module (`uwtools.cli`) could access them as `uwtools.api.eclfow.ECFLOW_PORT_MAX` and `uwtools.api.ecflow.ECFLOW_PORTMIN` and this `import` statement wouldn't be needed. It doesn't _really_ matter, but limiting our use of non-API references here as much as possible helps keep the interface between the CLI and the rest of the codebase as clean as possible.

src/uwtools/ecflow.py
```diff
@@ -402,6 +417,264 @@ def realize(
...
+            f"SSL directory {ssl_dir} exists but is missing required file(s): {missing}. "
+            "Provide all required files or remove the directory to allow regeneration."
+        )
+        raise UWError(msg)
+    log.info("Creating SSL directory %s", ssl_dir)
+    ssl_dir.mkdir(parents=True, exist_ok=True)
+    _ssl_generate_key(ssl_dir / _SSL_KEY)
+    _ssl_generate_cert(ssl_dir / _SSL_CERT, ssl_dir / _SSL_KEY)
+    _ssl_generate_dhparam(ssl_dir / _SSL_DHPARAM)
+    log.info("SSL credentials written to %s", ssl_dir)
```
```suggestion
    existing = [f for f in _SSL_FILES if (_SSL_DIR / f).exists()]
    if len(existing) == len(_SSL_FILES):
        log.info("Using existing SSL certificates in %s", _SSL_DIR)
        return
    if existing:
        missing = sorted(set(_SSL_FILES) - set(existing))
        msg = (
            f"SSL directory {_SSL_DIR} exists but is missing required file(s): {missing}. "
            "Provide all required files or remove the directory to allow regeneration."
        )
        raise UWError(msg)
    log.info("Creating SSL directory %s", _SSL_DIR)
    ssl_dir.mkdir(parents=True, exist_ok=True)
    _ssl_generate_key(_SSL_DIR / _SSL_KEY)
    _ssl_generate_cert(_SSL_DR / _SSL_CERT, _SSL_DIR / _SSL_KEY)
    _ssl_generate_dhparam(_SSL_DIR / _SSL_DHPARAM)
    log.info("SSL credentials written to %s", _SSL_DIR)
```
I couldn't think of a reason to keep the `ssl_dir` variable, but please ignore this suggestion if I missed a reason why we need it.

src/uwtools/ecflow.py
```diff
@@ -402,6 +417,264 @@ def realize(
...
+
+
+def _ssl_generate_key(path: Path) -> None:
+    """
+    Generate a 2048-bit RSA private key (no password) at ``path``.
```
maddenp-cu, line 462:
```suggestion
    Generate a 2048-bit RSA private key (no password) at path.
```
In docstrings outside `uwtools.api`, there's no need to use the Sphinx/RST double-backtick syntax, because these docstrings will never be rendered to HTML. The extra syntax just IMO makes the docstrings harder for developers to read, and to maintain.

I'm making this suggestion just on this line, but I think it would be better to apply it throughout this PR -- but I'd like to know if @christinaholtNOAA and @elcarpenterNOAA concur. I did a grep and see that we do have this Sphinx/RST syntax in a few other places, and I could remove those separately if we agree that we should be consistent about using that syntax only in API docstrings.

Thoughts?

src/uwtools/tests/api/test_ecflow.py
```diff
@@ -8,6 +8,50 @@
 from uwtools.exceptions import UWError
 
 
+def test_api_ecflow_server():
+    path = Path("foo")
+    with patch.object(ecflow, "_server") as mock_server:
+        result = ecflow.server(config=path, port=12345, insecure=False, report=True)
+    mock_server.assert_called_once_with(config=path, port=12345, insecure=False, report=True)
+    assert result is True
```
maddenp-cu, line 16:
```suggestion
def test_api_ecflow_server():
    path = Path("foo")
    with patch.object(ecflow, "_server") as _server:
        result = ecflow.server(config=path, port=12345, insecure=False, report=True)
    _server.assert_called_once_with(config=path, port=12345, insecure=False, report=True)
    assert result is True
```
When possible, `uwtools` style is to name mocks directly after the objects they mock.

Here and below...

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+        ecflow._provision_ssl()
+
+
+def test_ecflow__provision_ssl__creates_dir_and_files(logged, tmp_path):
+    ssl_dir = tmp_path / ".ecflowrc" / "ssl"
+
+    def fake_run(cmd, **_kwargs):
+        # Emulate openssl by creating the file named after its "-out" argument.
+        out = Path(cmd.split("-out ", 1)[1].split()[0])
+        out.write_text("x")
```
maddenp-cu:
```suggestion
        Path(cmd.split("-out ", 1)[1].split()[0]).touch()
```
Equivalent but simpler?

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+        raises(UWError, match="Failed to generate SSL certificate"),
+    ):
+        ecflow._ssl_generate_cert(cert_path, key_path)
+
+
+def test_ecflow__ssl_generate_dhparam__success(tmp_path):
+    # DH-parameter generation is slow and entropy-dependent, so (unlike the key and cert tests,
+    # which exercise real openssl) mock the shell call and assert the umask-prefixed command. The
+    # resulting 0600 permissions are verified for real by the key and cert tests, which share this
+    # umask mechanism.
```
maddenp-cu, line 786: This is a good point. My tests showed `dhparam` taking 16-20 seconds, so we probably do not want to do that in our unit tests. We don't need to prove that `openssl` works, only that we invoked it correctly.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+    path = tmp_path / "dh2048.pem"
+    with patch.object(ecflow, "run_shell_cmd", return_value=(True, "")) as mock_cmd:
+        ecflow._ssl_generate_dhparam(path)
+    mock_cmd.assert_called_once()
+    cmd = mock_cmd.call_args[0][0]
+    assert cmd.startswith("umask 0077 && ")
```
maddenp-cu:
```suggestion
    assert cmd.startswith("umask 0077 && openssl dhparam")
```
We could stand to be slightly more precise here.

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+
+
+def test_ecflow__ssl_generate_key__failure(tmp_path):
+    path = tmp_path / "server.key"
+    with (
+        patch.object(ecflow, "run_shell_cmd", return_value=(False, "error")),
+        raises(UWError, match="Failed to generate SSL private key"),
+    ):
+        ecflow._ssl_generate_key(path)
+
```
maddenp-cu, line 758:
```suggestion
    assert not path.is_file()
```

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+def test_ecflow__ssl_generate_cert__failure(tmp_path):
+    cert_path = tmp_path / "server.crt"
+    key_path = tmp_path / "server.key"
+    with (
+        patch.object(ecflow, "run_shell_cmd", return_value=(False, "error")),
+        raises(UWError, match="Failed to generate SSL certificate"),
+    ):
+        ecflow._ssl_generate_cert(cert_path, key_path)
+
```
maddenp-cu, line 780:
```suggestion
    assert not cert_path.is_file()
    assert not key_path.is_file()
```

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+
+def test_ecflow__ssl_generate_dhparam__failure(tmp_path):
+    path = tmp_path / "dh2048.pem"
+    with (
+        patch.object(ecflow, "run_shell_cmd", return_value=(False, "error")),
+        raises(UWError, match="Failed to generate DH parameters"),
+    ):
+        ecflow._ssl_generate_dhparam(path)
+
```
maddenp-cu, line 804:
```suggestion
    assert not path.is_file()
```

src/uwtools/tests/test_ecflow.py
```diff
@@ -693,3 +694,574 @@ def test_ecflow_validate__suite_with_optional_properties():
...
+        patch.object(ecflow, "YAMLConfig", return_value=mock_cfg),
+        patch.object(ecflow, "validate") as mock_validate,
+        patch.object(ecflow, "_provision_ssl"),
+        patch.object(ecflow, "_ServerThread") as mock_thread_cls,
+        patch.object(ecflow.signal, "signal"),
+        patch.object(ecflow, "_server_wait"),
+    ):
+        mock_thread_cls.return_value.error = None
+        ecflow.server(config=config_path, port=3141)
+    mock_validate.assert_called_once_with(mock_cfg)
```
maddenp-cu: EDIT: I have mixed feelings about these tests, as there's a lot of duplication and more mocking than is necessary (see the suggestion below for an example of how the mocking could be simplified, but please don't apply it). The mocking may make the tests faster, which is generally good, and maybe having each test test just one thing might also be good. After thinking about it a little, I think I'm inclined to keep the tests as they are and see in the longer term how we like them.

What follows is my first batch of comments, just for context:

```suggestion
def test_ecflow_server__validates_config(tmp_path):
    path = tmp_path / "server.yaml"
    config = {"ecflow": {"server": {"ECF_HOME": str(tmp_path)}}}
    path.write_text(yaml.dump(config))
    with (
        patch.object(ecflow, "_ServerThread") as _ServerThread,  # noqa: N806
        patch.object(ecflow, "_server_wait"),
    ):
        _ServerThread.return_value.error = None
        ecflow.server(config=path, port=3141, insecure=True)
```
I think this simpler formulation tests the same. It's easy enough to write a valid (or invalid, if the test requires it) config to `tmp_path` and then let `ecflow.server()` read and act on it. In this test, and probably others, we don't want to start a server thread or wait for it, so I think it makes sense to continue mocking those. But, otherwise, I think we can mock a lot less.

Please consider whether many of the tests below can be simplified in similar ways.

This test avoid mocking `_provision_ssl()` by passing `insecure=True` but, below, test `test_ecflow_server__calls_provision_ssl()` tests that, so we could decide to mock that here, _not_ pass `insecure=True` here, and then make the assertion that `_provision_ssl()` gets called here, rather than having a separate test for it. It just seems to me like we have a lot of test code here with significant duplication. I feel like we could have a smaller number of simpler tests, each of which make multiple assertions. Probably some behavior is mutually exclusive in a way that will require multiple tests, but maybe there could be less.

maddenp-cu (commented):
This is looking good. I'm down to mostly nitpicks and style suggestions.

src/uwtools/api/ecflow.py
AnnaSmoot-NOAA: Yes, that's neater and done.

src/uwtools/strings.py
AnnaSmoot-NOAA: Yes. I considered that. I lean towards leaving them in `_STR` because they're all used as CLI tokens (subcommand/argument/flag names) in `cli.py`, which is where the rest of the CLI tokens are, including the ecflow subcommands (`realize` and `validate`). Moving to `_EC` would split the ecflow tokens across two containers. Also, `EC` already has a `server` entry that makes the ecflow config-schema key , so adding a CLI `server` would make 1 symbol for 2 different concepts. I'm happy to move them though.

src/uwtools/tests/test_schemas.py
AnnaSmoot-NOAA: That's a good call. Thank you! I've added constraint coverage:  a loop over all the string-typed `ECF_` keys, integer/minimum checks for `ECF_CHECKINTERVAL`, `ECF_INTERVAL` (≥1) and `ECF_TASK_THRESHOLD` (≥0), and the single-character `minLength`/`maxLength`/type checks for `ECF_MICRODEF`.

src/uwtools/cli.py
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: You're right. Thank you! I see no reason to keep it. Switched to using `_SSL_DIR`.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: That's a good point an I agree, Paul. I could use a fixture. That would shorten each test but test count remains the same to pinpoint failures.  I'm thinking something like: 

from types import SimpleNamespace

```
@fixture
def server_mocks():
    """
    Patch the collaborators of ecflow.server() and yield handles to the mocks.

    The default config is a minimal valid secure-server config; tests that need
    different data can reassign `m.cfg.data` before calling ecflow.server().
    """
    cfg = Mock()
    cfg.data = {"ecflow": {"server": {"ECF_HOME": "/ecf"}}}
    with (
        patch.object(ecflow, "YAMLConfig", return_value=cfg) as yamlconfig,
        patch.object(ecflow, "validate") as validate,
        patch.object(ecflow, "_provision_ssl") as provision_ssl,
        patch.object(ecflow, "_ServerThread") as thread_cls,
        patch.object(ecflow.signal, "signal") as signal,
        patch.object(ecflow, "_server_wait") as server_wait,
    ):
        thread = thread_cls.return_value
        thread.error = None
        yield SimpleNamespace(
            config_path=Path("/some/server.yaml"),
            cfg=cfg,
            yamlconfig=yamlconfig,
            validate=validate,
            provision_ssl=provision_ssl,
            thread_cls=thread_cls,
            thread=thread,
            signal=signal,
            server_wait=server_wait,
        )
```

Then the tests: 

```
def test_ecflow_server__validates_config(server_mocks):
    m = server_mocks
    ecflow.server(config=m.config_path, port=3141)
    m.validate.assert_called_once_with(m.cfg)


@mark.parametrize(
    "config", [{"ECF_HOME": "/ecf"}, YAMLConfig({"ECF_HOME": "/ecf"}), Path("s.yaml")]
)
def test_ecflow_server__accepts_config_types(server_mocks, config):
    ecflow.server(config=config, port=3141)
    server_mocks.yamlconfig.assert_called_once_with(config)


def test_ecflow_server__calls_provision_ssl(server_mocks):
    ecflow.server(config=server_mocks.config_path, port=3141)
    server_mocks.provision_ssl.assert_called_once()


def test_ecflow_server__insecure_skips_ssl(server_mocks):
    ecflow.server(config=server_mocks.config_path, port=3141, insecure=True)
    server_mocks.provision_ssl.assert_not_called()


def test_ecflow_server__secure_sets_ecf_ssl_env(server_mocks):
    ecflow.server(config=server_mocks.config_path, port=3141)
    rundir, env, port, insecure = server_mocks.thread_cls.call_args.kwargs["args"]
    assert rundir == Path("/ecf")
    assert env["ECF_SSL"] == "1"
    assert port == 3141
    assert insecure is False


def test_ecflow_server__insecure_unsets_ecf_ssl_env(server_mocks):
    m = server_mocks
    m.cfg.data = {"ecflow": {"server": {"ECF_HOME": "/ecf", "ECF_SSL": "1"}}}
    ecflow.server(config=m.config_path, port=3141, insecure=True)
    _, env, _, insecure = m.thread_cls.call_args.kwargs["args"]
    assert "ECF_SSL" not in env
    assert insecure is True


def test_ecflow_server__starts_thread_and_waits(server_mocks):
    m = server_mocks
    ecflow.server(config=m.config_path, port=3141)
    m.signal.assert_called_once_with(ecflow.signal.SIGINT, m.signal.call_args.args[1])
    m.thread.start.assert_called_once()
    m.server_wait.assert_called_once()
    m.thread.join.assert_called_once()


def test_ecflow_server__raises_on_thread_error(server_mocks):
    m = server_mocks
    m.thread.error = "ecflow_server failed on port 3141: boom"
    with raises(UWError, match="ecflow_server failed on port 3141"):
        ecflow.server(config=m.config_path, port=3141)

def test_ecflow_server__shutdown_terminates(server_mocks):
    m = server_mocks
    m.thread.port = 54321
    with patch.object(ecflow, "run_shell_cmd", return_value=(True, "")) as run:
        ecflow.server(config=m.config_path, port=54321)
        shutdown = m.signal.call_args.args[1]
        shutdown(2, None)
    m.thread.terminal.set.assert_called()
    cmd = run.call_args.kwargs["cmd"]
    assert "--ssl" in cmd
    assert "--port 54321" in cmd
    assert "--terminate=yes" in cmd


def test_ecflow_server__report_vars_include_config(server_mocks):
    m = server_mocks
    m.cfg.data = {"ecflow": {"server": {"ECF_HOME": "/ecf", "ECF_LOG": "my.log"}}}
    with patch.object(ecflow.socket, "gethostname", return_value="server.hostname.com"):
        ecflow.server(config=m.config_path, port=54321, report=True)
    assert m.server_wait.call_args.kwargs["report_vars"] == {
        "ECF_HOME": "/ecf",
        "ECF_LOG": "my.log",
        "ECF_HOST": "server.hostname.com",
        "ECF_SSL": "1",
    }
```

src/uwtools/strings.py
maddenp-cu: I see what you mean, and can see it both ways, too. I'm not bothered by `server` living in `EC` and being used both for the CLI and for the schema lookup: "server" in both contexts refers to an ecFlow server. And `insecure` seems particularly ecFlow-related: I can't think of what else that concept would apply to in `uwtools`.

I wonder if what this dilemma actually suggests is that multiple string containers are not a great idea. Outside the scope of this PR, of course, but maybe it's something we should consider.

Let's see if other reviewers have thoughts on this, then I think you should decide what's best.

src/uwtools/tests/test_ecflow.py
maddenp-cu: If that works, it's be great to see less boilerplate in each test. I think it would help readers concentrate on the important parts of the test.

maddenp-cu (approved):
I think the only open questions for me were 1. Whether to shift some strings from `STR` to `EC` (hopefully other reviewers will weigh in) and 2. Can some of the unit tests be shortened for less boilerplate (the fixture @AnnaSmoot-NOAA suggests might be a good solution).

But whatever decisions are made on those two items, I'm happy with this update if it doesn't change drastically. Nice work!

src/uwtools/strings.py
AnnaSmoot-NOAA: @christinaholtNOAA @elcarpenter What do you guys think about this discussion and using multiple string containers or not?

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: It's interesting that it's that variable. I'm curious about the risk you mention for "flake/timeout" in the tests. How does that manifest?

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: Can you not also remove the existing mocks to allow the actual command to run in a subprocess, though?

src/uwtools/tests/test_ecflow.py
christinaholtNOAA: I see here that this is likely sufficient.

src/uwtools/ecflow.py
christinaholtNOAA: I'm fine with removing them. I don't think it needs to be in this PR, though.

src/uwtools/strings.py
christinaholtNOAA: I'm okay either way.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/test_ecflow.py
maddenp-cu: I think we should not run `openssl dhparam`. We don't need to test `openssl` -- we can trust that it works correctly if called correctly.

I'm less concerned about CI than I am about long delays in test turnaround for developers locally, who (should) run the tests frequently while working. Some tests predating this PR already run IMO too long, and I've been looking for an opportunity to dig in and find out why. I'd hate to introduce new long-running tests.

BTW, Gemini has a good answer to "why is the openssl dhparam time so variable?":
> The time taken by `openssl dhparam` is highly variable because the command forces the CPU to search for a specific, mathematically complex type of prime number called a safe prime. Because there is no simple pattern to prime numbers, OpenSSL has to rely on a trial-and-error process, making the generation time highly unpredictable.

src/uwtools/tests/test_ecflow.py
maddenp-cu: @christinaholtNOAA I've written something similar elsewhere, just trying to coalesce the arguments here:

1. `openssl dhparam` takes too long to run, and we don't need to test that `openssl` works, just that we call it correctly. We could unmock the other two, but I don't see much point because...
2. If we were actually going to check some properties of the files created by `openssl`, then it would be worth letting them run. But since we're not doing that (and probably won't) and just testing file existence, I don't see much value in actually running `openssl`. Well, there's probably value in running it _once_, in the cheapest way possible, to ensure that it remains available in our virtual environment. I think we do have a test that actually runs it, though.

src/uwtools/strings.py
maddenp-cu: I vote for just keeping what's here. I may do a PR to combine the string containers on `main` some time, to avoid having to spend time deciding where a specific string belongs in the future.

src/uwtools/ecflow.py
maddenp-cu: I'll take care of whatever remains in a separate PR.

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: **1. The test can run long enough to be killed:** 
The dhparam test can run long enough to be killed. It works by randomly searching for special prime numbers so it takes a different amount of time to run each time. If one run goes past the test timeout, the test gets killed and the whole CI job fails, even though nothing is actually broken. Running it again usually passes, which is the typical sign of an unreliable or flaky test.

**2. Randomness and test time:** 
Generating the dhparam needs a lot of random data from the system. A fresh CI container may not have much available, so openssl can pause and wait for more. That can look like the test hanging for a long time.

**3. Parallel testing:** 
With parallel testing, this test could cause slow downs, and run even slower by all the tests competing on the same CPUs.   

**In practice:** 
CI fails now and then on PRs that have nothing to do with SSL, tests pass when you re-run them, and overall CI gets slower because the job has to wait for the slowest openssl run. None of these are real bugs in the code. They are reliability and speed problems

**Mocking this test doesn't cost us much:** 
The only real behavior unique to the dhparam step is that openssl writes a file with owner-only (0600) permissions. Those permissions come from the shared umask 0077 && prefix. The key and cert tests already run real openssl with that same prefix and check the real 0600 result, and they finish in about 0.05 to 0.07 seconds each, so the permission behavior is already tested for real. The dhparam test only needs to confirm that we call openssl with the correct umask prefix and output path, which the mock checks instantly and reliably.

Option: 
If we really want real coverage of the dhparam, we could add a separate test that runs real openssl but is marked as "slow" and left out of the normal CI run (for example, only run when a specific flag or environment variable is set). That way we still generate the real file once in a while without putting a slow, unpredictable step in every PR's CI. I've never done this before, but could be an option.

christinaholtNOAA (approved):

src/uwtools/tests/test_ecflow.py
AnnaSmoot-NOAA: Done.

src/uwtools/tests/api/test_ecflow.py
AnnaSmoot-NOAA: Done.

PR #933, opened 2026-06-17
Support ECF_SSL configuration for uw ecflow server

### LINKED ISSUE #924: Support configurable SSL certificate directory for uw ecflow server
**Description**

Allow users to configure a custom `ECF_SSL` certificate directory for `uw ecflow server`, instead of being limited to the default ecFlow location, `$HOME/.ecflowrc/ssl`.

Currently the path is hard-coded in `src/uwtools/ecflow.py`:
`_SSL_DIR = Path.home() / ".ecflowrc" / "ssl"`

`_provision_ssl()` always operates on `_SSL_DIR`: it reuses the three required files (`dh2048.pem`, `server.cert`, `server.key`). If all exist raise `UWError` if the directory is incomplete, and otherwise generates them there. There is no way to point the server at a different cert store. 

This feature would add an optional config key (e.g. `ECF_SSL_DIR`, or a dedicated `ssl:` block — naming TBD) that overrides the default. Example:
```
ECF_HOME: /path/to/run
ECF_SSL_DIR: /shared/certs/ecflow   # new: custom SSL cert directory
```

-When set, `_provision_ssl()` applies its existing reuse/error/generate logic to that directory (with `600` permissions on generated files).
-When unset, behavior is unchanged (default to `$HOME/.ecflowrc/ssl`).
-The `--insecure` path is unaffected.

This supports site-managed, shared, per-experiment, or read-only cert stores where the default home location isn't writable or appropriate.

**Acceptance Criteria**

-Default-only (current behavior): keep `$HOME/.ecflowrc/ssl` as the sole location. Rejected because it can't accommodate shared or read-only cert stores, and forces all experiments on a host to share one cert set.
-Symlinking into the default location: users could manually symlink custom certs into `$HOME/.ecflowrc/ssl`. Rejected as the primary solution because it's manual, error-prone, and doesn't compose well with per-experiment configs, though it may still be relevant to the client-side constraint below.

**Additional Context**

Follow-up from #868 / PR #923; the default-only behavior was an intentional MVP scope decision.

One constraint to resolve: `ecflow_client` (including job-side child commands) must locate the same `server.crt` to connect over SSL, and the client defaults to `$HOME/.ecflowrc/ssl`. A custom server-side directory only works end-to-end if the client can also find the cert (via `ECF_SSL` server-name conventions and/or symlinking into the default location). This should be confirmed and documented as part of the work.

AnnaSmoot-NOAA:
- Add `_check_ssl_named()` helper to verify that a named certificate triplet (`<host>.<port>.crt`/`.key`/`.pem`) exists in `$HOME/.ecflowrc/ssl` before starting the server
- Update `server()` to route to `_check_ssl_named()` when `ECF_SSL` is a `<host>.<port>` string, and to `_provision_ssl()` for the default case (boolean or absent)
- Fix inline comment and JSON schema description for `ECF_SSL`: the string form is a `<host>.<port>` prefix, not a certificate file path
- Fix docs (`yaml/ecflow.rst`, `cli/tools/ecflow.rst`): replace incorrect file-path examples with `<host>.<port>` examples; clarify that all cert files must reside in `$HOME/.ecflowrc/ssl`
- Update existing SSL string test to use `<host>.<port>` format; add tests for `_check_ssl_named()` (all files present, missing files)

Closes #924

**Synopsis**

Adds proper support for the `ECF_SSL` config key in `uw ecflow server`. When `ECF_SSL` is absent or true, the default certificate triplet (`server.crt`/`server.key`/`dh2048.pem`) is auto-provisioned in `$HOME/.ecflowrc/ssl`. When `ECF_SSL` is a `<host>.<port>` string, the named triplet (`<host>.<port>.crt`/`.key`/`.pem`) is verified to exist there before the server starts. When `ECF_SSL` is `false` or `--insecure` is given, SSL is disabled entirely. All cert files must reside in `$HOME/.ecflowrc/ssl`; no custom directory is supported by ecflow.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/docs/sections/user_guide/cli/tools/ecflow.rst
+++ b/docs/sections/user_guide/cli/tools/ecflow.rst
@@ -125,7 +125,7 @@ Examples
 
      uw ecflow server --config-file server.yaml --port 54321
 
-* To start a server using a custom SSL certificate directory:
+bi* To start a server using a custom SSL certificate directory:
 
   .. code-block:: text
 
@@ -140,7 +140,7 @@ Examples
          ECF_HOME: /path/to/run
          ECF_SSL_DIR: /shared/certs/ecflow
 
-  ``ECF_SSL_DIR`` overrides the default certificate location (``$HOME/.ecflowrc/ssl``) for both the server and any ``ecflow_client`` commands that inherit the environment, including job child commands. See :ref:`ecflow_workflows` for full server YAML documentation.
+  ``ECF_SSL_DIR`` overrides the default certificate location (``$HOME/.ecflowrc/ssl``) used by ``uw ecflow server`` to provision SSL certificates. See :ref:`ecflow_workflows` for full server YAML documentation.
 
 * To emit a JSON report of the server details to ``stdout``:
 
--- a/docs/sections/user_guide/cli/tools/ecflow.rst
+++ b/docs/sections/user_guide/cli/tools/ecflow.rst
@@ -125,6 +125,23 @@ Examples
 
      uw ecflow server --config-file server.yaml --port 54321
 
+* To start a server using a custom SSL certificate directory:
+
+  .. code-block:: text
+
+     uw ecflow server --config-file server.yaml
+
+  Where ``server.yaml`` contains:
+
+  .. code-block:: yaml
+
+     ecflow:
+       server:
+         ECF_HOME: /path/to/run
+         ECF_SSL_DIR: /shared/certs/ecflow
+
+  ``ECF_SSL_DIR`` overrides the default certificate location (``$HOME/.ecflowrc/ssl``) for both the server and any ``ecflow_client`` commands that inherit the environment, including job child commands. See :ref:`ecflow_workflows` for full server YAML documentation.
+
 * To emit a JSON report of the server details to ``stdout``:
 
   .. code-block:: text
@@ -137,13 +154,15 @@ Examples
 
      {
        "vars": {
+         "ECF_HOME": "/path/to/run",
          "ECF_HOST": "server.hostname.com",
          "ECF_PORT": "54321",
-         "ECF_SSL": "1"
+         "ECF_SSL": "1",
+         "ECF_SSL_DIR": "/shared/certs/ecflow"
        }
      }
 
-  ``ECF_SSL`` is included only when SSL security is enabled (i.e. when ``--insecure`` is not given).
+  ``ECF_SSL`` is included only when SSL security is enabled (i.e. when ``--insecure`` is not given). ``ECF_SSL_DIR`` is included only when set in the config.
 
 .. _cli_ecflow_validate_examples:
 
--- a/docs/sections/user_guide/cli/tools/ecflow.rst
+++ b/docs/sections/user_guide/cli/tools/ecflow.rst
@@ -125,11 +125,11 @@ Examples
 
      uw ecflow server --config-file server.yaml --port 54321
 
-* To start a server using a custom SSL certificate:
+* To start a server using a named certificate triplet:
 
   .. code-block:: text
 
-     uw ecflow server --config-file server.yaml
+     uw ecflow server --config-file server.yaml --port 8888
 
   Where ``server.yaml`` contains:
 
@@ -138,9 +138,9 @@ Examples
      ecflow:
        server:
          ECF_HOME: /path/to/run
-         ECF_SSL: /shared/certs/server.crt
+         ECF_SSL: myhost.8888
 
-  ``ECF_SSL`` accepts ``true`` to enable SSL using the default certificate location (``$HOME/.ecflowrc/ssl/server.crt``), or a path string to use a specific certificate file. See :ref:`ecflow_workflows` for full server YAML documentation.
+  ``ECF_SSL`` accepts ``true`` to enable SSL using the default certificate triplet (``server.crt`` / ``server.key`` / ``dh2048.pem``) in ``$HOME/.ecflowrc/ssl``, or a ``<host>.<port>`` prefix string to select a named certificate triplet (``<host>.<port>.crt`` / ``.key`` / ``.pem``) from the same directory. The ``--port`` value must match the port in the prefix when using a named triplet. See :ref:`ecflow_workflows` for full server YAML documentation.
 
 * To emit a JSON report of the server details to ``stdout``:
 
@@ -156,8 +156,8 @@ Examples
        "vars": {
          "ECF_HOME": "/path/to/run",
          "ECF_HOST": "server.hostname.com",
-         "ECF_PORT": "54321",
-         "ECF_SSL": "/shared/certs/server.crt"
+         "ECF_PORT": "8888",
+         "ECF_SSL": "myhost.8888"
        }
      }
 
--- a/docs/sections/user_guide/cli/tools/ecflow.rst
+++ b/docs/sections/user_guide/cli/tools/ecflow.rst
@@ -125,7 +125,7 @@ Examples
 
      uw ecflow server --config-file server.yaml --port 54321
 
-bi* To start a server using a custom SSL certificate directory:
+* To start a server using a custom SSL certificate:
 
   .. code-block:: text
 
@@ -138,9 +138,9 @@ bi* To start a server using a custom SSL certificate directory:
      ecflow:
        server:
          ECF_HOME: /path/to/run
-         ECF_SSL_DIR: /shared/certs/ecflow
+         ECF_SSL: /shared/certs/server.crt
 
-  ``ECF_SSL_DIR`` overrides the default certificate location (``$HOME/.ecflowrc/ssl``) used by ``uw ecflow server`` to provision SSL certificates. See :ref:`ecflow_workflows` for full server YAML documentation.
+  ``ECF_SSL`` accepts ``true`` to enable SSL using the default certificate location (``$HOME/.ecflowrc/ssl/server.crt``), or a path string to use a specific certificate file. See :ref:`ecflow_workflows` for full server YAML documentation.
 
 * To emit a JSON report of the server details to ``stdout``:
 
@@ -157,12 +157,11 @@ bi* To start a server using a custom SSL certificate directory:
          "ECF_HOME": "/path/to/run",
          "ECF_HOST": "server.hostname.com",
          "ECF_PORT": "54321",
-         "ECF_SSL": "1",
-         "ECF_SSL_DIR": "/shared/certs/ecflow"
+         "ECF_SSL": "/shared/certs/server.crt"
        }
      }
 
-  ``ECF_SSL`` is included only when SSL security is enabled (i.e. when ``--insecure`` is not given). ``ECF_SSL_DIR`` is included only when set in the config.
+  ``ECF_SSL`` is included only when SSL security is enabled (i.e. when ``--insecure`` is not given and ``ECF_SSL`` is not ``false`` in the config).
 
 .. _cli_ecflow_validate_examples:
 
--- a/docs/sections/user_guide/yaml/ecflow.rst
+++ b/docs/sections/user_guide/yaml/ecflow.rst
@@ -295,10 +295,10 @@ The ``uw ecflow server`` command reads configuration from an ``ecflow: server:``
 
 All keys in the ``server:`` block are passed as environment variables to ``ecflow_server``. ``ECF_HOME`` is the only required key.
 
-``ECF_SSL`` -- Optional. Controls SSL for the server. Accepts a boolean or a path string:
+``ECF_SSL`` -- Optional. Controls SSL for the server. Accepts a boolean or a ``<host>.<port>`` prefix string:
 
-- ``true`` (default when ``--insecure`` is not given): Enable SSL using the auto-provisioned certificate at ``$HOME/.ecflowrc/ssl/server.crt``.
-- A path string (e.g. ``/shared/certs/server.crt``): Enable SSL using the specified certificate file. The matching ``server.key`` and ``dh2048.pem`` must already exist alongside it.
+- ``true`` (default when ``--insecure`` is not given): Enable SSL using the auto-provisioned default certificate triplet (``server.crt`` / ``server.key`` / ``dh2048.pem``) in ``$HOME/.ecflowrc/ssl``.
+- A ``<host>.<port>`` string (e.g. ``myhost.8888``): Enable SSL using the named certificate triplet (``<host>.<port>.crt`` / ``.key`` / ``.pem``) in ``$HOME/.ecflowrc/ssl``. The files must already exist; they are not auto-generated.
 - ``false``: Disable SSL (equivalent to passing ``--insecure``).
 
 For example, to force use of a named certificate triplet when running on a static port:
--- a/docs/sections/user_guide/yaml/ecflow.rst
+++ b/docs/sections/user_guide/yaml/ecflow.rst
@@ -282,6 +282,32 @@ Multiple expand variables of the same length may be provided together:
        MEM: ["01", "02"]
        LABEL: ["ctrl", "pert"]
 
+Server Configuration
+--------------------
+
+The ``uw ecflow server`` command reads configuration from an ``ecflow: server:`` block:
+
+.. code-block:: yaml
+
+   ecflow:
+     server:
+       ECF_HOME: /path/to/run
+
+All keys in the ``server:`` block are passed as environment variables to ``ecflow_server``. ``ECF_HOME`` is the only required key.
+
+``ECF_SSL_DIR`` -- Optional. A path to a directory containing SSL certificate files (``dh2048.pem``, ``server.crt``, ``server.key``). When set, overrides the default location of ``$HOME/.ecflowrc/ssl``. For example:
+
+.. code-block:: yaml
+
+   ecflow:
+     server:
+       ECF_HOME: /path/to/run
+       ECF_SSL_DIR: /shared/certs/ecflow
+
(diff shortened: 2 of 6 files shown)


src/uwtools/ecflow.py
```diff
@@ -541,7 +544,8 @@ def server(
     validate(cfg)
     server_cfg = cfg.data[STR.ecflow][STR.server]
     if not insecure:
-        _provision_ssl()
+        ssl_dir = Path(v) if (v := server_cfg.get("ECF_SSL_DIR")) is not None else None
+        _provision_ssl(ssl_dir=ssl_dir)
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Fixed. Replaced the `umask 0077 && openssl … {path}` shell-string approach with `os.umask(0o077)` + `subprocess.run([list], capture_output=True)` .

docs/sections/user_guide/yaml/ecflow.rst
```diff
@@ -282,6 +282,32 @@ Multiple expand variables of the same length may be provided together:
...
+.. code-block:: yaml
+
+   ecflow:
+     server:
+       ECF_HOME: /path/to/run
+       ECF_SSL_DIR: /shared/certs/ecflow
+
+Because ``ECF_SSL_DIR`` is a native ecflow environment variable, both ``ecflow_server`` and ``ecflow_client`` (including job child commands) read certificates from this directory automatically. Jobs that inherit ``ECF_SSL_DIR`` from the server environment will locate certificates without any additional symlinking.
+
+When ``--insecure`` is given to ``uw ecflow server``, SSL is disabled entirely and ``ECF_SSL_DIR`` is ignored.
```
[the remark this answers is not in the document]
AnnaSmoot-NOAA: Done.

src/uwtools/resources/jsonschema/ecflow.jsonschema
```diff
@@ -809,6 +809,10 @@
               "minimum": 0,
               "type": "integer"
             },
+            "ECF_SSL_DIR": {
+              "description": "Custom SSL certificate directory; overrides the default $HOME/.ecflowrc/ssl. Recognized by ecflow_server and ecflow_client (including job child commands), so jobs inheriting this variable will also locate certificates here without symlinking.",
```
maddenp-cu:
```suggestion
              "description": "Custom SSL certificate directory; overrides the default $HOME/.ecflowrc/ssl",
```
A bit long otherwise, and I don't think the part about `ecflow_client` is relevant here, since this is a server-config block.

docs/sections/user_guide/cli/tools/ecflow.rst
```diff
@@ -125,6 +125,23 @@ Examples
...
+  Where ``server.yaml`` contains:
+
+  .. code-block:: yaml
+
+     ecflow:
+       server:
+         ECF_HOME: /path/to/run
+         ECF_SSL_DIR: /shared/certs/ecflow
+
+  ``ECF_SSL_DIR`` overrides the default certificate location (``$HOME/.ecflowrc/ssl``) for both the server and any ``ecflow_client`` commands that inherit the environment, including job child commands. See :ref:`ecflow_workflows` for full server YAML documentation.
```
maddenp-cu: Where did you find out about this variable? I'm searching the online documentation and find no mention of it. Searching the ecFlow codebase, the variable exists, but is only used in a couple of places that appear to be strictly testing-related. I'm not sure we should rely on an undocumented test-harness variable, if that's what this is.

`ECF_SSL`, on the other hand, is documented [here](https://ecflow.readthedocs.io/en/latest/glossary.html#term-ecflow_server) and can, per the docs, be used to point to a user-specified _certificate_, rather than a directory.

src/uwtools/ecflow.py
```diff
@@ -419,41 +421,57 @@ def realize(
...
-    Ensure SSL certificates exist in $HOME/.ecflowrc/ssl.
+    Return the absolute path to the openssl executable.
+
+    :raises UWError: If openssl is not found on PATH.
+    """
+    path = shutil.which("openssl")
+    if path is None:
+        msg = "openssl not found on PATH"
+        raise UWError(msg)
+    return path
```
```suggestion
def _openssl() -> Path:
"""
    Return the absolute path to the openssl executable.

    :raises UWError: If openssl is not found on PATH.
    """
    path = shutil.which("openssl")
    if path is None:
        msg = "openssl not found on PATH"
        raise UWError(msg)
    return Path(path)
```
`uwtools` convention is to use `pathlib.Path` whenever possible.

src/uwtools/ecflow.py
```diff
@@ -464,8 +482,16 @@ def _ssl_generate_key(path: Path) -> None:
...
-    if not success:
+    old_umask = os.umask(0o077)
+    try:
+        result = subprocess_run(  # noqa: S603
+            [_openssl(), "genrsa", "-out", str(path), "2048"],
+            capture_output=True,
+            check=False,
+        )
+    finally:
+        os.umask(old_umask)
```
maddenp-cu: The store-and-reset mechanism for the umask is nice.

Any reason we cannot continue to use `run_shell_cmd()`? We have very few direct `subprocess` calls in `uwtools` (in fact, the only other one currently is for running the ecFlow server, and this makes me think that I should look into whether we can use `run_shell_cmd()` there, too)? That function has special logic for doing additional logging when running in debug mode that, although they might not be immediately useful, could be in the future.

Same question for the other two `subprocess_run` statements below...

maddenp-cu: Actually, I think we should reconsider the store-and-reset treatment for umask, as it is not threadsafe: It sets the umask for the entire Python interpreter, which would impact the permissions on files and directories created by every thread. Since `uwtools` offers a Python API, users may be making calls in a threaded environment. The umask set by calling `umask` as part of the command passed to `subprocess`, though, would be confined to a single subshell and not affect anything else.

Was there a specific security concern related to the old mechanism that you were trying to address?

docs/sections/user_guide/yaml/ecflow.rst
```diff
@@ -282,6 +282,32 @@ Multiple expand variables of the same length may be provided together:
...
+``ECF_SSL_DIR`` -- Optional. A path to a directory containing SSL certificate files (``dh2048.pem``, ``server.crt``, ``server.key``). When set, overrides the default location of ``$HOME/.ecflowrc/ssl``. For example:
+
+.. code-block:: yaml
+
+   ecflow:
+     server:
+       ECF_HOME: /path/to/run
+       ECF_SSL_DIR: /shared/certs/ecflow
+
+Because ``ECF_SSL_DIR`` is a native ecflow environment variable, both ``ecflow_server`` and ``ecflow_client`` (including job child commands) read certificates from this directory automatically. Jobs that inherit ``ECF_SSL_DIR`` from the server environment will locate certificates without any additional symlinking.
```
christinaholtNOAA:
```suggestion
Because ``ECF_SSL_DIR`` is a native ecFlow environment variable, both ``ecflow_server`` and ``ecflow_client`` (including job child commands) read certificates from this directory automatically. Jobs that inherit ``ECF_SSL_DIR`` from the server environment will locate certificates without any additional symlinking.
```

christinaholtNOAA (approved):

maddenp-cu (changes requested):

src/uwtools/resources/jsonschema/ecflow.jsonschema
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py

docs/sections/user_guide/yaml/ecflow.rst
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: I chose `subprocess_run()` with an argument list as the security-conscious default. Passing a list avoids shell parsing and injection risk entirely without needing argument quoting.  The `os.umask()` store/restore was the standard Python pattern, from what I could tell, for ensuring the key file gets restricted permissions regardless of the caller's umask.
However, I agree with your points about `uwtools` consistency and and thread-safety. I think those are more important here. 
I've updated the code and switched all three calls to `run_shell_cmd()` and moved the umask into the shell command string itself (`umask 077 && openssl ...`).

docs/sections/user_guide/cli/tools/ecflow.rst
AnnaSmoot-NOAA: Thanks for questioning that. I had poor wording when I wrote the issue. `ECF_SSL_DIR` is  is not a native ecFlow variable. It's a uwtools config key name we threw out in a meeting to name `ECF_SSL_DIR` for clarity. I'm updating the PR and `ecflow.rst`.

maddenp-cu: It's not just the description of the variable: It's the fact that this is not, as far as I can tell, the supported way to select a custom SSL certificate in ecFlow. `ECF_SSL_DIR` is an undocumented variable used by ecFlow developers for testing. The supported way to select a custom SSL certificate is to set `ECF_SSL` to a path to a certificate, rather than setting it to `1`, which indicates that the certificate should be found in the default location.

I think we might have just gotten (un)lucky by discussing an environment variable name, `ECF_SSL_DIR`, that *also* turned out to be an actual environment variable used by ecFlow internally for testing.

I'd be happy to be convinced otherwise, but the code changes in the PR indicate to me that the feature we're implementing is relying on unsupported/undocumented behavior in ecFlow. I think we should revamp this to use the `ECF_SSL` variable: In our YAML, we could accept either a boolean or a string -- a false boolean to turn SSL off (also the default), a true boolean to turn it on using the default certificate location, and a string to turn SSL on using a custom certificate path.

Please let me know if I've misunderstood something.

docs/sections/user_guide/cli/tools/ecflow.rst
```diff
@@ -125,6 +125,23 @@ Examples
 
      uw ecflow server --config-file server.yaml --port 54321
 
+bi* To start a server using a custom SSL certificate directory:
```
maddenp-cu:
```suggestion
* To start a server using a custom SSL certificate directory:
```

src/uwtools/ecflow.py
```diff
@@ -419,41 +420,57 @@ def realize(
...
     exist, creates it and generates the required SSL files using openssl.
 
+    :param ssl_dir: Directory in which to provision certificates. Defaults to $HOME/.ecflowrc/ssl
+        when None.
     :raises UWError: If the SSL directory exists but is incomplete, or if certificate generation
         fails.
     """
-    existing = [f for f in _SSL_FILES if (_SSL_DIR / f).exists()]
+    cert_dir = ssl_dir if ssl_dir is not None else _SSL_DEFAULT_DIR
+    existing = [f for f in _SSL_FILES if (cert_dir / f).exists()]
```
maddenp-cu: I'm wondering now about this check and its relationship to the specification of a custom certificate: If, again, I'm not mistaken about the fact that `ECF_SSL` can point to a custom certificate file (not directory), and that that's the official/supported way to override the default behavior, then it would seem to follow that all ecFlow really requires is that single certificate file. Of course, when we generate a certificate for users, there are two other files created as a side effect, but now I'm unclear as to whether those are actually needed after the certificate is generated. If not, then I think we could simplify our check to look only for the certificate file. If my understanding is right, though, then a user who points us to a single certificate file may not (or may no longer) have the other two files, and that might be fine. So maybe we'll need to simplify this check to look for just the single certificate file -- either the one the user specifies, or the default one.

docs/sections/user_guide/cli/tools/ecflow.rst
AnnaSmoot-NOAA: Oh yes! I see what you mean about it being a pre-existing environment variable for testing.  I agree. 
I've revamped the PR to use `ECF_SSL` directly as documented. The YAML config now accepts a boolean or a string. I removed `ECF_SSL_DIR` from the schema and the implementation entirely.

docs/sections/user_guide/cli/tools/ecflow.rst
AnnaSmoot-NOAA: Done.

src/uwtools/ecflow.py
AnnaSmoot-NOAA: I'm looking at [Openssl.cpp](https://github.com/ecmwf/ecflow/blob/develop/libs/base/src/ecflow/base/Openssl.cpp)

1. It looks like `ECF_SSL` does NOT accept a file path. When the docs say "allows users to specify a custom certificate path", they mean a filename prefix, not an absolute path. Setting `ECF_SSL=/shared/certs/server.crt` would cause ecFlow to look for `$HOME/.ecflowrc/ssl//shared/certs/server.crt.crt`, which isn't right.

2. All three files are always required by the server: 
```
ssl_context_->use_certificate_chain_file(crt());
ssl_context_->use_private_key_file(key(), ...);
ssl_context_->use_tmp_dh_file(pem());
```

3. `ECF_SSL_DIR` is the only mechanism to change the certificate directory, via `certificates_dir()`. The "for testing" comment is an internal note, not a hard restriction.

So looking at this, my updated implementation (setting `ECF_SSL` to an absolute cert path) is broken. ecFlow would interpret it as a filename prefix and construct a nonsensical path. The original` ECF_SSL_DIR` approach was mechanically correct. The naming confusion was coincidental. The YAML key happened to share the name of an ecFlow env var, but the env var is the real mechanism ecFlow uses to override the cert directory.



I see two paths forward: 

1.Revert to `ECF_SSL_DIR`. It's the only mechanism ecFlow provides for this, and the "for testing" comment appears to reflect its origin rather than a limitation on use.
2. Drop the custom cert directory feature from this PR and keep scope limited to the default location.

maddenp-cu: Thanks for the analysis. That's unfortunate wording ("path") in the docs -- I think they should have used "filename".

In re: option 1, to me, the combination of the comments in the ecFlow code about `ECF_SSL_DIR` being for testing, the fact that is *is* only used for testing in the ecFlow codebase, _plus_ the fact that there is no documentation about this variable, suggests that we shouldn't use it.

But I have an idea for an option 3 that might be better than option 2: Support `boolean` and `str` `ECF_SSL`: `false` (also the default) to disable it, `true` to enable it with the defaults, and a `str` value to select a certificate _filename_ relative to `$HOME/.ecflowrc/ssl`, which seems to be what's officially supported. That's probably better than nothing for power users: They can have multiple certificates and can select from among them, but have no control over the directory.

What do you think?

docs/sections/user_guide/yaml/ecflow.rst
```diff
@@ -282,6 +282,38 @@ Multiple expand variables of the same length may be provided together:
        MEM: ["01", "02"]
        LABEL: ["ctrl", "pert"]
 
+Server Configuration
+--------------------
+
+The ``uw ecflow server`` command reads configuration from an ``ecflow: server:`` block:
```
maddenp-cu:
```suggestion
The ``uw ecflow server`` command reads configuration from an ``ecflow.server`` block:
```
We usually use this "dotted" notation for paths in UW YAML.

docs/sections/user_guide/yaml/ecflow.rst
```diff
@@ -282,6 +282,38 @@ Multiple expand variables of the same length may be provided together:
...
+
+.. code-block:: yaml
+
+   ecflow:
+     server:
+       ECF_HOME: /path/to/run
+
+All keys in the ``server:`` block are passed as environment variables to ``ecflow_server``. ``ECF_HOME`` is the only required key.
+
+``ECF_SSL`` -- Optional. Controls SSL for the server. Accepts a boolean or a path string:
```
maddenp-cu:
```suggestion
``ECF_SSL`` -- Optional. Controls SSL for the server. Accepts a boolean or a certificate-filename prefix string:
```

docs/sections/user_guide/yaml/ecflow.rst
```diff
@@ -282,6 +282,38 @@ Multiple expand variables of the same length may be provided together:
...
+
+- ``true`` (default when ``--insecure`` is not given): Enable SSL using the auto-provisioned certificate at ``$HOME/.ecflowrc/ssl/server.crt``.
+- A path string (e.g. ``/shared/certs/server.crt``): Enable SSL using the specified certificate file. The matching ``server.key`` and ``dh2048.pem`` must already exist alongside it.
```
maddenp-cu:
```suggestion
- A certificate-filename prefix string (e.g. ``myhost.8888``): Enable SSL using the specified certificate file. Files with the given prefix and the extensions `.crt`, `.key`, and `.pem` must exist under `$HOME/.ecflowrc/ssl/`.
```

maddenp-cu (commented):
There might be more updates related to the semantics of `ECF_SSL` needed, but these caught my eye.

docs/sections/user_guide/yaml/ecflow.rst
AnnaSmoot-NOAA: Done.

docs/sections/user_guide/yaml/ecflow.rst
AnnaSmoot-NOAA: Done.

docs/sections/user_guide/yaml/ecflow.rst
AnnaSmoot-NOAA: Done.

maddenp-cu (approved):
Thanks again for your patience and persistence implementing this tricky feature.

PR #943, opened 2026-06-30
Update server logic

maddenp-cu:
**Synopsis**

- Use new callback option added to `run_shell_cmd` in #942 to give server code access to the process in which `ecflow_server` is running so that it can be terminated directly instead of by invoking `ecflow_client` in a subprocess.
- Delegate more control to `ECF_` environment variables, simplifying the call to `ecflow_server`.
- Revamp and simplify `server` and `_server_start` functions by factoring out lengthy or repetitive logic to nested helper functions, making identifiers simpler and more consistent, and factoring out string literals to the `strings` module.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -84,7 +84,7 @@ def server(
     report: bool = False,
 ) -> None:
     """
-    Start an ecFlow server, optionally with SSL security enabled.
+    Start an ecFlow server, optionally with SSL security disabled.
 
     The server runs in the foreground until interrupted (e.g. via CTRL-C), then terminated.
 
@@ -96,29 +96,30 @@ def server(
     """
 
     def certsetup() -> None:
-        if not insecure and ssl is not False:
-            prefix = ssl if isinstance(ssl, str) else None
+        if not insecure and ssl_cfg is not False:
+            prefix = ssl_cfg if isinstance(ssl_cfg, str) else None
             try:
                 _ssl_check(prefix)
             except UWSSLCertificateError:
-                if ssl in [True, None]:
+                if ssl_cfg in [True, None]:
                     _ssl_provision()
 
     def terminate(_signum: int, _frame: FrameType | None) -> None:
-        log.info("Terminating")
+        log.debug("Terminating")
         thread.terminal.set()
-        proc = cast(Popen, thread.proc)
-        proc.terminate()
-        proc.wait()
+        thread.initial.wait()
+        if thread.proc:
+            thread.proc.terminate()
+            thread.proc.wait()
 
     config = YAMLConfig(config)
     config.dereference()
     validate(config)
     env = config.data[STR.ecflow][STR.server]
-    ssl = env.get(STR.ECF_SSL)
+    ssl_cfg = env.get(STR.ECF_SSL)
     certsetup()
-    ssl = ssl if isinstance(ssl, str) else "" if insecure or ssl is False else "1"
-    env.update({STR.ECF_HOST: socket.gethostname(), STR.ECF_SSL: ssl})
+    ssl_env = ssl_cfg if isinstance(ssl_cfg, str) else "" if insecure or ssl_cfg is False else "1"
+    env.update({STR.ECF_HOST: socket.gethostname(), STR.ECF_SSL: ssl_env})
     thread = _ServerThread(target=_server_start, args=[env, port])
     signal.signal(signal.SIGINT, terminate)
     thread.start()
@@ -508,6 +509,7 @@ class _ServerThread(Thread):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.error: str | None = None
+        self.initial = Event()
         self.port: int | None = None
         self.proc: Popen | None = None
         self.terminal = Event()
@@ -576,8 +578,9 @@ def _server_start(env: dict[str, str], port: int | None) -> None:
     """
 
     def callback(proc: Popen) -> None:
-        thread.proc = proc
         thread.port = port
+        thread.proc = proc
+        thread.initial.set()
 
     def complain(error: str, messages: str | None = None) -> None:
         thread.terminal.set()
@@ -611,6 +614,7 @@ def _server_start(env: dict[str, str], port: int | None) -> None:
         except OSError as e:
             complain(f"Failed to launch ecflow_server: {e}", thread.error)
         break
+    thread.initial.set()
 
 
 def _server_wait(thread: _ServerThread, insecure: bool, env: dict[str, str] | None) -> None:
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -119,7 +119,7 @@ def server(
     server_config.update({"ECF_HOST": socket.gethostname(), "ECF_SSL": ssl})
     os.environ.update(server_config)
     rundir = Path(server_config["ECF_HOME"])
-    thread = _ServerThread(target=_server_start, args=[rundir, port, insecure])
+    thread = _ServerThread(target=_server_start, args=[rundir, port])
     signal.signal(signal.SIGINT, shutdown)
     thread.start()
     _server_wait(thread, insecure, server_config if report else None)
@@ -651,7 +651,7 @@ def _ssl_check(prefix: str | None) -> None:
         if _SSL_DIR.is_dir() and not prefix:
             log.error("Provide these files or remove %s to automatically generate", _SSL_DIR)
         raise UWSSLCertificateError
-    log.info("Using SSL certificates %s in %s", ", ".join(fns), _SSL_DIR)
+    log.debug("Using SSL certificates %s in %s", ", ".join(fns), _SSL_DIR)
 
 
 def _ssl_generate_cert(path: Path, key_path: Path) -> None:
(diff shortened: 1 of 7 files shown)


src/uwtools/tests/api/test_ecflow.py
```diff
@@ -6,6 +6,7 @@
 from uwtools.api import ecflow
 from uwtools.config.formats.yaml import YAMLConfig
 from uwtools.exceptions import UWError
+from uwtools.strings import STR
```
maddenp-cu, line 9: `ECF_` string literals were factored out to the `strings` module.

src/uwtools/utils/processing.py
```diff
@@ -64,7 +64,7 @@ def run_shell_cmd(
     proc = Popen(cmd, **kwargs)  # noqa: S603
     if callback:
         callback(proc)
-    output, _ = proc.communicate()
+    output, _ = proc.communicate()  # blocking call
```
maddenp-cu, line 67: Just a reminder for future readers that this is the statement in `run_shell_cmd` that blocks waiting for the subprocess call to return.

src/uwtools/ecflow.py
```diff
@@ -85,10 +84,9 @@ def server(
     report: bool = False,
 ) -> None:
     """
-    Start an ecFlow server, optionally with SSL security enabled.
+    Start an ecFlow server, optionally with SSL security disabled.
```
maddenp-cu, line 87: Because enabled is the default, disabled is an option.

src/uwtools/ecflow.py
```diff
@@ -97,48 +95,35 @@ def server(
...
-    def shutdown(_signum: int, _frame: FrameType | None) -> None:
-        log.info("Terminating")
+    def certsetup() -> None:
+        if not insecure and ssl_cfg is not False:
+            prefix = ssl_cfg if isinstance(ssl_cfg, str) else None
+            try:
+                _ssl_check(prefix)
+            except UWSSLCertificateError:
+                if ssl_cfg in [True, None]:
+                    _ssl_provision()
```
maddenp-cu, line 105: Factored out as a helper to make the flow of logic in the main function body cleaner.

src/uwtools/ecflow.py
```diff
@@ -97,48 +95,35 @@ def server(
...
-        # ecFlow enables SSL if ECF_SSL is set to any value, so unset it for an insecure server.
-        env.pop("ECF_SSL", None)
-    # Server variables to echo back when reporting: all config-supplied ECF_* values plus the
-    # runtime-determined host and SSL setting. ECF_PORT is added once the port is known.
-    report_vars = {**cfg_vars, "ECF_HOST": socket.gethostname()}
-    if ssl_on:
-        report_vars["ECF_SSL"] = ecf_ssl if isinstance(ecf_ssl, str) else "1"
-    thread = _ServerThread(target=_server_start, args=[rundir, env, port, insecure])
-    signal.signal(signal.SIGINT, shutdown)
+        thread.initial.wait()
```
maddenp-cu, line 110: To avoid a race condition, do not proceed until the server-startup logic signals that it is initially complete, whether that means that the server is running, or that it has given up trying to start the server. In either case, the server-startup logic will not make any changes to `thread.proc` after it sets `thread.initial`, so it will be safe for the termination logic here to look at `thread.proc` and decide how to proceed.

src/uwtools/ecflow.py
```diff
@@ -97,48 +95,35 @@ def server(
...
+        if thread.proc:
+            thread.proc.terminate()
+            thread.proc.wait()
+
+    config = YAMLConfig(config)
+    config.dereference()
+    validate(config)
+    env = config.data[STR.ecflow][STR.server]
+    ssl_cfg = env.get(STR.ECF_SSL)
```
maddenp-cu, line 119: `ssl_cfg` is the `ECF_SSL` value as read from the config, which may be `None`.

src/uwtools/ecflow.py
```diff
@@ -97,48 +95,35 @@ def server(
...
+    certsetup()
+    ssl_env = "" if insecure or ssl_cfg is False else ssl_cfg if isinstance(ssl_cfg, str) else "1"
```
maddenp-cu, line 121: `ssl_env` is the `ECF_SSL` value to export as an environment variable. Its potential values are all strings, and are different from what is accepted in the YAML config.

src/uwtools/ecflow.py
```diff
@@ -569,92 +555,87 @@ def _openssl() -> Path:
     return Path(path)
 
 
-def _server_report(port: int, report_vars: dict[str, str] | None) -> None:
+def _server_report(port: int, env: dict[str, str] | None) -> None:
```
maddenp-cu, line 558: The `env` value is what gets passed both to `_server_report`, here, to be formatted as JSON and printed to `stdout`. It's also what gets passed to `_server_start` (where it is merged with `os.environ` to create the environment for `ecflow_server`) and to `_server_wait`. It's the same data in all cases now, so has a single name.

src/uwtools/ecflow.py
```diff
@@ -569,92 +555,87 @@ def _openssl() -> Path:
...
+        if messages:
+            for line in messages.split("\n"):
+                log.error(line)
+
+    def post(proc: Popen) -> None:
+        thread.port = port
+        thread.proc = proc
+        thread.initial.set()
+
+    cmd = ["ecflow_server"]
```
maddenp-cu, line 592: The `ecflow_server` command is simpler now because it relies on environment variables instead of command-line switches.

src/uwtools/ecflow.py
```diff
@@ -677,7 +658,7 @@ def _ssl_check(prefix: str | None) -> None:
         if _SSL_DIR.is_dir() and not prefix:
             log.error("Provide these files or remove %s to automatically generate", _SSL_DIR)
         raise UWSSLCertificateError
-    log.info("Using SSL certificates %s in %s", ", ".join(fns), _SSL_DIR)
+    log.debug("Using SSL certificates %s in %s", ", ".join(fns), _SSL_DIR)
```
maddenp-cu, line 664: In practice, this struck me as unnecessary info for most users. They can run `uw --verbose` for this kind of debugging info.

src/uwtools/tests/test_ecflow.py
```diff
@@ -61,7 +61,7 @@ def server_mocks():
     different data can reassign `m.cfg.data` before calling ecflow.server().
     """
     cfg = Mock()
-    cfg.data = {"ecflow": {"server": {"ECF_HOME": "/ecf"}}}
+    cfg.data = {"ecflow": {"server": {STR.ECF_HOME: "/ecf"}}}
```
christinaholtNOAA, line 64: Do we typically use the `STR` instance for test data input? Do we want to start doing that? Not doing it might give us more awareness of breaking changes. Just a thought.

Perhaps we'd find it anyway given that the only reason to change these particular variables would come down from ecflow itself.

christinaholtNOAA (approved):

src/uwtools/tests/test_ecflow.py
maddenp-cu: It's probably a mix at the moment, but there are lots of instances of `STR.` references in the tests already. My feeling at the moment is that the benefit single-source-of-truth strings outweighs the benefit of having a test point out that that single source of truth is wrong. If it's wrong, it'd be nice if there were exactly one place to fix it. And if a unit test became the singe source of truth, it seems like that just moves the source of knowledge.

christinaholtNOAA: Reasonable. :)

elcarpenterNOAA (approved):

src/uwtools/ecflow.py
```diff
@@ -97,48 +95,35 @@ def server(
...
+    config = YAMLConfig(config)
+    config.dereference()
+    validate(config)
+    env = config.data[STR.ecflow][STR.server]
+    ssl_cfg = env.get(STR.ECF_SSL)
+    certsetup()
+    ssl_env = "" if insecure or ssl_cfg is False else ssl_cfg if isinstance(ssl_cfg, str) else "1"
+    env.update({STR.ECF_HOST: socket.gethostname(), STR.ECF_SSL: ssl_env})
+    thread = _ServerThread(target=_server_start, args=[env, port])
+    signal.signal(signal.SIGINT, terminate)
```
AnnaSmoot-NOAA, line 124: `server()` now mutates the parsed config dict. Do you think this may leak into later logic/tests if the same config object is reused? Do you think a copy before mutation would be helpful?  `env = dict(config.data[STR.ecflow][STR.server])`

src/uwtools/ecflow.py
```diff
@@ -569,92 +555,87 @@ def _openssl() -> Path:
...
     while not thread.terminal.is_set():
-        port = thread.port
-        if port:
+        if port := thread.port:
             try:
                 _client(port, insecure).ping()
             except RuntimeError as e:
                 if "Failed to connect" not in str(e):
-                    raise
+                    raise UWError("Could not start server on port %s" % port) from e
```
AnnaSmoot-NOAA, line 638: With this change in exception mapping, do you think we lose any useful diagnostic detail from ecFlow/client exceptions here?

AnnaSmoot-NOAA (approved):

src/uwtools/ecflow.py
maddenp-cu: A copy would be the safe choice, will do. Good call.

src/uwtools/ecflow.py
maddenp-cu: I'm not sure (the ecFlow messages can be a bit obscure), but to be on the safe side I'll add debug logs of the text of the original exception so that a user troubleshooting with `--verbose` can see them.

PR #946, opened 2026-07-08
ecFlow support updates

maddenp-cu:
**Synopsis**

Gather various updates from work on the end-to-end demo for separate review, to avoid a potentially much larger diff in a later PR.

- In `src/setup.py`, exclude `ecflow` from the list of PyPI packages installed via `pip install`, as it is not published to PyPI. It will be up to `venv` users to ensure that ecFlow is installed an on `PYTHONPATH`.
- Update the server termination procedure to halt, then checkpoint, then terminate the server, following the official advice [here](https://ecflow.readthedocs.io/en/5.16.0/ug/user_manual/ecflow_server/stopping_ecflow_servers.html), but using the client API instead of the CLI.
- Simplify the `script` block in an ecFlow task block to support a generic `body`.
- Update tests and docs.

**Type**

- [x] Documentation

**Impact**

**Checklist**


docs/sections/user_guide/cli/tools/ecflow/workflow_output/workflow/data_prep/fetch.ecf
```diff
@@ -1,12 +1 @@
...
 echo Fetching data...
-if [[ $? -ne 0 ]]; then
-   ecflow_client --msg="***JOB $ECF_NAME ERROR RUNNING J-SCRIPT ***"
-   ecflow_client --abort
-   exit 1
-fi
-
-%manual
-Script to run fetch
-%end
```
maddenp-cu, line 12: The code that produces the `.ecf` files is no longer this opinionated: It will honor a `task_<name>.script.manual` config key and add a manual block to the `.ecf` file, but this particular config YAML does not specify that -- it only has a `body`. The rest of the content here (except for `echo Fetching data...`, which is given as `body`) would now be an application-specific choice, either to be added to `body` or via a `.h` file pointed to by `script.include.entry` or `script.include.exit`.

Similarly in `process.ecf`, below.

docs/sections/user_guide/cli/tools/ecflow/workflow_output/suite.def
maddenp-cu, line 1: This file is generated by `make docs` but is not included in RST, so doesn't need to be committed.

docs/sections/user_guide/cli/tools/ecflow/realize-exec-dir.cmd
```diff
@@ -1,4 +1,4 @@
 rm -f suite.def
-uw ecflow realize --config-file ecflow.yaml --output-dir .
+uw ecflow realize --config-file forecast.yaml --output-dir .
```
maddenp-cu, line 2: I was getting confused about the difference between `ecflow.yaml` and `ecflow-workflow.yaml`, so renamed them to `forecast.yaml` and `workflow.yaml`, respectively. (And `ecflow-bad.yaml` to just `bad.yaml` for consistency.)

src/uwtools/tests/test_schemas.py
```diff
@@ -1058,32 +1063,72 @@ def test_schema_ecflow_suitedef_execution():
...
+    # script may be an object with body:
+    assert not errors({"body": "run.sh"})
+    # script as object with optional keys is valid:
+    assert not errors(
+        {
+            "batchargs": {"walltime": "00:30:00"},
+            "body": "run.sh",
+            "includes": {"entry": ["head.h"], "exit": ["tail.h"]},
+            "manual": "This is the manual",
+        }
```
maddenp-cu, line 1078: This is a decent preview of what the updated `task_<name>.script` block looks like. `body` could also be a multi-line YAML string, and of course `batchargs` supports many more options.

src/uwtools/ecflow.py
```diff
@@ -105,11 +103,17 @@ def certsetup() -> None:
...
         if thread.proc:
-            thread.proc.terminate()
+            assert thread.port
+            c = _client(thread.port, insecure)
+            log.info("Halting")
+            c.halt_server()
+            log.info("Checkpointing")
+            c.checkpt()
+            log.info("Terminating")
+            c.terminate_server()
```
maddenp-cu, line 116: This is a safer method for shutting down the server, ensuring that a checkpoint gets written before termination.

src/uwtools/ecflow.py
```diff
@@ -180,23 +184,21 @@ def __str__(self):
...
-            outpath.write_text(content)
+        if self._scripts:
+            for subpath, content in self._scripts.items():
+                assert subpath.startswith("/")
+                outpath = Path(str(path) + subpath).with_suffix(".ecf")
+                outpath.parent.mkdir(parents=True, exist_ok=True)
+                log.debug("Writing ecf script: %s", outpath)
+                outpath.write_text(content + "\n")
+        else:
+            log.debug("No scripts are configured for this workflow.")
```
maddenp-cu, line 200: Essentially the same logic, slightly simplified.

src/uwtools/ecflow.py
```diff
@@ -267,16 +266,16 @@ def _add_node(  # noqa: PLR0912
                     add_items(node.add_limit, subconfig)
                 case STR.meters:
                     add_items(node.add_meter, subconfig)
-                case STR.repeat:  # Only one repeat is allowed per node.
```
maddenp-cu, line 270: I'm happy to revert this if there's disagreement, but I didn't see see the value of this comment. What is the reader supposed to understand about this line after reading this comment? Maybe there's some information that needs to be captured somewhere in a comment, but it didn't make sense to me here.

Similarly for the other removed comments below. I kept the one whose meaning was clear to me.

src/uwtools/ecflow.py
```diff
@@ -328,104 +326,6 @@ def _add_workflow_components(self) -> None:
                 case STR.suites:
                     self._expand_block(subconfig, name, Suite, self._d)
 
-    def _create_ecf_script(self, config: dict, task: Task) -> None:
```
maddenp-cu, line 331: This method is re-written as `_prepare_ecf_script`. It doesn't really _create_ the script, it just assembles the content of the script in memory for writing later. In particular, the docstring (`Write the ecf script...`) was inaccurate. Maybe a nitpick, but I hope the new formulation is clearer.

src/uwtools/ecflow.py
```diff
@@ -468,24 +366,26 @@ def _expand_block(
             }
             self._add_node(**args)
 
-    def _jobscheduler(self, account: str, execution: dict, rundir: Path | str) -> JobScheduler:
+    def _prepare_ecf_script(self, config: dict, task: Task) -> None:
```
maddenp-cu, line 368: This replaces `_create_ecf_script()`.

src/uwtools/ecflow.py
```diff
@@ -604,7 +504,14 @@ def post(proc: Popen) -> None:
         log.debug("Trying to start server on port %s", port)
         env[STR.ECF_PORT] = str(port)
         try:
-            success, output = run_shell_cmd(cmd=cmd, cwd=cwd, env=env, quiet=True, callback=post)
+            success, output = run_shell_cmd(
+                cmd=cmd,
+                cwd=cwd,
+                env=env,
+                start_new_session=True,
```
maddenp-cu, line 510: `start_new_session=True` is necessary to prevent the server thread from seeing and responding to ctrl-c from a user. That SIGINT should be caught an handled by the main thread, which then halts, checkpoints, and terminates the server gracefully.

docs/sections/user_guide/cli/tools/ecflow/realize-exec-dir.cmd
christinaholtNOAA: I'm actually having a hard time telling a meaningful difference between the two, even with the new names. Are these two showing us different features? Are they both needed? Maybe we could just keep workflow.yaml?

src/uwtools/ecflow.py
christinaholtNOAA: These were particularly useful as I was sorting out the different cases, but I think they were just notes to myself and are fine to remove. They helped me figure out which ones should be lists of added attributes vs just one.

src/uwtools/ecflow.py
christinaholtNOAA: I'd say it's clearer, alright.

christinaholtNOAA (approved):
I'll let you decide what to do with the 2 docs YAMLs. I'm fine either way. 

This looks great, and I appreciate you cutting it off here for review.

docs/sections/user_guide/cli/tools/ecflow/realize-exec-dir.cmd
maddenp-cu: Yeah, I was also wondering if they should be combined. Since many other doc changes will be needed on my docs-and-demo branch, I'll tackle this there. I only did as much as I had to here to get everything to work together.

src/uwtools/ecflow.py
```diff
@@ -180,23 +184,21 @@ def __str__(self):
...
-            outpath.parent.mkdir(parents=True, exist_ok=True)
-            log.debug("Writing script: %s", outpath)
-            outpath.write_text(content)
+        if self._scripts:
+            for subpath, content in self._scripts.items():
+                assert subpath.startswith("/")
+                outpath = Path(str(path) + subpath).with_suffix(".ecf")
+                outpath.parent.mkdir(parents=True, exist_ok=True)
+                log.debug("Writing ecf script: %s", outpath)
+                outpath.write_text(content + "\n")
```
AnnaSmoot-NOAA, line 198: And then line 388: `self._scripts[task.get_abs_node_path()] = contents`

In practice, `get_abs_node_path()` on a task that's been added to a suite should always return an absolute path like `/suite_name/family_name/task_name`. However, with the `assert` disabled by `python -O`, I think it becomes a no-op. If anyone runs uwtools with `-O`, this check silently disappears. If something goes wrong, the user gets `AssertionError` with just a stack trace. 
Also, the code does `Path(str(path) + subpath)` not `Path / Path`. If `get_abs_node_path()` ever returned a relative path, the output would be silently wrong (a file in the wrong place), not an error.

I think what you have is fine, but I wonder if something like this would be safer and/or more consistent with how other errors are raised, like `UWConfigError` in `_node()` or `_expand_block()`: 

```
for subpath, content in self._scripts.items():
    if not subpath.startswith("/"):
        raise UWConfigError(
            f"Expected an absolute ecFlow node path, got: {subpath!r}"
        )
    outpath = Path(str(path) + subpath).with_suffix(".ecf")
    ...
```

src/uwtools/ecflow.py
```diff
@@ -468,24 +366,26 @@ def _expand_block(
...
-        return JobScheduler.get_scheduler(resources)
+        directives = ""
+        if self._scheduler:
+            props = {STR.scheduler: self._scheduler, **config[STR.batchargs]}
+            scheduler = JobScheduler.get_scheduler(props)
+            directives = "\n".join(scheduler.directives)
+        fmt = lambda k: "\n\n".join(
+            f"%include {x}" for x in config.get(STR.includes, {}).get(k, [])
+        )
+        includes_entry, includes_exit = map(fmt, [STR.entry, STR.exit])
```
AnnaSmoot-NOAA, line 383: In other parts of this file, inner functions are used freely (`complain`, `post`, `terminate`, etc. inside `server()`. I don't think this is something that needs fixed, I just wonder if something like this would be more consistent: 

```
def fmt(k):
    return "\n\n".join(
        f"%include {x}" for x in config.get(STR.includes, {}).get(k, [])
    )
includes_entry, includes_exit = map(fmt, [STR.entry, STR.exit])
```

AnnaSmoot-NOAA (approved):

elcarpenterNOAA (approved):

src/uwtools/ecflow.py
maddenp-cu: I pushed a change that makes this safer, using only `pathlib` calls and no string concatenation.

src/uwtools/ecflow.py
maddenp-cu: I usually prefer a `lambda` to a `def` for a single-expression inner function because it's more concise and can be defined right next to the call site without breaking the visual flow of the code (a `def` will get wrapped in blank lines by the formatter, so either introduces that visual break or must be placed apart from the call site to avoid it.)

In terms of overall codebase consistency, I'd wager we have far more `labmda`s than inner `def`s, though that might not be obvious from looking just at this module. Inner `def`s are required for multi-statement logic, so do appear here and there.

Python is somewhat hostile to `lambda`s, and to functional programming in general, but I've never found the arguments very persuasive.

AnnaSmoot-NOAA: Sounds good to me.

PR #947, opened 2026-07-09
Make 'account' directive optional

maddenp-cu:
**Synopsis**

In a recent discussion, I think we decided that the `account` batch directive should be optional, because there is a default value on some platforms and users should not be forced to specify it explicitly. The `batchargs` schema already agrees with this: Only the `walltime` directive is required. But code in `uwtools.scheduler` was behaving as if `account` were required.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/scheduler.py
+++ b/src/uwtools/scheduler.py
@@ -149,7 +149,7 @@ class JobScheduler(ABC):
             for x in fields(_DirectivesRequired)
             if getattr(_DirectivesRequired, x.name) not in self._props
         ]:
-            msg = "Missing required directives: %s" % ", ".join(missing)
+            msg = "Missing required directive(s): %s" % ", ".join(missing)
             raise UWConfigError(msg)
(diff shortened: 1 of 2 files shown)


src/uwtools/scheduler.py
```diff
@@ -149,7 +149,7 @@ def _validate_props(self) -> None:
             for x in fields(_DirectivesRequired)
             if getattr(_DirectivesRequired, x.name) not in self._props
         ]:
-            msg = "Missing required directives: %s" % ", ".join(missing)
+            msg = "Missing required directive(s): %s" % ", ".join(missing)
```
maddenp-cu, line 152: I kept the singular/plural general in case we ever expand the set of required directives.

elcarpenterNOAA (approved):

christinaholtNOAA (approved):
I think this is the right way to go for portability.

PR #949, opened 2026-07-13
Handle ECF_HOME mkdir failure

maddenp-cu:
**Synopsis**

I noticed that, if the `ECF_HOME` directory cannot be created, e.g. due to permissions errors, then `uw ecflow server` would not exit cleanly, but would have to be manually killed, as it would not even respond to ctrl-c.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/ecflow.py
+++ b/src/uwtools/ecflow.py
@@ -495,10 +495,14 @@ def _server_start(env: dict[str, str], port: int | None) -> None:
 
     cmd = ["ecflow_server"]
     cwd = Path(env[STR.ECF_HOME])
-    cwd.mkdir(parents=True, exist_ok=True)
     env = {**os.environ, **env}
-    thread = cast(_ServerThread, current_thread())
     static = port is not None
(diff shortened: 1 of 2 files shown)


src/uwtools/ecflow.py
```diff
@@ -495,10 +495,14 @@ def post(proc: Popen) -> None:
...
-    cwd.mkdir(parents=True, exist_ok=True)
     env = {**os.environ, **env}
-    thread = cast(_ServerThread, current_thread())
     static = port is not None
+    thread = cast(_ServerThread, current_thread())
+    try:
+        cwd.mkdir(parents=True, exist_ok=True)
+    except Exception as e:  # noqa: BLE001
+        complain(f"Failed to create ECF_HOME directory {cwd}", str(e))
+        return
```
maddenp-cu, line 505: Catching the exception and calling `complain` ensures that `thread.terminal` gets set so that the main thread also exits and does not keep waiting for the server to start.

elcarpenterNOAA (approved):

PR #951, opened 2026-07-15
Add uwtools.api.config.compose_to_dict

maddenp-cu:
**Synopsis**

To complement `uwtools.api.config.realize_to_dict()`, add `uwtools.api.config.compose_to_dict()`, which likewise returns a `dict` object, without writing anything either to `stdio` or to a file.

In the course of testing this, I tried a couple of namelist files and found in interesting issue, likely an undiagnosed bug. Given

`a.nml`
```
&a foo=1 bar="baz" /
```

and

`b.nml`
```
&a bar="qux" /
```

Old behavior:
```
>>> from uwtools.api.config import compose
>>> c = compose(configs=["a.nml", "b.nml"], output_file=os.devnull)
>>> c.as_dict()
{'a': Namelist([('foo', 1), ('bar', 'qux')])}
```

Note the appearance of the `f90nml` `Namelist` object in the `dict`. 

New behavior:
```
>>> from uwtools.api.config import compose
>>> c = compose(configs=["a.nml", "b.nml"], output_file=os.devnull)
>>> c.as_dict()
{'a': {'foo': 1, 'bar': 'qux'}}
```

I think there were potentially two issues:

1. With no output filename or output format specified, `compose` defaulted to `YAMLConfig` as an output-type class, and `YAMLConfig.as_dict` had no logic to convert `Namelist` objects, they appeared in the `dict`.
2. The logic in `NMLConfig.as_dict` to convert `Namelist` objects wasn't recursive, so wouldn't have helped even if it had been called.

This PR deals with issue 1 by replacing the abstract `as_dict` in the `Config` base class with a concrete one, and removing the concrete implementations from child classes, as the parent's suffices; and with issue 2 by making `as_dict` call a recursive helper.

**Type**

**Impact**

**Checklist**


elcarpenterNOAA (approved):

christinaholtNOAA (approved):
Looks good.

PR #956, opened 2026-07-27
API update

maddenp-cu:
**Synopsis**

- Expose `uwtools.drivers.stager.FileStager` via the API as `uwtools.api.driver.FileStager` to support apps creating external drivers using this abstract mixin.
- Expose `uwtools.utils.processing.run_shell_cmd` via the API as `uwtools.api.utils.run_shell_cmd`, and make the docstring more useful for end users.
- Update the documentation, including adding an API page for the new `uwtools.api.utils`.

**Type**

- [x] Documentation

**Impact**

**Checklist**


elcarpenterNOAA (approved):

src/uwtools/drivers/stager.py
```diff
@@ -10,7 +10,7 @@
 
 class FileStager:
     """
-    A base class for tasks that stage files.
+    An abstract base class for tasks that stage files.
```
christinaholtNOAA: We had referred to this as a mixin in the past. Is that a more apt description than abstract base class since this defines concrete collections that can be used directly in drivers?

christinaholtNOAA (approved):
One very minor semantics comment that I'll let you make the call on, but otherwise this looks great!

src/uwtools/drivers/stager.py
maddenp-cu: Yes, thank you! It's categorically _not_ abstract, I misread it. Fixing...

PR #969, opened 2026-08-27
Publish `uwtools` releases to PyPI

elcarpenterNOAA:
**Synopsis**
- build a PyPI wheel as part of the release workflow 
- publish release wheels to PyPI using Trusted Publishing 
- add CI build tooling to create wheel 

Testing: I did a test on my fork ([see PR here](https://github.com/elcarpenterNOAA/workflow-tools/pull/5)) and successfully published a package to TesPyPI, which you can [try out here.](https://test.pypi.org/manage/project/uwtools/release/2.20.0/) I installed it locally, and confirmed it was working and that `ecflow` was excluded from `uw --help`.

Documentation still needs to be written, but I don't feel I'm in a position to write it until it actually exists on PyPI.

**Type**

- [ ] Documentation

**Impact**

**Checklist**

### DIFF
--- a/.github/scripts/pypi-package-build.sh
+++ b/.github/scripts/pypi-package-build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -eux
+
+. $(dirname ${BASH_SOURCE[0]})/common.sh
+ci_conda_activate
+unset CONDEV_SHELL
+set -ux
+python -m build --no-isolation --wheel src
--- a/.github/scripts/pypi-package-build.sh
+++ b/.github/scripts/pypi-package-build.sh
@@ -3,5 +3,6 @@
 . $(dirname ${BASH_SOURCE[0]})/common.sh
 ci_conda_activate
 unset CONDEV_SHELL
+export PYPI_PROJECT_NAME=unified-workflow-tools
 set -ux
 python -m build --no-isolation --wheel src
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,12 +19,12 @@ jobs:
         run: .github/scripts/make-docs.sh
       - name: Format Check
         run: .github/scripts/format-check.sh
-      - name: Make PyPI Package
-        run: .github/scripts/pypi-package-build.sh
       - name: Make Package
         run: .github/scripts/make-package.sh
       - name: Publish
         run: .github/scripts/publish.sh
+      - name: Make PyPI Package
+        run: .github/scripts/pypi-package-build.sh
       - name: Publish PyPI Package
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,7 +17,13 @@ jobs:
         run: .github/scripts/make-docs.sh
       - name: Format Check
         run: .github/scripts/format-check.sh
+      - name: Make PyPI Package
+        run: .github/scripts/pypi-package-build.sh
       - name: Make Package
         run: .github/scripts/make-package.sh
       - name: Publish
         run: .github/scripts/publish.sh
+      - name: Publish PyPI Package
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: src/dist/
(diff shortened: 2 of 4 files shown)


.github/scripts/pypi-package-build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -eux
+
+. $(dirname ${BASH_SOURCE[0]})/common.sh
+ci_conda_activate
+unset CONDEV_SHELL
```
elcarpenterNOAA, line 5: `setup.py` omits `install_requires` when `CONDEV_SHELL` is set, and we need that info to get the `Requires-Dist` data.

.github/workflows/release.yaml
```diff
@@ -9,6 +9,8 @@ jobs:
   release:
     name: Release
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
```
elcarpenterNOAA, line 13: This enables GitHub OIDC for Trusted Publishing.

.github/scripts/pypi-package-build.sh
maddenp-cu: I'm pretty ignorant about how wheels are built: If `install_requires` is enabled, does the wheel still contain only the `uwtools` code, and not the code from dependency packages? I'd hope that the `uwtools` wheel just contains the `uwtools` code, plus _metadata_ that tells it what other wheels to install along with it.

.github/workflows/release.yaml
```diff
@@ -17,7 +19,13 @@ jobs:
         run: .github/scripts/make-docs.sh
       - name: Format Check
         run: .github/scripts/format-check.sh
+      - name: Make PyPI Package
+        run: .github/scripts/pypi-package-build.sh
```
maddenp-cu: Does the PyPI package build run any tests? I guess not. Would it be a good idea to move this step below the conda `Make Package` step, in that the conda package build _will_ run all the tests and so, if it fails, this workflow won't bother building the wheel? I think the workflow is already _safe_ in that it won't publish to PyPI if the conda build/publish fails, but it might save a little bit of CI time to not even try to build the wheel if we can't build the conda package.

maddenp-cu (approved):
If anything goes wrong at release time, it's easy to remove conda packages from our `ufs-community` channel and re-run the release workflow without incrementing versions. Not so easy with PyPI, of course. 🤞

.github/workflows/release.yaml
elcarpenterNOAA: Good call. I moved it.

.github/scripts/pypi-package-build.sh
elcarpenterNOAA: Correct, the wheel contains the `uwtools` package, and `install_requires` adds the metadata.

src/setup.py
```diff
@@ -24,14 +24,15 @@
...
+name_dist = os.environ.get("PYPI_PROJECT_NAME", name_conda)
 name_py = name_conda.replace("-", "_")
 
 # Define basic setup configuration.
 
 kwargs = {
     "entry_points": {"console_scripts": ["uw = %s.cli:main" % name_py]},
     "include_package_data": True,
-    "name": name_conda,
+    "name": name_dist,
```
elcarpenterNOAA, line 35: @maddenp-cu I am accepting defeat on getting the `uwtools` name on PyPI, so I have had to make a few changes: lines 27 and 35 of this file, plus a line in `pypi-package-build.sh`. 

The added `name_dist` should still be `uwtools` for conda, but `unified-workflow-tools` for PyPI. 

I tested this locally by building and installing the wheel, and by doing a `make package`, and all was well. I did not republish to TestPyPI since that would have required creating a new project there, but I can do that if it would be helpful.

.github/scripts/pypi-package-build.sh
```diff
@@ -0,0 +1,8 @@
...
+export PYPI_PROJECT_NAME=unified-workflow-tools
```
elcarpenterNOAA, line 6: New project name. 😿

src/setup.py
maddenp-cu: Probably your previous tests are sufficient. If anything goes wrong during deploy, we can manually remove the conda package from our `ufs-community` channel and re-run the pipeline with potential fixes. We just won't do the manual GitHub release until both conda and PyPI sides look good.

PR #973, opened 2026-08-31
Atomic improvements and null driver executable

maddenp-cu:
**Synopsis**

1. Support `null` as a value for `executable` in driver configs, to support all-Python drivers.
2. Give the `atomic` context manager a way to avoid creating the final file at all.

In re: item 1, there are reasons for requiring a `null` value instead of just omitting the key:

- In the case of `<driver>.execution.executable`, `executable` is the only required key in the `execution:` block, so making it optional would make the whole `execution:` block potentially also optional. We could consider that, but:
- Having a `null` value signals intent, while a missing `executable` key (or a missing `execution:` block) looks potentially like an omission.
- There are other places where a `null` value might be useful in the future to signal that certain behavior is unwanted. Currently, the `uwtools` mechanism for this is the `!remove` tag, but this must potentially be repeated many times in an override config if the same optional value (e.g. `account` under many task subblocks in a Rocoto block) should be suppressed. Compare to a case where one might set `user.account` to a real account name in a base config and have many `account: '{{ user.account }}'` references to it in Rocoto blocks. In this case, an override config would only need to override the single `user.account` value, setting it to `null`. Of course, supporting Python code would need to take appropriate action when seeing a `null` value in supported contexts -- as is done in this PR for `executable: null`. So allowing `executable: null` here is a proof-of-concept exercise (as well as potentially improving a real use case in `uw-aigfs`, but that's a separate PR).

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -376,10 +376,7 @@ class Driver(Assets):
         if self.config[STR.execution][STR.executable] is None:
             msg = "Drivers not specifying 'executable' must override the run() method"
             raise UWNotImplementedError(msg)
-        if self._batch:
-            yield self._run_via_batch_submission()
-        else:
-            yield self._run_via_local_execution()
+        yield self._run_via_batch_submission() if self._batch else self._run_via_local_execution()
 
     @task
     def runscript(self):
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -379,7 +379,8 @@ class Driver(Assets):
             else:
                 yield self._run_via_local_execution()
         else:
-            yield None
+            msg = "Drivers not specifying 'executable' must override the run() method"
+            raise UWNotImplementedError(msg)
 
     @task
     def runscript(self):
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -376,8 +376,8 @@ class Driver(Assets):
         taskname = self.taskname(STR.run)
         yield taskname
         if self.config[STR.execution][STR.executable] is None:
-            msg = "%s: Drivers with no 'executable' must implement a custom run() method" % taskname
-            yield blocker(msg)
+            msg = "%s must define 'executable' or implement custom run() method"
+            yield blocker(msg % self.__class__.__name__)
         yield self._run_via_batch_submission() if self._batch else self._run_via_local_execution()
 
     @task
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -373,7 +373,13 @@ class Driver(Assets):
         A run.
         """
         yield self.taskname(STR.run)
-        yield self._run_via_batch_submission() if self._batch else self._run_via_local_execution()
+        if self.config.get(STR.execution, {}).get(STR.executable):
+            if self._batch:
+                yield self._run_via_batch_submission()
+            else:
+                yield self._run_via_local_execution()
+        else:
+            yield None
 
     @task
     def runscript(self):
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -31,7 +31,7 @@ from uwtools.scheduler import JobScheduler
 from uwtools.strings import STR
 from uwtools.utils.file import writable
 from uwtools.utils.processing import run_shell_cmd
-from uwtools.utils.tasks import blocker
+from uwtools.utils.tasks import poison
 
 if TYPE_CHECKING:
     from datetime import datetime, timedelta
@@ -376,7 +376,7 @@ class Driver(Assets):
         yield self.taskname(STR.run)
         if self.config[STR.execution][STR.executable] is None:
             msg = "%s must define 'executable' or implement custom run() method"
-            yield blocker(msg % self.__class__.__name__)
+            yield poison(msg % self.__class__.__name__)
         yield self._run_via_batch_submission() if self._batch else self._run_via_local_execution()
 
     @task
--- a/src/uwtools/drivers/driver.py
+++ b/src/uwtools/drivers/driver.py
@@ -373,14 +373,13 @@ class Driver(Assets):
         A run.
         """
         yield self.taskname(STR.run)
-        if self.config.get(STR.execution, {}).get(STR.executable):
-            if self._batch:
-                yield self._run_via_batch_submission()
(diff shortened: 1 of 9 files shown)


src/uwtools/drivers/driver.py
```diff
@@ -373,7 +373,13 @@ def run(self):
         A run.
         """
         yield self.taskname(STR.run)
-        yield self._run_via_batch_submission() if self._batch else self._run_via_local_execution()
+        if self.config[STR.execution][STR.executable] is None:
+            msg = "Drivers not specifying 'executable' must override the run() method"
+            raise UWNotImplementedError(msg)
```
maddenp-cu: A driver is now free to set `executable: null` in its config (YAML `null` parses to Python `None`), but if it does so it must override the `run()` method with an implementation that does the necessary work in Python instead of delegating to a runscript.

src/uwtools/drivers/driver.py
```diff
@@ -373,6 +374,9 @@ def run(self):
         A run.
         """
         yield self.taskname(STR.run)
+        if self.config[STR.execution][STR.executable] is None:
+            msg = "%s must define 'executable' or implement custom run() method"
+            yield blocker(msg % self.__class__.__name__)
```
maddenp-cu: A driver is now able to set `executable: null` in its config, but if it does so it must override the `run()` method with an implementation that does the necessary work in Python instead of delegating to a runscript.

src/uwtools/tests/drivers/test_driver.py
```diff
@@ -445,9 +445,11 @@ def test_Driver_run(batch, driverobj, node):
...
     executable = Path(driverobj.config["execution"]["executable"])
     executable.touch()
-    with patch.object(driverobj, "_run_via_batch_submission", return_value=node) as rvbs:
-        with patch.object(driverobj, "_run_via_local_execution", return_value=node) as rvle:
-            driverobj.run()
+    with (
+        patch.object(driverobj, "_run_via_batch_submission", return_value=node) as rvbs,
+        patch.object(driverobj, "_run_via_local_execution", return_value=node) as rvle,
+    ):
+        driverobj.run()
```
maddenp-cu, line 452: Just slightly simplified syntax with a single `with`.

src/uwtools/utils/file.py
```diff
@@ -59,8 +59,9 @@ def atomic(path: Path) -> Iterator[Path]:
     with NamedTemporaryFile(dir=path.parent, prefix="%s." % path.name) as ntf:
         tmp = Path(ntf.name)
     yield tmp
-    log.debug("Atomically renaming %s -> %s", str(tmp), str(path))
-    tmp.rename(path)
+    if tmp.is_file():
+        log.debug("Atomically renaming %s -> %s", str(tmp), str(path))
+        tmp.rename(path)
```
maddenp-cu, line 65: Make the atomic rename contingent on the temp file existing. This gives the caller the chance to detect an error and remove the temp file, preventing the final file from ever being created. Usually our expectation is that execution doesn't make it to the `rename()` call because e.g. a batch job was terminated or a process was otherwise suddenly killed. But, as in `filecopy_hsi()`, below, a failed `subprocess` task could also be a reason to avoid renaming the temp file to the final-file name.

src/uwtools/utils/tasks.py
```diff
@@ -141,7 +141,10 @@ def filecopy_hsi(src: str, dst: Path, check: bool = True):
     dst.parent.mkdir(parents=True, exist_ok=True)
     with atomic(dst) as tmp:
         cmd = f"{STR.hsi} -q get '{tmp}' : '{src}'"
-        _, output = run_shell_cmd(cmd, taskname=taskname)
+        success, output = run_shell_cmd(cmd, taskname=taskname)
+        if not success:
+            log.error("Failed to copy %s via HSI", src)
+            tmp.unlink(missing_ok=True)
```
maddenp-cu, line 147: If the `hsi` command fails, any output it created is not to be trusted. Unlinking the file here now prevents `atomic()` from trying to do the final rename.

src/uwtools/utils/tasks.py
```diff
@@ -245,6 +248,25 @@ def hardlink(
...
+def link_target(path: Path | str):
+    """
+    An existing file, link, or directory.
+
+    :param path: Path to the file, link, or directory.
+    :param context: Optional additional context for the file.
+    """
+    path = _local_path(path)
+    yield "Target %s" % path
+    yield Asset(path, path.exists)
```
maddenp-cu, line 261: Just moved in the module.

src/uwtools/utils/tasks.py
```diff
@@ -245,6 +248,25 @@ def hardlink(
...
+
+
+@external
+def poison(taskname: str):
+    yield taskname
+    yield Asset(None, lambda: False)
```
maddenp-cu, line 267: A trivial but useful task to "poison" a dependency chain, preventing any task that depends on it, or any task that depends on any task that depends on it, etc., from getting to its action code.

src/uwtools/tests/utils/test_tasks.py
```diff
@@ -320,6 +325,18 @@ def test_utils_tasks_link_target(tmp_path, wrapper):
...
+def test_utils_tasks_poison():
+    node = tasks.poison(taskname="Unfulfilled requirement")
+    assert not node.ready
+    assert node.ref is None
+
+
+def test_utils_tasks__bad_scheme():
+    with raises(UWConfigError) as e:
+        tasks._bad_scheme(path="foo://x/y/z", scheme="foo")
+    assert str(e.value) == "Scheme 'foo' in 'foo://x/y/z' not supported"
```
maddenp-cu, line 337: We were missing a direct test for the function now tested here.

src/uwtools/drivers/driver.py
```diff
@@ -373,6 +374,9 @@ def run(self):
         A run.
         """
         yield self.taskname(STR.run)
+        if self.config[STR.execution][STR.executable] is None:  # e.g. from a YAML null
+            msg = "%s must define 'executable' or implement custom run() method"
+            yield poison(msg % self.__class__.__name__)
```
maddenp-cu, line 379: A driver is now able to set `executable: null` in its config, but if it does so it must override the `run()` method with an implementation that does the necessary work in Python instead of delegating to a runscript.

christinaholtNOAA (approved):
LGTM.

PR #976, opened 2026-09-04
Constrain schema references to package resources

maddenp-cu:
**Synopsis**

A potential user pointed out the possibility of a kind of data exfiltration via `$ref` references to schema files that contain relative paths, e.g. a series of `..` path components that traverse the filesystem, out of the package-resources directory toward the root, then down to files in arbitrary directories.

This PR addresses this by checking that package-resource references, before they are returned to the caller, remain inside the package-resources directory.

**Type**

**Impact**

**Checklist**

### DIFF
--- a/src/uwtools/config/jinja2.py
+++ b/src/uwtools/config/jinja2.py
@@ -296,6 +296,7 @@ def _deref_render(val: str, context: dict, local: dict | None = None) -> str:
     :return: The rendered value (potentially unchanged).
     """
     if not any(marker in val for marker in ("{{", "{%", "{#")):
+        deref_debug("Rendered", val)
         return val
     env = _register_filters(Environment(undefined=StrictUndefined))
     template = env.from_string(val)
--- a/src/uwtools/config/validator.py
+++ b/src/uwtools/config/validator.py
@@ -74,7 +74,7 @@ def internal_schema_file(schema_name: str) -> Path:
 
     :param schema_name: Name of uwtools schema to validate the config against.
     """
-    return resource_path("jsonschema") / f"{schema_name}.jsonschema"
+    return resource_path(f"jsonschema/{schema_name}.jsonschema")
 
 
 def validate(schema: dict, desc: str, config: JSONValueT) -> bool:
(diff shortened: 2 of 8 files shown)


src/uwtools/config/validator.py
```diff
@@ -209,14 +204,17 @@ def _resolver(schema: dict) -> RefResolver:
...
 
-    return cast(RefResolver, RefResolver.from_schema(schema, handlers={"urn": retrieve}))
+def _schema(uri: str) -> dict:
+    name = uri.rsplit(":", maxsplit=1)[-1]
+    schemaroot = resource_path("jsonschema")
+    path = resource_path(f"jsonschema/{name}.jsonschema")
+    if not path.resolve().is_relative_to(schemaroot.resolve()):
+        raise UWError("Invalid schema URI: %s" % uri)
+    text = path.read_text()
+    return cast(dict, json.loads(text))
```
maddenp-cu, line 217: This factored-out helper enforces the constraint that scheme references resolve to paths under the `jsonschema/` package-resources directory. A separate check, in `resource_path()` (see below), provides a more general constraint on the location of _any_ package-resource request.

src/uwtools/utils/file.py
```diff
@@ -125,8 +126,12 @@ def resource_path(suffix: str = "") -> Path:
     :param suffix: A subpath relative to the location of the uwtools resource files. The prefix path
         to the resources files is known to Python and varies based on installation location.
     """
-    with resources.as_file(resources.files("uwtools.resources")) as prefix:
-        return prefix / suffix
+    root = resources.files("uwtools.resources")
+    with resources.as_file(root) as prefix:
+        path = prefix / suffix
+        if not path.resolve().is_relative_to(prefix.resolve()):
+            raise UWError("Resource reference '%s' is outside package resources" % suffix)
```
maddenp-cu, line 133: This check ensures that any package-resource request only returns a path within the package-resources directory.

src/uwtools/config/jinja2.py
```diff
@@ -296,6 +296,7 @@ def _deref_render(val: str, context: dict, local: dict | None = None) -> str:
     :return: The rendered value (potentially unchanged).
     """
     if not any(marker in val for marker in ("{{", "{%", "{#")):
+        deref_debug("Rendered", val)
```
maddenp-cu, line 299: Unrelated, but the PR that sped up the unit tests led to a change in debug logging that showed up via `make docs`. This line restores the previous behavior.

elcarpenterNOAA (approved):

PR #982, opened 2026-09-15
Add --all switch to uw rocoto iterate

maddenp-cu:
**Synopsis**

Add a `--all` switch, mutually exclusive with `--task <name>`, to `uw rocoto iterate`, which ultimately sets the `-a` (aka `--all`) switch on `rocotorun`, requesting execution of all workflow tasks. This eliminates the need to define a special depends-on-everything task to use with the existing `--task` switch to drive the whole workflow to completion. This addition prompted some changes to the way workflow state is tracked and reported by the `iterate` action, as there's now no single state that describes the whole workflow, which might have tasks in several states (`SUBMITTING`, `RUNNING`, `COMPLETED`, etc.) at once. To deal with this, the workflow is now either in `active` or `inactive` state, depending on whether it has any tasks that are still eligible to run.

I successfully drove an AIGFS workflow to completion on Ursa using the code on this branch.

**Type**

- [x] Documentation

**Impact**

**Checklist**


src/uwtools/rocoto.py
```diff
@@ -133,17 +134,21 @@ def __del__(self):
             self._con.close()
 
     def iterate(self) -> bool:
-        state = self._state
-        while state not in self._states[self.State.INACTIVE]:
+        state_type = self._state_type
```
maddenp-cu, line 137: Instead of tracking individual Rocoto task states, which don't make sense with `--all`, track _types_ of states, i.e. activate states, transient states, and inactive states. Update the terminology `state` -> `state_type` to harmonize with this change.

src/uwtools/rocoto.py
```diff
@@ -162,14 +167,18 @@ def _cursor(self) -> sqlite3.Cursor | None:
...
+        return data
 
     @property
     def _query_stmt(self) -> str:
-        return "select state from jobs where taskname=:taskname and cycle=:cycle order by id desc"
+        terms = ["cycle=:cycle"]
+        if self._task:
+            terms.append("taskname=:taskname")
+        constraints = " and ".join(terms)
+        return f"select state from jobs where {constraints} order by id desc"  # noqa: S608
```
maddenp-cu, line 181: Make a more selective SQL query when `--task` is specified.

src/uwtools/rocoto.py
```diff
@@ -181,31 +190,68 @@ def _report(self) -> None:
...
+        if self._all:
+            return f"Rocoto tasks for cycle {self._cycle}: %s"
+        return f"Rocoto task '{self._task}' for cycle {self._cycle}: %s"
+
+    def _state_to_state_type(self, state: str) -> _RocotoIterator.State:
+        for state_type in self.State:
+            if state in self._states[state_type]:
+                return state_type
+        msg = f"Unexpected state: {state}"
+        raise AssertionError(msg)
```
maddenp-cu, line 208: This is synonymous with calling `assert`, and is appropriate here because this should never happen, and is an internal error if it does.

src/uwtools/rocoto.py
```diff
@@ -181,31 +190,68 @@ def _report(self) -> None:
...
+                ROCOTO.ERROR,
+                ROCOTO.STUCK,
+                ROCOTO.SUCCEEDED,
+            ],
+            self.State.TRANSIENT: [
+                ROCOTO.CREATED,
+                ROCOTO.DYING,
+                ROCOTO.STALLED,
+                ROCOTO.SUBMITTING,
+            ],
```
maddenp-cu, line 254: Factored out string literals to `uwtools.strings.ROCOTO`.

AGENTS.md
```diff
@@ -0,0 +1,8 @@
+# Coding Agent Instructions
```
maddenp-cu, line 1: Added this to help keep agents on track, and to avoid repeating the same advice. We can add to this as we discover other behavioral invariants we'd like to enforce.

AGENTS.md
```diff
@@ -0,0 +1,8 @@
...
+
+- Always consider your work incomplete until the command `make format && make test` succeeds. The developer will start the agent in an environment in which those commands are available; if you find that not to be the case, report it to the developer.
+- Follow existing conventions, idion, styles, etc. in the codebase. If you'd like to deviate from existing conventions, discuss it with the developer.
```
elcarpenterNOAA:
```suggestion
- Follow existing conventions, idioms, styles, etc. in the codebase. If you'd like to deviate from existing conventions, discuss it with the developer.
```

src/uwtools/rocoto.py
```diff
@@ -181,31 +190,67 @@ def _report(self) -> None:
...
-        return f"Rocoto task '{self._task}' for cycle {self._cycle}: %s"
+    def _state_type(self) -> _RocotoIterator.State | None:
+        def get_state_type(states: list[str]) -> _RocotoIterator.State | None:
+            if not states:
+                return None
+            state_types = set(map(self._state_to_state_type, states))
+            for state_type in [self.State.ACTIVE, self.State.TRANSIENT]:
+                if state_type in state_types:
+                    return state_type
+            return self.State.INACTIVE
```
elcarpenterNOAA: Just out of curiosity, why'd you go with an inner function?

elcarpenterNOAA (approved):

src/uwtools/rocoto.py
maddenp-cu: Good question. I think it emerged from refactoring: It might have been a local lambda first, then maybe a function with a very short name, etc. Since it wasn't needed outside the enclosing function, nothing forced it out, and the unit tests of the enclosing function covered it. It doesn't benefit from being a closure, though, so maybe it would make sense to factor it out.

AGENTS.md
```diff
@@ -0,0 +1,8 @@
+# Coding Agent Instructions
+
+- Always consider your work incomplete until the command `make format && make test` succeeds. The developer will start the agent in an environment in which those commands are available; if you find that not to be the case, report it to the developer.
+- Follow existing conventions, idioms, styles, etc. in the codebase. If you'd like to deviate from existing conventions, discuss it with the developer.
+- If test failures appear to be caused by sandboxing or environment limitations, report them to the developer instead of adding workaround code. Developers will run tests independently and will tell coding agents if any outstanding issues remain.
+- Never issue state-changing `git` commands without authorization from the developer.
+- Order functions in modules, methods in classes, keys in dictionaries, items in lists, etc. lexicographically unless there is a specific reason for ordering them some other way.
```
christinaholtNOAA: Should this one also include a stipulation to comment on alternative ordering?

christinaholtNOAA (approved):
Nice change! I am happy to see the new AGENTS.md, too. I might pick it up in some other projects. :)

AGENTS.md
maddenp-cu: Good idea, will add.

